PR #14, opened 2015-03-17
BZ1201853: Huge amount ClassNotFoundException WARNs on Project build

manstis:
@wmedvede @mswiderski 

This PR is for https://bugzilla.redhat.com/show_bug.cgi?id=1201853.

It sets the "package name white-list" to the default package created for new projects: https://github.com/droolsjbpm/kie-wb-common/commit/8125e298cb4f65787e688659afe58189cff70b85#diff-5043480836f5a8381d4aeced0fb90d8fR342

It also uses the "package name white-list" for filtering classes

(a) Added to the DataModelOracle - which contains the meta-data for classes available for authoring:

https://github.com/droolsjbpm/kie-wb-common/commit/8125e298cb4f65787e688659afe58189cff70b85#diff-37f1dae81fc151bee95c9e1ee6831b71R98

(b) Validated for integrity (i.e. the class can be loaded) - the source of the WARNs mentioned in the BZ:

https://github.com/droolsjbpm/kie-wb-common/commit/8125e298cb4f65787e688659afe58189cff70b85#diff-45fdd57129750eb9eb41a4a2f753303fR186

Ultimately the cause of the BZ is the User adding jbpm-executor as a dependency (see https://bugzilla.redhat.com/show_bug.cgi?id=1201853#c9). Even with the "package name white list" KIE's KieModuleMetaData.Factory.newKieModuleMetaData(...) still resolves all classes in the KIE Project from all direct and indirect dependencies in the pom... so "package name white-list" makes some aspects quicker but not all. 

https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java#L1214

### DIFF
--- a/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/cache/LRUProjectDataModelOracleCache.java
+++ b/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/cache/LRUProjectDataModelOracleCache.java
@@ -1,9 +1,7 @@
 package org.kie.workbench.common.services.datamodel.backend.server.cache;
 
 import java.io.IOException;
-import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import javax.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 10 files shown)


mswiderski:
looks good to me. 

One thing we might want to look at is the KieModuleMetaData is to add an option to provide some sort of filtering to skip some packages as well. Same as the builder does.  Wdyt?

Anyway I guess that's less critical as KieModuleMetaData does not create instances of all the classes, it just collect their names, if I understood the logic right.

manstis:
@mswiderski 

Thanks for looking Maciej. I'll merge this PR.

KieModuleMetaData is harder (if not impossible to fix). This uses Aether to resolve all dependencies for a pom and then iterates all dependencies extracting meta-data for all classes. Even if we change this to use the "package name white-list" the pom/dependency resolution is the biggest time-sink and there's little we can do - it delegates to org.eclipse.aether for dependency resolution.

manstis:
For information, jbpm-executor has ~85 dependencies (resolved by Aether) and ~16,000 classes available for authoring... no wonder adding it as a dependency stresses the workbench.

PR #30, opened 2015-05-20
BPMSPL-160: DRL text editor improvements.

manstis:


porcelli:
http://github.com/droolsjbpm/kie-wb-common/commit/c8643a624

PR #31, opened 2015-05-20
Fix off-by-one error in NewResourcesMenu.

mbarkley:
Found this small error while poking around. I believe `getMenuItemsWithoutProject` was only meant to filter out the first element of `items`.


manstis:
http://github.com/droolsjbpm/kie-wb-common/commit/145e78d7d

manstis:
Thanks Max. 

I had a look too and the getMenuItemsWithoutProject() method is flawed anyway (even with your fix).

Method setup() tries to ensure NewProjectHandler is always at the top of the "New.." menu. This is fine in the case of kie-wb and kie-drools-wb as we always have a NewProjectHandler class; however if another project using kie-wb-common did not have a NewProjectHandler then items.get(0) isn't going to be the NewProjectHandler and what is returned from getMenuItemsWithoutProject() is going to be wrong.

If you care fixing that for us too it'd be great and you get your name in lights twice! ;)

Cheers,

Mike

mbarkley:
Micheal,

I don't mind fixing this at all :-)

PR #32, opened 2015-05-20
Return all items if there is no NewProjectHandler.

mbarkley:
This is a follow up from [this PR](https://github.com/droolsjbpm/kie-wb-common/pull/31#issuecomment-103962460).


manstis:
Thanks Max.

http://github.com/droolsjbpm/kie-wb-common/commit/7d64be7e4

PR #34, opened 2015-05-26

wmedvede:
Project class loader caching to optimize datamodeler reading/writing operations

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
@@ -29,6 +29,10 @@
     </dependency>
     <dependency>
       <groupId>org.drools</groupId>
+      <artifactId>drools-compiler</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
       <artifactId>drools-workbench-models-datamodel-api</artifactId>
     </dependency>
     <dependency>
@@ -48,6 +52,10 @@
     <dependency>
       <groupId>org.guvnor</groupId>
       <artifactId>guvnor-services-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.guvnor</groupId>
+      <artifactId>guvnor-project-builder</artifactId>
     </dependency>
     <dependency>
       <groupId>org.jboss.errai</groupId>
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceHelper.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceHelper.java
@@ -30,6 +30,8 @@ import javax.inject.Named;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.output.ByteArrayOutputStream;
+import org.drools.compiler.kie.builder.impl.InternalKieModule;
+import org.drools.core.util.ClassUtils;
 import org.guvnor.common.services.project.model.Project;
 import org.guvnor.common.services.project.model.Package;
 import org.guvnor.common.services.project.service.ProjectService;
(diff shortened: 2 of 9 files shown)


manstis:
Hi Walter,

Does this PR really bring any gains?

https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/LRUBuilderCache.java#L71 will invalidate a Builder in LRUBuilderCache whenever any file (in the corresponding project) that invalidates the existing caches; e.g. pom.xml, kmodule.xml, *.java, *.class etc (see my email) changes.

Therefore whenever one of these files changes the existing Builder in LRUBuilderCache will be destroyed and re-created. This leads to an entry in LRUProjectDependenciesClassLoaderCache being re-created too. So your new cache will not be "hit" whenever the pom.xml changes.

It will be "hit" whenever requests for the Project Dependencies Class Loader is requested; but your comment in the corresponding cache states you added it to improve performance due to the "...the project ClassLoader may change with ever incremental build..." - this is only true if the pom.xml changes in which case the Builder (and hence LRUProjectDependenciesClassLoaderCache entry) will be destroyed and rebuilt anyway.

Finally, do you solely rely on the LRUBuilder cache invalidating a Builder to effectively invalidate your LRUProjectDependenciesClassLoaderCache? i.e. the Builder is destroyed and re-created? Is it not safer to also ```@Observe``` the InvalidateDMOProjectCacheEvent event?

Am I missing something?

Cheers,

Mike

wmedvede:
Hello Mike, I agree, the mechanism is not perfect :).

I'll try to explain a bit better.

In previous version I've calculated the project class loader with a code like this (basically relaying on the KieModuleMetadata).

    private ClassLoader getProjectClassLoader(KieProject project) {

        final KieModule module = builderCache.assertBuilder(project).getKieModuleIgnoringErrors();
        final ClassLoader classLoader = KieModuleMetaData.Factory.newKieModuleMetaData(module).getClassLoader();
        return classLoader;
    }


But, I've found that when the KieModuleMetadataImpl is initialized, then the maven dependencies are also calculated. So, every time I had to access the project class loader dependencies calculation was performed, with the subsequent delay.

So I've just done some simple reasoning:
Every time a .java or pom.xml file is changed, then the LRUBuilderChache will invalidate it's entry for the given project.

After this, no matter if a full build or incremental build is performed, a new Builder will be crated for the given project. BUT, since a new Builder is created, even the incremental builder will do a full build.
If you see the ResourceChangeObserver, all the methods invoked on the incrementalBuilder instance, will internally execute the full build, since the Builder instance is new.

Now, given that when when the Buidler builds the project we are calculating the KieModuleMetadata I use this invocation to store the dependencies class loader in the cache. With the hope that this can help in the future. I've only cached the part of the class loader that has the dependencies, since eventually other resources like DRL files that has a real incremental build (and not creates a new Builder), can generate classes that I could need in the project class loader in a later stage.

So, next time I need the project class loader, I go to the cache an take the part of the class loader that has the dependencies, and then the current project classes from the KieModule.

see:

https://github.com/wmedvede/kie-wb-common/blob/data-modeler-6.3-project-classloader-optimization/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceHelper.java#L509

Given that it's just an optimization I've just added some lines of code in the builder, relaying on current implementation and not on the InvalidateDMOProjectCacheEvent. So I just let the code we have run, and collect info a some point of interest.

I guess it works, the bad thing is that whenever a .java file is modified, in the end we have a full build, and then the dependencies calculation. So there are not to much benefits. Except in some scenarios: (2 and 3 are important)

1) when for example I save the file, then I need the project class loader both a save time, and indexing time. With this mechanism only the first invocation will have to wait for the dependencies calculation.

2) other scenario is when I have to switch form the source tab (in edition mode) to the editor tab, and viceversa. The user doesn't save anything but I still need to go to the server and do some parsing calculation. With this mechanism I can get the project classloader with no delay.  (of course assuming no other user has changed the pom.xml, or .java files)

3) a similar scenario is when the user modified the java source, and now wants to save the file. Before saving I need to go to the server and try to parse the file and try to see if the package name, or class name has changed. If so, I'll ask the user for a file rename, etc.

But I agree, it's not perfect. I guess the power solution will be to cache the dependencies calculation and when the maven build is called then try to use the cached calculation. If the pom.xml file changes, then we should calculate the dependencies again.

manstis:
http://github.com/droolsjbpm/kie-wb-common/commit/85ad82a37

PR #37, opened 2015-06-03
GUVNOR-2190 : Ace editor DRL syntax hilighting for the source tab.

Rikkola:
Had to move the DrlEditor class to kie-wb-common. PR to drools-wb here 
https://github.com/droolsjbpm/drools-wb/pull/36


manstis:
http://github.com/droolsjbpm/kie-wb-common/commit/c39653349

PR #41, opened 2015-06-09

wmedvede:
Silly commit to make the properties and classes tables to be repainted when the view loses the focus


manstis:
http://github.com/droolsjbpm/kie-wb-common/commit/3052f6fb8

PR #42, opened 2015-06-09
Lock support for project editor

csadilek:
Hi @manstis, @Rikkola,

This PR adds lock support to the project editor. The locks are fine-grained so that one user editing the dependencies doesn't stop others from modifying imports or the kmodule settings.

This is a minimally invasive change and doesn't affect the existing project editor design i.e. it doesn't convert it to a workbench editor or use nested editors. It does work fine and since we need a solution soon it seems we can postpone the bigger refactoring, if still desired.

Cheers,
Christian


Rikkola:
https://github.com/droolsjbpm/kie-wb-common/commit/fc743b7cfc81f97cbd64bf0e687d04dcf702fa31

PR #43, opened 2015-06-11
Data modeller 6.3 domain editors

wmedvede:
This PR adds plug-ability for client side domain editors, and server side handlers to enable by default settings for the different domains. 
Not definitive since there will be likely modifications when the mechanims is refactored to use UF tooling windows. But it'll still be good to be integrated into master prior migration to BS3, etc.


wmedvede:
forgot to squash.....

PR #46, opened 2015-06-12
BZ-1207162: Keep Register Server dialog open if connection fails

porcelli:


manstis:
https://github.com/droolsjbpm/kie-wb-common/commit/2bf2e8a1db63a98a7238d63120527e882cfdf806

PR #50, opened 2015-06-25
Dynamically Load Additional Build Options in Project Screen

mbarkley:
This PR adds extensibility to the `ProjectScreenPresenter` so that other modules can implement additional build options in the build drop down.

The motivation for this change comes from LiveSpark, where we want to offer to different kinds of build/deployment for production and development.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java
@@ -195,7 +195,7 @@ public class ProjectScreenPresenter
         this.buildOptions = view.getBuildOptionsButton();
 
         makeMenuBar();
-        
+
         reloadRunnable = new Runnable() {
 
             @Override
@@ -203,7 +203,7 @@ public class ProjectScreenPresenter
                 ProjectScreenPresenter.this.reload();
             }
         };
-        
+
         titleProvider = new TitleProvider() {
 
             @Override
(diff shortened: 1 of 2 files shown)


kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java
```diff
@@ -199,6 +213,73 @@ public String getTitle() {
         };
     }
 
+    private void configureBuildExtensions( Project project, DropdownButton buildDropdownButton ) {
```
manstis: Hi Max, 

I'm intrigued with this code.. why do you destroy ```@Dependent``` scoped ```BuildOptionExtension```?

This isn't the only place in KIE Workbench we use BeanManager to get implementations of an interface and wonder if we need to change these to safe-guard against whatever you're doing this here for too?

Cheers,

Mike

manstis: I assume it's to avoid a memory leak when the Project Editor is closed and re-opened and ```@Dependent``` scoped beans are recreated by ```getBuildExtensions()```?

mbarkley: Hi Michael,

The short answer is that any `@Dependent` beans looked up with the bean manager will never be destroyed unless explicitly done by the caller. This can lead to weird problems (aside from just memory leaks) if these beans have `@Observes` methods, fire events, et cetera.

In this case, I chose to destroy the `@Dependent` beans right away for simplicity, but another option that is probably more appropriate (though more complicated) in most places is to hold references to the looked-up beans and destroy them when the bean that looked them up is destroyed.

Cheers,
Max

manstis:
http://github.com/droolsjbpm/kie-wb-common/compare/e784193...c077fa2

PR #52, opened 2015-06-29
BZ1236100: Message Console does not get shown in the AuthoringPerspectiveNoContext perspective

manstis:


manstis:
https://github.com/droolsjbpm/kie-wb-common/commit/92dcc209a

PR #53, opened 2015-07-10
Kie builder optimizations

wmedvede:
The optimization includes the caching of the PomModel for a given project.
The PomModel will be reused whenever it's possible for the KieBuilder creation, this will reduce the KieBuilder initialization procedure given that the pom.xml parsing with the subsecuent dependencies calculation will be skipped.


manstis:
http://github.com/droolsjbpm/kie-wb-common/commit/418978c77

PR #63, opened 2015-08-03
BPMSPL-100: refactor to add the tag selector on Technical & Business View 

pefernan:


Rikkola:
http://github.com/droolsjbpm/kie-wb-common/commit/0bd6d0811

PR #64, opened 2015-08-23
Fix indexing tests by disabling Git daemon and ssh

psiroky:
* this fixes a lot of test failures in drools-wb
   (when running the tests in parallel)


Rikkola:
https://github.com/droolsjbpm/kie-wb-common/commit/af6e6e0de7918215bcd589af614ccf745cd4b04f

PR #65, opened 2015-08-24
Bug 1254907 - Items of project explorer are not refreshed automatically

Rikkola:
Few goals here. 
1. Fix the bug
2. Making the code more unit testable
3. Splitting the responsibilities in BaseViewPresenter to smaller units. The logic for what gets shown and when we update is getting out of hand. I hope the changes help more than confuse. 

This PR depends on a guvnor PR and a PR in kie-wb-distributions depends on this.


Rikkola:
This was needed for the release so merged in.
https://github.com/droolsjbpm/kie-wb-common/commit/a85573710b986444dd14f2ada4b3efce377b737b

PR #66, opened 2015-08-24
Bug 1250349 - Exceptions are not logged in server.log file if not abl…

Rikkola:
…e to connect to kie-server


Rikkola:
https://github.com/droolsjbpm/kie-wb-common/commit/25064d52b95b7ec2edc02906d088f4625b080154

PR #67, opened 2015-08-24

wmedvede:
- i18n added for the Persistence Descriptor Editor


psiroky:
@wmedvede just out of curiosity, what is the reason to create a PR and then self-merge it few minutes later? I am seeing PRs as way for others to review things, which does not happen this way (at least not before the changes land in the branch). 

I know that PR is more visible (comparing to direct push), so there is higher chance someone notices the changes and looks at them. Maybe this is worth creating the PR.

PR #75, opened 2015-09-04
Make business/technial views of PEX consistent

jhrcek:
..and add some IDs for selenium testing

The only substantial change is reordering of Project explorer's controls in BusinessViewWidget.ui.xml to be consistent with TechnicalViewWidget.ui.xml (see screenshot before and after change https://ctrlv.cz/Z3dA)

In addition to that I added some DOM IDs that will make maintenance of our selenium tests easier. It's much easier for us to maintain relatively stable locators ala "div[id^='pex_content']" than having to update brittle locators like "div div[aria-hidden=true]+div>div>div>div>div>div:not([aria-hidden=true])" after every change done to the view.


jhrcek:
If this PR is acceptable, please also cherry pick to 6.3.x for product testing.

manstis:
*kie-wb-common*
(master) https://github.com/droolsjbpm/kie-wb-common/commit/eb9e55279

(6.3.x) https://github.com/droolsjbpm/kie-wb-common/commit/8902d646b

manstis:
Thanks @jhrcek

PR #77, opened 2015-09-04
DROOLS-733 - update status of kie container info if opened when kie s…

mswiderski:
…erver instance connects to controller

should go into both master and 6.3.x


manstis:
*kie-wb-common*
(master) https://github.com/droolsjbpm/kie-wb-common/commit/c8a069306

(6.3.x) https://github.com/droolsjbpm/kie-wb-common/commit/32d310a3b

manstis:
Thanks @mswiderski

PR #78, opened 2015-09-04
Hide docks when you put data modeler in background

ederign:
Depends on: https://github.com/uberfire/uberfire/pull/166


manstis:
DataModelerWorkbenchFocusEvent with a "hidden" property? Surely a "focus" event should have a "focus" property or similar?

PR #80, opened 2015-09-07
BZ1202551: Test Scenario result is affected by previous XLS Decision Table validation

manstis:


manstis:
@wmedvede Hi Walter.. you're probably one of the few people around that have looked at the ```Builder``` code. Toni is on PTO, so I wonder if you could merge this PR please. Thanks.

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java
```diff
@@ -526,9 +530,16 @@ public KieContainer getKieContainer() {
...
-            ReleaseId releaseId = kieModule.getReleaseId();
-            KieContainer kieContainer = kieServices.newKieContainer( releaseId );
+            // Do not retrieve the KieContainer with KieServices.newKieContainer(releaseId) since this looks-up the KieModule to
+            // create the KieContainer from KieRepository. This holds the most recent KieModule (for the ReleaseId) that was built with
+            // kieBuilder.buildAll() which *may* be a KieModule created during asset validation and hence will lack many assets.
+            // See https://bugzilla.redhat.com/show_bug.cgi?id=1202551
+            final KieModule kieModule = kieBuilder.getKieModule();
+            final ReleaseId releaseId = kieModule.getReleaseId();
+            final org.drools.compiler.kie.builder.impl.KieProject kieProject = new KieModuleKieProject( (InternalKieModule) kieBuilder.getKieModule(), null );
+            final KieContainer kieContainer = new KieContainerImpl( kieProject,
```
mswiderski: @manstis , wouldn't it be better to provide method on kieServices that skips the KieRepository look up instead of using here the implementation classes (KieModuleKieProject, KieContainerImpl)? Just wondering as it might be useful in other places/use cases, wdyt?

manstis: @mswiderski Hi.. to be honest I first created a unit test in drools-compiler showing the issue; and was going to ask @mariofusco to have ```KieBuilder.buildAll()``` skip deployment to ```KieRepository``` but I thought the impact might be too great. However perhaps adding a new method to create a ```KieContainer``` from a ```KieModule``` would be a better option (as you suggest). Let's see what @mariofusco thinks on his return.... OK if I reassign this PR to you (now you're involved!)?

mswiderski: sure thing

PR #88, opened 2015-09-22
BZ1202765: NullPointerException in EnumFileIndexer if fact used in enumeration does not exist

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1202765

Fixed logging of errors.


Rikkola:
http://github.com/droolsjbpm/kie-wb-common/commit/fe65bbb2b

PR #89, opened 2015-09-23
BZ1263713: Opening .gitignore in Administration perspective never finishes

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1263713


manstis:
The problem was that any file prefixed with a "dot" was assumed to be one of the workbench's meta-data files and an attempt was made in ```KieService``` to resolve the "principle file" instead. In valid use-cases (e.g. ```.gitignore``` etc) the "principle file" does not exist which lead to attempts to load the ```MetaData``` object to fail leading to an incomplete ```Overview``` object being returned to the client leading to the View not opening and the ```BusyIndicator``` to remain on screen.

porcelli:
Tests are failing: https://gist.github.com/porcelli/e4e138f3e1e8de78c279

manstis:
@porcelli There was a missing "." file in the PR now duly added.. please re-try.

porcelli:
https://github.com/droolsjbpm/kie-wb-common/commit/182ba0ef306ef2aca9f5f63a87c4620d2fefb932

PR #93, opened 2015-09-25
BZ1263118: 'Repository View' doesn't remember the last visited place

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1263118

The changes made fall into three categories:

1) When ```TECHNICAL_CONTENT``` is requested in ```ProjectExplorerContentResolver``` re-use the ```lastContent.getLastFolderItem().getItem()``` where applicable (we used to always fall back to ```null``` which forced re-load from the project root).

2) We used to pass ```ActiveOptions``` around all over the place, cloning and overwriting content until our hearts content. This lead to ```ActiveOptions``` state depending on the code-path. IMO, this led to inconsistencies. I've replaced the general munging of different instances with a single ```@ApplicationScoped``` CDI bean holding a single instance of ```ActiveOptions``` shared between all applicable consumers.

3) ```TreeNavigator``` was throwing some ```NPEs``` in Chrome Dev Console which was also leading to subsequent code to not run, which could lead to inconsistent state in the UI (e.g. menu items not being updated correctly). I've added a bunch of defensive programming checks to simply avoid NPEs being logged (smells bad, but since this screen is set for another re-write soon I thought not to waste too much time on it). I didn't experience any change in behaviour when re-trying the ```TreeNavigator```.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-backend/src/main/java/org/kie/workbench/common/screens/explorer/backend/server/FolderListingResolver.java
+++ b/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-backend/src/main/java/org/kie/workbench/common/screens/explorer/backend/server/FolderListingResolver.java
@@ -16,6 +16,9 @@
 
 package org.kie.workbench.common.screens.explorer.backend.server;
 
+import java.util.List;
+import javax.inject.Inject;
+
 import org.guvnor.common.services.project.model.Package;
 import org.guvnor.common.services.project.model.Project;
 import org.kie.workbench.common.screens.explorer.model.FolderItem;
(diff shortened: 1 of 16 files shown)


manstis:
@porcelli I hate this screen!

manstis:
I need to cherry-pick to ```6.3.x``` and ```6.2.x``` (nice, what's the chances of that going well!?!)

Rikkola:
I have had my battles with the PE screen myself. I think this area might be worth rewriting at some point. It is really hard to add unit tested code to this area.

porcelli:
http://github.com/droolsjbpm/kie-wb-common/commit/356aa4288

manstis:
*kie-wb-common*
(6.3.x) http://github.com/droolsjbpm/kie-wb-common/commit/adc55abff

manstis:
*kie-wb-common*
(6.2.x) http://github.com/droolsjbpm/kie-wb-common/commit/71f0a14bd

PR #100, opened 2015-10-02
Fix screens following migration to Bootstrap3

manstis:


manstis:
*kie-wb-common*
(master) http://github.com/droolsjbpm/kie-wb-common/commit/dec71a9c1

manstis:
Thanks @manstis!

PR #102, opened 2015-10-04
JBPM-4777: +Responsive design fix for data modeller after BS3 upgrade

csadilek:
This is a responsive design improvement for data modeller on master (after BS3 upgrade and docks resizing) (not needed on 6.3.x).

It makes sure the data modeller renders fine in case of minimal horizontal space. The domain container panel is then placed underneath the data object panel and remains usable.

Before: (https://dl.dropboxusercontent.com/u/16288770/before.png)
After: (https://dl.dropboxusercontent.com/u/16288770/after.png)


wmedvede:
Was merged into master

PR #104, opened 2015-10-05
GUVNOR-2276: GuvnorDefaultEditorViewImpl: LockManager tries to re-initialise View

manstis:
See https://issues.jboss.org/browse/GUVNOR-2276


Rikkola:
https://github.com/droolsjbpm/kie-wb-common/commit/f55d7a593
I threw a unit test there, so that in the future when somebody who likes to refactor with a heavy hand does changes (=me) does not break the fix.

manstis:
*kie-wb-common*
(6.3.x) http://github.com/droolsjbpm/kie-wb-common/commit/a1301acc5

See https://bugzilla.redhat.com/show_bug.cgi?id=1270831

PR #107, opened 2015-10-14
GUVNOR-2268 : Test Scenarios: Switch to using uberfire-extensions DatePicker

Rikkola:
Deleted a pile of dead code. 
Built drools-wb, the jbpm projects and kie-wb-distributions after removing these and they all passed.

This PR in drools-wb needs to go in before this.
https://github.com/droolsjbpm/drools-wb/pull/76


manstis:
*kie-wb-common*
(master) http://github.com/droolsjbpm/kie-wb-common/commit/120004eb7

manstis:
Thanks @Rikkola !

PR #111, opened 2015-10-16
Fixes around GWT compilation for showcase

psiroky:
The diff is quite big, because I had to fix formatting in the pom :(


manstis:
My personal view is -1.

These "temporary" web-apps proved useful during the BS2->BS3 upgrade but I question their use now.

psiroky:
@manstis I am definitely for removing them. But not everyone seemed to be OK with that. I want to at least fix them since they are already in the codebase.

Rikkola:
The benefits are: 
-We see the GWT compile errors in guvnor/kie-wb-common and not in drools-wb
-Once a year somebody runs the webapp and it saves time

Still in the end I think we would end up saving more time if we did not have the time consuming webapp builds in guvnor and kie-wb-common. Removing these would be a step towards faster builds and tests.

psiroky:
Closing this one and will create new PR that will remove the showcase completely as that seems to be what most people agree with.

PR #112, opened 2015-10-16
BZ1272363: Building Indicator is not shown when performing 'Build & Deploy'

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1272363

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/pom.xml
+++ b/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/pom.xml
@@ -192,6 +192,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-testing-utils</artifactId>
+      <scope>test</scope>
+    </dependency>
+
     <!-- GWT and GWT Extensions -->
     <dependency>
       <groupId>com.google.gwt</groupId>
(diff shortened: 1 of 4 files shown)


manstis:
master and 6.3.x please

manstis:
@mswiderski 

I've assigned this PR to you since you're probably most familiar with the "Build Options" button.

mswiderski:
@manstis this was actually done on purpose to not show the busy indicator as build and install and build and deploy are rather long running and async tasks - starts build process as this is only for managed repos. So now when this is enabled when are these should be hidden?

manstis:
Hi @mswiderski I'm glad I assigned this PR to you; as IDK it was by design.

The "Busy" popup is closed in the callback to ```AssetManagementService.buildProject(...)``` which IIRC triggers the processes backing "Build & Install" and "Build & Deploy". These processes have the potential to be long(er) running but the invocation of ```AssetManagementService.buildProject(...)``` should be quick.

If you feel we should continue to not show the "Busy" popup for these actions feel free to say so and I'll revert much of the PR - to only really include the tests.

mswiderski:
wondering if the busy indicator will not actually hide the notification massage [1] as it's closed after notification is sent. In general I believe we should have just one of these - either notify that build process started or show busy indicator.
Personally I opt for the notification because the busy indicator is bit misleading as it says building while the building has not yet started... wdyt?

[1] https://github.com/manstis/kie-wb-common/blob/BZ1272363/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java#L712

manstis:
Fair enough; I'll modify this PR accordingly.

manstis:
@mswiderski OK, I removed surplus "hide" of the "Busy" popup in the "Install" and "Build" commands and modified the tests to check whether Notifications had been raised instead. I'll comment the BZ that the reported behaviour is by design and the reasoning behind it.

PR #115, opened 2015-10-20
BZ1192831: User with no privileges for Repository can view and modify assets in that Repository

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1192831

The last piece of the equation was that Users could see (Search) assets in Repositories that did not explicitly have Groups assigned; but were contained in an OU to which the User lacked access. This change filters authorized Repositories first on the OUs to which the User has access and then the Repositories contained within those OUs.


Rikkola:
https://github.com/droolsjbpm/kie-wb-common/commit/c00e1caed

PR #118, opened 2015-10-20
Delete showcase as it is not needed anymore

psiroky:
* the showcase was primarily created to easy BS3
   upgrade which is now mostly done.


manstis:
It's a +1 from me - but, as discussed on IRC, best get a second opinion
before merging.

psiroky:
I've talked to @csadilek and @Rikkola and they both agreed to remove the showcase.

csadilek:
Thanks!

master: d29dfdeb98facc811e85c056a7228d026524fbc0

cristianonicolai:
I might be missing something, but I do think the showcase apps give us a really nice and isolated way to test UI related features and fixes. For example, I can test and develop my UI without having to build the entire workbench, which is really time consuming. Now that we'll have more and more PR's to review, I don't want to spend a lot of time rebuilding everything to do a simple check. I do understand that the showcase apps creates a maintenance and build time problem, so I just want to raise the point for us to think how we can address it. Maybe what we need is less repositories, but still have showcases in there, or it is simply me that need to know how to be more efficient on handling all of these modules :)

Rikkola:
We still have the showcase in drools-wb. What I do is work on guvnor+kie-wb-common+drools-wb. Here drools-wb has the webapp showcase. Now the designer uses guvnor+kie-wb-common+jbpm-designer. Here jbpm-designer has the showcase. Same for jbpm-form-modeler, optaplanner-wb and so on. I don't think anyone uses kie-wb when adding new features.

Or this is at least how we planned things would work, but...
At least optaplanner-wb needs the editors from drools-wb. I'm suspect that the other might do the same. So it might be possible to merge kie-wb-common and drools-wb. So we can have one less repo. As a dev who does most of his changes into these two it would help if there was only one.

PR #119, opened 2015-10-21
BZ1192831: User with no privileges for Repository can view and modify assets in that Repository

manstis:
PR for 6.3.x branch.


Rikkola:
https://github.com/droolsjbpm/kie-wb-common/commit/e6769e21b85071a2eccde573b5e1550a87029fbe

PR #120, opened 2015-10-21
BZ-1211257 - Comments always show the 'current' time

wmedvede:
MVP implementation of some missing part and the BZ is fixed.
DateFormatHelper class was added in order to keep the CommentLinePresenter as independent of GWT as possible, but even so the client side formatting is done by using GWT classes. Btw, I think it's enough for this BZ. 

Of course latter PR+cherry-pick into master will be needed.


kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/util/DateFormatHelper.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.widgets.client.util;
+
+import java.util.Date;
+
+import com.google.gwt.i18n.client.DateTimeFormat;
+
+public class DateFormatHelper {
```
manstis: It'd be good to use the Workbench "date" formats settings. WDYT?

```
private final static String DATE_FORMAT = ApplicationPreferences.getDroolsDateFormat();
private final static DateTimeFormat FORMATTER = DateTimeFormat.getFormat( DATE_FORMAT );
```

wmedvede: I think that as soon the current format is changed, there will be people at some other place of the earth blaming about it. But trying be positive, this is what we have.

if we use DATE_FORMAT then something like this will be shown: "21-Oct-2015", but it seems to be worth to show also the hour/min, so
if we use DATE_TIME_FORMAT then something like this will be shown: "21-Oct-2015 06:20:14"
Se that the hour is shown in 1..12 format, ideally 18 should be shown instead of 06.

Summarizing:

Since I don't know where the current value for the DATE_TIME_FORMAT is set, or if it has some collateral effects on other workbench areas if changed, since it seems to use a system property "drools.datetimeformat" and the prefix "drools" makes me worry about if this is used for some important drools thing.

So, if YOU change the DATE_TIME_FORMAT or DATE_FORMAT default value to something that we can use, and warranty it doesn't break any other "drools" related thing, it's ok for me to use this value.

manstis: Well argued! I'll leave @Rikkola to review (and +1) since he knows this area of the code like his own son!

kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/discussion/CommentLineViewImpl.java
```diff
@@ -53,14 +43,29 @@
...
+    @Override
+    public void setAuthor( String author ) {
+        this.author.setText( author );
+    }
 
-        this.author.setText(record.getAuthor() + ":");
-        this.comment.setText("\"" + record.getNote() + "\"");
-        this.date.setText(
-                DateTimeFormat.getFormat(DateTimeFormat.PredefinedFormat.DATE_TIME_SHORT).format(
-//                                new Date(record.getTimestamp())
```
Rikkola: It is best to remove all commented code before doing commit. The next guy looking at this will think this is needed and is afraid to remove it. Then years from now we have a pile commented code that everybody is afraid to touch, because there might be a hidden message.

That being said. I try to commit commented code with TODOs all the time, but these reviews stop me ;)

kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/util/DateFormatHelper.java
Rikkola: We actually have, or used to have, two date format settings.

#1 The drools date format we get from Application Preferences is the one used by DRL source.

#2 Then we had another format for what the date pickers in the editors use. This is because the users wanted the default drools format for DRL, to make the maintaining the rules easier (if you have two rule bases in the same knowledge session with different date/time you get trouble), but the users were still using the editors with a language that we do not even have i18n for, for example Finnish. They wrote all the rules with DSL.

The code here is the third use case. Showing "logs" to users. 

I would propose we use the browser defaults for anything that is shown to the web-user ("logs" and editors) and then we use the application preferences only for DRL. 

@manstis I know it is possible to override the browser language. Does that change the date and time format?

kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/discussion/CommentLinePresenter.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2014 JBoss Inc
```
Rikkola: 2015, but this is not enough to reject the PR.

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/discussion/CommentLinePresenterTest.java
```diff
@@ -0,0 +1,81 @@
...
+    private CommentLineView view;
+
+    @InjectMocks
+    CommentLinePresenter presenter;
+
+    @Test
+    public void testVisualization() {
+
+        CommentLineView view = mock( CommentLineView.class );
+        CommentLinePresenter presenter = new CommentLinePresenterWithNOGWTCode( view );
```
Rikkola: Good trick to get around the GWT.create.

Rikkola:
+1

csadilek:
Thanks!

6.3.x: 7a51dcdf7468363426f68ec773589ec1b54f3b65

*Don't forget to send it to master as well ;)*

PR #124, opened 2015-10-30
BZ1275325: WAS: indexing fails with ContextNotActiveException

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1275325

Implementation of EJB was flawed; it was neither an "no interface" bean, nor a "local interface" bean.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/KieServerWBControllerAdminImpl.java
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/KieServerWBControllerAdminImpl.java
@@ -15,7 +15,6 @@
 
 package org.kie.workbench.common.screens.server.management.backend;
 
-import java.util.concurrent.Executor;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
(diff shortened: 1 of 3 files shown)


mswiderski:
@manstis just a question as I ran into it locally when trying your early bits - when I opened project explorer I got a popup window with NoSuchMethodError referring to SimpleAsyncExecutorService. Though it might have been only my local issue though I wanted to double check that is not a general one.

manstis:
@mswiderski I've ran kie-drools-wb on EAP, Wildfly and EAP with these changes and on all occasion navigated to Authoring->Project Explorer->an asset->opened the asset.. Each installation was fresh - but we don't persist any information relating to SimpleAsyncExecutorService, so I wouldn't have expected problems. Let me just try kie-wb on EAP with these changes, in case I've missed something somewhere that was not caught as a compilation error.

manstis:
@mswiderski Hi, I ran kie-wb on EAP and generally fiddled around on Project Explorer. No errors thrown.  Project Explorer uses the [SimpleAsyncExecutorService](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-backend/src/main/java/org/kie/workbench/common/screens/explorer/backend/server/ExplorerServiceImpl.java#L231) to persist it's state; however it's usage hasn't been modified by any of these PRs.

mswiderski:
@manstis then I +1 this :) great job!

csadilek:
6.3.x: 0bbad9a1d9e84e6bc32d5a742e1dce16f060b196

PR #125, opened 2015-10-30
BZ1275325: WAS: indexing fails with ContextNotActiveException

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1275325

Implementation of EJB was flawed; it was neither an "no interface" bean, nor a "local interface" bean.


csadilek:
master: 132627dd0ac70338760feb870526ef5cce42af62

PR #126, opened 2015-11-02
JBPM-4818 - fetch controller info from system properties

sutaakar:


manstis:
@mswiderski Can you please review this?

sutaakar:
Just noticed an error in there, will update PR.

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/ServerManagementServiceImpl.java
```diff
@@ -98,8 +99,8 @@
     private KieServerControllerAdmin controllerAdmin;
     private KieServerControllerStorage controllerStorage;
 
-    private String controllerUser = "kieserver";
-    private String controllerPassword = "kieserver1!";
+    private String controllerUser = System.getProperty(KieServerConstants.CFG_KIE_CONTROLLER_USER, "kieserver");
+    private String controllerPassword = System.getProperty(KieServerConstants.CFG_KIE_CONTROLLER_PASSWORD, "kieserver1!");
```
mswiderski: seems like the wrong constant is used here as the controller user/password should be used from kie server to connect to controller and not the other way around. was this the issue you found @sutaakar?

sutaakar: Yes, just rewritten PR to use ControllerUtils.getUser()/ControllerUtils.getPassword()

sutaakar:
PR updated

mswiderski:
could we add a test to make sure the changes via system properties are taken into consideration? I guess by mocking injection points we should be able to intercept the invocation. wdyt?

sutaakar:
I think it is good idea, will prepare some test.

sutaakar:
@mswiderski Added username/password check as part of improved testStartContainer.

mswiderski:
this looks very good to me +1

csadilek:
Thanks!

master: 97570320ab1d3b9dee21a146aae8db906e031840

PR #128, opened 2015-11-02
BZ1274349: Asset Management: Branch selector partially hidden

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1274349

Original PR for 6.3.x: https://github.com/droolsjbpm/kie-wb-common/pull/127

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/BranchSelectorViewImpl.java
+++ b/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/BranchSelectorViewImpl.java
@@ -75,7 +75,7 @@ public class BranchSelectorViewImpl
 
     @Override
     public void clear() {
-        button.clear();
+        menu.clear();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


csadilek:
@manstis this is the test I would add which works fine for me: https://gist.github.com/csadilek/1bd5438f6a004a30d347

If you agree I can simply amend it to this commit.

manstis:
@csadilek Yeah, that's fine. I was trying to assert the widgets in the ```DropDownMenu```.

csadilek:
@manstis I've updated the gist with a test that checks widgets are added when addBranch is called: https://gist.github.com/csadilek/1bd5438f6a004a30d347

Like you said bs3 text widget doesn't play nicely, but this works around it. In any case with these tests it should be good enough. wdyt?

manstis:
Looks good... but it'd only work for adding a single branch. Also the clear
test could assert the menu has zero items. WDYT?

manstis:
Let's merge and be done with :-)

csadilek:
@manstis menu is a mock in this test, so to test that clear works on the menu we need to write a test for the GWT BS3 class DropDownMenu. Adding another test that add multiple branches. Then I will push.

csadilek:
Thanks.

Pushed with tests amended:

master: 21c4cbc6b799818c143a2b8f61e09597d8391418

manstis:
Thanks mate, looks good.

PR #131, opened 2015-11-11
[BZ#1272363] added more tests for this bug fix

jomarko:


manstis:
@Rikkola Could you please take a look at this.

Rikkola:
+1
Just as a note. The ProjectScreenPresenter is getting bloated, it is over 1000 lines long. This change actually reduces the amount of lines. The next change on this area (who ever does the work, this is not targeted at @jomarko ) should take the build functionality from ProjectScreenPresenter and make a new class for it.

csadilek:
Thanks, @jomarko!

msater: eb8586bf5140cfae0d671249a77c34c610da51c8

PR #146, opened 2015-11-17
BZ1193484: Roles added to Projects are ignored by Business Central. Fix injection of ProjectService

manstis:
Injection of ```ProjectService``` failed at runtime. I changed to ```ProjectService<?>``` which worked OK locally on Wildfly however have elected to use the concrete interface ```KieProjectService``` to avoid any potential (unforeseen) issues on WebSphere.


ederign:
+1

PR #149, opened 2015-11-18
GUVNOR-2293: Support by-passing NewResourceView from NewResourceHandler

manstis:
https://issues.jboss.org/browse/GUVNOR-2293

The "New Project" menu item now routes direct to the "New Project Wizard". Additional changes were made to move validation of the project name to the ```POMEditorPanel``` used by the Wizard; as the previous sole point of validation (```NewResourceView```) is now by-passed.

### DIFF
--- a/kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/handlers/DefaultNewResourceHandlerTest.java
+++ b/kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/handlers/DefaultNewResourceHandlerTest.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.widgets.client.handlers;
+
+import javax.enterprise.event.Event;
+
+import com.google.gwt.core.client.Callback;
+import com.google.gwt.user.client.ui.IsWidget;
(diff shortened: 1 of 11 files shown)


manstis:
See also https://github.com/droolsjbpm/guvnor/pull/196

manstis:
@psiroky Hi Petr, it looks like the slave is out of space. Anything you can do?

mbiarnes:
+1 - the checks have passed now

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/handlers/DefaultNewResourceHandlerTest.java
```diff
@@ -0,0 +1,186 @@
...
+    private Caller<KieProjectService> projectServiceCaller;
+    private ValidationService validationService;
+    private Caller<ValidationService> validationServiceCaller;
+    private PlaceManager placeManager;
+    private Event<NotificationEvent> notificationEvent;
+    private BusyIndicatorView busyIndicatorView;
+
+    @Before
+    public void setup() {
+        packagesListBox = mock( PackageListBox.class );
```
Rikkola: If you use @RunWith(GwtMockitoTestRunner.class), you could add the @Mock annotations for the fields and leave these out. But this is just what I personally like, since it makes the setup() method shorter.

Rikkola:
+1

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/handlers/DefaultNewResourceHandlerTest.java
manstis: This is true, but I want "fresh" mocks for each test. That's why I set them up in ```@Before```.

I'd expect ```@Mock``` to re-use the same mock for all tests - but correct me if I am wrong?

Rikkola: Good point. I actually never thought about that, but @Mock actually does reset for each test.

Rikkola: Hi @Mock, you probably don't get pinged enough.

csadilek:
Thanks.

master: 898b1db687f830a381181110c66d0ee049fc7623

PR #153, opened 2015-11-24
Prevent constraint fail when there is no identity on request scope

ederign:
Prevent constraint fail when there is no identity on request scope
=================================================

Motivation
--------------
Recently, Kie Uberfire Social had received a PR creating a pluggable constraints mechanism for social timelines.

All the current implementations depend on User injection in order to apply the constraints to a social event, however, in the startup of app server social creates an in-memory cache for social events.

In that moment, there is no User injected on request scope, so if we apply this constraint to the timeline throws a ContextNotActiveException.

Solution
-----------

In order to prevent this, I've created a new class called UserCDIContextHelper. In that class, I check if there is a request scope for the user in runtime. The only way that I've found to do that is that ugly hack:

      public boolean thereIsALoggedUserInScope() {
        try {
             if ( identity == null ) {
                 return false;
             }
             checkIfThereIsAValidUserOnRequestScope();
         } catch ( ContextNotActiveException c ) {
            return false;
         }
         return true;
     }
 
    private void checkIfThereIsAValidUserOnRequestScope() {
         identity.getIdentifier();
    }

This seems like a lack of CDI 1.0 feature. If someone has a better implementation to that I'm open to change it.

Child PR
------------
https://github.com/droolsjbpm/kie-uberfire-extensions/pull/16

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-social-home-page/kie-wb-common-social-home-page-backend/src/main/java/org/kie/workbench/common/screens/social/hp/security/SocialEventOUConstraint.java
+++ b/kie-wb-common-screens/kie-wb-common-social-home-page/kie-wb-common-social-home-page-backend/src/main/java/org/kie/workbench/common/screens/social/hp/security/SocialEventOUConstraint.java
@@ -15,20 +15,19 @@
  */
 package org.kie.workbench.common.screens.social.hp.security;
 
-import java.util.ArrayList;
-import java.util.Collection;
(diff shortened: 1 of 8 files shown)


manstis:
+1 to merge (we do similar checks in ApplicationScopedProducer in kie-[drools]-wb).

csadilek:
Thanks.

master: 393bd3e24e4595d538964431858791d92410e85a

PR #154, opened 2015-11-24
Prevent constraint fail when there is no identity on request scope and Null pointer exception when a repo related event are not associated with a project

ederign:
I've put 2 commits in this Pr because they are related. If you want, ping me and I will split them.


csadilek:
+1

csadilek:
6.3.x: 6a136051f0506439bf33b3fb53d8c4e343e183ff and 77b7285fe636ddede1d62d499612fbc8cab5b2f7

PR #155, opened 2015-11-24
(6.3.x) BZ1284782: Full Text search throws exception in special case

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1284782


manstis:
I'll ask if this should be on 6.3.x before requesting it's merged.

triceo:
Not right now. I've proposed this for the first maintenance update.

manstis:
Ok it's your call I guess - but it means "full text" search is quite broken in 6.2.0 as the chances of a term matching something in Designer's "global" area is, IMO, reasonably high.

porcelli:
https://github.com/droolsjbpm/kie-wb-common/commit/d3a4940359fb936b936fe9069f21c674f2d98f35

PR #161, opened 2015-11-27
(Temporary?) fix for failing tests...

manstis:
This PR is for review.. it fixes the failing tests on 6.3.x but IDK if the fix just masks the issue :(

@ederign @porcelli Any opinions?

### DIFF
--- a/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/pom.xml
+++ b/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/pom.xml
@@ -116,11 +116,23 @@
     <dependency>
       <groupId>org.guvnor</groupId>
       <artifactId>guvnor-project-backend</artifactId>
+      <exclusions>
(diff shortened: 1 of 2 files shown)


manstis:
I don't want to fiddle with the (possible) rogue commit: https://github.com/droolsjbpm/kie-uberfire-extensions/commit/17bcc2019582cc7e0a3454539abd51088e59c1c6 as it does "special" things for WAS...

ederign:
@manstis I thought about that yesterday, but I'm a little bit afraid of masking the real issue because the error is not a Social error but a "File system is closed error" when this tests creates a social event that will be persisted.

manstis:
Closed as this masks the issue.

ederign:
@manstis  the question is: is that a FS issue or a error in test FS setup?

ederign:
@manstis one think that I've noticed yesterday is that they use:

fs.forceAsDefault(); -> For a .SimpleFileSystemProvider but in the FS is closed is a JGit one.

(investigating).

PR #165, opened 2015-12-01
BPMSPL-251: [GSS-RFE] Need settings to disable background build when opening Project Authoring

manstis:
Expose configuration setting "org.kie.build.disable-project-explorer" as a System Property.


manstis:
See https://github.com/droolsjbpm/kie-docs/pull/19 for documentation.

PR #170, opened 2015-12-01
Guvnor 2309, 2310, 2315

csadilek:
**Contains 3 commits**

*View/Layout changes only*

GUVNOR-2309: Comments section in overview uses inefficient layout
GUVNOR-2310: When commenting on an unlocked asset, the first (few) char(s) of the comment are removed
GUVNOR-2315: Improved fix so the users current asset type collapse state stays in place in case the project explorer gets refreshed (i.e. new file gets added or lock state changes)

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/business/BusinessViewWidget.java
+++ b/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/business/BusinessViewWidget.java
@@ -16,25 +16,17 @@
 package org.kie.workbench.common.screens.explorer.client.widgets.business;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


csadilek:
@manstis as discussed, can you take a look, please?

manstis:
+1 to merge (we could arguably have a test for the collapses change; but IDK if GwtMockito will play well with the View implementation)

csadilek:
@manstis that's true, but for your original fix as well ;). If you want to try and write a test for it we can hold back the fix for GUVNOR-2315. I just added it here as I shouldn't have merged the original one without trying it out first and now want to fix it ;). It's somewhat on the edge for requiring a test...but we can try.

manstis:
Happy for you to merge "as is"

csadilek:
@manstis tbh I am not sure if collapsed by default is better than expanded thinking about it. Wdyt?

manstis:
See the JIRA; I think collapsed by default is better.. especially
considering projects larger than our silly examples.

manstis:
+1 to merge.

csadilek:
4f5ee268ea68f8bb6643b5bb9917c09a627fb844 62880b27307e07dabb20c682e16fdd33d885f15a 4f5ee268ea68f8bb6643b5bb9917c09a627fb844

PR #190, opened 2016-01-06
Fixes after upgrade to IP BOM 6.0.0.Final

psiroky:
* the additional dependeny is need to make WELD tests happy.
   IP BOM 6.0.0.Final contains different version of Picketlink
   (the one used in EAP) and it probably does not depend on the
   persistence-api itself, but in fact requires the EMF at some
   injection point

Needs to be merged together with: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/123


manstis:
+1 to merge.

psiroky:
Closing as it is now needed (the Picketlink downgrade had to be reverted).

PR #197, opened 2016-01-14
GUVNOR-2290 - New search menu builder and renamed find to asset search

cristianonicolai:
Depends on https://github.com/uberfire/uberfire/pull/274


manstis:
Sweet! +1 to merge.

@cristianonicolai .... now this brings me to a related topic (and yes, here probably isn't the place to ask, but since you're elbow-deep in this code already....) "Contextual Search" is sort of broken - The "Task List" Perspective sets up "local contextual search" however the other perspectives don't reset "local contextual search" to their needs. So, in essence, once you've gone to the "Task List" all other Perspectives contextual search is broken. Really we need each perspective to setup contextual search.. unfortunately they don't.. Would you mind taking look?

There are various BZs relating to issues with "Search" (I suspect at least some are due to the above):-

* https://bugzilla.redhat.com/show_bug.cgi?id=1283873
* https://bugzilla.redhat.com/show_bug.cgi?id=1283864
* https://bugzilla.redhat.com/show_bug.cgi?id=1283880

(If not don't worry.. if you're worrying at all!!)

csadilek:
master: e177e280b8d4dee454fe6b305137be7024cdb359

PR #232, opened 2016-02-02
GUVNOR-2358 : Make generic validator more reusable.

Rikkola:
This will be used in optaplanner-wb for example.


kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorFileSystemProvider.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private String getDestinationPath( final org.uberfire.java.nio.file.Path path ) {
+        return path.toUri().toString().substring( getRootPathLength() );
+    }
+
+    public String getDestinationPath() {
+        return resourcePath.toURI().substring( getRootPathLength() );
+    }
+
+    public boolean projectExists() {
```
manstis: Is this really a defunct method? Project is passed in the Constructor (and I assume it should not be null). Perhaps a ```PortablePreCondition.notNull( "project", project)``` check in the Constructor and remove this method?

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorFileSystemProvider.java
```diff
@@ -0,0 +1,105 @@
...
+                             supportingFileFilters );
+    }
+
+    public void write() {
+
+        //Add resource to be validated first as:-
+        // - KieBuilder fails fast on some compilation issues
+        // - KieBuilder validates resources in the order they were added
+        // - We want to catch errors for the resource being validated first
+        kieFileSystem.write( getDestinationPath(),
```
manstis: ```getDestinationPath()``` will fail if ```project == null``` - hence my comment below (about ```projectExists```) is even more likely to be plausible.

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/DefaultGenericKieValidator.java
```diff
@@ -0,0 +1,60 @@
...
+                                       final KieProjectService projectService ) {
+        this.ioService = ioService;
+        this.projectService = projectService;
+    }
+
+    public List<ValidationMessage> validate( final Path resourcePath,
+                                             final InputStream resource,
+                                             final DirectoryStream.Filter<org.uberfire.java.nio.file.Path>... supportingFileFilters ) {
+
+        return new Validator( new ValidatorFileSystemProvider( resourcePath,
```
manstis: (You'll need to read my comments relating to ```ValidatorFileSystemProvider``` first).

Perhaps move the check for ```project == null``` to before here and return an empty collection if it is null (like the old code effectively did).

manstis:
Seems OK - other than the potential for NPE relating to ```project``` - perhaps a few more tests to cover that scenario may be helpful?

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorFileSystemProvider.java
Rikkola: Good question. It is there because the original code had it. 
Only reason for handling null projects is that you might have a resource file outside of a project. User hits the validation and turns out we return no errors == validate success. 
I think your way is good enough. We can't expect the validator to work out side of project.

manstis: The old way "failed fast" - https://github.com/droolsjbpm/kie-wb-common/pull/232/files#diff-eaf6905b87bef1baa96563bef1f2935fL98 You now fail a lot later - after the potential for a NPE to be throw ;)

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/DefaultGenericKieValidator.java
Rikkola: That is what this code does too. We check the project exists before running the the validation in Validator.validate().
But you are right. The goal here is make reusable bits, so ValidatorFileSystemProvider should not be usable in a way that can cause a NPE.
And your comments made me think we should just fail fast if the asset is not in a project.

Rikkola:
I'll add the test for when project is null.

Rikkola:
After a chat with @manstis in IRC. We decided to keep the old way. If there is no usable project we return no errors.

Rikkola:
Added a test and better handling for project==null.

manstis:
@Rikkola hehe, we could discuss our own preferences for where ``NoProjectException``` should be thrown (or used entirely) forever.. but I'm not going to press my personal preference on what you've done - it works and there are tests... 

+1 to merge.

Rikkola:
Well personally I would fail if the project is null in the constructor and report back to the UI. Kind of like what you suggested at one point. 
But this changes the behavior from what it used to be. The validation will most likely change in the future so we can talk about this again. 
Anyway thanks for the feedback. Nice to have comments about the implementation.

csadilek:
master: 30f51a70b35ca0fda78d0ed4fedd05ead808af47

Rikkola:
thanks

PR #242, opened 2016-02-09
GUVNOR-2376 : KSession selector should fire value changed when kbase is selected and when the default ksession is selected

Rikkola:


manstis:
@Rikkola Just a thought, that it might have been better to fire the event here: https://github.com/Rikkola/kie-wb-common/blob/GUVNOR-2376-ksession-selector/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/widget/KSessionSelectorViewImpl.java#L101 instead of where you do now.. that way if ```KSessionSelectorViewImpl.setSelectedValue(..)``` didn't *match* anything the event would not get fired either? 

At the moment if ```KSessionSelectorViewImpl.setSelectedValue(..)``` doesn't actually set the value on the ```ListBox``` (IDK, e.g. a duff value was passed) the event is still fired.. IDK if this matters.

Rikkola:
Heh, I thought about putting this into the Impl, but being lazy didn't want to do the test for it.

But your point about the value not being set does mean it is better to move the firing of the event to Impl and test it. I could also test the presenter and check the value is set before calling view.fireValueChanged(), but that sounds less practical.

Rikkola:
@manstis Test for the Implementation now. Wasn't that hard. GwtMockito handled this quite well, usually I have had to make huge setups.

manstis:
@Rikkola There's still potential for the event to be fired even if nothing was selected. I'd have gone for something like this: https://gist.github.com/manstis/847452da0902350cf7d6

But, since we control population of the dropdowns it's probably a low risk issue (and IDK if raising unnecessary events is even likely to cause issues elsewhere.. I doubt it TBH.

manstis:
+1 to merge (avoiding further procrastination)

Rikkola:
My first try was something like that. The reason for the current way is that even if the value does not change. We fire the event and this way whatever is visible in the dropdowns gets populated into the model (test for the presenter hopefully protect against this). The current way keeps if-conditions out of the view and lets the presenter just say "make it so".

manstis:
+1 to merge.

csadilek:
master: a05ab5719bea2a13fb8ef2bea75362354a163bd3

PR #259, opened 2016-02-15
GUVNOR-2348: Dropdown button in dependencies list instead of double clicking

paulovmr:


paulovmr:
@manstis Could you review this PR? I don't know if you haven't looked at it yet, but seems that it was forgotten.

manstis:
@Rikkola since you wrote the original code would you mind looking at @paulovmr's changes?

Rikkola:
+1 looks nice and clean

PR #260, opened 2016-02-16
[6.4.x] cherry-picked changes needed by Optaplanner workbench

Rikkola:


manstis:
Any (good) reason why you didn't squash the 3 commits for GUVNOR-2358?

Rikkola:
They were 3 different commits in master. So I just cherry-picked each commit one by one. 
Because the needs were not clear during development I had to make two more commits after the first 2358. I can squash them for 6.4.x if needed.

manstis:
@Rikkola ok, thanks for the explanation (that if I did not ask someone else would!)

+1 to merge.

csadilek:
@Rikkola @manstis I missed this when we merged it on master so it's fine now but in my opinion we should use unique commit messages in the future. When browsing history it's quite confusing to see the exact same commit message multiple times. I would always add a more details explanation on subsequent lines explaining what this commit adds to the previous one with the same title.

csadilek:
6.4.x: 

078d8ac6c9254945744e662b19243af842cda4ba
859515a112e6765d26ce71f4deca09a2dab6ff0c
af9dad91ea912556ea1b7e1564347e6bc1a81657
63b408d47065879d3ae925a90ac799745ff82c20
ec9600e08e09f7a5d3dd888e829d4f7943a5e389

PR #263, opened 2016-02-17
BPMSPL-252: [GSS-RFE] Ability to check duplicate of ReleaseId (GroupId, ArtifactId and Version) with existing projects. Updates following QE feedback.

manstis:
See https://issues.jboss.org/browse/BPMSPL-252

This PR includes updates requested by QE following their testing. It also moves a couple of WebBeans (WAS) unfriendly ```<? extends Project>``` from ```guvnor``` to ```<KieProject>``` in ```kie-wb-common```.

Branch name retained (yes, I know @psiroky it's not ideal) to be consistent and support reconciliation of all changes made for the feature.


manstis:
Part of an ensemble:-

https://github.com/droolsjbpm/guvnor/pull/252
https://github.com/droolsjbpm/kie-wb-common/pull/263

kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/project/ProjectServiceImplResolvePackagesTest.java
```diff
@@ -43,6 +44,17 @@
     private BeanManager beanManager;
     private Paths paths;
 
+    @BeforeClass
+    public static void setupSystemProperties() {
+        //These are not needed for the tests
```
triceo: I wonder if we shouldn't externalize these somehow, so that they can be shared across tests? We only have few and already there is duplication.

psiroky: It might not be a bad idea to disable these directly in the POM (in surefire config), and let specific tests enable them if needed (only very small number of tests will actually need these daemons).

manstis: @triceo @psiroky Yes, yes.. but ```guvnor``` and ```kie-wb-common``` are not alone here.. I'd propose keeping this "as is" here (otherwise BPMSPL-252 will never get finished) and raise a new BZ/JIRA to disable daemons across the whole uberfire/droolsjbpm/kie test world that we tackle as a separate concern...

psiroky: Agreed @manstis. I was just pointing out one of the possible solutions. 

I've created https://issues.jboss.org/browse/GUVNOR-2410 to track that.

Rikkola:
+1

PR #266, opened 2016-02-18
GUVNOR-2426 : KSessionSelector default ksession value can be defaultKieSession or null

Rikkola:


manstis:
I've looked at this PR at least twice and I still don't understand it :-( Could you please explain the issue it addresses to me?

PR #270, opened 2016-02-19
Fix dependency management

cristianonicolai:
Delay calling service to reload dependencies until changes are saved.


cristianonicolai:
@Rikkola @manstis same fix as we sent to 6.4.x branch, please have a look.

manstis:
+1 to merge.

PR #271, opened 2016-02-23
BZ1310893: Full Text search returns only 10 results even if it hits more than 10

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1310893

The issue was that we did not consider "start index" or "page size" when building the response.


Rikkola:
+1

PR #272, opened 2016-02-23
[6.4.x] BZ1310893: Full Text search returns only 10 results even if it hits more than 10

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1310893

(cherry picked from commit de9381b47cd28f68ab1ae93dc9ad1aa2dcd17081)

This is the corollary of https://github.com/droolsjbpm/kie-wb-common/pull/271


Rikkola:
+1

PR #273, opened 2016-02-23
[6.3.x] BZ1303534: [GSS](6.2.z): Full Text search returns only 10 results even if it hits more than 10

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1303534

(cherry picked from commit de9381b47cd28f68ab1ae93dc9ad1aa2dcd17081)

This is the corollary of https://github.com/droolsjbpm/kie-wb-common/pull/271 and https://github.com/droolsjbpm/kie-wb-common/pull/272 for 6.3.x.


Rikkola:
+1

PR #274, opened 2016-02-23
GUVNOR-2414: When deleting a file from a project, an error message is shown even though the file is succesfully deleted.

paulovmr:


paulovmr:
This PR also resolves GUVNOR-2419: After deleting from one project and switching to another, it is not possible to return to the first one.

Depends on: https://github.com/uberfire/uberfire-extensions/pull/218

manstis:
@paulovmr Great stuff!

+1 to merge.

PR #276, opened 2016-02-24
Add test coverage for BZs 983540, 983528

zkrejcov:


manstis:
@Rikkola One for you to review (please)!

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/KSessionsPanel.java
```diff
@@ -50,6 +50,11 @@ public KSessionsPanel(
         view.setPresenter(this);
     }
 
+    // for testing purposes
+    void setLockRequiredEvent(javax.enterprise.event.Event<LockRequiredEvent> lockRequired) {
+        this.lockRequired = lockRequired;
+    }
+
```
Rikkola: Since the lockRequired is injected into a field. You can pass it in the constructor instead of using this method.
Just to keep methods only used for testing out of the code and this way we inject everything the same way. The original author did the mistake of just adding a field injection, it should already be there.

Rikkola:
Good work.

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/KSessionsPanel.java
zkrejcov: I considered that, but then rather decided to keep the existing code as-is, to avoid any possible disruptions - my understanding of the code isn't optimal.
I can change it of course.

Rikkola:
+1
Thank you for the tests.

psiroky:
Merged into 6.4.x.

PR #280, opened 2016-02-29
GUVNOR-2419: After deleting from one project and switching to another, it is not possible to return to the first one.

paulovmr:
@manstis The change that we discussed here: https://github.com/droolsjbpm/kie-wb-common/pull/275/files#r53944301

I've tested it on Windows, and it works.


manstis:
+1 to merge.

manstis:
Thanks!

csadilek:
6.4.x: ded8d1bba61dd53794b535b9cab2f750516ed6cc

PR #291, opened 2016-03-07
GUVNOR-2452: use of a different Div widget in order to avoid ambiguity

porcelli:


csadilek:
+1

csadilek:
master: 7e35de6b964eadc93922c929d8d424237f98dac8

manstis:
@porcelli, @csadilek I image this was caused by more than one CDI Bean matching an injection point for ```DIV``` (e.g. we have custom controls extending ```DIV``` too). This type of issue is, IMO, only going to become more common as we move away from ```Widget``` to HTML based controls. We already have a number of cases where we've had to introduce qualifiers on some screens to ensure correct injection point match. Is there anything we can do, as a design pattern, to help avoid these situations? IDK perhaps encourage each module to define a qualifier and ensure *all* injection points in a module use the qualifier (this would mean we'd need to write producers too). Then my next comment would be is there anything we can do to avoid the boiler plate - all scenarios so far (of which I am aware) have been on the ```View``` and increasingly ```@Templated``` Views. Is it plausible to either have Errai automagically inject qualifiers on ```@Templated``` Views and generate a Producer? Or should we consider a step-aside from pure CDI annotations and have a ```@InjectLocal``` annotation that only resolves to beans in the same module? Or should we just communicate a best practice (qualifier and producer factory).. WDYT?

porcelli:
@manstis for sure @csadilek is the best to answer this, but my take on this is that with the `correct` use of Errai UI and proper Errai wrapers for elements we won't be traped by this issue.

csadilek:
@manstis @porcelli This will NOT be an issue going forward when we're moving to our new DOM based wrappers. Those can't be subclassed and won't cause ambiguity. The issue here is that DIV (gwtbootstrap3) is in fact a widget type and has numerous subtypes. So, this is a temporary solution as we didn't want to switch to the new DOM wrappers and retest everything last night.

csadilek:
@manstis on the second point, do you have an example of cases where we had to use qualifiers and would have rather not to. We can definitely step outside CDI and add custom features if they make sense i.e. ```@InjectLocal``` (although we don't actually know what module a type was discovered in, but there might be a way to make this work).

manstis:
@csadilek [This](https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtree-editor/drools-wb-guided-dtree-editor-client/src/main/java/org/drools/workbench/screens/guided/dtree/client/widget/palette/GuidedDecisionTree.java) is first example that came to mind and the use [here](https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtree-editor/drools-wb-guided-dtree-editor-client/src/main/java/org/drools/workbench/screens/guided/dtree/client/widget/palette/GuidedDecisionTreePalette.java#L77).. if it's not envisaged to be a problem ongoing as we move away from ```Widgets``` then let's move along.. nothing to see here!

PR #292, opened 2016-03-08
Adds default constructor.

Rikkola:
Without this CDI fails to make an instance of the class and when running the UI the users sees a "loading" popup that never goes away.


manstis:
+1 to merge.

Rikkola:
Thank you.
@porcelli do we need this for 6.4.x?

porcelli:
@Rikkola thx for the catch.. in theory no (well I executed kie-wb soooo many times local and never faced such issue). It looks like an Errai 4 behavior issue.

Rikkola:
Ok. That explains how it got in like this in the first place :)

csadilek:
In CDI, all normal scoped beans need to be proxyable. Errai already steps outside CDI to not strictly require a default no-arg constructor. If another constructor allows passing in null parameters it will use that instead. This *sometimes* worked (by accident) in Errai 3 and caused numerous issues in cycle detection and resolution.

So, you have the following options:

- Make the bean dependent scoped (usually the better choice for UI anyway)
- Make the constructor save to pass null arguments
- Add a default no-arg constructor

@Rikkola @porcelli @manstis

PR #294, opened 2016-03-08
GUVNOR-2452 - test enhancement to not fail in case of order of elemen…

mswiderski:
…ts is different

@porcelli could you take a look


porcelli:
+1

csadilek:
master: 51d3b17828225d5c6944974178a262b537f459da

PR #295, opened 2016-03-08
[6.4.x] GUVNOR-2452 - test enhancement to not fail in case of order of elemen…

mswiderski:
…ts is different

(cherry picked from commit 3ded67433c929b28307d3d66d6764dfb96666a5e)


csadilek:
6.4.x: dcb1d719affdb3d6ee4bb496b6e4a98dc41d809c

PR #301, opened 2016-03-10
[master] In exec.server UI, switch from KeyUpHandler to ChangeHandler.

zkrejcov:
See also GUVNOR-2386 - keyUp tends to mess up focus. This is
a simple copy of the fix mention in GUVNOR-2386.

(cherry picked from commit 2b4f1acad53964a2937f3ee11c0cbce5e32affa2)


porcelli:
+1

porcelli:
https://github.com/droolsjbpm/kie-wb-common/commit/7ffae35f29a06fc14e87f240a95b840f68a2ffd0

manstis:
A word of caution... ```KeyUpHandler``` does not itself mess with anything... The uses of ```KeyUpHandler``` in this PR remain "client side" meaning whenever a key is pressed the effect stays in the browser (e.g. it clears some styling etc).. The example cited for https://issues.jboss.org/browse/GUVNOR-2386 is different.. the ```KeyUpHandler``` leads to an asynchronous call the the server to validate the new field value and if invalid the ```Widget``` effectively re-loaded. This lead to the manifestation of some user input being "lost"... so whilst use of ```ValueChangeHandler``` *can* be an improvement it doesn't necessarily always need replace use of ```KeyUpHandler``` :)

zkrejcov:
Good to know, thanks manstis! 
The fact is, the issue in GUVNOR-2386 was repeated here in (at least) one of the cases. It seemed better to simply default to the ChangeHandler now in all the cases to avoid possible repetion at a later date...

PR #308, opened 2016-03-10
GUVNOR-2457: remove docs reference from empty pages (it may return in

porcelli:
the future when we have a better idea how it would work). Also adapts
achors to repospond to "add" actions.


csadilek:
6.4.x: 512e0ce1964794f83efa23d652ded014bd4ad1ca

PR #309, opened 2016-03-10
GUVNOR-2457: remove docs reference from empty pages (it may return in

porcelli:
the future when we have a better idea how it would work). Also adapts
achors to repospond to "add" actions.


csadilek:
master: 7bc4499aaede8692b56a9071eec3e13fab999d35

PR #321, opened 2016-03-19
GUVNOR-2352 : Adding of manual dependency does not work correctly

Rikkola:
The issue was that the model presentented in the dependency grid was not the same as the one in the POM model class. This was because we list both the dependencies from the pom and the transient dependencies of these dependencies in the grid. Because of this updates or adding new dependencies was not working correctly. The fix for this is I made `EnchancedDependencies` to be shown on the grid, these are both "normal" from the pom's deps and transient deps of these deps. Then `EnhancedDependenciesManager` is responsible for keeping the grid and the pom model in sync.

`EnchancedDependencies` also includes the packageNames the dependency contains. These are used by the editor's white list features. 

Since the `EnhanceDependencyManager` is now responsible for synching. I added a `NewDependencyPopup` for the grid. Before we would hit "new" and this added an empty line in to the grid. Now to prevent the user from adding 10 empty lines (or partially filled in lines) and hitting save, the user has to first will in the popup form with valid GAV.  

https://github.com/droolsjbpm/guvnor/pull/292

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/DependencyLoader.java
+++ b/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/DependencyLoader.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.screens.projecteditor.client.forms.dependencies;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+
+import org.guvnor.common.services.project.model.Dependency;
+import org.jboss.errai.common.client.api.Caller;
+import org.jboss.errai.common.client.api.ErrorCallback;
+import org.jboss.errai.common.client.api.RemoteCallback;
+import org.kie.workbench.common.services.shared.dependencies.DependencyService;
+import org.kie.workbench.common.services.shared.dependencies.EnhancedDependencies;
+import org.kie.workbench.common.services.shared.dependencies.NormalEnhancedDependency;
+
+@Dependent
+public class DependencyLoader {
+
+    private final List<Dependency> updateQueue = new ArrayList<>();
+
+    private final Caller<DependencyService>   dependencyService;
+    private       EnhancedDependenciesManager manager;
+
+    @Inject
+    public DependencyLoader( final Caller<DependencyService> dependencyService ) {
+        this.dependencyService = dependencyService;
+    }
+
+    public void init( final EnhancedDependenciesManager manager ) {
+        this.manager = manager;
+        this.updateQueue.clear();
+    }
+
+    public void load() {
+        if ( !updateQueue.isEmpty() ) {
+            loadFromServer();
+        } else {
+            returnDefault();
+        }
+    }
+
+    private void returnDefault() {
+        EnhancedDependencies enhancedDependencies = new EnhancedDependencies();
+        for ( Dependency dependency : updateQueue ) {
+            enhancedDependencies.add( new NormalEnhancedDependency( dependency,
+                                                                    new HashSet<String>() ) );
+        }
+
+        updateQueue.clear();
+
+        manager.onEnhancedDependenciesUpdated( enhancedDependencies );
+    }
+
+    private void loadFromServer() {
+
+        dependencyService.call( new RemoteCallback<EnhancedDependencies>() {
+                                    @Override
+                                    public void callback( final EnhancedDependencies result ) {
+                                        onLoadSuccess( result );
+                                    }
+                                },
+                                new ErrorCallback<Object>() {
+                                    @Override
+                                    public boolean error( final Object o,
+                                                          final Throwable throwable ) {
+
+
+                                        returnDefault();
+                                        return false;
+                                    }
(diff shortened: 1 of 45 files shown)


manstis:
@Rikkola, I'll look in detail on Monday but I wonder whether you meant to have "enchanted" classes as well as "enhanced" (see your onEnchanted... methods)?

Rikkola:
Heh, I don't.

manstis:
@Rikkola OK, please let me know when you want me to review this...

Rikkola:
@manstis ok now. Disenchanted the code.

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/DependencyLoader.java
```diff
@@ -0,0 +1,103 @@
...
+        this.dependencyService = dependencyService;
+    }
+
+    public void init( final EnhancedDependenciesManager manager ) {
+        this.manager = manager;
+        this.updateQueue.clear();
+    }
+
+    public void load() {
+        if ( !updateQueue.isEmpty() ) {
```
manstis: Should this be "not empty"? It is empty when ```DependencyLoader``` is initialised, and ```returnDefault()``` (which would be called in response to the first ```load()```) clears it too. I'd have thought after initialisation the first operation would be to ```loadFromServer()```?

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/NewDependencyPopupViewImpl.java
```diff
@@ -0,0 +1,164 @@
...
+import org.gwtbootstrap3.client.ui.constants.ValidationState;
+import org.kie.workbench.common.screens.projecteditor.client.resources.ProjectEditorResources;
+import org.uberfire.ext.widgets.common.client.common.popups.BaseModal;
+import org.uberfire.ext.widgets.common.client.common.popups.footers.ModalFooterOKCancelButtons;
+
+public class NewDependencyPopupViewImpl
+        extends BaseModal
+        implements NewDependencyPopupView {
+
+    private static Binder uiBinder = GWT.create( Binder.class );
```
manstis: I believe, although I haven't seen it communicated to BSIG, that "new Views" should be using ```errai-ui``` however since this is a component of a larger UI it's probably best to keep its implementation consistent... so this comment is FYI (re: future Views for completely new items).

manstis: @csadilek Was you planning on communicating the proposal to use ```errau-ui``` for new Views (in 7.x onwards)?

manstis:
@Rikkola A couple of comments.. nothing serious.

Can you please explain ```DependencyLoader``` a little? Are you returning a basic list of dependencies to the Project Editor "on startup" first and then enriching with more data asynchronously (as retrieval is slow?) I see your enqueue changes etc but would just like to understand it a bit more..

Rikkola:
At the moment it needs to load and then show. So no "quick" show. If we load the dependencies in on the background then we need to figure out how to slide in the transient dependencies, since they are shown in the same list. 
One option would be to open an accordion for transient deps. If they are not loaded we show a loader animation.
Another it that we make the grid manage this. Currently it updates everything in one bump. This will be tricky so voting for accordion.

manstis:
@Rikkola I'm not asking for lazy enrichment.. I just don't understand what ```DependencyLoader``` does! can you explain (and why it holds a queue?)

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/DependencyLoader.java
Rikkola: `DependencyLoader` only loads what is in the queue. The `EnhancedDependenciesManager` fills the queue for the first load.
Initilialisation only sets everything up. This happens when the project editor screen is opened. `load()` does the actual work when the tab is first opened or when content changes.

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/NewDependencyPopupViewImpl.java
Rikkola: I remember hearing that we are moving, but wasn't sure when. There used to be some difficulties running GWT UiBinder and Errai UI side by side, but I guess they have been fixed and I can use Errai UI from now.
But you are right, better to start ASAP if that is the future.

Rikkola:
@manstis alrighty. Keep on asking, I'll answer ;-)
 
I'm using a queue mainly to unify the load for a single or multiple dependencies. At first, you load all the dependencies that are in the pom.xml. Then when the user adds dependencies you add them in one by one. So instead of having the init load and then incremental load I just made a single load.

Also planned the future a bit. If we add facets or plugins, then those will most likely add more than one dependency in and this makes it possible to feed the dependencies in and also it makes possible to add the lazy load.

Another thing that is isolated in the `DependencyLoader` is failure while loading. The `EnhancedDependenciesManager` does not need to care about failure since the loader manages it. And again in the future this will need to load information about the repositories defined in the pom.xml and any repositories defined upstream. The same that you did for the checks if a GAV is already used. Since currently we only check the local repository. Didn't add the checks in now since it will increase the, already slowish, load time.

manstis:
@Rikkola thanks for the explanation.

+1 to merge.

csadilek:
master: 87cb2612655f225dea7a6b551afaa02f262d0bc1

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/forms/dependencies/NewDependencyPopupViewImpl.java
csadilek: Yes, going forward (7.x onwards) we want all new UI features developed in Errai UI.

PR #323, opened 2016-03-21
BZ-1318987 - Execution server controller doesn't notify Kie server wh…

mswiderski:
…en creating/deleting container


mswiderski:
@porcelli @sutaakar could you please take a look?

mswiderski:
anyone else is willing to take a look?

csadilek:
Changes look good to me. The only comment I have is that if create should start and delete should stop the container using the ```kieServerInstanceManager``` then that would be worth verifying in the tests as it seems specific to this subtype of ```KieServerControllerAdminImpl```. New to this code and happy to be corrected though.

+1

mswiderski:
create is always in stopped state so it's not starting and it's covered with this test https://github.com/droolsjbpm/kie-wb-common/pull/323/files#diff-d0923c03d778166502c09757a915ebb1R149

csadilek:
Ah I see, you're calling startContainer in the test which calls notifyKieServersOnCreateContainer on this subtype. So, scratch my comment. Happy to merge if there are no objections.

sutaakar:
looks good to me, +1

csadilek:
master: 49aa207ecc8854e4c0c228282ff9b87e935bcb7e

PR #325, opened 2016-03-22
BZ-1313782:  Error on startup: Unable to bootstrap project org.kie.example:project1:1.0.0-SNAPSHOT

pefernan:


Rikkola:
We usually force the user to fill in the project name. In some rare cases when it does not exists we use artifactId. Because of this it makes sense to use it as project name.
+1

csadilek:
6.4.x: 32c9dc0bb249bef433da1ae5e6cdf1bb4f87d73a

PR #330, opened 2016-03-23
GUVNOR-2473: test class moved to right place

jomarko:


manstis:
+1 to merge (assuming Jenkins is green... and I see no reason for it not to be)

csadilek:
master: e4505e951207c980f5bf99bb712bb03e6eedde95

PR #331, opened 2016-03-23
GUVNOR-2473: test class moved to right place

jomarko:


manstis:
+1 to merge (assuming Jenkins is green... and I see no reason for it not to be)

csadilek:
6.4.x: 3c17566b92cac691aee42f621f66b45090fa23b3

PR #338, opened 2016-04-06
[6.3.x] BZ-1322241 - [GSS](6.2.z) ServerImpl cannot be cast to KieServerInsta…

mswiderski:
…nce after migration from 6.1 to 6.2

@porcelli could you take a look?


porcelli:
+1

PR #343, opened 2016-04-17
Refactor Weld-based tests + fixes after Weld 1.x -> Weld 2.x upgrade

psiroky:
* guvnor-test-utils now contains simple junit runner
   (WeldJUnitRunner) which takes care of starting and
   stopping the Weld behind the scenes. Using the runner
   results in much less duplication in setup code

@manstis, @Rikkola could you please take a look?


kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/builder/ResourceChangeIncrementalBuilderTest.java
```diff
@@ -157,7 +117,7 @@ public void testResourceAdded() throws Exception {
 
         //Perform incremental build
         buildChangeListener.addResource( resourcePath );
-
+        waitForIncrementalBuildResults(buildResultsObserver);
```
psiroky: @manstis, @Rikkola, I am not exactly sure why this was not needed before. If I understand correctly the incremental build happens async on background, so the test actually needs to wait for the results to become available? (removing the wait method results in null returned by `buildResultsObserver.getIncrementalBuildResults()`. I must be overlooking something here...

manstis: @psiroky.. are you sure the old tests actually ran asynchronously? I'll look more tomorrow.. but that's my suspicion.

psiroky: I am definitely _not_ 100 % sure, it is just how I understood it from the code. I simply upgraded from Weld 1.x -> Weld 2.x, so I would have hoped that should not cause the difference.

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/PackageDataModelServiceTest.java
```diff
@@ -39,28 +42,17 @@
...
- * Tests for DataModelService
- */
+@RunWith(WeldJUnitRunner.class)
 public class PackageDataModelServiceTest {
 
     private final SimpleFileSystemProvider fs = new SimpleFileSystemProvider();
+    @Inject
     private BeanManager beanManager;
+    @Inject
     private Paths paths;
```
Rikkola: Really liking how it is just possible to inject these.

Rikkola:
WeldJUnitRunner is a good idea. Another area that is hard to test at the moment are the client side presenters. Some sort of Runner might also help there. Not to mention the base presenter our editors use could also use some refactoring to be more testable. The current setup is quite long.

psiroky:
Yeah, I find this idea of different custom runners very appealing. It is then very easy to capture shared setup/configuration code without the need of complex test class hierarchy (e.g. multiple abstract classes with shared code).

kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/builder/ResourceChangeIncrementalBuilderTest.java
manstis: @psiroky I suspect the test class ```org.kie.workbench.common.services.backend.TestIncrementalBuilderExecutorManagerFactoryImpl``` defined in ```beans.xml``` is no longer being used as the ```@Alternative```? So the test is probably *better* now - if it's using an Async Executor vs mocked synchronous execution.

kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/builder/ResourceChangeIncrementalBuilderTest.java
```diff
@@ -47,68 +46,36 @@
...
     private BuildResultsObserver buildResultsObserver;
+
+    @Inject
     private BuildService buildService;
+
+    @Inject
     private KieProjectService projectService;
 
+    @Inject
+    private org.guvnor.common.services.builder.ResourceChangeIncrementalBuilder buildChangeListener;
```
manstis: @psiroky This is probably injecting the *real* ```ResourceChangeIncrementalBuilder``` rather than the mocked "test" one that ran synchronously.

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/PackageDataModelServiceTest.java
manstis: +1 will make writing tests (somewhere in between Unit and Integration) much easier - i.e. the tests aren't mocked to death but also aren't full integration tests.. :)

manstis:
+1 to merge (when the WELD changes go en masse)

kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/builder/ResourceChangeIncrementalBuilderTest.java
psiroky: Thanks @manstis, nice catch! I believe the reason for that is different behavior of Weld 1.x vs. Weld 2.x when it comes to alternatives. Weld 2.x won't inject alternative into different bean module (e.g. jar) if it does not have the @Priority annotation (yeah, took me a while to figure that one out -- with help of @winklerm  and @rsynek).

Anyway, I would be more than fine with fixing this. Using the async executor brings us closer to the "real" app, but at the same time is also another place where things can go wrong for these tests.

What do you think would be better for this case?

manstis: @psiroky I'm in favour of being closer to the *real* app.. so let's keep the ```waitForIncrementalBuildResults(..)```.. could you please tidy up (i.e. delete!) the old test alternative?

psiroky: OK, will remove the class and all references to it.

manstis:
@psiroky Still failing....

psiroky:
Yeah, guvnor needed bump as well (and least I hope that was the case).

PR #349, opened 2016-04-29
BZ-1328663: selection list of fact fields are ordered randomly (not dictionary order) in GDT and test scenario editor

paulovmr:


manstis:
+1 to merge.

Thanks Paulo, good job!

csadilek:
@paulovmr looks like the client and server-side SortHelpers are identical, or am I missing something? 

If they are, it's better to share that code to avoid duplication. That's the biggest benefit we have from using gwt ;). @manstis

manstis:
@csadilek, you know I had a quick look earlier and noticed the exact same thing... when returning later to comment on the PR I completely forgot... you're absolutely right (of course)

manstis:
@paulomvr, somewhere here would be a good place for it: https://github.com/droolsjbpm/kie-wb-common/tree/master/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-api/src/main/java/org/kie/workbench/common/services/datamodel/model (or do all sorting client side?)

paulovmr:
@manstis @csadilek Thanks for the comments! I will evaluate this (use only one class or do all sorting client side) first thing monday.

paulovmr:
@csadilek @manstis Hi, I moved the SortHelper (and its tests) to https://github.com/droolsjbpm/kie-wb-common/tree/master/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-api/src/main/java/org/kie/workbench/common/services/datamodel and now it is shared between backend and client.

@manstis I think it makes more sense to maintain that sorting call in the backend, because it is more centralized and, otherwise, I would have to call it in more places. But, if you see any other reason to change its location, I can consider it. :)

manstis:
+1 to merge.

PR #354, opened 2016-05-04
[6.3.x] BZ-1324141 - ServerImpl cannot be cast to KieServerInstance after mig…

mswiderski:
…ration from 6.1 to 6.2 - fixed identifier handling, return migrated path to load migrated kie server content

@sutaakar could you please take a look?

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/VFSKieServerControllerStorage.java
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/VFSKieServerControllerStorage.java
@@ -16,6 +16,7 @@
 package org.kie.workbench.common.screens.server.management.backend;
 
 import java.math.BigInteger;
+import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/VFSKieServerControllerStorage.java
```diff
@@ -193,14 +194,19 @@ private boolean isHex( final String endpoint ) {
...
-    protected void migrate(Path path) {
+    protected Path  migrate(Path path) {
         if (!path.toString().endsWith("-info.xml")) {
             try {
                 final ServerRef serverRef = (ServerRef) xs.fromXML(ioService.readAllString(path));
 
+                String serverId = serverRef.getName();
+                if (serverId == null || serverId.isEmpty()) {
+                    URL url = new URL(serverRef.getUrl());
+                    serverId = "Unnamed Kie Server at " + url.getHost();
```
sutaakar: What about putting just hostname as server id?
Server id needs to be configured also as system property on Kie server side and it is sent as part of URL to workbench when Kie server starts. Whitespaces in name could make problems in URL.

mswiderski: we can't protect from whitespace as name can have it as well. Leaving only host name could be confusing as it might not be informative enough what it is.

mswiderski: fixes applied, thanks @sutaakar

mswiderski:
this is for one off patch of 6.2.z so if someone could take a look that would be great

sutaakar:
+1
Tried it locally and it seems to work.
There just have to be specific migration order followed:
- After migration there has to be started Workbench first while Kie server are turned off and opened Server management screen. 
If Kie server is started before internal migration is done then this server is registered in Workbench, causing errors in migration which is run afterwards.
Also Kie server and Workbench has to have proper configuration of system properties - according to docs.

porcelli:
+1

PR #355, opened 2016-05-10
Authoring->Project Authoring->New Item->Data Object: Only accepting latin characters as input

wmedvede:


wmedvede:
@manstis 
Hello Mike, this is the PRE-PR that I've added more info via email. Just for verification if you agree with the changes mainly at the Builder. Would you mind take a look?

NOT FOR MERGING

manstis:
Hi Walter, I guess our use of ```Path.toUri()``` introduces the issue; as ```URI``` are ```ASCII-7``` encoded. I wonder whether replacing ```projectPrefix``` and it's use with ```Path.relativize(projectRoot)``` would give the same results w/o the String munging and URI decoding? Just a thought to perhaps give a quick look at?

wmedvede:
Hi @manstis ,  wdyt about something like this:

private String buildKieFSDestinationPath( final Path resource ) {
 +        final Path relativePath = projectRoot.relativize( resource );
 +        return relativePath.toString();
 +    }

( relies on the toString() that do the job of getting the decoded string value..., but this is something we are aleady doing around the platform when we do path.getFileName().toString() )

manstis:
@wmedvede Hi.. yes, IMO, that's much better (more robust.. we eliminate some ```String``` munging that caused the encoding/decoding issues).. if you can write a few tests we can get this merged. Thanks.

wmedvede:
@manstis 

Ok, let's see what the QE/Edson/Alessandro Lazarotti says about this in 6.4.x. For master I think we can add it and ask the different "editor / screen owners" to check.

I'll close this PR by now since it's a PRE-PR

PR #357, opened 2016-05-13
Data modeller filter extension

mbarkley:
This PR introduces a mechanism for filtering out Java classes in a project from being loaded by the DataModelerService. This feature is necessary for LiveSpark, where we generate Java classes with GWT dependencies that cannot be class-loaded (and these classes are not data models so there is no reason we would want them to be loaded as such).

There are also two commits with some minor house-keeping (removing unused private fields/methods, and fixing some compiler warnings).

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
@@ -18,13 +18,18 @@ package org.kie.workbench.common.screens.datamodeller.backend.server;
 
 import java.io.ByteArrayInputStream;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


mbarkley:
Hi @wmedvede , could you please review this for me?

wmedvede:
Hi Max,
I think it should work well. The only thing that you must keep in mind is that when a SourceFilter is applied, Non-filtered classes should not point to Filtered classes. 

i.e. if we have that A has a reference to B, and B is filtered, then A should also be filtered. For example if you create this with the data modeler:
public class A extends B {

}
And we filter B excluding it from the result, and A is in the result I guess this can lead to troubles. And same thing apply for the KieBuilder filtering. I'm sure you already considered that, but just to be aware. 

For livespark if I don't remember wrong, the filter should filter things like XXXFormModel.java, or XXXRestService.java, but other filters that may require inspect all the object graph may be very complex to do.

@mbarkley  @csadilek 

Other than this, for me it's ok.
+1

mbarkley:
@wmedvede That's a good point you've brought up. As it currently exists someone could make a filter that breaks the data modeller, but as you pointed out preventing this would be lots of work so I think it is an acceptable flaw. Especially because this feature should only be used to prevent the DataModelerService from loading types that aren't data models. Since data models should only ever reference other data models (or built-in types) we shouldn't run into dependency-related issues in practice.

csadilek:
+1, thanks guys.

PR #358, opened 2016-05-16
Add tests for extension checking

zkrejcov:
... and also
- fix a small issue with extension checking
- remove a bit of code duplication
- refactor code so that this part is testable


manstis:
+1 to merge.

PR #362, opened 2016-05-23
JBPM-5115: Context of Project Editor is often not updated after importing project using ExamplesWizard

manstis:
See https://issues.jboss.org/browse/JBPM-5115

I changed the import of Examples to include the first imported Project in the ```ProjectContextChangeEvent``` event fired upon completion. I tested a few times with ```kie-drools-wb``` with clean installs and the newly imported Project was correctly selected in the Project Explorer.


Rikkola:
+1

PR #364, opened 2016-05-24
Migrate QE test for Source tab read-only mode

yurloc:
This test replaces QE Selenium test for [BZ 1169777 - Source tab inconsistencies - editing allowed](https://bugzilla.redhat.com/show_bug.cgi?id=1169777). The expected result of this BZ is that "analyst" role(s) are not allowed to edit Data Object's Java source directly. Compared to existing Selenium test, this unit test doesn't provide 1:1 coverage because it doesn't work with concrete roles. So any changes in [policy file](https://github.com/droolsjbpm/kie-wb-distributions/blob/7e3c0f3/kie-drools-wb/kie-drools-wb-webapp/src/main/resources/workbench-policy.properties#L63) won't be detected by this test while Selenium test would fail with an analyst user if the restriction was removed from the policy file.

I believe this test is useful because it adds some missing (not 100% sure) coverage, I'm just not sure if we can remove the original Selenium test. @jhrcek wdyt?

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/test/java/org/kie/workbench/common/screens/datamodeller/client/SourceEditorReadOnlyTest.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/test/java/org/kie/workbench/common/screens/datamodeller/client/SourceEditorReadOnlyTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.screens.datamodeller.client;
+
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


manstis:
@wmedvede Could you please review?

jhrcek:
@yurloc Since we don't have any comprehensive test for the access restrictions specified in workbench-policy.properties file I would me most happy if we actually implemented something like one or two dedicated negative UI tests for these (e.g login with analyst and visit all the places in workbench, try doing everything that should be forgiven and verify it can't be done). Then we could delete these several other of the existing isolated negative tests & speed the tests up & increase the coverage at the same time.

@manstis Are there any plans for reimplementing the access control restriction implemented via workbench-policy file? E.g. as a part of work done by @dgutierr security management? Or will this way of controlling access to features be preserved?

manstis:
@jhrcek AFAIK @dgutierr 's new ACL stuff provides the means to configure permissions for any *protected resource* - what is considered a protected resource is up to (collectively) "us" - so it's plausible the new ACL stuff would support a tab in an editor to be a protected resource. I assume the Data Modeller already has custom code to handle selection of the "Source" tab. It may need revising to fit David's changed ACL. Perhaps @dgutierr or @wmedvede can add their 2c too!

jhrcek:
:-1: Based on discussion with @dgutierr today it's clear that the current way of controlling access to this feature will be superseded shortly, so I think that tests in this PR will not have value for the future  ( @wmedvede please comment if I'm wrong).

manstis:
@yurloc @jhrcek @dgutierr Is this PR still relevant? Does it need revising for the new ACL stuff?

dgutierr:
I think it needs to be revisited as the ACL APIS have changed quite a lot.

As you can see in the following diff: 

https://github.com/droolsjbpm/kie-wb-common/commit/60047a8fa78e86dd5a8d05c29f5636504bf03cac#diff-5166e0a37add3f6506d1c15648308d73

KieWorkbenchACL no longer exists. You need to use AuthorizationManager instead.

tomasdavidorg:
After discussion with @jhrcek, I will rewrite this test.

For now, please close this pull request.

PR #365, opened 2016-05-24
RHBPMS-1367: [GSS] (6.3.z) unnecessary property name and values are shown in ServerManagementBrowser with locale=ja

paulovmr:


manstis:
@paulovmr @mbiarnes Please make sure you inform Angela about these - otherwise the next time Michael fetches the translations from Zanata theses will be overwritten with the ones in Zanata, that contain the erroneous character.

paulovmr:
@manstis I reassigned the JIRA issue to Angela, asking for her to replicate these changes in Zanata. Thanks for the warning.

PR #366, opened 2016-05-24
[6.4.x] RHBPMS-1367: [GSS] (6.3.z) unnecessary property name and values are shown in ServerManagementBrowser with locale=ja

paulovmr:


manstis:
@paulovmr @mbiarnes Please make sure you inform Angela about these - otherwise the next time Michael fetches the translations from Zanata theses will be overwritten with the ones in Zanata, that contain the erroneous character.

psiroky:
@manstis I am wondering what is the translation process? Should we be directly modifying these non-English files, or should we do that via the Zanata UI and then just fetch the changes?

manstis:
@psiroky Well..... it's debatable; there are (at least) three approaches that would work:-

1) We ask the "translation team" to fix in Zanata, then we fetch and commit to github once they've made the changes (this is probably the more "process correct" solution)
2) We fix in Zanata, then follow the same steps as [1]. Whilst this is OK for "simple" fixes as is this case - there's a rogue character in the translated text - I'd like to stay away from it as we'd have another "mini-process" to follow (i.e. we could not do this when the actual translation is wrong, and would have to follow [1]).
3) We fix in github and ask the "translation team" to fix in Zanata (this is what's happened here; and the risk is that if the fixes are not made in Zanata before our next ```mvn zanata:pull-modules``` our "local" (to github) changes are overwritten.

psiroky:
@manstis thanks for the details! What I am after is to make sure the changes won't get overwritten with the next zanata:pull. I guess for these simple changes, we could do it directly in the Zanta UI? And the bigger ones would be done by the translation team. That way we would avoid any manual step like notifying Zanata team about the changes we did (and hoping they do them on the server as well). WDYT?

manstis:
@psiroky I tend to agree; it's probably more expeditious for us to fix simple issues that require no translation skills (e.g. removing / or ' or {0} - three scenarios we've recently had) in Zanata directly (in fact I've done so for @mbiarnes in the past). If the issue requires some linguistic dexterity then we'd need to reassign the JIRA to Angela (or somebody on her team) and wait for the change to be made in Zanata.

paulovmr:
@manstis Thanks for the process clarification!

I reassigned the JIRA issue to Angela, asking for her to replicate these changes in Zanata.

PR #369, opened 2016-05-26
GUVNOR-2522 : NPE after entering authoring perspective / creating new project

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2522

Related:
https://github.com/droolsjbpm/guvnor/pull/319

There is no need to pass the ProjectContext clone or the real one around. We can just inject the real thing.


manstis:
+1 to merge.

PR #373, opened 2016-05-30
JBPM-5030: Access control based on the new UF's authz subsystem

dgutierr:
Changes required to support the new UF's security APIs

This PR depends on uberfire/uberfire#321


kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/PerspectiveIds.java
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.workbench.client;
+
+public interface PerspectiveIds {
```
tomasdavidorg: @dgutierr Why do you use interface here? Would it be possible possible to use enum instead, wdyt? Thanks.

dgutierr: I think an Interface is fine for defining constants.

yurloc: Firstly, I also think an enum is probably not usable for this job. Another thing is that constants interface is considered an anti-pattern by some. The only disadvantage of using an interface for defining constants is that you can't prevent people from implementing it. If that happens, the constants become a public interface / contract of the implementing class, which is potentially "dangerous" (from maintenance / refactoring point of view). The other option is to use a constants class, which you can make final (non-extendable), making it safer to use in this regard.

I personally prefer using constants classes over interfaces and I think it's a matter of taste to some extent. I don't think that using an interface in this situation is a big problem. Just wanted to share my thoughts.

In case someone is interested in more details, there is a good (and short) article on this topic on DZone https://dzone.com/articles/constants-in-java-the-anti-pattern-1.

tomasdavidorg: @dgutierr @yurloc Ok. Thanks for explanations.

porcelli:
+1 (after UF merge)

csadilek:
Jenkins please retest this.

PR #375, opened 2016-06-02
kie-server: added ServerInstanceConnected and ServerInstanceDisconnec…

mswiderski:
…ted events for fine grained control

this is to fix the kie-wb-common (server management) after droolsjbpm-integration changes merge. More tests and code will come with remove of jbpm runtime as this is just to not fail the build


mrietveld:
@mswiderski Thanks! 

+1 when the tests pass.

PR #377, opened 2016-06-03
GUVNOR-2541: Guided Decision Table Editor: Add support to 'Save all'

manstis:
See https://issues.jboss.org/browse/GUVNOR-2541

This adds generic support to ```KieMultipleDocumentEditor``` for sub-classes to provide a "Save all" mechanism to Users.

### DIFF
--- a/kie-wb-common-widgets/kie-wb-metadata-widget/pom.xml
+++ b/kie-wb-common-widgets/kie-wb-metadata-widget/pom.xml
@@ -62,6 +62,14 @@
       <artifactId>errai-security-server</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.jboss.errai</groupId>
+      <artifactId>errai-ui</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.errai</groupId>
+      <artifactId>errai-ioc</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.uberfire</groupId>
       <artifactId>uberfire-client-api</artifactId>
     </dependency>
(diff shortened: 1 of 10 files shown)


zkrejcov:
Thanks for the added javadocs!

manstis:
Part of an ensemble:

https://github.com/droolsjbpm/kie-wb-common/pull/377
https://github.com/droolsjbpm/drools-wb/pull/190

csadilek:
Looks good. One question: Should the "save all" menu item be enabled/disabled based on the lock state similar to the save/rename/delete menu items? [1]

Now that you've added the path to the UpdatedLockStatusEvent this seems doable. So, if one of the assets was locked by someone else "save all" would be disabled.

Wdyt?

[1] https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/menu/BasicFileMenuBuilderImpl.java#L518

manstis:
@csadilek The "regular" menu items act on single assets; hence if the asset is locked it makes sense to prevent saving, renaming, deleting etc. "Save all" however acts on *all* the assets registered to the ```KieMultipleDocumentEditor``` only one (of a few) of which may be locked. At the moment the code in the PR skips are documents that are "read-only". The state of this flag is currently maintained by sub-classes or ```KieMultipleDocumentEditor``` (for example my use in Drools uses a ```LockManager``` to control the state - but it does not have to be limited to this, atm it's up to the concrete implementation to decide what makes a "document" read-only). That said, perhaps I need to think about moving ```LockManager``` into ```KieMultipleDocumentEditor``` instead of letting sub-classes do their own thing.... hmmm... food for thought for the weekend.

csadilek:
@manstis understood. Just wanted to discuss it. I guess it comes down to the question if we're ok with "saveAll" to potentially only save "some" files, not all of them. 

Thinking about this more, it actually is fine as there can't be any unsaved changes for a locked file (with the current pessimistic locking).

So, +1. Let me know if you agree then I will merge.

PR #391, opened 2016-06-22
GUVNOR-2559: Guided Decision Table Editor: Improve appearance (for Red Hat Summit) - Some colour changes and only show FileName not full Path

manstis:
Tart and fart really; to make some colours more consistent with the rest of the workbench.


csadilek:
+1

PR #394, opened 2016-06-27
Fix WorkbenchTreeProvider bug and remove repeating code

tomasdavidorg:
Looks like there was a bug on line 92 in WorkbenchTreeProvider.java and I remove some repeating code.

If you think the test makes no sense, I will remove it.


manstis:
@dgutierr Hi, could you please review.

dgutierr:
This looks nice!

+1 

Thanks @tomasdavidorg

PR #395, opened 2016-06-27
latest Zanata translation changes in kie-wb-common from 2016-06-27

kiereleaseuser:


kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/resources/org/kie/workbench/common/screens/server/management/client/resources/i18n/Constants_ja.properties
```diff
@@ -30,10 +30,10 @@ ContainerStatusEmptyView.Documentation=ドキュメント
 ContainerStatusEmptyView.LearnMoreMessage=このトピックについての詳細はこちらで調べられます
 ContainerStatusEmptyView.NoRemoteServers=リモートサーバーがありません
 ContainerStatusEmptyView.Refresh=更新
-ContainerView.ArtifactId=アーティファクト ID:
+ContainerView.ArtifactId=アーティファクト ID\:
```
manstis: @paulovmr @psiroky Does this \ cause problems? IIRC there was chat about it in the past.

paulovmr: @manstis Not anymore. This problem turned out to be a bug on Errai *.properties parser, which is already fixed. :)

psiroky: AFAICT this was fixed by Errai 3.2.4.Final (e.g. it will now filter out this `\`). However, it would definitely be worth verifying manually in the UI.

manstis:
+1 to merge.

mbiarnes:
@manstis @paulovmr @bbrodt 
I thought this is soloved with the new errai version?
maybe the binaries are still not available?
I will remove the "\"

paulovmr:
@mbiarnes The "\" are still there, but the problem will no longer happen, because the parser was fixed to work with escaped ":" (this case) and "=".

PR #397, opened 2016-06-29
RHBPMS-3997: [GSS](6.3.x) Project dependencies are not retained in BPM Suite 6.3.0

karreiro:
In fact, the dependencies were being saved correctly. However, in the dependency page load, a filter was removing the dependencies that had scope different from "compile". As "compile" is the default scope, this PR makes dependencies without scope be handled as a compile dependency.


manstis:
@Rikkola could you please review?

manstis:
Looks like good analysis; personally I don't like passing magic values to methods ie "null" and would prefer another overloaded method.. but that's personal preference!

karreiro:
@manstis Thanks :)

Something like this:
```
// org.guvnor.common.services.project.model.Dependencies
public Collection<GAV> getCompileScopedGavs() {
    return getGavs( "compile", null );
}
```
right?

karreiro:
Done:
- https://github.com/droolsjbpm/guvnor/pull/331
- https://github.com/droolsjbpm/guvnor/pull/332

Rikkola:
Yep getCompileScopedGavs is better. While null is actually same as "compile", since null defaults to "compile". Good to handle the nulls inside the Dependencies class like you did.

+1

PR #398, opened 2016-06-29
[6.4.x] RHBPMS-3997: [GSS](6.3.x) Project dependencies are not retained in BPM Suite 6.3.0

karreiro:
In fact, the dependencies were being saved correctly. However, in the dependency page load, a filter was removing the dependencies that had scope different from "compile". As "compile" is the default scope, this PR makes dependencies without scope be handled as a compile dependency.


karreiro:
ok to test.

Rikkola:
+1

PR #411, opened 2016-07-27
GUVNOR-2621: DataModelOracle: Does not handle annotations with member properties of type Class e.g. @XmlType

manstis:
See https://issues.jboss.org/browse/GUVNOR-2621

The solution contained in this PR is to use the ```toString()``` representation of any non-portable values.

```Class``` on the client has very little use, as it cannot be instantiated. I tested with ```drools-wb-webapp``` and Data Modeller and rules (with, for example, EVENTS) continue to work fine.. All tests pass (I had to change two assertions from ``` Role.Type.EVENT``` to ```"EVENT"``` however we do not use this to determine whether a Type is an event).


manstis:
@pefernan @wmedvede Could you please review?

pefernan:
@manstis looks good to me.
+1

manstis:
@csadilek Please do not merge yet; I want to see how your Errai fix (for marshaling Class) may affect this first. Thanks.

PR #416, opened 2016-08-03
JBPM-5233: Data source management

wmedvede:
- #1 iteration of the data sources management, adds the following main components:

    - DataSourceManagementPerspective: entry point to the ds mgmt interface.
    - DataSourdeDefEditor: edits a data source definition.
    - NewDataSourceDefWizard: creates a new data source definition
    - DriverDefEditor: edits a db driver definition.
    - NewDriverDefWizard: creates a new driver definition.
    - DataSourceDefExplorerScreen: entry point for visualizing the existing data sources and drivers.

    - kie-wb-datasource-mgmt-dbcp: a portable implementation of the data source management system.
    - kie-wb-datasource-mgmt-wildlfy: a Wildlfy based implementation of the data source management system.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
@@ -26,6 +26,8 @@ import org.jboss.errai.common.client.api.Caller;
 import org.jboss.errai.common.client.api.RemoteCallback;
 import org.kie.workbench.common.screens.datamodeller.client.pdescriptor.ClassRow;
 import org.kie.workbench.common.screens.datamodeller.client.pdescriptor.ClassRowImpl;
+import org.kie.workbench.common.screens.datamodeller.client.widgets.datasourceselector.DataSourceInfo;
+import org.kie.workbench.common.screens.datamodeller.client.widgets.datasourceselector.DataSourceSelector;
 import org.kie.workbench.common.screens.datamodeller.client.pdescriptor.PropertyRow;
 import org.kie.workbench.common.screens.datamodeller.client.resources.i18n.Constants;
 import org.kie.workbench.common.screens.datamodeller.client.type.PersistenceDescriptorType;
@@ -37,8 +39,10 @@ import org.kie.workbench.common.screens.datamodeller.model.persistence.Transacti
 import org.kie.workbench.common.screens.datamodeller.service.DataModelerService;
 import org.kie.workbench.common.screens.datamodeller.service.PersistenceDescriptorEditorService;
 import org.kie.workbench.common.screens.datamodeller.service.PersistenceDescriptorService;
+import org.kie.workbench.common.services.shared.preferences.ApplicationPreferences;
 import org.kie.workbench.common.widgets.client.resources.i18n.CommonConstants;
 import org.kie.workbench.common.widgets.metadata.client.KieEditor;
+import org.uberfire.annotations.Customizable;
 import org.uberfire.backend.vfs.ObservablePath;
 import org.uberfire.client.annotations.WorkbenchEditor;
 import org.uberfire.client.annotations.WorkbenchMenu;
@@ -51,6 +55,7 @@ import org.uberfire.lifecycle.OnClose;
 import org.uberfire.lifecycle.OnFocus;
 import org.uberfire.lifecycle.OnMayClose;
 import org.uberfire.lifecycle.OnStartup;
+import org.uberfire.mvp.Command;
 import org.uberfire.mvp.ParameterizedCommand;
 import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.workbench.model.menu.Menus;
@@ -67,6 +72,8 @@ public class PersistenceDescriptorEditorPresenter
 
     private PersistenceDescriptorType type;
 
+    private DataSourceSelector dataSourceSelector;
+
     private Caller<PersistenceDescriptorEditorService> editorService;
 
     private Caller<PersistenceDescriptorService> descriptorService;
@@ -80,6 +87,7 @@ public class PersistenceDescriptorEditorPresenter
     @Inject
     public PersistenceDescriptorEditorPresenter( final PersistenceDescriptorEditorView baseView,
                                                  final PersistenceDescriptorType type,
+                                                 @Customizable final DataSourceSelector dataSourceSelector,
                                                  final Caller<PersistenceDescriptorEditorService> editorService,
                                                  final Caller<PersistenceDescriptorService> descriptorService,
                                                  final Caller<DataModelerService> dataModelerService ) {
@@ -87,9 +95,11 @@ public class PersistenceDescriptorEditorPresenter
         this.view = baseView;
         baseView.setPresenter( this );
         this.type = type;
+        this.dataSourceSelector = dataSourceSelector;
         this.editorService = editorService;
         this.descriptorService = descriptorService;
         this.dataModelerService = dataModelerService;
+        view.showDataSourceSelector( dataSourceIntegrationEnabled() );
     }
 
     @OnStartup
@@ -279,6 +289,31 @@ public class PersistenceDescriptorEditorPresenter
     public void onJTADataSourceChange() {
         ensurePersistenceUnit();
         getContent().getDescriptorModel().getPersistenceUnit().setJtaDataSource( view.getJTADataSource() );
+    }
+
+    @Override
+    public void onSelectDataSource() {
+        dataSourceSelector.setGlobalSelection();
+        dataSourceSelector.show( new ParameterizedCommand<DataSourceInfo>() {
+            @Override
+            public void execute( DataSourceInfo dataSourceInfo ) {
+                ensurePersistenceUnit();
+                if ( dataSourceInfo.isDeployed() ) {
+                    getContent().getDescriptorModel().getPersistenceUnit().setJtaDataSource(
+                            dataSourceInfo.getJndi() );
+                    view.setJTADataSource( dataSourceInfo.getJndi() );
+                }
+            }
+        }, new Command() {
+            @Override
+            public void execute() {
+                //do nothing
+            }
+        } );
+    }
+
+    private boolean dataSourceIntegrationEnabled() {
+        return ApplicationPreferences.getBooleanPref( "persistence-descriptor-editor-options.integrate-datasources" );
     }
 
     @Override
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/handlers/NewJavaFileTextHandler.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/handlers/NewJavaFileTextHandler.java
@@ -44,6 +44,7 @@ import org.uberfire.ext.editor.commons.client.validation.ValidatorWithReasonCall
 import org.uberfire.ext.widgets.common.client.callbacks.HasBusyIndicatorDefaultErrorCallback;
 import org.uberfire.ext.widgets.common.client.common.BusyIndicatorView;
 import org.uberfire.ext.widgets.common.client.common.popups.errors.ErrorPopup;
+import org.uberfire.mvp.Command;
 import org.uberfire.workbench.type.ResourceTypeDefinition;
 
 @ApplicationScoped
@@ -115,7 +116,6 @@ public class NewJavaFileTextHandler extends DefaultNewResourceHandler {
         Map<String, Object> params = new HashMap<String, Object>( );
         for ( ResourceOptions options : resourceOptions ) {
             params.putAll( options.getOptions() );
-            options.restoreOptionsDefaults();
         }
 
         dataModelerService.call( getSuccessCallback( presenter ),
@@ -147,4 +147,11 @@ public class NewJavaFileTextHandler extends DefaultNewResourceHandler {
         } ).isJavaFileNameValid( javaFileName + ".java" );
     }
 
+    @Override
+    public Command getCommand( NewResourcePresenter newResourcePresenter ) {
+        for ( ResourceOptions options : resourceOptions ) {
+            options.restoreOptionsDefaults();
+        }
+        return super.getCommand( newResourcePresenter );
+    }
 }
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/widgets/jpadomain/JPADataObjectEditor.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/widgets/jpadomain/JPADataObjectEditor.java
@@ -29,7 +29,9 @@ import com.google.gwt.user.client.ui.Widget;
 import org.gwtbootstrap3.client.ui.constants.ButtonType;
 import org.kie.workbench.common.screens.datamodeller.client.command.DataModelCommand;
 import org.kie.workbench.common.screens.datamodeller.client.command.DataModelCommandBuilder;
+import org.kie.workbench.common.screens.datamodeller.client.command.ValuePair;
 import org.kie.workbench.common.screens.datamodeller.client.handlers.DomainHandlerRegistry;
+import org.kie.workbench.common.screens.datamodeller.client.handlers.jpadomain.JPADomainHandler;
 import org.kie.workbench.common.screens.datamodeller.client.handlers.jpadomain.util.SequenceGeneratorValueHandler;
 import org.kie.workbench.common.screens.datamodeller.client.model.DataModelerPropertyEditorFieldInfo;
 import org.kie.workbench.common.screens.datamodeller.client.resources.i18n.Constants;
@@ -42,10 +44,10 @@ import org.kie.workbench.common.services.datamodeller.core.Annotation;
 import org.kie.workbench.common.services.datamodeller.core.DataObject;
 import org.kie.workbench.common.services.datamodeller.core.ObjectProperty;
 import org.kie.workbench.common.services.datamodeller.core.impl.AnnotationImpl;
-import org.uberfire.ext.widgets.common.client.resources.i18n.CommonConstants;
 import org.uberfire.ext.properties.editor.client.fields.BooleanField;
 import org.uberfire.ext.properties.editor.client.fields.TextField;
 import org.uberfire.ext.properties.editor.model.PropertyEditorCategory;
+import org.uberfire.ext.widgets.common.client.resources.i18n.CommonConstants;
 import org.uberfire.mvp.Command;
 
 @Dependent
@@ -56,6 +58,8 @@ public class JPADataObjectEditor
     private static Map<String, DataModelerPropertyEditorFieldInfo> propertyEditorFields = new HashMap<String, DataModelerPropertyEditorFieldInfo>();
 
     private JPADataObjectEditorView view;
+
+    private JPADomainHandler handler;
 
     private static final String EOL = "</BR>";
 
@@ -71,6 +75,9 @@ public class JPADataObjectEditor
         super( handlerRegistry, dataModelerEvent, commandBuilder );
         this.view = view;
         view.init( this );
+        List<String> domainQuery = new ArrayList<String>( );
+        domainQuery.add( getDomainName() );
+        handler = (JPADomainHandler) handlerRegistry.getDomainHandlers( domainQuery ).get( 0 );
     }
 
     @PostConstruct
@@ -103,6 +110,7 @@ public class JPADataObjectEditor
         if ( dataObject != null ) {
             updateEntityField( dataObject.getAnnotation( JPADomainAnnotations.JAVAX_PERSISTENCE_ENTITY_ANNOTATION ) );
             updateTableNameField( dataObject.getAnnotation( JPADomainAnnotations.JAVAX_PERSISTENCE_TABLE_ANNOTATION ) );
+            updateAuditedField( dataObject.getAnnotation( JPADomainAnnotations.HIBERNATE_ENVERS_AUDITED ) );
             setReadonly( getContext() == null || getContext().isReadonly() );
         }
         loadPropertyEditor();
@@ -167,6 +175,25 @@ public class JPADataObjectEditor
         }
     }
 
+    @Override
+    public void onAuditedFieldChange( String newValue ) {
+        if ( getDataObject() != null ) {
+
+            Boolean doAdd = Boolean.parseBoolean( newValue );
+            DataModelCommand command;
+            if ( doAdd ) {
+                List<ValuePair> defaultValues = new ArrayList<>( );
+                defaultValues.add( new ValuePair( "targetAuditMode", "NOT_AUDITED" ) );
+                command = commandBuilder.buildDataObjectAddAnnotationCommand( getContext(),
+                        getName(), getDataObject(), JPADomainAnnotations.HIBERNATE_ENVERS_AUDITED, defaultValues );
+            } else {
+                command = commandBuilder.buildDataObjectRemoveAnnotationCommand( getContext(),
+                        getName(), getDataObject(), JPADomainAnnotations.HIBERNATE_ENVERS_AUDITED );
+            }
+            command.execute();
+        }
+    }
+
     //convenient method for facilitating testing
     protected void doEntityFieldChange( boolean isPersistable, boolean generateIdenfitier ) {
 
@@ -195,6 +222,15 @@ public class JPADataObjectEditor
         }
     }
 
+    private void updateAuditedField( Annotation annotation ) {
+        clearAuditedField();
+        if ( annotation != null ) {
+            if ( propertyEditorFields.containsKey( JPADataObjectEditorView.AUDITED_FIELD ) ) {
+                updatePropertyEditorField( JPADataObjectEditorView.AUDITED_FIELD, annotation, "true" );
+            }
+        }
+    }
+
     protected void loadPropertyEditor() {
         view.loadPropertyEditorCategories( getPropertyEditorCategories() );
     }
@@ -208,7 +244,9 @@ public class JPADataObjectEditor
 
         category.withField( createEntityField() );
         category.withField( createTableNameField() );
-
+        if ( handler.isDataObjectAuditEnabled() ) {
+            category.withField( createAuditedField() );
+        }
         return categories;
     }
 
@@ -228,6 +266,15 @@ public class JPADataObjectEditor
                 TextField.class,
                 Constants.INSTANCE.persistence_domain_objectEditor_table_field_help_heading(),
                 Constants.INSTANCE.persistence_domain_objectEditor_table_field_help() );
+    }
+
+    private DataModelerPropertyEditorFieldInfo createAuditedField() {
+        return createField( Constants.INSTANCE.persistence_domain_objectEditor_audited_field_label(),
+                JPADataObjectEditorView.AUDITED_FIELD,
+                "false",
+                BooleanField.class,
+                Constants.INSTANCE.persistence_domain_objectEditor_audited_field_help_heading(),
+                Constants.INSTANCE.persistence_domain_objectEditor_audited_field_help() );
     }
 
     private DataModelerPropertyEditorFieldInfo createField( String label, String key, String currentStringValue, Class<?> customFieldClass ) {
@@ -251,6 +298,7 @@ public class JPADataObjectEditor
     public void clear() {
         clearEntityField();
         clearTableNameField();
+        clearAuditedField();
     }
 
     private void clearEntityField() {
@@ -259,6 +307,12 @@ public class JPADataObjectEditor
 
     private void clearTableNameField() {
         updatePropertyEditorField( JPADataObjectEditorView.TABLE_NAME_FIELD, null, "" );
+    }
+
+    private void clearAuditedField() {
+        if ( propertyEditorFields.containsKey( JPADataObjectEditorView.AUDITED_FIELD ) ) {
+            updatePropertyEditorField( JPADataObjectEditorView.AUDITED_FIELD, null, "false" );
+        }
     }
 
     private void updatePropertyEditorField( String fieldId, Annotation currentValue, String currentStringValue ) {
--- a/kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceDef.java
+++ b/kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceDef.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.screens.datasource.management.model;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+@Portable
+public class DataSourceDef {
+
+    String uuid;
+
+    String name;
+
+    String jndi;
+
+    String connectionURL;
+
+    String driverUuid;
+
+    String user;
+
+    String password;
+
+    public DataSourceDef() {
+    }
+
+    public String getUuid() {
+        return uuid;
+    }
+
+    public void setUuid( String uuid ) {
+        this.uuid = uuid;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName( String name ) {
+        this.name = name;
+    }
+
+    public String getJndi() {
+        return jndi;
+    }
+
+    public void setJndi( String jndi ) {
+        this.jndi = jndi;
+    }
+
+    public String getConnectionURL() {
+        return connectionURL;
+    }
+
+    public void setConnectionURL( String connectionURL ) {
+        this.connectionURL = connectionURL;
+    }
+
+    public String getDriverUuid() {
+        return driverUuid;
+    }
+
+    public void setDriverUuid( String driverUuid ) {
+        this.driverUuid = driverUuid;
+    }
+
+    public String getUser() {
+        return user;
+    }
+
+    public void setUser( String user ) {
+        this.user = user;
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public void setPassword( String password ) {
+        this.password = password;
+    }
+
+    @Override
+    public String toString() {
+        return "DataSourceDef{" +
+                "uuid='" + uuid + '\'' +
+                ", name='" + name + '\'' +
+                ", jndi='" + jndi + '\'' +
+                ", connectionURL='" + connectionURL + '\'' +
+                ", driverUuid='" + driverUuid + '\'' +
+                ", user='******'" +
+                ", password='******'" +
+                '}';
+    }
+
+    @Override
+    public boolean equals( Object o ) {
+        if ( this == o ) {
+            return true;
+        }
+        if ( o == null || getClass() != o.getClass() ) {
+            return false;
+        }
+
+        DataSourceDef that = ( DataSourceDef ) o;
+
+        if ( uuid != null ? !uuid.equals( that.uuid ) : that.uuid != null ) {
+            return false;
+        }
+        if ( name != null ? !name.equals( that.name ) : that.name != null ) {
+            return false;
+        }
+        if ( jndi != null ? !jndi.equals( that.jndi ) : that.jndi != null ) {
+            return false;
+        }
+        if ( connectionURL != null ? !connectionURL.equals( that.connectionURL ) : that.connectionURL != null ) {
+            return false;
+        }
+        if ( driverUuid != null ? !driverUuid.equals( that.driverUuid ) : that.driverUuid != null ) {
+            return false;
+        }
+        if ( user != null ? !user.equals( that.user ) : that.user != null ) {
+            return false;
+        }
+        return !( password != null ? !password.equals( that.password ) : that.password != null );
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = uuid != null ? uuid.hashCode() : 0;
+        result = ~~result;
+        result = 31 * result + ( name != null ? name.hashCode() : 0 );
+        result = ~~result;
+        result = 31 * result + ( jndi != null ? jndi.hashCode() : 0 );
+        result = ~~result;
+        result = 31 * result + ( connectionURL != null ? connectionURL.hashCode() : 0 );
+        result = ~~result;
+        result = 31 * result + ( driverUuid != null ? driverUuid.hashCode() : 0 );
+        result = ~~result;
+        result = 31 * result + ( user != null ? user.hashCode() : 0 );
+        result = ~~result;
+        result = 31 * result + ( password != null ? password.hashCode() : 0 );
+        result = ~~result;
+        return result;
+    }
+}
--- a/kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceStatus.java
+++ b/kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceStatus.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.screens.datasource.management.model;
+
+public enum DataSourceStatus {
+
+    /**
+     * A data source that was just created but never looked up but the lookup operation.
+     */
+    NEW,
+
+    /**
+     * A data source that was looked up by at least one client.
+     */
+    REFERENCED,
+
+    /**
+     * A data source in stale mode may continue serving connections but it's not safe to do it. Components that uses
+     * a data source may register themselves as listeners to be notified about data source status changes, and in this
+     * way know when a data source enters in STALE mode. If this is the case, it's recommended to do a new lookup
+     * of the data source to get a safe reference.
+     *
+     */
+    STALE
+
+}
--- a/kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DriverDefInfo.java
+++ b/kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DriverDefInfo.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.screens.datasource.management.model;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.uberfire.backend.vfs.Path;
+
+@Portable
+public class DriverDefInfo extends DefInfo {
+
+    private DriverDeploymentInfo deploymentInfo;
+
+    public DriverDefInfo() {
+    }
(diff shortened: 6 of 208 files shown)


wmedvede:
@manstis @csadilek 

Hello Mike, Christian,
This is the data sources management PR related to kie-wb-commons. Would you mind take a look?.

Thanks!!
Walter.

csadilek:
@wmedvede failing tests on jenkins is related: https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-common-pullrequests/369/testReport/junit/org.kie.workbench.common.screens.datasource.management.client.explorer.common/DefExplorerContentTest/testDataSourceDefClick/

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
```diff
@@ -282,6 +291,31 @@ public void onJTADataSourceChange() {
...
+                    getContent().getDescriptorModel().getPersistenceUnit().setJtaDataSource(
+                            defInfo.getDeploymentInfo().getJndi() );
+                    view.setJTADataSource( defInfo.getDeploymentInfo().getJndi() );
+                }
+            }
+        }, new Command() {
+            @Override
+            public void execute() {
+                //do nothing
+            }
```
csadilek: Could use Lambda expressions for Commands and ParameterizedCommands to make this shorter: () -> {}

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceStatus.java
```diff
@@ -0,0 +1,40 @@
...
+    NEW,
+
+    /**
+     * A data source that was looked up by at least one client.
+     */
+    REFERENCED,
+
+    /**
+     * A data source in stale mode may continue serving connections but it's not safe to do it. Components that uses
+     * a data source may register them self as listeners to be notified about data source status changes, and in this
```
csadilek: Components that use a data source may register themselves as listeners...

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceStatus.java
```diff
@@ -0,0 +1,40 @@
...
+     * way know when a data source enters in STALE mode. If this is the case, it recommended to do a new lookup
```
csadilek: it's recommended... (just nit-picking, really appreciate the javadocs ;))

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DriverDefInfo.java
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.screens.datasource.management.model;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.uberfire.backend.vfs.Path;
+
+@Portable
+public class DriverDefInfo extends DefInfo {
```
csadilek: Are instances of all these classes sent between the client and the server? If not, they should not be ```@Portable```. Errai needs to generate a marshaller for every portable type. So, the fewer the better in general.

csadilek:
Nice work! Need to interrupt and continue tomorrow after @manstis had a look as well, it's a very big changeset ;).

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
wmedvede: Agree to use them in new stuff, in fact I've used them at some places but not in all new places to be honest, for this editor that runs in both 7.x and 6.x I'd prefer to keep this java 7 like code style.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DriverDefInfo.java
wmedvede: yes this class us used both at client side by the driver definition editor, and the new driver wizard, and at server side. I think that in general I was careful with tagging @Portable classes, but whaterver thing you find please let me know.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
manstis: I agree with @wmedvede .. if code is also destined for 6.5.x using JDK1.8 features makes cherry-picking impossible.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
```diff
@@ -282,6 +291,31 @@ public void onJTADataSourceChange() {
...
+        } );
+    }
+
+    private boolean dataSourceIntegrationEnabled() {
+        return ApplicationPreferences.getBooleanPref( "persistence-descriptor-editor-options.integrate-datasources" );
```
manstis: It'll be advantageous to make this a constant somewhere (and to be honest I'd be tempted to move ```org.livespark.backend.server.preferences.LiveSparkPreferencesLoader``` from LiveSpark to here.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/handlers/NewJavaFileTextHandler.java
```diff
@@ -147,4 +148,11 @@ public void callback( final Boolean response ) {
         } ).isJavaFileNameValid( javaFileName + ".java" );
     }
 
+    @Override
+    public Command getCommand( NewResourcePresenter newResourcePresenter ) {
+        for ( ResourceOptions options : resourceOptions ) {
```
manstis: I assume this code is "guilty" of making the "options" in the New Data Object modal hide "is auditable" when you click on the "OK" button and before the Data Modeller appears. Shame we cannot reset options elsewhere (e.g. when the New Resource modal is shown). See [here](https://www.youtube.com/watch?v=pichHuPwA2s&feature=youtu.be) @ 4:12

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/widgets/jpadomain/JPADataObjectEditor.java
```diff
@@ -167,6 +175,25 @@ public void onTableNameChange( String newValue ) {
         }
     }
 
+    @Override
+    public void onAuditedFieldChange( String newValue ) {
+        if ( getDataObject() != null ) {
+
+            Boolean doAdd = Boolean.TRUE.toString().equals( newValue );
```
manstis: ```Boolean.parseBoolean(...)``` is probably better (as it handles case insensitivity). Also ```doAdd``` could be a ```boolean``` not a ```Boolean```.. but is being a bit picky ;-)

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceDef.java
```diff
@@ -0,0 +1,161 @@
...
+
+    String jndi;
+
+    String connectionURL;
+
+    String driverUuid;
+
+    String user;
+
+    String password;
```
manstis: @csadilek @wmedvede Do we, should we, have any concerns about passing (unencrypted) passwords across the wire? Are we happy to promote (do we promote?) use of SSL for BxMS in scenarios involving open networks etc?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/TestResult.java
```diff
@@ -0,0 +1,50 @@
...
+    public boolean isTestPassed() {
+        return testPassed;
+    }
+
+    public void setTestPassed( boolean testPassed ) {
+        this.testPassed = testPassed;
+    }
+
+    public String getMessage() {
+        return message;
```
manstis: I suspect this will be a nightmare to i18n? If you have a well defined list of messages it can be better to use an enum and decode the enum to a message client-side (it looks like ```org.kie.workbench.common.screens.datasource.management.backend.service.DataSourceDefEditorServiceImpl``` has a pretty well defined set of potential messages)

manstis: ```org.uberfire.ext.widgets.common.client.callbacks.DefaultErrorCallback``` does something similar (but uses ```@Portable``` exceptions rather than an enum).

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/core/DeploymentOptions.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package org.kie.workbench.common.screens.datasource.management.backend.core;
+
+/**
+ * This class models the available deployment options that can be used for deploying data sources and drivers.
+ */
+public class DeploymentOptions {
+
+    enum DeploymentMode { CREATE_DEPLOYMENT, CRATE_OR_RESYNC_DEPOLYMENT }
```
manstis: Typo ```CREATE_OR_RESYNC_DEPLOYMENT```

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/core/DriverDeploymentCache.java
```diff
@@ -0,0 +1,41 @@
...
+package org.kie.workbench.common.screens.datasource.management.backend.core;
+
+import java.util.List;
+
+import org.kie.workbench.common.screens.datasource.management.model.DataSourceDeploymentInfo;
+import org.kie.workbench.common.screens.datasource.management.model.DriverDef;
+import org.kie.workbench.common.screens.datasource.management.model.DriverDeploymentInfo;
+
+/**
+ * Specialized cache for the storage of the driver deployments.
```
manstis: What's so specialized about it vs (for example) having a sub-class of  ```org.guvnor.common.services.backend.cache.LRUCache```?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/core/DriverDeploymentCacheEntry.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.kie.workbench.common.screens.datasource.management.model.DataSourceDeploymentInfo;
+import org.kie.workbench.common.screens.datasource.management.model.DriverDef;
+
+/**
+ * Represents a registration record in the DriverDeploymentCache
+ */
+public class DriverDeploymentCacheEntry {
```
manstis: If this is an entry in a cache; it's probably going to be useful to implement ```equals()``` and ```hashcode()```?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefEditorServiceImpl.java
```diff
@@ -0,0 +1,357 @@
...
+        DataSourceDefEditorContent editorContent = new DataSourceDefEditorContent();
+        String content = ioService.readAllString( Paths.convert( path ) );
+        DataSourceDef dataSourceDef = DataSourceDefSerializer.deserialize( content );
+        editorContent.setDataSourceDef( dataSourceDef );
+        editorContent.setProject( projectService.resolveProject( path ) );
+        return editorContent;
+    }
+
+    @Override
+    public Path save( final Path path,
```
manstis: You do a lot of discrete operations in this method; is it worthy of synchronisation?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefEditorServiceImpl.java
```diff
@@ -0,0 +1,357 @@
...
+
+            //create the datasource file.
+            ioService.write( nioPath, content, new CommentedOption( optionsFactory.getSafeIdentityName() ) );
+            fileCreated = true;
+
+            runtimeManager.deployDataSource( dataSourceDef, DeploymentOptions.create() );
+
+        } catch ( Exception e1 ) {
+            logger.error( "It was not possible to create data source: {}", dataSourceDef.getName(), e1 );
+            if ( fileCreated ) {
```
manstis: Nice tidy-up!

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefEditorServiceImpl.java
```diff
@@ -0,0 +1,357 @@
...
+        return testConnection( dataSourceDef, null );
+    }
+
+    @Override
+    public TestResult testConnection( DataSourceDef dataSourceDef, Project project ) {
+
+        TestResult result = new TestResult( false );
+
+        if ( isEmpty( dataSourceDef.getConnectionURL() ) ) {
+            result.setMessage( "A valid connection url is required" );
```
manstis: You'd have read my comments about i18n!

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefQueryServiceImpl.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private Collection<DriverDefInfo> resolveDrivers( final Path path ) {
+
+        final org.uberfire.java.nio.file.Path nioPath = Paths.convert( path );
+        final List<DriverDefInfo> result = new ArrayList<>( );
+
+        try {
+            final DirectoryStream<org.uberfire.java.nio.file.Path> stream = ioService.newDirectoryStream( nioPath,
+                    entry -> Files.isRegularFile( entry ) &&
```
manstis: There's a ```LinkedFilter``` and ```DotFileFilter``` in guvnor that *could* be used here.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefQueryServiceImpl.java
```diff
@@ -0,0 +1,251 @@
...
+                            !entry.getFileName().toString().startsWith( "." ) &&
+                            entry.getFileName().toString().endsWith( DRIVER_FILE_TYPE ) );
+
+            stream.forEach( file -> {
```
manstis: LOL.. so your argument for not using lamdas instead of anonymous classes was that you would not be able to cherry-pick to other branches that were not on JDK1.8.... hmmmmm.. this sort of change things ;-)

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefQueryServiceImpl.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private Collection<DataSourceDefInfo> resolveDataSources( final Path path ) {
+
+        final org.uberfire.java.nio.file.Path nioPath = Paths.convert( path );
+        final List<DataSourceDefInfo> result = new ArrayList<>( );
+
+        try {
+            final DirectoryStream<org.uberfire.java.nio.file.Path> stream = ioService.newDirectoryStream( nioPath,
+                    entry -> Files.isRegularFile( entry ) &&
```
manstis: Likewise re: ```LinkedFilter``` and ```DotFileFilter```

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DriverDefEditorServiceImpl.java
```diff
@@ -0,0 +1,297 @@
...
+        DriverDefEditorContent editorContent = new DriverDefEditorContent();
+        String content = ioService.readAllString( Paths.convert( path ) );
+        DriverDef driverDef = DriverDefSerializer.deserialize( content );
+        editorContent.setDriverDef( driverDef );
+
+        return editorContent;
+    }
+
+    @Override
+    public Path save( final Path path,
```
manstis: Am I missing something this looks soooooo like ```DataSourceDefEditorServiceImpl```? Shame both cannot extend a base generic class and just implement the specifics for "DataSource" and "Driver" serialization/deserialization...

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/util/URLConnectionFactory.java
```diff
@@ -0,0 +1,90 @@
...
+                throw new Exception( "Class: " + driverClass + " must extend: " + Driver.class.getName() );
+            } else {
+                driver = (Driver) clazz.newInstance();
+            }
+        } catch ( ClassNotFoundException e ) {
+            throw new Exception( "Driver class: " + driverClass + " was not found.", e );
+        }
+    }
+
+    public Connection createConnection() throws SQLException {
```
manstis: I might find it later; but what closes/releases the connections?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/test/java/org/kie/workbench/common/screens/datasource/management/service/DefExplorerQueryServiceTest.java
```diff
@@ -0,0 +1,397 @@
...
+        repo_o3_1 = new RepositoryMock( "repo_o3_1", "repo_o3_1" );
+        repo_o3_2 = new RepositoryMock( "repo_o3_2", "repo_o3_2" );
+        repo_o3_3 =  new RepositoryMock( "repo_o3_3", "repo_o3_3" );
+        o3.getRepositories().add( repo_o3_1 );
+        o3.getRepositories().add( repo_o3_2 );
+        o3.getRepositories().add( repo_o3_3 );
+        organizationalUnits.add( o3 );
+    }
+
+    private class RepositoryMock implements Repository {
```
manstis: A *real* ```Mock``` with ```when( repo.getAlias() ).thenReturn( alias )``` etc would have been far less code!

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/datasource/DataSourceDefEditor.java
```diff
@@ -0,0 +1,389 @@
...
+    private DataSourceDefType type;
+
+    private Caller<DataSourceDefEditorService> editorService;
+
+    private Caller<DataSourceRuntimeManagerClientService> dataSourceManagerClient;
+
+    private DataSourceDefEditorContent editorContent;
+
+    @Inject
+    protected SavePopUpPresenter savePopUpPresenter;
```
manstis: Hmmmmmmmm.. constructor injection and field injection.. always nice to see ;-) It must make the unit tests fun to write?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/driver/DriverDefEditor.java
```diff
@@ -0,0 +1,391 @@
...
+    private DriverDefType type;
+
+    private Caller<DriverDefEditorService> editorService;
+
+    private Caller<DataSourceRuntimeManagerClientService> dataSourceManagerClient;
+
+    private DriverDefEditorContent editorContent;
+
+    @Inject
+    protected SavePopUpPresenter savePopUpPresenter;
```
manstis: You might want to move to constructor injection.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/driver/DriverDefEditor.java
```diff
@@ -0,0 +1,391 @@
...
+
+    @Inject
+    protected DeletePopUpPresenter deletePopUpPresenter;
```
manstis: You might want to move to constructor injection.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/driver/DriverDefEditor.java
```diff
@@ -0,0 +1,391 @@
...
+import org.uberfire.workbench.events.NotificationEvent;
+import org.uberfire.workbench.model.menu.MenuFactory;
+import org.uberfire.workbench.model.menu.Menus;
+
+import static org.uberfire.ext.editor.commons.client.menu.MenuItems.*;
+
+@Dependent
+@WorkbenchEditor( identifier = "DriverDefEditor",
+        supportedTypes = { DriverDefType.class } )
+public class DriverDefEditor
```
manstis: Much like your service implementations; the ```@WorkbenchEditor``` implementation for both Drivers and Data Sources are very similar (especially the concurrency support). You might want to consider consolidating (in the future).

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/driver/DriverDefMainPanelViewImpl.html
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+
+        <div id="driver-class-form-group" class="form-group" data-field="driver-class-form-group">
+            <label class="control-label" data-i18n-key="driverClass"></label>
+            <input class="form-control" data-field="driver-class"/>
+            <span data-field="driver-class-help" class="help-block" style="visibility: hidden;"></span>
+        </div>
+
+        <div id="group-id-form-group" class="form-group" data-field="group-id-form-group">
+            <label class="control-label" data-i18n-key="groupId"></label>
```
manstis: I noticed in your video this is showing as "Group Name" (or it might have been in the Wizard)... you might want to ensure accuracy in the "source" Zanata files before the translation teams translate incorrectly.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContentViewImpl.html
```diff
@@ -0,0 +1,78 @@
...
+              <a href="#" data-field="add-new-datasource">
+                <i class="fa fa-plus"></i>
+                <span data-i18n-key="AddDataSource">Add DataSource</span>
+              </a>
+            </div>
+
+            <div class="ds-mgmt-element-navigation list-group" data-field="datasources-list-group">
+              <a href="#" class="list-group-item">
+                <i class="fa fa-fw fa-folder-o"></i>
+                DataSource A
```
manstis: Is this list overwritten programmatically?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContentViewImpl.html
```diff
@@ -0,0 +1,78 @@
...
+            <div class="ds-mgmt-add-new-element">
+              <a href="#" data-field="add-new-driver">
+                <i class="fa fa-plus"></i>
+                <span data-i18n-key="AddDriver">Add Driver</span>
+              </a>
+            </div>
+
+            <div class="ds-mgmt-element-navigation list-group" data-field="drivers-list-group">
+              <a href="#" class="list-group-item">
+                <i class="fa fa-server"></i>
```
manstis: Same with this list.. did this HTML come from UX and they put place-holders for lists?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContent.java
```diff
@@ -0,0 +1,151 @@
...
+            placeManager.goTo( view.createPlaceRequest( dataSourceDefInfo.getPath() ) );
+        }
+    }
+
+    protected void onDriverItemClick( DriverDefInfo driverDefInfo ) {
+        placeManager.goTo( view.createPlaceRequest( driverDefInfo.getPath() ) );
+    }
+
+    protected DefItem createItem() {
+        return itemInstance.get();
```
manstis: @mbarkley Hi "king of disposal", IIRC we'd need to destroy these ourselves?

manstis: @wmedvede Depending on ^^ I don't see these are destroyed anywhere.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/project/ProjectSelector.java
```diff
@@ -0,0 +1,179 @@
...
+import com.google.gwt.user.client.ui.FlowPanel;
+import org.guvnor.common.services.project.model.Project;
+import org.guvnor.structure.organizationalunit.OrganizationalUnit;
+import org.guvnor.structure.repositories.Repository;
+import org.gwtbootstrap3.client.ui.AnchorListItem;
+import org.kie.workbench.common.screens.explorer.client.resources.i18n.ProjectExplorerConstants;
+import org.kie.workbench.common.screens.explorer.client.widgets.dropdown.CustomDropdown;
+import org.kie.workbench.common.screens.explorer.client.widgets.navigator.NavigatorBreadcrumbs;
+
+public class ProjectSelector extends Composite {
```
manstis: Oeeee.... not ```errai-ui``` or ```UiBinder```... hmmmmm.. @csadilek Do we have *any* permitted exceptions for new development?

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/project/ProjectSelector.java
```diff
@@ -0,0 +1,179 @@
...
+                this.projects.setText( activeProject.getProjectName() );
+            } else {
+                this.projects.setText( ProjectExplorerConstants.INSTANCE.nullEntry() );
+            }
+            for ( final Project project : projects ) {
+                this.projects.add( new AnchorListItem( project.getProjectName() ) {{
+                    addClickHandler( new ClickHandler() {
+                        @Override
+                        public void onClick( ClickEvent event ) {
+                            onOrojectSelected( project );
```
manstis: Typo.. ```onProjectSelected```

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/util/UIUtil.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.google.gwt.dom.client.Element;
+import com.google.gwt.dom.client.Style;
+import org.gwtbootstrap3.client.ui.constants.ValidationState;
+import org.uberfire.ext.widgets.common.client.common.StyleHelper;
+import org.uberfire.ext.widgets.common.client.common.popups.YesNoCancelPopup;
+import org.uberfire.ext.widgets.common.client.resources.i18n.CommonConstants;
+import org.uberfire.mvp.Command;
+
+public class UIUtil {
```
manstis: This could/should move into Uberfire (see ```org.uberfire.ext.widgets.common.client.common.StyleHelper```

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/util/PopupsUtil.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.guvnor.common.services.shared.validation.model.ValidationMessage;
+import org.gwtbootstrap3.client.ui.constants.ButtonType;
+import org.kie.workbench.common.widgets.client.popups.validation.ValidationPopup;
+import org.uberfire.ext.widgets.common.client.common.popups.YesNoCancelPopup;
+import org.uberfire.ext.widgets.common.client.resources.i18n.CommonConstants;
+import org.uberfire.mvp.Command;
+
+@ApplicationScoped
+public class PopupsUtil {
```
manstis: It would be great if this moved to [here](https://github.com/uberfire/uberfire/tree/master/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/popups)

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/resources/org/kie/workbench/common/screens/datasource/management/client/resources/i18n/DataSourceManagementConstants.properties
```diff
@@ -0,0 +1,93 @@
...
+NewDataSourceDefWizard.DataSourceCreatedMessage=Data source: {0} was successfully created
+NewDataSourceDefWizard.DataSourceCreateErrorMessage=Data source was not created due to the following error: {0}
+NewDataSourceDefWizard.WizardStartErrorMessage=Wizard initialization failed, it was not possible to load driver definitions: {0}
+NewDataSourceDefWizard.FileExistsErrorMessage=File already exists: {0}
+
+DataSourceDefPageViewImpl.pageTitle=Data source info
+
+DriverDefMainPanelViewImpl.name=Name
+DriverDefMainPanelViewImpl.driverClass=Driver Class Name
+DriverDefMainPanelViewImpl.groupId=Group Name
```
manstis: Found it! This should obviously be "Group ID".. best to correct before this ends up in Zanata.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-wildfly/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/integration/wildfly/WildflyBaseClient.java
```diff
@@ -0,0 +1,187 @@
...
+        return outcome.contains( "failed" );
+    }
+
+    public void safeClose( final Closeable closeable ) {
+        if ( closeable != null ) {
+            try {
+                if ( logger.isDebugEnabled() ) {
+                    logger.debug( "starting ModelControllerClient connection close" );
+                }
+                boolean disableClose = Boolean.valueOf( System.getProperty( "disableClose" ) );
```
manstis: Is this a "KIE" setting? If so it'd be better to prefix with some name-space (like most other System Properties we define)

manstis:
@wmedvede OK, I'm finished reviewing.

I've commented inline most of the time; however generally:-

1) There don't appear to be any Unit Tests for anything on the back-end; screens' services, Driver provisioning, Data Source provisioning etc.. Are these to follow when you return from holiday?

2) You use ```IsWidget``` in your Views. We should try to use ```IsElement``` and ```UberElement<T>``` instead (should not be too difficult since most of your views use ```errai-ui```.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/PersistenceDescriptorEditorPresenter.java
wmedvede: I think that at this moment this something "in the middle" because of this early iteration. And probably the e.g. "persistence-descriptor-editor-options.integrate-datasources" will disappear when we have a higher level configuration of the LS application. I don't, just o say something, there should be a LS application configuration "tab/screen/editor" or whatever place were the user just says, I want to use this DS, and the persistence.xml descriptor should automatically configured behind the scene. 

But to be honest I couldn't find a better place to setup this at this point, and keeping the Preferences loader in Livespark, at least by now, it's a warranty that we don't contaminate the other workbenches.

In my opinion we can keep this for this iteration.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContent.java
mbarkley: @manstis @wmedvede I think in this case changing the `itemInstance` from `Instance` to `ManagedInstance` would suffice, since the `DefExplorerContent` is dependent scoped.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/project/ProjectSelector.java
csadilek: This is a custom widget that is being used in an Errai UI templated bean. Ideally, we don't want to use any widget types anymore and use DOM wrappers instead. @wmedvede can you add a TODO for the second iteration of this to change all your Errai UI beans to not rely on widgets? @manstis Yes, I think we should only allow widgets where we can't easily provide an alternative i.e. some of the gwtbootstrap3 components that we use programmatically (which are widgets).

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceDef.java
csadilek: @manstis @wmedvede without SSL, our login screen sends the password in plain text which is much worse than the DS password ;).

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/handlers/NewJavaFileTextHandler.java
wmedvede: @manstis 
the effect you say is because the default options is setup at the same time the object is created and this causes the 4:12 "effect", works well if the options are reset when the dialog is about to be opened in the getCommand. It's now fixed

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-api/src/main/java/org/kie/workbench/common/screens/datasource/management/model/DataSourceDef.java
wmedvede: This is something we'll have to address at some point. I don't know how other WB pieces works, e.g., when the user logg-in by I can imagine that if the workbench is not installed with SSL then the pasword will be sent in plain text along the wire. 
But, yes, we should go back on this in next iterations.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefQueryServiceImpl.java
wmedvede: I said that lamdas are good for the new stuff, and this is an example of something completely new that will live in 7.x. The case of the PersistenceDescriptorEditor above is different because it's something that we'll need to keep both in 7.x and 6.x ;). And I've also said that I didn't used the lambas in all places, even for the new stuff. Btw, sooner or later it'll become a standard I guess.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/util/URLConnectionFactory.java
wmedvede: yes, clients of this class typically connection pools are responsible for releasing connections. This class just create them.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/datasource/DataSourceDefEditor.java
wmedvede: good catch, thanks

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContentViewImpl.html
wmedvede: yes, this is just for having a more clear idea about what the list-group has inside. It's also good for copy pasting this piece into an html editor and looking how it looks like. But the list elements are then added programatically, so Data Source A is something that the user will never see.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContentViewImpl.html
wmedvede: this list is also the populated programmatically

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/project/ProjectSelector.java
wmedvede: I think it doesn't look so bad for this iteration given that 1) this component doesn't need to much ui layout, and 2) I'm not sure this component will survive... since it seems like all the navigation mechanism, etc., will be changed.
Additionally, I believe that the Idea of having this separate project navigation in the data sources perspective is something that won't survive. In my opinion, in case we keep the idea of having data sources per project, then you shoud be able to crate the data source from some point in the given project. I mean, it's not so good to have to havigate to a separate perspective for having the chance of creating a data source for a given project. I think this are very likely to be changed. So I think it's ok for this iteration.

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-backend/src/main/java/org/kie/workbench/common/screens/datasource/management/backend/service/DataSourceDefEditorServiceImpl.java
wmedvede: @manstis I don't see what synchronization you suggest for this case. The method basically gets the String with the json containing the data source definition, reads the project, and returns the result

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/editor/driver/DriverDefMainPanelViewImpl.html
wmedvede: Fixed

kie-wb-common-screens/kie-wb-common-datasource-mgmt/kie-wb-common-datasource-mgmt-client/src/main/java/org/kie/workbench/common/screens/datasource/management/client/explorer/common/DefExplorerContent.java
wmedvede: @manstis @mbarkley 
fixed, thanks!

wmedvede:
@manstis 
Hello Mike, with reference to 1) yes, there are missing tests for back-end side. And will have to be completed after holidays.
With reference to 2) I would prefer to apply this type of changes after holidays too, since the UI that we have at this moment works. 

@csadilek @manstis 
To summarize, I think some suggestions were applied and some other ones will have to wait. So if there aren't any other suggestion from my side this is "ready" to merge.

csadilek:
+1 (give that work on this continues later i.e. more tests added).

PR #417, opened 2016-08-05
Fix the build after data source management merge.

mbarkley:


csadilek:
+1

csadilek:
This fixes the failing optaplanner-wb and drools-wb builds.

manstis:
It might be worth checking kie-wb deployed to Wildfly or something to check we don't need to filter in assemblies.

csadilek:
@manstis we are verify that before merging.

csadilek:
Verified that deployment is fine. Merging, deploying snapshots and triggering builds.

PR #419, opened 2016-08-09
Fix i18ed label in User perspective

jhrcek:
cherry-pick fix from 6.4.x

For explanation see older PR https://github.com/droolsjbpm/kie-wb-common/pull/257


manstis:
+1 to merge (when green). Thanks @jhrcek

PR #420, opened 2016-08-09
DROOLS-1235 MBean naming rework

tarilabs:
* reflecting internal api / implementation of KieContainerImpl
constructor change, requiring a container Id (normally a new
KieContainerImpl is called via KieServices so transparently, but in this
case is explicitly required manual management).

ref: https://github.com/droolsjbpm/drools/pull/866/files#diff-e543bc3de6e92024156f45b7159043d0L106
ref: droolsjbpm/drools#866

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.function.Predicate;
 
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
@@ -528,7 +529,7 @@ public class Builder {
             final KieModule kieModule = kieBuilder.getKieModule();
             final ReleaseId releaseId = kieModule.getReleaseId();
             final org.drools.compiler.kie.builder.impl.KieProject kieProject = new KieModuleKieProject( (InternalKieModule) kieBuilder.getKieModule(), null );
-            final KieContainer kieContainer = new KieContainerImpl( kieProject,
+            final KieContainer kieContainer = new KieContainerImpl( UUID.randomUUID() + "_bz1202551" , kieProject,
                                                                     KieServices.Factory.get().getRepository(),
                                                                     releaseId );
             return kieContainer;


kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java
```diff
@@ -528,7 +529,7 @@ public KieContainer getKieContainer() {
             final KieModule kieModule = kieBuilder.getKieModule();
             final ReleaseId releaseId = kieModule.getReleaseId();
             final org.drools.compiler.kie.builder.impl.KieProject kieProject = new KieModuleKieProject( (InternalKieModule) kieBuilder.getKieModule(), null );
-            final KieContainer kieContainer = new KieContainerImpl( kieProject,
+            final KieContainer kieContainer = new KieContainerImpl( UUID.randomUUID() + "_bz1202551" , kieProject,
```
etirelli: @manstis why is the container here instantiated directly, instead of using the factory method?

@tarilabs I don't understand why concatenate the "_bz..." string with the random UUID?

manstis: @etirelli What Factory method in particular? You should read the [comment](https://github.com/tarilabs/kie-wb-common/blob/2ce87ab33944fd2009c15bdfcd020a81237cebb8/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java#L525) in the code above that call to understand *why* we're getting a container this way. IDK if there's an alternative to avoid looking up a ```KieContainer``` from the ```KieRepository```?

etirelli: @manstis sorry, the comment was not showing in the diff and I did not check the full file. I see it now. Thanks!

tarilabs: @etirelli this ID will never be reflected back in the KieServices internal map of created container Ids, given this container is created manually using the internal api / directly the implementation. I've run the test locally on this git repo with this change, and it's okay (includes originally a test covering this BZ fix). The reason I preferred to put an ad-hoc ID marker however, shall ever in the future be found any problem with these containers which are managed manually; otherwise by simply looking at a log would be impossible to understand, as the id would be completely random indeed.

manstis:
Unless there's any strong objections.. this gets a +1 from me to merge (when green).

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java
etirelli: Thanks for the explanation.

manstis:
@tarilabs Why close?

tarilabs:
Sorry @etirelli  @manstis , I'm just realizing the reflecting internal api / implementation `new KieContainerImpl()` is being called also in Optaplanner, and who know what I'll found out next. At this point, I think I will manage the case internally (anyway is being a deliberate internal call) and having the ID with custom marker + uuid done directly there. Possibly I'll mark constructor deprecated, so can be managed on all other projects in due time.

PR #422, opened 2016-08-10
PLANNER-620 Workbench: CH: Support difficulty comparator definition

mcimbora:
Preview, further changes may come.

https://issues.jboss.org/browse/PLANNER-620 defines a requirement to support creating comparator objects for existing data objects (ones annotated with `@PlanningEntity`). The comparator is created in OptaPlanner WB and added as a nested static class of the associated data object. This PR provides a way to define nested classes & methods for data object.


mcimbora:
OptaPlanner WB PR: https://github.com/droolsjbpm/optaplanner-wb/pull/90

csadilek:
@Rikkola would you mind taking a look at this one as well. You're probably more familiar with this than me?

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/command/DataModelCommandBuilder.java
```diff
@@ -167,4 +168,18 @@ public DataObjectNameChangeCommand buildDataObjectNameChangeCommand( final DataM
...
+                                                                                 final DataObject dataObject,
+                                                                                 final JavaClass nestedClass) {
+        return new DataObjectAddNestedClassCommand( context, source, dataObject, notifier, nestedClass );
+    }
+
+    public DataObjectRemoveNestedClassCommand buildDataObjectRemoveNestedClassCommand( final DataModelerContext context,
+                                                                                 final String source,
+                                                                                 final DataObject dataObject,
+                                                                                 final JavaClass nestedClass) {
+        return new DataObjectRemoveNestedClassCommand( context, source, dataObject, notifier, nestedClass );
```
Rikkola: Minor issue. Code is not aligned. Not big thing.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/command/DataObjectAddNestedClassCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.kie.workbench.common.screens.datamodeller.client.DataModelerContext;
+import org.kie.workbench.common.screens.datamodeller.events.ChangeType;
+import org.kie.workbench.common.screens.datamodeller.events.DataObjectChangeEvent;
+import org.kie.workbench.common.services.datamodeller.core.DataObject;
+import org.kie.workbench.common.services.datamodeller.core.JavaClass;
+
+public class DataObjectAddNestedClassCommand extends AbstractDataModelCommand {
+
+    private JavaClass nestedClass;
```
Rikkola: Another minor issue. If you always set the nestedClass in the constrcutor, best to make this final. So when the next guy comes in and adds a constructor he doesn't accidentally miss this.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/command/DataObjectAddNestedClassCommand.java
```diff
@@ -0,0 +1,43 @@
...
+                                            DataModelChangeNotifier notifier, JavaClass nestedClass ) {
+        super( context, source, dataObject, notifier );
+        this.nestedClass = nestedClass;
+    }
+
+    @Override
+    public void execute() {
+        dataObject.addNestedClass( nestedClass );
+        DataObjectChangeEvent event = new DataObjectChangeEvent( ChangeType.NESTED_CLASS_ADD_CHANGE,
+                getContext().getContextId(), getSource(), getDataObject(), null, null, nestedClass );
```
Rikkola: Ok, I think this is more serious and I know you are just reusing existing code, not making it. We have this all over our code base and every now and then this causes bugs. So when ever anyone runs into it, best to fix it.

There is a long line of parameters here. I took a quick look and the `DataObjectChangeEvent` uses some sort of builder. So you can do `new DataObjectChangeEvent().withCurrentDataObject( getDataObject() ).withNewValue( nestedClass );`. Not sure what the desing patterns is, I prefer Builder for these situations, but since this is already in we can use it.

The problem with long constructors is, hard to know what value is what here. Another problem is that we are passing in nulls and that is the path to NPEs.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/command/DataObjectRemoveNestedClassCommand.java
```diff
@@ -0,0 +1,43 @@
...
+import org.kie.workbench.common.services.datamodeller.core.JavaClass;
+
+public class DataObjectRemoveNestedClassCommand extends AbstractDataModelCommand {
+
+    private JavaClass nestedClass;
+
+    public DataObjectRemoveNestedClassCommand( DataModelerContext context, String source, DataObject dataObject,
+                                               DataModelChangeNotifier notifier, JavaClass nestedClass ) {
+        super( context, source, dataObject, notifier );
+        this.nestedClass = nestedClass;
```
Rikkola: Again I would make the nestedClass final and make sure nestedClass is not null here with: `this.nestedClass = PortablePreconditions.checkNotNull( "nestedClass", nestedClass );`

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/command/DataObjectRemoveNestedClassCommand.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public void execute() {
+        dataObject.removeNestedClass( nestedClass );
+        DataObjectChangeEvent event = new DataObjectChangeEvent( ChangeType.NESTED_CLASS_REMOVE_CHANGE,
+                getContext().getContextId(), getSource(), getDataObject(), null, nestedClass, null );
```
Rikkola: Now the null and nestedClass changed places. This makes me think the `DataObjectEvent` could use a clean up and the `DataObjectRemove/AddNestedClasCommands` should be split to remove the need for nulls. Took a look a the codes and I see several "todos", so somebody is planning to do this. Right now I suggest you don't touch it. Just use the `.withParemeterName()` and `checkNotNull` for the values you care about.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/codegen/GenerationEngine.java
```diff
@@ -172,6 +173,14 @@ public void generateAttribute(GenerationContext generationContext, ObjectPropert
         generateSubTemplate(generationContext, template);
     }
 
+    public void generateMethod( GenerationContext generationContext, Method method, String template) throws Exception {
+        generateSubTemplate(generationContext, template);
+    }
```
Rikkola: Can Method be removed? 
Same goes for `JavaClass` in the `generateNestedClass` method below.

Rikkola:
As I talked to @mcimbora in IRC. Went through this with a fine comb. guvnor and kie-wb-common are shared modules, so we should be extra careful and do some clean up every time we run into anything that smells bad.

mcimbora:
@Rikkola I updated the PR. I added `JavaFileIndexerExtension` to allow client apps to define the index as well. This is due to the fact the Comparator class is defined in OptaPlanner WB. I also added new methods to DomainHandler interfaces (one to process events on client side - when the planner data editor is not constructed yet, we'd miss the updates if we relied on observer method, other one to process data object when model is loaded - it initializes Comparator object of a data object).

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/codegen/GenerationEngine.java
mcimbora: It's referenced by a template, see `java_nested_class.vm`

mcimbora: It's referenced by a template, see `java_nested_class.vm`

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/command/DataObjectAddNestedClassCommand.java
mcimbora: I updated the code as suggested. I agree it's more obvious what the code does now. I can fix all the data model commands in a separate PR.

Rikkola:
Ok. Thank you.
+1

wmedvede:
Hello @mcimbora , I wasn't included in this thread, not a problem, also I believe the PR was merged during my holidays. But I could see some coded added here during the review of  https://github.com/droolsjbpm/kie-wb-common/pull/467, and I found something that adds some problems, and potential regressions.

It's ok that you add an "auto-generated" inner class if needed. But, when the java source is being upated, all existing inner classes (the generated by you, but eventually any other existing inner class) are being deleted. But only the generated by you is regenerated. Since  we must maintain the round-trip between "user added stuff" and "WB added stuff", we can only remove/overwrite the code we know that is generated by us.
see:

https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/JavaRoasterModelDriver.java#L707

I think that for this stuff we should apply the same strategy as the one we are working on for the planer generated methods.

PR #423, opened 2016-08-14
Adding new forms engine.

pefernan:
Merge after:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/252

Adding forms engine I:
- Added API
- Layout Generation
- Dynamic Form Engine*

Adding forms engine II:
- Added Form Editor
- Added Form Widgets

Adding forms engine III:
- Fixing issues generating dynamic-forms with ListBoxes that user SelectDataProvider

Adding forms engine IV:
- Added CRUD component
- Fixing small issues


pefernan:
Test's are also failing on master. I'll take a look tomorrow.

Rikkola:
Ok over 300 files. Is this all just being moved from location A to B?

csadilek:
jenkins retest this.

pefernan:
@Rikkola yes,  it was placed on the livespark repo.  It has to be moved here to be used by other components.

psiroky:
Should we perhaps migrate the whole git history? Or was it decided not to do so for some reason?

csadilek:
@psiroky not needed. This is only part of the LS work...related to forms which needs to be re-usable in other projects.

psiroky:
@csadilek great, thanks for the explanation!

mbarkley:
Jenkins, please retest this.

mbarkley:
@pefernan Normally I'd add inline comments where appropriate, but there are so many files that looking at the diff in GitHub is too slow so I've aggregated all my notes here.

First I'll say great stuff, Pere! I've tried out LS with the changes in this and related branches and it's all working quite well.

That being said, here are some issues I noticed:

* There are some types in your test cases that you call mocks, but they are not really mocks: DynamicFormRendererMock, MockFieldManager, FormFieldProviderMock, and FormHandlerMock. These are actually testable subclasses of the types being tested (you never mock something you are testing; you mock the types you are not testing). A more accurate naming convention would be to use the prefix "Test" or "Testable" (i.e. TestFieldManger or TestableFieldManager).

* The PictureWidgetDriver has a significant amount of JavaScript code and no tests. I think there's enough going on there to be at least a single test. Also if you ever need to modify this type in the future, I think it would be worth while to try and reduce the amount of JS (probably by using JSInterop), but that's less important than there being some test coverage.

* The DynamicFormRendererTest.doBind method has some verify statement inside a callback. The test would be more reliable if you verified that this callback is invoked (it could just never be invoked and the test would pass).

* The FormDefinitionSerializerImplTest.doSerializationTest method uses `assertNotSame` to compare integers, but ought to use `assertNotEquals` (the former is less reliable since the parameters will be auto-boxed to Integers, at which point they could be different object references while both representing the same int).

* The FormEditorPresenterTest has a couple blocks with commented out code that ought to be deleted.

csadilek:
@romartin if you haven't already done so, it would be great if you could take a look at this tomorrow to see it all works for you. 

@Rikkola this will be worked on more in the next couple of months, but it would be great if you could take a look as well.

Thanks.

pefernan:
@csadilek I think that @romartin is on vacation until september. Maybe we can wait to merge it until then, I can spend the time fixing the comments Max provided.

Rikkola:
We try to make a separate class for the MVP View interface. I see here it was usually in the Presenter. Wouldn't change this for this PR, just something to keep in mind in the future.

Over all I think the code is even better than the existing kie-wb-common code. Short, clean and easy to read. Test coverage is not the best. We should try to raise the coverage when new features go in or bugs are fixed, but if this code is just moved from one location to another ( = QA has tested it and probably has some UI tests ) then I don't see a reason to block the PR because of this. With the exception @mbarkley brought up. The JavaScript should have at least one test so who ever works on this code in the future has an example and a starting point.

Another thing that I might want to bring up is the location of the codes. Currently they are in the root as kie-wb-common/kie-wb-common-forms. kie-wb-common is split into 3 sections screens, services and widgets. I believe all this should go into screens ( kie-wb-common/kie-wb-common-screens/kie-wb-common-forms ).

Rikkola:
So @manstis was ok with this location. I still feel it might be best to place it in screens, but I can live with this location and then we change it later if possible.

pefernan:
Added some changes as @mbarkley suggested, pending add testing to PictureWidgetDriver

mbarkley:
+1 (Since Pere is starting PTO now, better to merge now and get the last test when he returns.)

csadilek:
Thanks @mbarkley and @Rikkola for the reviews. I will go ahead an merge this as it will get us unstuck. Also, the quality is rather high for work in progress and other projects (except Stunner and LiveSpark) are currently not affected.

+1

PR #424, opened 2016-08-15
PLANNER-630 Workbench: Solver configuration names contain redundant 'solver' suffix

mcimbora:
Use suffix of given resource type instead of determining it by the last occurrence of '.' character. This is motivated by the fact that solver configuration filenames use 'solver.xml' suffix, which leads to undesired behavior.


mcimbora:
The test failures are unrelated.

mcimbora:
The same have been observed e.g. in https://github.com/droolsjbpm/kie-wb-common/pull/423

csadilek:
@mcimbora can you please add a test for ```getBaseFileName```. The test failures on master are already investigated. Thanks.

mcimbora:
@csadilek Thanks for looking at this, I updated the PR.

csadilek:
+1

PR #429, opened 2016-08-29
JavaSourceVisitor - add null check for method return type

mcimbora:
Constructors are included in the method list, however their return type is null. Avoid NPEs by adding null check.


csadilek:
@mcimbora could you add a test for this, please? Seems this class isn't unit tested, but it's a good opportunity to start ;).

mcimbora:
@csadilek Sure, will do.

mcimbora:
@csadilek I updated the test coverage of `JavaSourceVisitor`.

csadilek:
thanks @mcimbora, +1.

manstis:
@mcimbora I noticed the same issue today (before seeing this PR). Could you please update the JIRA I raised https://issues.jboss.org/browse/GUVNOR-2653 accordingly?

PR #461, opened 2016-09-08
Fixed Issue moving fields on layout

pefernan:
This fixes a issue that made unable to move form fields inside the layout.


manstis:
The change is subtle. It'd probably be a good idea to add a test for such subtleties.

pefernan:
@manstis test added.

manstis:
Thanks @pefernan 

+1 to merge.

PR #462, opened 2016-09-09
GUVNOR-2673 Default ExceptionHandler that pops up any uncaught RuntimeException

mcimbora:


mcimbora:
Use common Uberfire DefaultErrorCallback to handle uncaught exceptions. This removes the need to define empty error callback when using `org.jboss.errai.common.client.api.Caller`.

manstis:
-1 (following a quick review.. I'd expressly asked for uberfire's ```DefaultErrorCallback``` *NOT* to be replaced... Need to look at this (and the others in more detail - Monday).

csadilek:
@manstis DefaultErrorCallback is still here, unchanged. This should not result in a change in behaviour.

manstis:
@mcimbora @csadilek Can this be merged? IIRC I saw on IRC some comment that ```@UncaughtExceptionHandler``` was not working as expected should an ```Exception``` be thrown in a ```@AfterInitialisation``` annotated method?

mcimbora:
@manstis Let's wait until I discuss the issue with @mbarkley. There are changes that might be affected by the issue, e.g. https://github.com/droolsjbpm/kie-wb-common/pull/462/files#diff-0a8208309d24b6d59e8a16a878504b13R160

mcimbora:
@manstis I talked about this with @mbarkley and the exceptions thrown from RPC endpoints should indeed be caught. I checked this once more and it works as expected (the faulty reproducer I provided before threw the exception at a different spot). This can be merged if there are no more questions.

manstis:
+1 to merge.

Thanks for the explanation (and pointers) @csadilek

PR #464, opened 2016-09-13
GUVNOR-2508 : When importing external projects, the dependency are not taken and at application server also

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2508


manstis:
+1 to merge (given the agreed solution is to show a "Loading..." popup) :-)

Rikkola:
For one we need this. The load times can get high. I tested with a jar that does not even exist and this still helps.
Could not reproduce the issue and this is the fix we ended up agreeing on :)

PR #466, opened 2016-09-14
PLANNER-536 Workbench solver editor: bendable score combobox selection should allow filling in hard and soft level size

mcimbora:
Depends on https://github.com/droolsjbpm/optaplanner/pull/220.

The PR introduces concept of custom methods. Apart from generated methods (getters/setters/equals etc.) we might need to define custom methods in a data object required by upstream project (optaplanner engine in this case). For this purpose `@CustomMethod` annotation was added (method-level), which tells `JavaRoasterModelDriver` to load the method when transforming source to the data object.

Other changes
- Add MethodFilter
- Rename SourceFilter to JavaTypeFilter
- Add CommandBuilder methods to allow clients to manipulate with methods in a data object

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-api/src/main/java/org/kie/workbench/common/screens/datamodeller/service/DataModelerService.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-api/src/main/java/org/kie/workbench/common/screens/datamodeller/service/DataModelerService.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import org.guvnor.common.services.shared.metadata.model.Metadata;
 import org.guvnor.common.services.shared.validation.model.ValidationMessage;
 import org.jboss.errai.bus.server.annotations.Remote;
+import org.kie.workbench.common.screens.datamodeller.model.DataModelerError;
 import org.kie.workbench.common.screens.datamodeller.model.EditorModelContent;
 import org.kie.workbench.common.screens.datamodeller.model.GenerationResult;
 import org.kie.workbench.common.screens.datamodeller.model.TypeInfoResult;
@@ -38,6 +39,7 @@ import org.kie.workbench.common.services.datamodeller.driver.model.AnnotationSou
 import org.kie.workbench.common.services.datamodeller.driver.model.AnnotationSourceResponse;
 import org.kie.workbench.common.services.shared.project.KieProject;
 import org.uberfire.backend.vfs.Path;
+import org.uberfire.commons.data.Pair;
 
 @Remote
 public interface DataModelerService {
@@ -75,13 +77,15 @@ public interface DataModelerService {
             final boolean saveCurrentChanges, final String source, final DataObject dataObject,
             final Metadata metadata );
 
-    void delete( final Path path, final DataObject dataObject, final String comment );
+    void delete( final Path path, final String comment );
 
     GenerationResult refactorClass( final Path path, final String newPackageName, final String newClassName );
 
     List<ValidationMessage> validate( String source, final Path path, DataObject dataObject );
 
     TypeInfoResult loadJavaTypeInfo( final String source);
+
+    GenerationResult loadDataObject( final Path projectPath, final String source, final Path sourcePath );
 
     List<PropertyType> getBasePropertyTypes();
 
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
@@ -17,16 +17,11 @@
 package org.kie.workbench.common.screens.datamodeller.backend.server;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.StringTokenizer;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
-import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
 import javax.enterprise.inject.Any;
@@ -59,7 +54,7 @@ import org.kie.workbench.common.screens.datamodeller.backend.server.file.DataMod
 import org.kie.workbench.common.screens.datamodeller.backend.server.handler.DomainHandler;
 import org.kie.workbench.common.screens.datamodeller.events.DataObjectCreatedEvent;
 import org.kie.workbench.common.screens.datamodeller.events.DataObjectDeletedEvent;
-import org.kie.workbench.common.screens.datamodeller.events.DataObjectSavedEvent;
+import org.kie.workbench.common.screens.datamodeller.events.DataObjectRenamedEvent;
 import org.kie.workbench.common.screens.datamodeller.model.DataModelerError;
 import org.kie.workbench.common.screens.datamodeller.model.EditorModelContent;
 import org.kie.workbench.common.screens.datamodeller.model.GenerationResult;
@@ -79,9 +74,9 @@ import org.kie.workbench.common.services.datamodeller.core.ElementType;
 import org.kie.workbench.common.services.datamodeller.core.PropertyType;
 import org.kie.workbench.common.services.datamodeller.core.impl.DataObjectImpl;
 import org.kie.workbench.common.services.datamodeller.core.impl.PropertyTypeFactoryImpl;
+import org.kie.workbench.common.services.datamodeller.driver.FilterHolder;
 import org.kie.workbench.common.services.datamodeller.driver.ModelDriver;
 import org.kie.workbench.common.services.datamodeller.driver.ModelDriverException;
-import org.kie.workbench.common.services.datamodeller.driver.SourceFilter;
 import org.kie.workbench.common.services.datamodeller.driver.impl.JavaRoasterModelDriver;
 import org.kie.workbench.common.services.datamodeller.driver.impl.ProjectDataModelOracleUtils;
 import org.kie.workbench.common.services.datamodeller.driver.impl.UpdateInfo;
@@ -146,7 +141,7 @@ public class DataModelerServiceImpl
     private Event<DataObjectDeletedEvent> dataObjectDeletedEvent;
 
     @Inject
-    private Event<DataObjectSavedEvent> dataObjectSavedEvent;
+    private Event<DataObjectRenamedEvent> dataObjectRenamedEvent;
 
     @Inject
     private RefactoringQueryService queryService;
@@ -177,25 +172,11 @@ public class DataModelerServiceImpl
     private Instance<DomainHandler> domainHandlers;
 
     @Inject
-    @Any
-    private Instance<SourceFilter> filtersInstance;
-
-    private Collection<SourceFilter> filters;
+    private FilterHolder filterHolder;
 
     private static final String DEFAULT_COMMIT_MESSAGE = "Data modeller generated action.";
 
     public DataModelerServiceImpl() {
-    }
-
-    @PostConstruct
-    private void setup() {
-        filters = StreamSupport.stream( filtersInstance.spliterator(), false ).collect( Collectors.toList() );
-    }
-
-    @PreDestroy
-    private void tearDown() {
-        filters.forEach( filter -> filtersInstance.destroy( filter ) );
-        filters.clear();
     }
 
     @Override
@@ -296,12 +277,6 @@ public class DataModelerServiceImpl
             editorModelContent.setCurrentProjectPackages( serviceHelper.resolvePackages( project ) );
             editorModelContent.setDataModel( resultPair.getK1() );
             editorModelContent.setDataObject( resultPair.getK1().getDataObject( className ) );
-
-            Iterator<DomainHandler> it = domainHandlers != null ? domainHandlers.iterator() : null;
-            while ( it != null && it.hasNext() ) {
-                it.next().processDataObject( resultPair.getK1().getDataObject( className ), resultPair.getK1() );
-            }
-
             editorModelContent.setDataObjectPaths( resultPair.getK2().getClassPaths() );
 
             editorModelContent.setOriginalClassName( className );
@@ -356,9 +331,8 @@ public class DataModelerServiceImpl
 
             ModelDriver modelDriver = new JavaRoasterModelDriver( ioService,
                                                                   Paths.convert( defaultPackage.getPackageMainSrcPath() ),
-                                                                  true,
                                                                   classLoader,
-                                                                  filters );
+                                                                  filterHolder );
             ModelDriverResult result = modelDriver.loadModel();
             dataModel = result.getDataModel();
 
@@ -400,9 +374,9 @@ public class DataModelerServiceImpl
         }
     }
 
-    private Pair<DataObject, List<DataModelerError>> loadDataObject( final Path projectPath,
-                                                                     final String source,
-                                                                     final Path sourcePath ) {
+    public GenerationResult loadDataObject( final Path projectPath,
+                                            final String source,
+                                            final Path sourcePath ) {
 
         if ( logger.isDebugEnabled() ) {
             logger.debug( "Loading data object from projectPath: " + projectPath.toURI() );
@@ -415,20 +389,20 @@ public class DataModelerServiceImpl
 
             project = projectService.resolveProject( projectPath );
             if ( project == null ) {
-                return new Pair<DataObject, List<DataModelerError>>( null, new ArrayList<DataModelerError>() );
+                return new GenerationResult( null, null, new ArrayList<DataModelerError>() );
             }
 
             ClassLoader classLoader = classLoaderHelper.getProjectClassLoader( project );
-            JavaRoasterModelDriver modelDriver = new JavaRoasterModelDriver( ioService, null, false, classLoader, filters );
+            JavaRoasterModelDriver modelDriver = new JavaRoasterModelDriver( ioService, null, classLoader, filterHolder );
             ModelDriverResult driverResult = modelDriver.loadDataObject( source, Paths.convert( sourcePath ) );
 
             if ( !driverResult.hasErrors() ) {
                 if ( driverResult.getDataModel().getDataObjects().size() > 0 ) {
                     dataObject = driverResult.getDataModel().getDataObjects().iterator().next();
                 }
-                return new Pair<DataObject, List<DataModelerError>>( dataObject, new ArrayList<DataModelerError>() );
+                return new GenerationResult( source, dataObject, new ArrayList<DataModelerError>() );
             } else {
-                return new Pair<DataObject, List<DataModelerError>>( null, serviceHelper.toDataModelerError( driverResult.getErrors() ) );
+                return new GenerationResult( source, null, serviceHelper.toDataModelerError( driverResult.getErrors() ) );
             }
 
         } catch ( Exception e ) {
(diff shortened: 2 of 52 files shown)


mcimbora:
@wmedvede Walter, would you mind having a look at this if you have some time? Thanks!

mcimbora:
Looking at the failures.

mcimbora:
Jenkins retest this

wmedvede:
Hello @mcimbora , I think it's ok. 
But just one thing came up to my mind now that I was reviewing the code (sorry that I didn't note it when we chatted). Instead of using the CustomMethod annotation, I think it'll be better to use the javax.annotation.Generated annotation instead.

 Basically because:
1) In the end the @CustomMethod  is just a marker annotation to identify that a method that was generated by us. 
2) if we deliver code with the CustomMethod annotation, then we may add problems to the people that wants to compile the generated code outside the WB.
3) as soon the user changes the optaplanner parameters you will always rewrite the method.

In general all the generated stuff by the data modeler could have the @Generated, or as a minimum at the class level we could add the Generated annotation. But since we had the requirement of supporting roundtrip between client modifications and our generated code, i.e., the client should be able to modify the class in the IDE, push it into the WB, and continue working, or even modify the class in the source tab, I believe it could be a bit confusing to tag everything as Generated and let the rountrip at the same time.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/MethodFilterImpl.java
```diff
@@ -0,0 +1,32 @@
...
+package org.kie.workbench.common.services.datamodeller.driver.impl;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.jboss.forge.roaster.model.Method;
+import org.kie.workbench.common.services.datamodeller.core.CustomMethod;
+import org.kie.workbench.common.services.datamodeller.driver.MethodFilter;
+
+@ApplicationScoped
+public class MethodFilterImpl implements MethodFilter {
```
wmedvede: @mcimbora I think It'll be better to provide the MethodFilter implementation you need in the optaplanner-wb, and we'll try to load only load the method you are interested in.

mcimbora:
Jenkins retest this

mcimbora:
@wmedvede Thanks for the comments, the PR has been updated. I moved the method filtering implementation to optaplanner-wb (https://github.com/droolsjbpm/optaplanner-wb/pull/106) so that other clients can provide their own filtering scheme. I'll send livespark PR once this is integrated.

wmedvede:
Hello @mcimbora, could you please take a look at my comments in the already merged https://github.com/droolsjbpm/kie-wb-common/pull/422. 
Thanks.

csadilek:
@mcimbora @wmedvede did you guys have a chat about this? What's the status here? ;)

mcimbora:
@csadilek Correct, we had a meeting regarding this. Quite a lot of changes were required, I'm adding unit tests ATM. I'll update the issue when ready.

mcimbora:
Jenkins retest this

mcimbora:
@wmedvede I updated the PR.
New changes overview:
- switch to using `@Generated` annotations on method & nested class level
- add filters for method & nested classes, include all filters in `FilterHolder` class
- add tests for method & nested class update
- introduce `Type` and `Parameter` classes (currently for methods only) that allow holding information about type arguments
- introduce `DataObjectException` class. Used by backend `DomainHandler` to indicate the custom processing of a data object failed. This is useful when downstream wb apps add a custom objects (e.g. comparator object in OptaPlanner WB) and there is a failure when loading the object (e.g. someone mangles auto-generated code). At this point of time only source editor should be available.

wmedvede:
Hello @mcimbora I've tried to start looking into your PR this afternoon, but I lost all afternoon due to build erros, related, dashbuilder, console, etc. I'll stat tomorrow with this PR.

mcimbora:
@wmedvede No problem, thanks!

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
```diff
@@ -297,11 +279,20 @@ protected EditorModelContent constructContent( Path path,
             editorModelContent.setDataModel( resultPair.getK1() );
             editorModelContent.setDataObject( resultPair.getK1().getDataObject( className ) );
 
+            List<DataModelerError> domainHandlerErrors = new ArrayList<>();
+
             Iterator<DomainHandler> it = domainHandlers != null ? domainHandlers.iterator() : null;
             while ( it != null && it.hasNext() ) {
-                it.next().processDataObject( resultPair.getK1().getDataObject( className ), resultPair.getK1() );
```
wmedvede: Hello @mcimbora , I'm not sure if we should do this processing here, and introduce the ComparatorObject. 
If all the information/metadata we need is modeled by using annotations, and the parsed object has the inner classes with the given annotations, then I believe we should be able to work in the UI with the metadata that we already parsed. i.e. The parent class + it's annotations, and the inner class + it's annotations.
I believe that this CompartorObject with the list of ObjectPropertyPath is basically a data structure to facilitate the UI generation, but not sure if is something we should return as part of the DataObject.
wdyt?

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/JavaRoasterModelDriver.java
```diff
@@ -123,22 +124,22 @@ public JavaRoasterModelDriver() {
...
+        this();
+        this.filterHolder = filterHolder;
     }
 
     public JavaRoasterModelDriver( IOService ioService,
                                    Path javaRootPath,
-                                   boolean recursiveScan,
                                    ClassLoader classLoader,
-                                   Collection<SourceFilter> filters ) {
+                                   FilterHolder filterHolder ) {
```
wmedvede: I believe that to have an @ApplicationScoped FiterHolder to keep centralized the CDI collection of all filters is good, but probably we can define an interface FilterHolder and then let a FilterHolderImpl be an @ApplicationScoped implementation. In this way we can keep the JavaRoasterModerDriver independent from CDI components.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
wmedvede: Related to this, I believe that to use an annotation to define the fieldPaths items might also help the parsing.

Currently we have this.
@ComparatorDefinition(fieldPaths = { "java.math.BigInteger:name=asc" })

but probably it's worth to explore this alternative. (just a suggestion)

@ComparatorDefinition( fieldPaths = { @ComparatorField( type="java.math.BigInteger", name="field1", order = "asc" ),  @ComparatorField( type="java.lang.String", name = "field2", order ="desc" )

wmedvede:
Hello @mcimbora a quick comment regarding the DomainHandler.postEventProcessing method.
I don't remember if the  was introduced in a previous PR, and don't know all the execution flows that you need to react on. But I belive that if the DomainHandler.postCommandProcessing( DataModelCommand command ); method is enough for your use case, we should us it instead of the new one. 
¿?

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
mcimbora: @wmedvede I think you're right. The idea was to parse the object property paths defined in `@ComparatorDefinition` annotation only once and encapsulate the result in a separate object. Note the comparator object uses `ObjectProperty` instances internally, i.e. it additionally loads the object properties from the data model. I like the idea of having specialized `@ComparatorField` annotations which would avoid such parsing. The data model can be accessed from object editor, so the object properties could be loaded on demand from there.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/JavaRoasterModelDriver.java
mcimbora: +1 this makes sense.

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
mcimbora: +1 good idea. I think we just need a bit more granularity here as nested object fields can be used. e.g.:

`@ComparatorDefinition:`
+ ComparatorFieldPath[]


`@ComparatorFieldPath:`
+ ComparatorField[]
+ boolean ascending


`@ComparatorField:`
+ String name
- String type

wmedvede: Yes, I think it'll be good to try this strategy. In the end (at least with current requirements) what we do in the data modeler is generate the java class + some methods + and finally add metadata to the generated class by using annotations. 
I think the code you already have for building the ComparatorObject, etc, can be refactored to mange annotations, but  remains very similar.
In some places, when I need to manage a complex annotation with a lot of parameters, etc, I found useful to create a Hander that encapsulates the annotation management (addding, removing value paris, etc) and in the UI I've just call the Handler when I want to manage the annotation value pairs, etc. Just a suggestion that could be good for you to use at some point, e.g., for facilitating adding/removing  the ComparatorFields to the ComparatorAnnotation, etc.
Here's a couple of examples that might worth to take a look.

https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/handlers/jpadomain/util/SequenceGeneratorValueHandler.java

https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/handlers/jpadomain/util/RelationshipAnnotationValueHandler.java

wmedvede: I'm not sure if we need this extra @ComparatorFieldPath, since in the end what you need to know is wich fields should be used for the comparator. BUT, I don't  know all the planner internals, so it's just a suggestion.

mcimbora:
Regarding `postEventProcessing` method, it was introduced because not all changes that might affect the comparator object are done via commands, e.g. deleting object property. I'll have a look if I can refactor the code to use the commands as well if it's worth the complexity introduced.

mcimbora:
Hi @wmedvede I updated the PR together with https://github.com/droolsjbpm/optaplanner-wb/pull/106. If needed, we can set up a meeting to guide you through the changes, wdyt?

wmedvede:
Thanks Matej, I'll go through the code this afternoon and let you know.

Regards,
Walter.

2016-10-12 19:33 GMT+02:00 Matej Čimbora <notifications@github.com>:

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-api/src/main/java/org/kie/workbench/common/screens/datamodeller/service/DataModelerService.java
```diff
@@ -52,6 +52,8 @@
 
     DataModel loadModel(final KieProject project);
 
+    DataObject getDataObject( final String typeName, final Path path );
```
wmedvede: Hello @mcimbora , I could find that this method is used in the PlannerDomainHander in optaplanner-wb. Please see my comments related to the saving of the data object that is being done there. If this method is no longer needed I believe we should remove it.

wmedvede:
Hello @mcimbora, for some reason the PRs are not always properly updated... If I go the the "changed" files list in the PR, the DataObjectBrowser.java is not there any more, but it's modified in your branch, and also in the local copy I've got from your branch. I had this annoying behavior in some of my PRs too.
Anyway, 
If you look at here:

https://github.com/mcimbora/kie-wb-common/blob/PLANNER-536/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/widgets/editor/DataObjectBrowser.java#L405

I'll appreciate to not add this executePostEventProcessing( DataModelerEvent event ) here, since the data object browser is a component to just show the list of properties and eventually add/remove properties. And at the moment we also don't had the need to force all the DomainHandlers to implement a postEventProcessing method.

I believe that if you need to react upon the DataModelerEvent then you can add just add an @Observer at the PlannerDomainHandler and just do the same that you are doing in the executePostEventProcessing.

The PlannerDomainHandler could look something like this:

void postEventProcessing( @Observes DataModelerEvent event) {
}

And you can just leave basically the same programming you already have there. (eventually doing some filtering if you are interested in some particular changes, etc.)

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/JavaRoasterModelDriver.java
```diff
@@ -418,26 +421,45 @@ public DataModel createModel() {
...
-                                nestedJavaClass.addInterface( interfaceDefinition );
+            if ( nestedTypes != null ) {
+                for ( JavaSource nestedType : nestedTypes ) {
+                    if ( isAccepted( nestedType ) ) {
+                        if ( nestedType instanceof JavaClassSource ) {
+                            JavaClassImpl nestedJavaClass = new JavaClassImpl( "", nestedType.getName(), DriverUtils.buildVisibility( nestedType.getVisibility() ) );
+                            dataObject.addNestedClass( nestedJavaClass );
+                            if ( javaClassSource.getInterfaces() != null ) {
+                                for ( String interfaceDefinition : ( (JavaClassSource) nestedType ).getInterfaces() ) {
+                                    if ( !"java.io.Serializable".equals( interfaceDefinition ) ) {
```
wmedvede: Hello @mcimbora I believe that if we can avoid this harcoded filtering here and let some upper level component just do nothing or not process the Serializable interface it can be good.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/JavaRoasterModelDriver.java
```diff
@@ -854,6 +922,101 @@ private void addAnnotationArrayMemberValue( AnnotationSource annotationSource, A
...
+        }
+    }
+
+    private void addMethod( JavaClassSource javaClassSource, org.kie.workbench.common.services.datamodeller.core.Method method, ClassTypeResolver classTypeResolver ) throws ClassNotFoundException {
+        MethodSource<JavaClassSource> methodSource = javaClassSource.addMethod();
+        methodSource.setName( method.getName() );
+        methodSource.setReturnType( buildMethodReturnTypeString( method.getReturnType(), classTypeResolver ) );
+        methodSource.setParameters( buildMethodParameterString( method.getParameters(), classTypeResolver ) );
+        methodSource.setBody( method.getBody() );
+        methodSource.setVisibility( org.jboss.forge.roaster.model.Visibility.PUBLIC );
```
wmedvede: Likely the Method may also implement HasVisibility and we can take it from there

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/util/NamingUtils.java
```diff
@@ -152,6 +152,41 @@ public static String createQualifiedName( String packageName, String className )
         }
     }
 
+    public static String createJavaGetter( String name, String className ) {
```
wmedvede: I don't understand why this methods for generating code were moved from the generation class  to here

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
```diff
@@ -529,12 +530,20 @@ public void callback( final Path response ) {
...
     }
 
-    private RemoteCallback<Path> getRenameSuccessCallback( final RenamePopUpPresenter.View renamePopupView ) {
+    private RemoteCallback<Path> getRenameSuccessCallback( final RenamePopUpPresenter.View renamePopupView, String packageName, String newFileName ) {
         return new RemoteCallback<Path>() {
             @Override
             public void callback( final Path targetPath ) {
                 renamePopupView.hide();
                 view.hideBusyIndicator();
+                dataModelerEvent.fire( new DataObjectChangeEvent()
```
wmedvede: I believe that we need to rise this event upon rename, we should try to do it in the servcie method instead

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
```diff
@@ -887,7 +896,7 @@ private void rename( final boolean saveCurrentChanges ) {
                                        public void execute( final FileNameAndCommitMessage details ) {
                                            view.showBusyIndicator( org.kie.workbench.common.widgets.client.resources.i18n.CommonConstants.INSTANCE.Renaming() );
 
-                                           modelerService.call( getRenameSuccessCallback( renamePopUpPresenter.getView() ),
+                                           modelerService.call( getRenameSuccessCallback( renamePopUpPresenter.getView(), context.getDataObject().getPackageName(), details.getNewFileName() ),
```
wmedvede: This invocation context.getDataObject().getPackageName() introduces a NPE in cases when a file that wasn't parsed properly is being edited.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/util/NamingUtils.java
mcimbora: Body of the `compare` method is generated on the client, based on user selection https://github.com/droolsjbpm/optaplanner-wb/pull/106/files#diff-a201a393e09971228568a8a6dee7eef9R144. I wanted to reuse the code which already exists, however exposing GenerationTools as a part of gwt module seemed to me as an overkill.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/driver/impl/JavaRoasterModelDriver.java
mcimbora: @wmedvede +1

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
mcimbora: @wmedvede Makes sense. I'll introduce new `DataObjectRenamed` event for this purpose.

mcimbora:
@wmedvede I updated the PR with the latest suggestions.

wmedvede:
Hello @mcimbora  @csadilek , +1 from my side.

csadilek:
Nice work @mcimbora and thanks for the thorough review @wmedvede.

+1

PR #469, opened 2016-09-14
GUVNOR-2636: [Decision Table Spreadsheet] uploading new version causes error

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2636


manstis:
I don't believe we need concern ourselves with archiving from "Project Explorer" as this simply allows the User to download a ZIP file for a package in the project. 

I would be more concerned about updating [this](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-default-editor/kie-wb-common-default-editor-client/src/main/java/org/kie/workbench/common/screens/defaulteditor/client/editor/NewFileUploader.java) and (unfortunately) [this](org.uberfire.ext.widgets.core.client.editors.defaulteditor) and (equally unfortunately) [this](https://github.com/uberfire/uberfire/blob/master/uberfire-server/src/main/java/org/uberfire/server/BaseUploadServlet.java)

This last part is arguably out of the scope of the original JIRA.. so my advice would be to close this PR (as we don't really need the changes to "Project Editor") and raise a new JIRA for a comparable fix for "(Generic) File Upload". WDYT?

karreiro:
Creating a new JIRA for that sounds like a good idea for me too. Thanks, @manstis :-)

karreiro:
Jenkins, please retest this.

PR #477, opened 2016-09-21
RHBRMS-265 - Kie-Server: No UI feedback after changing container version

mswiderski:
@porcelli could you please have a look?


porcelli:
+1

PR #482, opened 2016-09-23
GUVNOR-2682: Invalid DRL file is successfully validated after reloading of Business Central

manstis:
See https://issues.jboss.org/browse/GUVNOR-2682

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorBuildService.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorBuildService.java
@@ -124,6 +124,11 @@ public class ValidatorBuildService {
                 builder.build();
             }
             final Builder clone = builder.clone();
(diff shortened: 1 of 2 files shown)


manstis:
@karreiro Hi, would you mind taking a look please.

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorBuildService.java
```diff
@@ -124,6 +124,11 @@ public ValidatorBuildService( final @Named("ioStrategy") IOService ioService,
                 builder.build();
```
karreiro: Instead of removing the resource. What do you think about doing something like this?
```
resultBuilder.add( builder.build().getMessages() );
```

karreiro: Ohhh! Right, thanks for the explanation :-)

PR #483, opened 2016-09-23
GUVNOR-2453: Project Editor: <artifactId>:<groupId>:<version> should be written <groupId>:<artifactId>:<version>

manstis:
See https://issues.jboss.org/browse/GUVNOR-2453


csadilek:
+1

PR #484, opened 2016-09-23
GUVNOR-2702: add more flexibility to name a containerId

porcelli:


manstis:
Let's wait for Jenkins Gods to turn green before merge...

PR #487, opened 2016-09-23
GUVNOR-2703: adding user's feedback on exec server operations.

porcelli:
Depends on #477


porcelli:
@mswiderski please review ;)

mswiderski:
tested locally and I see fly over in case there is an error on kie server so it looks good to me.

One thing (not directly related) the notification popup on container card does not work for me... though I couldn't figure out what might caused it.

so +1 for this change

porcelli:
jenkins please retest this

porcelli:
jenkins please retest this

PR #488, opened 2016-09-27
GUVNOR-2707: Guided Decision Table Editor: Overview Page throws an exception unless a Version is explicitly selected

manstis:
See https://issues.jboss.org/browse/GUVNOR-2707

Retrieve the version for the "Overview Screen" from the ```VersionRecordManager``` instead of from the active "Document". This is how regular ```KieEditors``` pass the Version to the "Overview Screen".


csadilek:
+1

PR #506, opened 2016-10-12
Stunner - Review of git ignores.

romartin:
Review of git ignores in stunner folders - feedback from @manstis . Thanks!


manstis:
@romartin Hi, thanks. Merged "as is" however you may also need to include an additional ignore in ```kie-wb-common``` root [pom](https://github.com/droolsjbpm/kie-wb-common/blob/master/.gitignore#L22) for Zanata related files (when this is first pushed to/pulled from Zanata for i18n). Please feel free to submit as another PR :-)

PR #507, opened 2016-10-13
server-ui-backend: direct dependency on kie-remote-common is not needed

psiroky:
* the code does not import any classes from there and it is being
   brought in transitively via kie-server-client


mswiderski:
+1 to merge once build completes

PR #518, opened 2016-10-21
RHBPMS-4373: Singleton runtime strategy is always shown in container configuration

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4373

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/widget/config/process/ProcessConfigPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/widget/config/process/ProcessConfigPresenter.java
@@ -102,7 +102,7 @@ public class ProcessConfigPresenter {
     }
 
     private void setupView( final ProcessConfig processConfig ) {
-        this.view.setContent( ClientRuntimeStrategy.convert( processConfig.getRuntimeStrategy(), view.getTranslationService() ).getValue( view.getTranslationService() ),
+        this.view.setContent( ClientRuntimeStrategy.valueOf( processConfig.getRuntimeStrategy() ).getValue( view.getTranslationService() ),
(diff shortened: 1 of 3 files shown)


mswiderski:
@karreiro this does not seem to work as expected in case there are many containers in single server - looks like the change is applied to all containers and can't be changed per container.

karreiro:
@mswiderski Hi Maciej :-) I'm not sure if I misunderstood, but I think it *is* working, see:
a

mswiderski:
can you check if you can change it after container was created? that's where I found it not working as expected.

karreiro:
@mswiderski ohh! ok :-) see:
b

mswiderski:
still happens for me, the difference is I am running in kie-wb on wildfly and not hosted mode. Though not sure if that should change anything.

Another thing I noticed is that same issue is for the other drop down merge mode - I mean original issue - value always come to KEEP ALL as default no matter which one was selected.

is this the only change I need?

mswiderski:
ok, I think I know where the issue is - it works for new one but it does not work for those that were created before this change. Could it be some backward compatibility issue?

karreiro:
Hi

I've tried to reproduce the problem with these steps:

Creating the Container without my fix:
c1

And then, trying to open the container that I created, with my fix:
c2

I could not reproduce (I'm using Wildfly). Am I doing something different?

Thanks.

mswiderski:
on top of you shown what I did was to change the container after starting workbench with the fix. And that then was changing for all and for some it then disabled the save button (it was greyed out)

karreiro:
Hi @mswiderski 

Thanks for the explanation. Unfortunately I could not reproduce yet :-(

First of all, I created the containers (using the workbench *without the fix*):
c1

After that, I updated the containers (using the workbench *with the fix*):
c2

Am I missing something?

Thank you.

mswiderski:
@karreiro in that case it must be something on my end that causes this weird behavior. Since you tried all what I did and couldn't reproduce I think we should merge this to have it available for wider audience (QE too).

Thanks a lot for all the tests and sorry for taking lots of your time for this. 

+1 from me

karreiro:
@mswiderski Thank you for reporting, Maciej. :-)

PR #523, opened 2016-10-25
GUVNOR-2728 : Dependency validation issues after upgrading from BxMS 6.1 to 6.3.3

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2728


Rikkola:
Whoap. One small bug to fix.

PR #530, opened 2016-10-28
GUVNOR-2729 : [Test Scenario] Run of a testscenario causes duplication of available Knowledge Bases

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2729


karreiro:
+1

PR #539, opened 2016-11-03
Ignore Stunner Showcase webapp

manstis:
Stunner Showcase causes a dependency of kie-wb-common on drools-wb, which is wrong.

See BSIG mailing list for details.


krisv:
+1

PR #540, opened 2016-11-03
PLANNER-686: Workbench: Move Planner-specific code from kie-wb-common to optaplanner-wb

mcimbora:
Move PermissionTreeSetup & WorkbenchTreeProvider classes to kie-wb, as they directly reference downstream content (Planner).


mcimbora:
Part of:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/309
https://github.com/droolsjbpm/kie-wb-common/pull/540
https://github.com/droolsjbpm/optaplanner-wb/pull/118
https://github.com/droolsjbpm/kie-wb-distributions/pull/396

manstis:
@wmedvede @ederign IIRC you two probably know the most about "Docks".. could you please review?

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/PermissionTreeSetup.java
```diff
@@ -1,100 +0,0 @@
...
-        this.workbenchTreeProvider = workbenchTreeProvider;
-        this.perspectiveTreeProvider = perspectiveTreeProvider;
-        this.orgUnitTreeProvider = orgUnitTreeProvider;
-        this.repositoryTreeProvider = repositoryTreeProvider;
-        this.projectTreeProvider = projectTreeProvider;
-
-    }
-
-    public void configureTree() {
-        perspectiveTreeProvider.setPerspectiveName(HOME, i18n.HomePage());
```
wmedvede: Hello @mcimbora I don't see why this code is specific to the planner-wb. If this configurations are removed
e.g.  perspectiveTreeProvider.setPerspectiveName(HOME, i18n.HomePage());
when the user goes into the security configurations and uses the comobobox that lists the available perspectives, the user will see the perspective id instead of the i18n label.
I'll appreciate if @dgutierr  can confirm if this can simply be removed and all the security setting continues working well.

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/PermissionTreeSetup.java
```diff
@@ -1,100 +0,0 @@
...
-        perspectiveTreeProvider.setPerspectiveName(DATASET_TASKS, i18n.Tasks());
-        perspectiveTreeProvider.setPerspectiveName(PROCESS_DASHBOARD, i18n.Process_Dashboard());
-
-        // Exclude some perspectives
-        perspectiveTreeProvider.excludePerspectiveId("AuthoringPerspectiveNoContext"); /* kie-wb-distributions */
-        perspectiveTreeProvider.excludePerspectiveId("FormDisplayPerspective"); /* jbpm-console-ng */
-        perspectiveTreeProvider.excludePerspectiveId("Drools Tasks"); /* jbpm-console-ng */
-        perspectiveTreeProvider.excludePerspectiveId("Experimental Paging"); /* jbpm-console-ng */
-        perspectiveTreeProvider.excludePerspectiveId("StandaloneEditorPerspective"); /* uberfire */
-        perspectiveTreeProvider.excludePerspectiveId("WiresTreesPerspective"); /* uberfire */
```
wmedvede: Hello @mcimbora , same comments here. I don't know if this can be removed. I'll appreciate if @dgutierr  can confirm if this can be removed from the WB.

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/WorkbenchTreeProvider.java
```diff
@@ -1,104 +0,0 @@
...
-    }
-
-    @Override
-    public void loadChildren(PermissionNode parent, LoadOptions options, LoadCallback callback) {
-
-        if (parent.propertyEquals(NODE_TYPE, NODE_ROOT)) {
-            List<PermissionNode> result = new ArrayList<>();
-
-            PermissionLeafNode node1 = createPermissionLeafNode(RepositoryFeatures.CONFIGURE_REPOSITORY, i18n.ConfigureRepositories(), i18n.ConfigureRepositoriesHelp());
-            PermissionLeafNode node2 = createPermissionLeafNode(EDIT_SOURCES, i18n.DataModelerEditSources(), i18n.DataModelerEditSourcesHelp());
```
wmedvede: @mcimbora This permision "EDIT_SOURCES" is required by the data modeller editor. If removed the "Source" edition tab will fail.

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/PermissionTreeSetup.java
mcimbora: Hi @wmedvede, I was more concerned about `perspectiveTreeProvider.setPerspectiveName(PLANNER_ADMIN, i18n.PlannerAdministration());`. Basically any project that depends on kie-wb-common that's listed here introduces this problem (e.g. DROOLS_ADMIN). IMO kie-wb-commom shouldn't have any knowledge of its downstream projects.

As to the i18n problem I moved the messages to kie-wb-distributions, see https://github.com/droolsjbpm/kie-wb-distributions/pull/396/files#diff-9b47ca212829c50d06e14bd7eeb7bd4aR63 & https://github.com/droolsjbpm/kie-wb-distributions/pull/396/files#diff-900331082308c1bf3a53a62f18f082e7R163

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/docks/AuthoringWorkbenchDocks.java
```diff
@@ -76,18 +73,6 @@ public AuthoringWorkbenchDocks( final UberfireDocks uberfireDocks,
     public void perspectiveChangeEvent( @Observes UberfireDockReadyEvent dockReadyEvent ) {
         currentPerspectiveIdentifier = dockReadyEvent.getCurrentPerspective();
         if ( authoringPerspectiveIdentifier != null && dockReadyEvent.getCurrentPerspective().equals( authoringPerspectiveIdentifier ) ) {
-            if ( authorizationManager.authorize( WorkbenchFeatures.PLANNER_AVAILABLE, sessionInfo.getIdentity() ) ) {
```
wmedvede: Hello @mcimbora the data modeler related docks, i.e. drools-domian, jpa-domain, advanced-domian, and planner-domian are enabled/disabled if:
1) at least one .java file is open and the current selected editor is the data modeler. (if a java file is open, but the current selected editor is the drl editor, then the docks are removed from the docks bar)
2) the "editor" tab is selected for the current selected .java file. (note that if the user clicks on the source tab, then the docks disappears)
3) for the particular case of the planner dock an extra condition apply, the dock is available only when the feature PLANNER_AVAILABLE is enabled.

I'm afraid that the code you moved to the planner-wb only takes into account 3). So I believe it won't work well according to 1 and 2. The planner dock is being managed here because we need to ensure 1,2, and 3. Whatever refactoring you do must ensure that the wb continues working with the same semantic.

mcimbora: Thanks for the explaination @wmedvede! I'll check the behavior & fix the code if necessary.

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/PermissionTreeSetup.java
wmedvede: Hello @mcimbora , sorry I haven't seen that you moved this permissions setup to kie-wb-distributions.
I believe that I can make sense, but I see that introduces a duplication of the Permissions setup both in kie-wb, and kie-drools-wb. I'm not saying this is wrong, but I'd prefer that @manstis and @dgutierr takes a look at it.

dgutierr: @mcimbora The reason to go for kie-wb-common was to avoid duplication of shared code as @wmedvede suggested as well to keep this initialization code in one single place. Yet, I agree some of this code is related to downstream projects so the decision was a trade-off. I'd vote for leave it as that, wdyt?

mcimbora: @dgutierr @wmedvede @manstis I agree that some duplication has been introduced here, yet I think this is a good opportunity to separate kie-wb-common & optaplanner-wb once for all. I see kie-wb & kie-drools-wb already contain a lot of duplicated code. However if we don't want to make things any worse in this regard I propose an alternative approach:

There are 2 classes that reference planner (linked below). The initialization code could stay where it is right now, however we can extract creation of docks/nodes into provider classes. Additionally this would avoid issues based on data modeller selection Walter referenced earlier in the review.
 
https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/WorkbenchTreeProvider.java#L85
https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/docks/AuthoringWorkbenchDocks.java#L102

At the end we would have WorkbenchAuthoringDockProvider & WorkbenchPermissionLeafNodeProvider classes in both kie-wb-common & optaplanner-wb. The classes linked above would then leverage `Instance<*Provider>` to initialize dock/permission node lists.

Wdyt?

mcimbora: Hmm this is indeed not so easy, https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/docks/AuthoringWorkbenchDocks.java#L76 needs to be checked with each `UberfireDockReadyEvent` to see if we've got the planner permission...

mcimbora:
Closing as this blocks multiple smaller issues that will be addressed separately. Will reopen once a better solution is found.

PR #541, opened 2016-11-03
JBPM-5395 - Stunner - get list of ruleflow groups from server

tsurdilo:


manstis:
My biggest gripe is that this appears to be a copy and paste of a ```NamedQuery``` from jbpm-designer.

tsurdilo:
@manstis it is the same query used in jbpm-designer. both do the same thing. if there is a better way to do this let me know. thanks.

tsurdilo:
@manstis i have refactored the query to make it a lot less jbpm-designer'ish but currently cannot run the kie-wb war on wildfly10 to test. as soon as i'm able to i will amend the pr. thanks

tsurdilo:
@manstis refactored the query - let me know if this is better for you now. thanks.

manstis:
@tsurdilo I guess we'll have to live with the duplication for now.. I assume either you, @jlindop or @romartin are to address such duplication when Stunner's BPMN specifics move to Designer (or jbpm-wb). There are however a few other minor issues (relating to license headers) remaining.

tsurdilo:
@manstis ok looking at license headers now. Both designers share properties for nodes that do the same thing. The queries used in jbpm-designer were using the kie api already so I'm not sure if its needed to re-implement this logic if it already worked. The editors do not share the property panel impl so I guess some duplication will exist for a while.

tsurdilo:
@manstis  added license header to new test class and updated formatting of others

kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/main/java/org/kie/workbench/common/services/refactoring/backend/server/query/standard/FindRuleFlowNamesQuery.java
```diff
@@ -0,0 +1,177 @@
...
+import org.uberfire.paging.PageResponse;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.net.URI;
+import java.util.*;
+
+@ApplicationScoped
+public class FindRuleFlowNamesQuery extends AbstractFindQuery implements NamedQuery {
```
tsurdilo: yes it has because it also has a ruleflow-group property for the business rule task. It's query however is within the jbpm-designer code and stunner does not have dependencies to jbpm-designer which i dont think it should so I created a query in kie-wb-commons which then can be reused by other components if needed

tsurdilo: I'm not sure why you keep repeating the copy/paste thing. i have changed the query to not include the jbpm-designer-specific string concatenation and use a map pagerow instead so they are not the same but even if they were I don't see how that would be bad or something :). the queries are reusable components i guess and the query in jbpm-designer was in its codebase (should have created it in kie-wb tobegin with i guess).

PR #542, opened 2016-11-03
JBPM-5394: Removing wrong dependencies to drools-wb assets.

romartin:
Hi @csadilek @manstis @Rikkola ,
Here is the fix for the showcase modules. In fact the dependency to the drools-wb was not necessry, it was my mistake, sorry guys. I have done a complete review for the dependencies, fixed some and hope no more issues due to this will appear... sorry for the inconveniences again.
See https://issues.jboss.org/browse/JBPM-5394
Thanks!


Rikkola:
@romartin Not a problem. We used to have this kind of issues all the time and they would only get picked up when we started releasing. Now it looks like the automated builds protect us.

jlindop:
+1

PR #544, opened 2016-11-08
KIE forms jBPM integration:

pefernan:
- Cleaned up unnecessary dependency


csadilek:
+1 (this should also fix the build failure on master.)

PR #545, opened 2016-11-08
GUVNOR-2739: Guided Decision Table Graph Editor: Add ability to create new Decision Tables from within editor

manstis:
See https://issues.jboss.org/browse/GUVNOR-2739

This adds a "New" button to the "Documents" menu. 

Buttons are disabled until a command is set for the button; whence it becomes enabled.


manstis:
Part of an ensemble:

- https://github.com/droolsjbpm/kie-wb-common/pull/545
- https://github.com/droolsjbpm/drools-wb/pull/333

csadilek:
+1

PR #546, opened 2016-11-08
JBPM-5401 - Stunner - Remove background settings and shape dimensions categories for all nodes

tsurdilo:
@jlindop commenting rather than removing as we might want to add this back for future milestone


jlindop:
Tested and looks good.
+1

tsurdilo:
will redo

PR #547, opened 2016-11-09
JBPM-5391: Use of Iterable instead Collection for the command API and related components. Fixed some bugs as well.

romartin:
Hey @csadilek , 
As commented before, here are some fixes for the commands stuff. Finally I replaced the use of Collections everywhere, that in fact should be Lists in order to guarantee order, by the use of Iterables on the different public methods, I think it's the most generic we can do. 
So each implementation ( whatever it's a command registry or it's a command mamanger ), have to guarantee that the iteration order for the commands is the expected one ( internally they can use whatever implmentation to handle the collection such as list stacks etc, only need to guarantee order on public api - I have added javadocs about this as well ). 
Please can you verify this and lemme know your thoughts? Thanks in adavance!!


romartin:
Closing this PR as this raised an interesting refactor missing in commands. Will provide another PR for this fix as well soon. Thanks!

PR #550, opened 2016-11-09
JBPM-5402 - Stunner - rename General Settings to General - and moving…

tsurdilo:
… fielddef label to constants

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-basicset/kie-wb-common-stunner-basicset-api/src/main/java/org/kie/workbench/common/stunner/basicset/definition/BasicConnector.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-basicset/kie-wb-common-stunner-basicset-api/src/main/java/org/kie/workbench/common/stunner/basicset/definition/BasicConnector.java
@@ -37,6 +37,8 @@ import javax.validation.Valid;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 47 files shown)


jlindop:
Tested and looks good. Thanks for fixing the typos in constants earlier.
+1

manstis:
Out of interest how is ```@FieldDef``` i18n'ed? I assume the "label" is used in the UI; however we cannot substitute the constants for GWT ResourceBundles and our regular i18n mechanism as (again) I assume Form Modeller needs the annotations at code-generation (compile) time and not runtime. @pefernan Can you advise?

pefernan:
@manstis at this point they aren't i18n'ed. This is the first version of the dynamic engine and it's going to be refactored as soon as I have the M1 done. If you want we can have a chat about that, I'll be happy to expalin in detail how it works and I can get some advice from you for the new version ;)

manstis:
@pefernan Hi, I was just curious; now I know at the moment they are not i18n'ed. 

I guess in the future ```@FieldDef``` may either replace ```label``` with ```label-key``` and a lookup performed at runtime; or to preserve backwards compatibility a ```label-key``` property added to the annotation in addition to ```label```. 

Assuming you use Errai UI this could be achieved with generated ```TranslationKey``` implementations based on ```label-key``` and ```TranslationService``` (or if you're using GWT then similarly generated classes).

pefernan:
@manstis yeah the idea is to use the TranslationService for this if possible. 
At the begining this feature was intended as an "experiment" and now it's being used intensively, as I said it will require a big refactor.

PR #552, opened 2016-11-10
JBPM-5391: Removing use of collections on command manager/registry

romartin:
Hey @csadilek ,
As commented [before](https://github.com/droolsjbpm/kie-wb-common/pull/547), there is no need to have collections of commands on the command manager/registry as already exist the composie command type. 
So I have removed the BatchCommandmanager/Result and related stuff and just used a generic composite command implementation (  see CompositeCommandImpl ) to achieve same behavior, sounds good.
Now the command registry uses List types as well, so we can expect the right ordering.
You'll see on the RedoCommandHandler as well that there is no need right now to compare lists of commands, as the command for each execution is only a single one ( command can potentially be composite ). As using instance equality for commands, which already had a quick talk for this, the equals on the instance guarantees the right undo/redo in both regular/composite commands.
This changes really make the api more clear and easy to use, in fact most of this PR just removed classes and tests, so that's a good indicator. Tests are passing and diagrams loading fine on the UI, I can undo/redo all commands previously executed as well, so this fixes the last bug we found too.

Please could you review it?? Really thanks again!! :)


csadilek:
+1, nice, big improvement @romartin !

PR #560, opened 2016-11-11
JBPM-5397: Support for custom diagram factories. Added BPMN Diagram Factory for Project context.

romartin:
This feature was required in order to populate some of the BPMN process properties with specific values that are given from the attributes of the guvnor project currently being used.
So adding support for custom diagram factories is the most generic way to achieve that. Each "definition set" now is able to provide custom diagram factories which can populate the diagram instance with the given context attributes.
Added the concrete BPMN diagram factory implementation as well to achieve these results for the BPMN Definition Set.
See https://issues.jboss.org/browse/JBPM-5397
@jlindop As you need this feature, could you please verify this PR? Ping me if you need some help.
Thanks!


jlindop:
@romartin Thanks Roger. This looks good.
+1

PR #563, opened 2016-11-14
RHBRMS-2689: Unable to build project if kbase and ksession names are identical

manstis:
See https://issues.jboss.org/browse/RHBRMS-2689

See also https://github.com/droolsjbpm/drools/pull/1004 and https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/187

The Workbench also needs to use the "version-less" Name Space (for new Projects). Older Projects created in the Workbench are tied to 6.0.0 XSD and hence, if the fix for RHBRMS-2689 is needed in older versions, the old 6.0.0 XSD would need updating too :-(

See https://issues.jboss.org/browse/DROOLS-1023

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/kmodule/converters/KModuleConverter.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/kmodule/converters/KModuleConverter.java
@@ -34,7 +34,8 @@ public class KModuleConverter
     public void marshal(Object value, HierarchicalStreamWriter writer, MarshallingContext context) {
         KModuleModel kModule = (KModuleModel) value;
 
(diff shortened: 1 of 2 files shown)


manstis:
@ge0ffrey @Rikkola IDK if you want to pass comment on this too?

ge0ffrey:
All code, examples, docs, wb ui's, etc should definitely use the new versionless namespace. But for what's it's worth: any old artifacts that still use the old namespace will still work on drools, it just internally treats it as the versionless namespace, so migrating the wb database might be optional.

manstis:
Thanks @ge0ffrey I was testing @mariofusco fix to the XSD for https://issues.jboss.org/browse/RHBRMS-2689 and noticed the Workbench was creating Projects against a fixed (older) XSD and hence my testing was failing. "Old" Projects reference the 6.0.0 XSD and hence fail to validate against the newer (fixed) XSD.. at least this PR "stops the rot". I've commented the JIRA about the possibilities for older BxMS projects.

PR #574, opened 2016-11-22
JBPM-5420: Stunner - move BPMN specific field constants from kie-wb-common-stunner-basicset-api to kie-wb-common-stunner-bpmn-api

jlindop:


tsurdilo:
+1

PR #587, opened 2016-12-01
JBPM-5354: Remember the JPA and Advanced tool windows navigation context

wmedvede:


wmedvede:
Hello @manstis , would you mind review this PR?

Quick note: you'll see that some refactoring/improving was required due to some still missing MVP-itezation, etc. Since all the sibling widgets for all the domains are based on UiBinder, plus that we want this to be merged into 6.5.x, and 6.5.x uses a different errai version from master, etc, the two new widgets resulting form the refactoring were made also with UiBinder.

Thanks!

wmedvede:
Hello @manstis , yes, this Presenters are basically widgets that I expect to "add" (and I really do) in different higher level widgets. If they were errai-ui components then they should implement "IsElement" instead. But as said, I'd prefer to keep this two new widgets with UiBinder.

p.s. will try to find some more reviewers for my code then :)

PR #591, opened 2016-12-02
Fixes after JPA 2.0 -> 2.1 upgrade

psiroky:
Please don't merge yet. All the IP BOM 7.0.0.CR5 PRs need to be merged together.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
@@ -192,7 +192,7 @@
     <!-- persistence -->
     <dependency>
       <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
+      <artifactId>hibernate-jpa-2.1-api</artifactId>
     </dependency>
     <dependency>
       <groupId>org.hibernate</groupId>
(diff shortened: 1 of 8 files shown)


kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
```diff
@@ -216,6 +216,16 @@
...
+    <dependency>
+      <groupId>org.jboss.spec.javax.ejb</groupId>
+      <artifactId>jboss-ejb-api_3.2_spec</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.gwt</groupId>
+      <artifactId>gwt-user</artifactId>
+      <scope>test</scope>
+    </dependency>
```
psiroky: @manstis, @csadilek I am not exactly sure about this one. Without the `gwt-user` dep Weld complains that it can not find `JsType`. But should we really depend (even if for testing) on gwt-user in -backend module? Also I am not exactly sure why this changes with the JPA upgrade which does not seems to have anything to do with GWT?

manstis: @psiroky I tend to agree with you. Do you have a stack trace that might help pinpoint why it's expected?

@wmedvede Do you know why ```gwt-user``` may be required?

wmedvede: @manstis Current version works, but I don't know to be honest why this dependency seems to be required now.

PR #592, opened 2016-12-02
Add missing hashCode method for classes with overridden equals method

hasys:


hasys:
Hi @manstis, good point! I updated my PR with your notes.

PR #595, opened 2016-12-05
JBPM-5443: Stunner - enable setting of Name and Description propertie…

jlindop:
…s on all BPMN objects


tsurdilo:
tested. can set it . + 1

PR #598, opened 2016-12-07
JBPM-5438: Stunner - get DataTypes for Data I/O Editor & Process Vars from server

jlindop:


tsurdilo:
@jlindop I think StunnerFindDataTypesQuery should be renamed to FindDataTypesQuery and reside in kie-wb-common-refactoring-backend/ with the other queries (you should rename the one in jbpm-designer to DesignerXYZQuery (and update that where its currently used) so they don't have the same name. 

Also DataTypesQueryService would be better to be called FindDataTypesService (or similar) as it could be extended to run more than just a single query in future. 

just my 2c
otherwise +1

jlindop:
@tsurdilo Thanks for the suggestions - I've made those changes. The Designer query is already called DesignerFindDataTypesQuery, so there won't be a conflict.

tsurdilo:
@jlindop thanks

+1

PR #608, opened 2016-12-19
JBPM-5452: Case Modeller

manstis:
Preliminary changes for Case Modeller.

This is a work in progress (and as such should technically not be merged) however it contains sufficient changes to the BPMN2 related (un)marshaller and Annotation Processor to risk a wide divergence from ```master``` and a nightmare conflict scenario later.

The new code in this PR is not visible in ```kie-wb``` or ```kie-drools-wb```.


manstis:
Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/337

PR #611, opened 2016-12-21
GUVNOR-2810: Library: Default package should not have the _ character

paulovmr:


manstis:
@paulovmr Sorry, but should we not have some form of test for this!? I know it's such a petty change.. but it clearly caused problems for @Rikkola ;-) so we'd want to prevent it creeping back in (e.g. someone decided in the future to change ```projectGroupId``` to include "dots").

Rikkola:
How did it add the _ in? 
And was the end result _default folder in a org folder was the folder name actually org._default?

The UI shows both in the same way.

paulovmr:
@manstis I agree, I'll make one test to check if the preference value is a valid package.

@Rikkola My best guess is that [this line](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/project/ProjectSaver.java#L227) adds the "_" since `default` is a Java identifier. The result was a `_default` folder inside a `org` folder.

Rikkola:
@paulovmr true, that must be why. Well in that case the _ was a good thing. But I would still prefer something else ;-)

paulovmr:
@Rikkola Yes, I agree. :) I changed it to the same pattern that we have when creating a project through Authoring perspective (`<OU_NAME>.<PROJECT_NAME>` by default, the user can still change it on the Library preferences screen).

manstis:
Thanks @paulovmr

PR #615, opened 2016-12-21
[6.5.x] RHBPMS-4421: Container name should be set to GAV when artifact is selected

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4421

ff


karreiro:
@manstis @livthomas Could you take a look on this?

PR #623, opened 2017-01-03
GUVNOR-2765: Full screen editors on Library

paulovmr:
Depends on: https://github.com/uberfire/uberfire/pull/596

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/events/AssetDetailEvent.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/events/AssetDetailEvent.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.screens.library.client.events;
+
+import org.guvnor.common.services.project.model.Project;
+import org.uberfire.backend.vfs.Path;
+
+import static org.uberfire.commons.validation.PortablePreconditions.*;
+
+public class AssetDetailEvent {
+
+    private Project project;
+
+    private Path path;
+
+    public AssetDetailEvent() {
+    }
+
+    public AssetDetailEvent( final Project project,
+                             final Path path ) {
+        checkNotNull( "project", project );
+        checkNotNull( "path", path );
+
+        this.project = project;
+        this.path = path;
+    }
+
+    public Project getProject() {
+        return project;
+    }
+
+    public Path getPath() {
+        return path;
+    }
+
+    @Override
+    public boolean equals( final Object o ) {
+        if ( this == o ) {
+            return true;
+        }
+        if ( !( o instanceof AssetDetailEvent ) ) {
+            return false;
+        }
+
+        final AssetDetailEvent that = (AssetDetailEvent) o;
+
+        if ( getProject() != null ? !getProject().equals( that.getProject() ) : that.getProject() != null ) {
+            return false;
+        }
+        return !( getPath() != null ? !getPath().equals( that.getPath() ) : that.getPath() != null );
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = getProject() != null ? getProject().hashCode() : 0;
+        result = ~~result;
+        result = 31 * result + ( getPath() != null ? getPath().hashCode() : 0 );
+        result = ~~result;
+
+        return result;
+    }
+}
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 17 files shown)


kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/events/AssetDetailEvent.java
```diff
@@ -0,0 +1,76 @@
...
+
+        this.project = project;
+        this.asset = asset;
+    }
+
+    public Project getProject() {
+        return project;
+    }
+
+    public Path getAsset() {
```
manstis: IMO use of ```asset``` is inconsistent with all other existing code; where we freely refer to ```path```.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
```diff
@@ -0,0 +1,102 @@
...
+        this.placeManager = placeManager;
+        this.libraryBreadcrumbs = libraryBreadcrumbs;
+        this.projectDetailEvent = projectDetailEvent;
+        this.assetDetailEvent = assetDetailEvent;
+    }
+
+    public void assetSelected( @Observes AssetDetailEvent assetDetails ) {
+        final PathPlaceRequest pathPlaceRequest = generatePathPlaceRequest( assetDetails.getAsset() );
+
+        pathPlaceRequest.getPath().onDelete( () -> {
```
manstis: What about concurrent delete? Whilst ```LockManager``` provides optimistic locking in the UI; should somebody modify an file via SSH (clone repo, change, commit, push) then we still need to worry about concurrent deletes.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
```diff
@@ -0,0 +1,102 @@
...
+            placeManager.goTo( LibraryPlaces.PROJECT_SCREEN );
+            projectDetailEvent.fire( new ProjectDetailEvent( assetDetails.getProject() ) );
+        } );
+
+        pathPlaceRequest.getPath().onRename( () -> {
```
manstis: Same as for delete; re: concurrent rename.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
```diff
@@ -0,0 +1,102 @@
...
+            libraryBreadcrumbs.setupLibraryBreadCrumbsForAsset( assetDetails.getProject(),
+                                                                pathPlaceRequest.getPath().getFileName(),
```
manstis: Why not just pass ```pathPlaceRequest``` (or at most ```pathPlaceRequest.getPath()```).. then you won't need two parameters for the ```filename``` and ```path```.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
```diff
@@ -0,0 +1,102 @@
...
+                                                                pathPlaceRequest.getPath() );
+        } );
+
+        placeManager.goTo( pathPlaceRequest );
+        libraryBreadcrumbs.setupLibraryBreadCrumbsForAsset( assetDetails.getProject(),
+                                                            assetDetails.getAsset().getFileName(),
```
manstis: Likewise; re: two parameters.

paulovmr:
@manstis Hi, I made the changes you suggested. Thanks for the review!

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
```diff
@@ -0,0 +1,102 @@
...
+        this.placeManager = placeManager;
+        this.libraryBreadcrumbs = libraryBreadcrumbs;
+        this.projectDetailEvent = projectDetailEvent;
+        this.assetDetailEvent = assetDetailEvent;
+    }
+
+    public void assetSelected( @Observes AssetDetailEvent assetDetails ) {
+        final PathPlaceRequest pathPlaceRequest = generatePathPlaceRequest( assetDetails.getPath() );
+
+        pathPlaceRequest.getPath().onConcurrentDelete( parameter -> {
```
manstis: @paulovmr I meant to add the concurrent handlers in addition to the non-concurrent handlers. You appear to have removed the original non-concurrent handling?

paulovmr: @manstis Sorry, my mistake. Fixed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetPerspective.java
```diff
@@ -0,0 +1,116 @@
...
+
+        pathPlaceRequest.getPath().onConcurrentDelete( info -> {
+            newConcurrentDelete( info.getPath(),
+                                 info.getIdentity(),
+                                 () -> {  },
+                                 () -> goBackToProject( assetDetails.getProject() ) ).show();
+        } );
+        pathPlaceRequest.getPath().onDelete( () -> goBackToProject( assetDetails.getProject() ) );
+
+        pathPlaceRequest.getPath().onConcurrentRename( parameter -> updateBreadcrumbs( assetDetails.getProject(), pathPlaceRequest.getPath() ) );
```
manstis: OK. I think we're getting there (sorry for not spotting it earlier).

At the moment we attach a handler for concurrent rename to show a popup to give the User the option of ignoring the change. See [here](https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java#L253).

I wonder whether we should do the same here (like you do for concurrent deletes)?

However it does introduce an interesting issue ;-) 

```AssetPerspective``` will attach handlers for rename/delete etc as will ```BaseEditor```. The ones attached in ```AssetPerspective``` handle updating the perspective and the ones attached in ```BaseEditor``` handle updating the editor. Both "concurrent" handlers give the User the ability to ignore or refresh; so Users would be presented with the popups twice.

Alternatively we could have ```AssetPerspective``` simply perform the action of updating itself (and not prompt the User). However if the User chose to ignore the concurrent change in the Editor the perspective would still update. This may be the lessor of two evils (concurrent popups shown twice vs ```AssetPerspective``` always shows "current state"). This would mean changing the concurrent delete handler in this PR to not use ```newConcurrentDelete(..)``` popup but simply go back to the project (as a regular delete).

WDYT?

paulovmr: @manstis Nice catch :). I [added](https://github.com/uberfire/uberfire/pull/596/files#diff-d8b56b39c1f00334526c947c313fb044R291) some events on `BaseEditor`, so we can monitor whether the user chose to accept or ignore the concurrent action. Now I think the behavior is ok.

PR #627, opened 2017-01-09
RHBRMS-2730, RHBRMS-2729

lazarotti:


lazarotti:
@manstis, the only difference than that you showed me is adding the full stack trace also to logger.debug. Even adding the exception message to:

            final String msg = MessageFormat.format( ERROR_EXTERNAL_CLASS_VERIFICATION,
                                             toFQCN( packageName, className ),
                                             e.getMessage() );

... sometimes it is still not clear. In that jira even fixing the message it will show only "Underlying system error is Lme/lazarotti/SaySomething;" - however if added the stacktrace we will see in the stack: 
"Caused by: java.lang.ClassNotFoundException: me.lazarotti.SaySomething"

That's why I also added it.

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/ClassVerifier.java
```diff
@@ -76,9 +76,11 @@ private void verifyClass( final String packageName,
             }
 
         } catch (Throwable e) {
-            final String msg = MessageFormat.format( ERROR_EXTERNAL_CLASS_VERIFICATION,
-                                                     toFQCN( packageName, className ) );
+           final String msg = MessageFormat.format( ERROR_EXTERNAL_CLASS_VERIFICATION,
+                                             toFQCN( packageName, className ),
+                                             e.getMessage() );
```
hasys: Hi @lazarotti, this new formatting for the fast look like you are calling MessageFormat.*format*().toFQCN.().e.getMessage(), I'm not sure that this format is better, than previous.

lazarotti: It is important to catch and show the exception, or the user will not be able to identify what dependency/transitve dependency is missing. See the Jira https://issues.jboss.org/browse/RHBRMS-2730

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/ClassVerifier.java
```diff
@@ -76,9 +76,11 @@ private void verifyClass( final String packageName,
...
             logger.warn( msg );
+	    logger.debug( "This state is usually encountered when the Project references a class not on the classpath; e.g. in a Maven 'provided' scope or 'optional' dependency.", e);
```
lazarotti: ACK, i will send a new PR with tabulation fixed

lazarotti:
@manstis I am trying to figure out someway to simulate an exception during the class checking. I can't mock this with mockito as it does not work over final classes (java.lang.Class).

Can I add PowerMockito to the test suite ? It allows that. To understand  what I am trying to do see the test below (which does not work with Mockito):
http://pastebin.test.redhat.com/445192 (sure, the assert I will change to match an expected message with the error).
.
There is any other way to load a problematic class for verification in our test suite instead of mocking ?

manstis:
@lazarotti Hi, I have no problem with PowerMockito being introduced (we use it elsewhere already). However you could more simply mock ```kieModuleMetaData.getClass.getClass(..)``` to throw ```ClassNotFoundException``` - or at least that's what I was thinking!

lazarotti:
Closing this one, please review the new PR https://github.com/droolsjbpm/kie-wb-common/pull/636

PR #628, opened 2017-01-09
[6.5.x] RHBRMS-2730, RHBRMS-2729

lazarotti:


lazarotti:
@manstis, the only difference than that you showed me is adding the full stack trace also to logger.debug. Even adding the exception message to:

            final String msg = MessageFormat.format( ERROR_EXTERNAL_CLASS_VERIFICATION,
                                             toFQCN( packageName, className ),
                                             e.getMessage() );

... sometimes it is still not clear. In that jira even fixing the message it will show only "Underlying system error is Lme/lazarotti/SaySomething;" - however if added the stacktrace we will see in the stack: 
"Caused by: java.lang.ClassNotFoundException: me.lazarotti.SaySomething"

That's why I also added it.

lazarotti:
As I did different commits while creating the test, instead if rebasing this PR i create the one:
https://github.com/droolsjbpm/kie-wb-common/pull/635

PR #629, opened 2017-01-10
GUVNOR-2841 : Guided Decision Table: Restored dtable should not be read only

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2841

A bug I found while fixing a bug that was found when fixing a bug.

`KieMultipleDocumentEditor` has the one line fix. I split the `KieMultipleEditorest` to an abstract base to make the class a bit shorter. The only new test I actually added is `KieMultipleDocumentEditorRestoreTest.restoreMakesTheDocumentEditable()`


Rikkola:
@manstis or @karreiro if either of you could please take a look.

Rikkola:
Thank you

PR #630, opened 2017-01-10
JBPM-5489: Stunner - fix Data I/O for reusable subprocess

jlindop:


hasys:
Hi @jlindop,
looks and works good. Just some style things, but as I know we will fix all style related things later this week in one big PR, so +1

PR #631, opened 2017-01-11
GUVNOR-2845: Data modeller: Add new Java 8 Local Date/Time types

mcimbora:


mcimbora:
Part of:
droolsjbpm/droolsjbpm-knowledge#198
droolsjbpm/droolsjbpm-integration#789
droolsjbpm/kie-wb-common#631

@manstis Can you please have a look?

manstis:
@wmedvede Could you please review for @mcimbora thanks.

wmedvede:
Hello @manstis @mcimbora Hello guys,
I don't know exactly the drools internals, so I'll assume that new types are supported by the drools engine. 
e.g, if a user creates a pojo with e.g. a field of DateTime, and sets conditions, etc. using this field/type, is the engine prepared for supporting it??
Assuming yes from my side it's ok.

Hello @pefernan, if I don't remember wrong this new java types are not yet supported by the form modeller, so probalby we'll need a small adding there for manage this types. Could you please take a look at it?

manstis:
@wmedvede I assume Drools itself handles *any* class. 

However the ```DataModelOracle``` used by the editors (and I assume ```optaplanner-wb``` editors) only considers classes assignable from ```java.util.Date``` to be "date" types and hence use of a model with ```java.time.*``` classes will not be presented to Users as a "date picker". 

@mariofusco Can you confirm my understanding?

@mcimbora Is the above an issue for you?

pefernan:
@wmedvede nope, the new Form Modeler isn't supporting them, it's not a big deal adding this types to the "supported types list" on the forms side, the only problem is that this java.time types aren't available on GWT and don't know how they can be treated when binding the data to the widget.

mcimbora:
@manstis Not at this point of time. The idea is to be able to create OptaPlanner data model containing these date/time types. We create our data sets outside the workbench.

mcimbora:
@manstis PR updated, added support for OffsetDateTime.

Checked with employee rostering example which uses LocalDate/LocalDateTime fields in the domain model/planner rules https://github.com/droolsjbpm/kie-wb-playground/pull/4

PR #632, opened 2017-01-11
JBPM-5451: Stunner - Implement DefaultRoute property on ExclusiveGateway

jlindop:
@tsurdilo  The DefaultRoute dropdown for Exclusive Gateway now lists the names of the target nodes rather than the outgoing SequenceFlows. I've made sure this is backwards compatible with Designer by keeping the 'dg' attribute in the serialized bpmn with the value <SequenceFlowName : SequenceFlowID>.


tsurdilo:
@jlindop tested and works. I think changing from connection to the connected node name is fine, but imo this does not improve it by much, i'll try to explain
- we havent implemented this in the jbpm-designer but in bpmn2 there is also a visual notation of the default sequence flow, for example see: http://imgur.com/a/6RBxD (little slash mark one the "else" flow)

- imo we should show the conditional expression if one exists, then sequence flow name if one exists and the connecting node name as the last resort as they may chance many times during modeling of the process. 

- this property was initially on the sequence flow level and we moved it to the gateway properties in jbpm designer but maybe we could check with the ux team becaue imo the ideal scenario would be for users to make this selection as they are entering the condition expression on the gateway itself (maybe a checkbox or something similar)  

WDYT?

jlindop:
@tsurdilo I agree with you - I'll create a Jira to track this suggestion.  In this PR I wanted to reinstate the DefaultRoute property so that we're at least compatible with Designer.

tsurdilo:
+1

PR #635, opened 2017-01-12
[6.5.x] RHBRMS-2730, RHBRMS-2729 - Business Central does not show what

lazarotti:
dependency is missing that makes a class invalid for authoring


manstis:
Waiting on ACKs on https://issues.jboss.org/browse/RHBRMS-2730 before merging.

PR #642, opened 2017-01-20
uberfire-social-activities-api is used server-side and hence cannot be 'provided' scope

manstis:
@ederign Could you please approve; (probably) just a copy and paste error.


ederign:
Yes. It was a copy paste error. Sorry about this and thanks.

PR #649, opened 2017-01-24
[UF-443] - Header and Footer should support Errai UI

ederign:
depends on https://github.com/uberfire/uberfire/pull/573.

@romartin both showcases seems broken. 

https://gist.github.com/ederign/7ae37f75929569fd29c5f173eca9fc0a


romartin:
hey @ederign , well yes that exception at startup happens but does not hurt, both showcases should be running fine anyway... isnt' it? anyway, a good reminder, have to fix this, will do on a next PR that I have only for some bug fixings, if that's good for you.. :)

PR #651, opened 2017-01-25
JBPM-5609: Stunner - fix BpmnFileIndexerTest failure by increasing wa…

jlindop:
…it for indexing to finish.

This test was passing locally but failed on Jenkins. I can get the test to fail locally with the same error as reported on Jenkins by reducing the wait time, so I think increasing the wait time should fix the test on Jenkins.

See https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/kie-releases-7.0.x/job/031.Releasing_KIE_deployLocally-7.0.x/lastCompletedBuild/testReport/


tsurdilo:
@jlindop both 
Thread.sleep(5000);
and 
Thread.sleep(15000);
worked for me locally. I noticed that all other tests in commons that test indexing use 
Thread.sleep(5000);
so maybe it would be worth to see if the number of bpmn files in the BPMN_FILES is causing the extra delay and if splitting it up would help? If in future you add more bpmn files to test or other test uses more we might run into this again. wdyt?

jlindop:
@tsurdilo There are only 7 files being indexed. I've only seen this error reported once, and the test must have passed all the other times it ran, so maybe it's a one off. I think I'll close the PR and keep an eye out for it happening again. If it does, we can split the files into groups and/or increase the wait time.

psiroky:
It seems the obvious solution here is to fix the test so that it does not have to rely on Thread.sleep(). I know it's not easy, but IMO worth the effort.

PR #657, opened 2017-01-27
JBPM-5584: Stunner - add validation for Priority property

jlindop:


tsurdilo:
+1

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/connectors/Priority.java
```diff
@@ -58,6 +60,7 @@
     public static final transient String defaultValue = "";
 
     @Value
+    @Pattern(regexp = "^$|^[1-9][0-9]*", message = "Must be a positive integer")
```
manstis: Does zero not count as a valid value? Based on the test I assume not.

PR #661, opened 2017-01-31
Stunner - Added findbugs maven plugin execution

romartin:
Hey @hasys ,
Here it is the initial version you proposed for the findbugs maven plugin running on Stunner, when using the `full` profile as already discussed. Not sure if this finally have to be merged or it's still not decided, due latest discussions on our bsig mailing list... 
Anyway it's now configured with `maxRank=6`, so just a few failures were raised, and fixed, but that's an initial point :)

### DIFF
--- a/kie-wb-common-stunner/pom.xml
+++ b/kie-wb-common-stunner/pom.xml
@@ -139,6 +139,25 @@
           </executions>
         </plugin>
 
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>findbugs-maven-plugin</artifactId>
+          <configuration>
+            <maxRank>6</maxRank>
+            <effort>Max</effort>
+            <xmlOutput>true</xmlOutput>
+          </configuration>
+          <executions>
+            <execution>
+              <id>findbugs-check</id>
+              <goals>
+                <goal>check</goal>
+              </goals>
+              <phase>compile</phase>
+            </execution>
+          </executions>
+        </plugin>
+ 
       </plugins>
 
     </pluginManagement>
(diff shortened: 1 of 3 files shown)


kie-wb-common-stunner/pom.xml
```diff
@@ -204,6 +223,21 @@
         <artifactId>maven-failsafe-plugin</artifactId>
       </plugin>
 
+      <plugin>
+        <groupId>org.pitest</groupId>
+        <artifactId>pitest-maven</artifactId>
+        <version>1.1.10</version>
```
romartin: @hasys not sure what does this do, but if really necesary, we probably should add this managed artifact into our KIE BOM....

manstis:
Yikes. The changes are on injected constructors... do we really need null checks in these scenarios. Sounds like the plugin has issues to me.

romartin:
@manstis yeahhh agree. So my thoughs are: we know that the CDI behaves this way and no NPE will happen, but being "syntax and language strict", an NPE could potentially occur... so that's why I didn't complain about this, but agree that not sure if makes sense refactoring now our constructors for this.

manstis:
Heh, let's hope we don't run this plugin against the whole workbench codebase!

romartin:
@manstis hehehehe I will take some holidays if so.. :P

hasys:
Hi @romartin, @manstis.

Thank you for review!

I guess in case of false positives it's better for now to disable findbugs and merge only real fixes (everything except pom.xml). I will try to find out another way how to integrate it without false positives.

romartin:
Ok @hasys , @manstis thanks, closing this PR for now.

triceo:
I believe @rsynek was experimenting with reasonable Findbugs settings on jBPM core, maybe he has some suggestions.

rsynek:
@hasys, @romartin,
please check the PR https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/382. The problem with CDI/constructors should be fixed.

PR #663, opened 2017-02-02
fix Windows specific test failures

barboras7:
When we run community tests against productized binaries on Windows, there are some Windows specific test failures. Most of these failures are caused by line separators and file paths. There will be some more fixes also in other repositories. 
This commit contains some minor changes in tests that replace \n separator by the OS specific separator.


hasys:
Jenkins retest please

wmedvede:
Hello @manstis , @hasys 
+1
Thanks

PR #667, opened 2017-02-03
JBPM-5428: Readonly option in the Checkbox component doesn't work.

pefernan:
Fixed other Forms minor issues:
- Fixing infinite loop switching Form to "PRETTY-MODE"
- Fixing Form unbind


pefernan:
@romartin can you take a look?

PR #669, opened 2017-02-04
JBPM-5630: Disable kbase fragments if last kbase was deleted

jomarko:


manstis:
@Rikkola Hi Toni, could you please take a look at @jomarko PR?

PR #678, opened 2017-02-09
JBPM-5646, JBPM-5647: Stunner - Add Start Signal and Timer Events

jlindop:
@tsurdilo The Start Signal event has a simple textbox for the SignalRef and I've created a Jira to cover getting a list of Signals for the user to select from. Also note that the icons for these 2 events are copies of the NoneStartEvent icon because we don't have the final icons yet.


jlindop:
@tsurdilo Closing as discussed.

PR #680, opened 2017-02-12
GUVNOR-2877: Include project root path in references query to avoid duplicate results

mcimbora:
Also includes GUVNOR-2878: Index: Incorrect query results when using special characters in the query


wmedvede:
+1

PR #686, opened 2017-02-15
RHBPMS-4586: Unable to create container with name which includes underscore (_)

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4586

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/service/SpecManagementServiceCDI.java
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/service/SpecManagementServiceCDI.java
@@ -27,7 +27,7 @@ import org.kie.server.controller.impl.KieServerInstanceManager;
 import org.kie.server.controller.rest.RestSpecManagementServiceImpl;
 import org.kie.workbench.common.screens.server.management.service.SpecManagementService;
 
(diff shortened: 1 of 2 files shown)


karreiro:
@manstis Could you take a look? :-)

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/service/SpecManagementServiceCDI.java
```diff
@@ -108,13 +109,19 @@ private String validContainerIdWithSuffix( final String serverTemplateId,
...
+            if ( !( Character.isLetterOrDigit( c ) || allowedChar( c ) ) ) {
                 return false;
             }
         }
+
         return true;
     }
+
+    private boolean allowedChar( final char c ) {
+        final char[] allowedChars = { ':', '-', '.', '_' };
```
manstis: Ouch; do we have to construct this array each time the method is called (from a loop, that iterates the same number of times as identifier is characters long)...?

karreiro: I was thinking just about readability, but you're right, Mike, it does not worth in this case.

karreiro:
Thanks, Mike! :-)

jomarko:
@karreiro thanks for test covering a lot of 'special' characters.

PR #687, opened 2017-02-15
[6.5.x] RHBPMS-4586: Unable to create container with name which includes underscore (_)

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4586


karreiro:
@manstis Could you take a look? :-)

manstis:
@karreiro As an aside you can validate a ```String``` as a Maven ArtifactID (as implied by the JIRA) using [this](https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/validation/ValidationUtils.java#L72).

karreiro:
@manstis I'm not calling the method `isArtifactIdentifier` because in this case `:` is a valid character. So I'm using a similar regex.

karreiro:
Thanks, Mike! :-)

PR #689, opened 2017-02-20
GUVNOR-2856: [Guided Decision Table] Remove import throws exception

manstis:
See https://issues.jboss.org/browse/GUVNOR-2856

GWT's ```onBrowserEvent``` was throwing a RTE when the ```ButtonCell``` had no rendered content.

It is a general problem for the "Data Object imports" tab and not specific to Guided Decision Tables.

See https://github.com/droolsjbpm/kie-wb-common/pull/689/files?w=1 with whitespace diffs removed.


manstis:
@Rikkola @karreiro Would you mind reviewing please?

PR #693, opened 2017-02-22
PLANNER-741: Workbench: Display warning message when score type is changed in a Planning Solution and scoreHolder is referenced

mcimbora:
* Add DeleteValidator
* Add Imports to data objects

Part of:
- https://github.com/droolsjbpm/kie-wb-common/pull/693
- https://github.com/droolsjbpm/optaplanner-wb/pull/135

Imports are used in the optaplanner-wb PR, as type parameter in superclass definition of a data object sometimes uses only simple class name (fqn is necessary). This way the fqn can be derived using ClassTypeResolver, which accepts a collection of import strings.

I also noticed renamedClasses and deletedClasses of UpdateInfo are not used, perhaps it's a good time to get rid of the class completely?

@wmedvede As this affects the data modeller, would you mind having a look? Thank you in advance.


manstis:
There is HasImports and Import in drools/drools-workbench-models IIRC. I'll review better tomorrow when at PC.. I also think Paulo has written "delete validator" type interfaces and service already (also for renaming). I'll need to find them...

manstis:
https://github.com/droolsjbpm/drools/tree/master/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/imports

manstis:
https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-api/src/main/java/org/uberfire/ext/editor/commons/service/restrictor/DeleteRestrictor.java

wmedvede:
I'll review when refactored then

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/core/HasImports.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.services.datamodeller.core;
+
+import java.util.List;
+
+public interface HasImports {
```
wmedvede: @manstis @mcimbora With reference to this HasImports, I believe that if this interface is needed in the datamodeler models it's better to have it here. See that we have already a lot of datamodeler interfaces in the org.kie.wb.common.services.datamodeller.core package, e.g. HasAnnotations, HasMethods, HasVisibilty, etc.

manstis: @wmedvede I don't believe that's necessarily a good argument.. ```drools-workbench-models``` had ```HasImports``` long before Data Modeller existed. Perhaps the other interfaces you mention would be better in ```drools-workbench-models```? ;-) There are plenty more uses of ```HasImports``` than just Data Modeller.

manstis:
OK, lets me pragmatic about this.. there are plenty of ```Command``` interfaces kicking around (even in KIE land) and we don't to force use of a single one (I'm sure there are plenty more examples of _identical_ interfaces kicking around in other packages, libraries etc). So.. let's agree we can live with the duplication.

As for ```DeleteValidator``` @mcimbora version returns messages as to why a deletion cannot occur; whereas @paulovmr version just prevents it from happening with no feedback. Perhaps we could agree to simply raise a JIRA suggesting ```DeleteValidator``` and ```DeleteRestriction``` (plus the other similar interfaces for copy and rename) can one day be combined; and @paulovmr implementation simply vetos the operation if the messages collection is not empty... then we can all be happy and move on :-)

manstis:
Jenkins please retest this.

mcimbora:
@manstis Right, the behavior is a little bit different, as you explained. The restrictions are "hard constraints", whereas the validation should be seen as a "soft constraint", i.e. user should be still permitted to perform the action if they know what they're doing.  I'm happy to refactor & move the whole "validation UI" (incl. validation popup) to uberfire, will create issue for this.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/main/java/org/kie/workbench/common/services/datamodeller/core/HasImports.java
wmedvede: @manstis The "argument" is in favor of having the datamodeler api/model self contained.
I know that there are duplication of interfaces (and likely in other places of the platform too), and yes nothing of what I've done is perfect, but since I don't believe a mayor refactoring will be done at this moment, I still prefer to pay the price to have this interface within the datamodeler domian. But whatever thing you want to do guys is fine for me.

kie-wb-common-services/kie-wb-common-data-modeller-core/src/test/resources/org/kie/workbench/common/services/datamodeller/driver/package6/ImportsUpdateTest.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.services.datamodeller.driver.package6;
+
+import javax.annotation.Generated;
+
+public class ImportsUpdateTest
```
manstis: Is this class needed, or are we missing a test?

kie-wb-common-services/kie-wb-common-data-modeller-core/src/test/resources/org/kie/workbench/common/services/datamodeller/driver/package6/ImportsUpdateTestResult.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.services.datamodeller.driver.package6;
+
+import javax.annotation.Generated;
+import java.util.List;
+import java.util.ArrayList;
+
+public class ImportsUpdateTestResult
```
manstis: Is this class needed, or are we missing a test?

kie-wb-common-services/kie-wb-common-data-modeller-core/src/test/resources/org/kie/workbench/common/services/datamodeller/driver/package6/ImportsUpdateTest.java
mcimbora: @manstis Yes, the class is used by JavaRoasterModelDriverTest, https://github.com/droolsjbpm/kie-wb-common/pull/693/files#diff-c86ec07f5404f9c29bf5c35c0adb185fR554

manstis:
Jenkins please retest this.

wmedvede:
Hello @manstis @mcimbora no more comments from my side.
Thanks

PR #698, opened 2017-02-27
Workaround for IOExceptions when generating forms-related code

psiroky:


psiroky:
@pefernan or @romartin could you please quickly check this?

pefernan:
@psiroky good point! Thanks for fixing that. +1

PR #702, opened 2017-03-01
guvnor ala integration (DON'T MERGE)

wmedvede:
- adds a pipeline for performing the project builds based on current Builder and KieBuidler build system.
    - makes the BuildService to use the pipeline
    - Adds a BuildInfoService to avoid spread use of the LRUBuilderCache

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/pom.xml
+++ b/kie-wb-common-services/kie-wb-common-services-backend/pom.xml
@@ -76,6 +76,14 @@
       <artifactId>guvnor-structure-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.guvnor</groupId>
+      <artifactId>guvnor-ala-spi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.guvnor</groupId>
+      <artifactId>guvnor-ala-source-git</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.jboss.errai</groupId>
       <artifactId>errai-bus</artifactId>
     </dependency>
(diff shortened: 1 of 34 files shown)


wmedvede:
Hello @manstis ,

I'll appreciate if you can take a look at this PR and let me know if you agree with the approach and whatever feedback you might have, since you are the best person for this piece of WB. I'll also add @porcelli  to the thread.

It basically contains the integration of the ALA system for making the WB "build" and "build & deploy" actions to pass through a pipeline.

Some notes: 

1) forget about the tests. (you'll see comments or commented stuff in the tests just for making the project compile, etc., but at this stage the most important is to check if we agree with the approach for the ala integration. So the tests are just commented.)

2) The main goal right now is to integrate the ala stuff into the wb and make the "build" and "build and deploy" actions to go throug a pipeline. Basically what we talked a week ago by hangout. I've talked with @porcelli  this week regarding the concern about the "distribution" of the build system, and at this stage the "local" approach is good. (i.e. we don't need to have something that can theoretically launch the build or run the pipeline in a dedicated cluster node, etc)

3) So we now basically will have a pipeline that takes care of the build. Internally this pipeline will still use the ad-hoc build based on the Builder/KieBuilder we have right now.

4) Components that uses the LRUBuidlerCache directly, will now go through the BuildInfoService.

5) There are still some polishment to do, but the idea right now is to validate the approach.

Thanks!

Regards,
Walter.

manstis:
@wmedvede There's quite a few related failing tests.. I assume we're only reviewing this PR "in principle"? Or was you fixing tests etc before we consider merge?

wmedvede:
@manstis yes, the idea is to do a conceptual review to be sure if we are aligned with the proposal. And then, if the approach is good enough, fix the tests plus some polishment.

kie-wb-common-services/kie-wb-common-services-backend/pom.xml
```diff
@@ -76,6 +76,14 @@
       <artifactId>guvnor-structure-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.guvnor</groupId>
```
manstis: ```guvnor-ala``` is built after ```kie-wb-common-services-backend``` so this won't work in a "clean" build environment.

kie-wb-common-services/kie-wb-common-services-backend/pom.xml
```diff
@@ -76,6 +76,14 @@
...
+      <artifactId>guvnor-ala-spi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.guvnor</groupId>
```
manstis: Likewise.. ```guvnor-ala``` is built after ```kie-wb-common-services-backend```

wmedvede:
Hello @manstis , thanks.
I'll cancel this PR and send the defintive one when finished.

PR #705, opened 2017-03-02
JBPM-5731: Stunner - Diagram related dock screens only appear on the authoring perspective.

romartin:
Hey @manstis , @mbarkley 
This is a fix for the Stunner dock screens. See [this Jira ticket](https://issues.jboss.org/browse/JBPM-5731).
@mbarkley I'm not able to build your appformer branch, having some compile issues... will keep trying for a while, anyway would be good if you can check this fix makes the dock screens appear on your showcase as well.
Thanks guys!


kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/pom.xml
```diff
@@ -193,6 +193,10 @@
       <groupId>org.kie.workbench.screens</groupId>
       <artifactId>kie-wb-common-library-api</artifactId>
     </dependency>
+    <dependency>
```
manstis: Oh dear :-(

The ```kie-wb-common-stunner-project-client``` module is built **after** ```kie-wb-common-screens```... therefore this won't work in a "clean" build environment as the module will not exist. I think we need to investigate a new way to attach Docks to the Authoring perspective. Possible an interface ```DockProvider``` that we inject ```Instance<DockProvider>``` into the Authoring Perspective and iterate implementations to get the actual dock. Then you could have implementations in ```kie-wb-common-workbench-client``` for the _normal_ docks and in implementation in ```kie-wb-common-stunner-project-client``` for the _Stunner_ docks.

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/resources/org/kie/workbench/common/workbench/KieDefaultWorkbenchClient.gwt.xml
```diff
@@ -19,6 +19,7 @@
   <inherits name="org.kie.workbench.common.widgets.KieWorkbenchWidgetsCommon"/>
   <inherits name="org.kie.workbench.common.screens.library.LibraryClient"/>
   <inherits name='org.gwtbootstrap3.extras.typeahead.Typeahead'/>
+  <inherits name="org.kie.workbench.common.stunner.project.StunnerProjectClient"/>
```
manstis: See my first comment. If we moved to the ```DockProvider``` approach we would not need to inherit this GWT module (which would be impossible anyway in a "clean" build environment).

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/pom.xml
mbarkley: Hi @manstis,

I don't think this is an issue. I just tried building this branch locally after wiping out all `org/kie` in my local Maven repo.

As far as I understand, if you build from the root folder and there are no dependency cycles, Maven will build the artifacts in the right order (even if that means building some submodules of kie-wb-screens, building some other stuff, and then coming back and building more submodules of kie-wb-screens).

manstis: @mbarkley Great; that was my only concern, which now appears unfounded.

PR #713, opened 2017-03-08
GUVNOR-2930 : Library: New Projects are always created into the first Repository

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2930


Rikkola:
@paulovmr Take a look if this breaks any planned functionality. I see there are default ou and repo in the settings, but not sure how they are supposed to work. Anyway without this there is no way to create a project in to a selected repository.

paulovmr:
@Rikkola The default OU and repo are only used to create a new OU and/or repo if none exist. Your fix is correct, the wrong behavior was right only in the past, when we didn't have the OU and repo selector at the header, and the library only dealt with one OU and repo.

Rikkola:
@paulovmr Ok. Makes sense. Should it still honor the defaults for empty empty workspaces?

paulovmr:
@Rikkola When the library is accessed, default OU and repo are created automatically, if none exists. So when a project is created, we will always have an OU and repo selected at the header too.

PR #715, opened 2017-03-09
GUVNOR-2909 : Library: Add branch selector on the header (along with the repository selector).

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2909

@paulovmr Hi, take a look please.


paulovmr:
@Rikkola Hi, the changes look good, just to minor comments:

* Some files weren't formatted, other were. I think would be good to format all modified files.
* Is there a reason why you changed some interface methods to `final`?

Rikkola:
@paulovmr Updating formats, rebasing with latest. 
And about final, I just have an obsession with final, not really needed for interfaces.

Rikkola:
@jomarko That works since the branch in repository is null by default and then the branch in event is again null. I'll set some value to `repository.getDefaultBranch()` just in case.

jomarko:
@Rikkola thanks for explanation.

PR #718, opened 2017-03-10
JBPM-5756: Case Modeller: Preview Panel is broken

manstis:
See https://issues.jboss.org/browse/JBPM-5756

The main change in this PR is moving ```CanvasCommandFactory``` to an interface and having two implementations of it; one for default Stunner/BPMN2 and the other for Case Management. These are now used in ```SessionPreviewImpl``` to get the correct implementation for the diagram being previewed.

There is a test for the selection of ```CanvasCommandFactory``` implementation.


manstis:
@romartin I'm following your lead with small PRs :-)

@hasys @romartin please review.. it's not as bad as it looks!

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/presenters/session/impl/SessionPreviewImpl.java
```diff
@@ -108,9 +118,15 @@ protected AbstractCanvasHandler getProxiedHandler() {
...
                     @Override
+                    protected CanvasCommandFactory getCanvasCommandFactory() {
+                        return SessionPreviewImpl.this.getCanvasCommandFactory();
+                    }
+
+                    @Override
                     protected void enableControls() {
                         zoomControl.enable(canvas);
-                        zoomControl.setMinScale(0).setMaxScale(1);
+                        zoomControl.setMinScale(0);
```
manstis: I split these into individual statements to prevent mocking the world and his friends in Unit Tests.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/presenters/session/impl/SessionPreviewImpl.java
```diff
@@ -199,9 +225,8 @@ public void onError(final ClientRuntimeError error) {
      * a higher width and darker color for the line stroke.
      */
     private void updateCanvasDecorator(final AbstractCanvas.View canvasView) {
-        canvasView
-                .setDecoratorStrokeWidth(2)
-                .setDecoratorStrokeAlpha(0.8)
-                .setDecoratorStrokeColor("#404040");
+        canvasView.setDecoratorStrokeWidth(2);
```
manstis: I split these into individual statements to prevent mocking the world and his friends in Unit Tests.

manstis:
Jenkins please retest this.

romartin:
hey @hasys , let you know that for fixing the CM preview and explorer there are still missing other fixes, that's just the first action taken, but more commits are expected for making those work fine in CM. So thanks for the feedback, well tested :) But we need a bit more time yet... 
BTW @manstis I'm running the showcase now as well and I see that the layouts have changed? It's the expected behavior or it's a bug? As you can see on above @hasys screenshot, tasks are in same row as subprocesses... and I'm not able now to put the task below a subprocess.. is this some new expected behavior?  thanks!

manstis:
@hasys Yes both are completely "by design" at the moment. Please see:- 

https://issues.jboss.org/browse/JBPM-5753 (connections)
https://issues.jboss.org/browse/JBPM-5755 (preview panel)

When https://issues.jboss.org/browse/JBPM-5752 is done it'll help identify what is needed for JBPM-5753.

manstis:
@hasys @romartin tells me he cannot add tasks to subprocesses.. so something is broken. Let me look into and fix whatever regression I've introduced before merging this PR!

romartin:
@manstis @hasys just realized I was using the reusable `subprocess instead` of the `ad-hoc` one, sorry. @manstis so reusable processes cannot contain tasks, right? I understand they appear and behave like a regular task in the CM layout then. If so, all works fine, sorry for the bad feedback :)

manstis:
@romartin @hasys Correct. "Reusable Sub-process" does not have children. I am not entirely clear on what node types we will be supporting in Case Modeller; but for now I am concentrating on functionality rather than detail or aesthetics. Here's a video showing current state locally: https://youtu.be/1ZX6RHDAiQw

hasys:
Hi @manstis, thank you for explanations I see now. Look good! :+1: 

@manstis, @romartin About styles: if I correct "unstylish" code caused by guys with Eclipse from other repositories. We need to synchronize Eclipse formatter to fix it's in the future. Unfortunately I exhausted all my time limits for formatters :( so for now, I guess, the easiest and the fastest way is to code formatter once in a while to format code if needed. Not ideal solution, but the easiest one :(

PR #720, opened 2017-03-13
Fix error in examples wizard by setting default branch to master.

mbarkley:


manstis:
@Rikkola Could you please confirm you're happy with the change? (You added the property for GUVNOR-2909).

PR #722, opened 2017-03-14
GUVNOR-2907 : Library: Show "Messages" screen with build details when the project is built.

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2907

@paulovmr Here is messages screen. Still working on the rest.


jhrcek:
Hello @Rikkola,
I'm interested how this will work in the Library. Will this make the panel appear when project is built or on other occasions as well? Also will there be a way to open this panel when user wants (like the "Messages" link in the top navigation bar we had in authoring), or will opening be controlled entirely by the application? I'm asking because I remember "spontaneously appearing" Messages panel was causing some problems in selenium tests in the past..

Rikkola:
@jhrcek The `MessageConsoleService` opens the Messages panel and checks this white list. This should not change when the panel shows up. I haven't seen the Messages button in the new wireframe so could be that it is not needed, but not sure when anyone would press it. Maybe if you close the panel and then regret and want to take a look.

The Build button is also moving, so could be that the Messages button moves next to it.

https://github.com/kiegroup/guvnor/blob/c0cd212bec07d718182d87d4308fb0cb1bfb08e7/guvnor-message-console/guvnor-message-console-client/src/main/java/org/guvnor/messageconsole/client/console/MessageConsoleService.java#L46-L45

paulovmr:
jenkins retest this

PR #723, opened 2017-03-14
JBPM-5773: Stunner - not able to add Start Signal and Start Timer events

pefernan:
@romartin can you take a look? It seems that there was no ShapeDefs for that nodes ;)


pefernan:
jenkins retest this.

PR #724, opened 2017-03-14
JBPM-5755: Preview Panel does not work

manstis:
See https://issues.jboss.org/browse/JBPM-5452


manstis:
@romartin @hasys Want to review?

The issue for Case Management's preview not working was not that the mutations were not command based (they are!) but the Preview Panel used a generic ```CanvasHandler``` that led to a generic handler method being invoked rather than the correct implementation for Case Management. Preview now uses a new instance of the correct ```CanvasHandler``` for the Definition Set associated with the Preview canvas.

hasys:
Hi @manstis,
Thank you, looks good. But when I played for some time with Case Modeller it starts to work very slow and after that I start to receive this error:
```
Unable to complete your request. The following exception occurred: Exception caught: (RangeError) : Failed to execute 'getImageData' on 'CanvasRenderingContext2D': Out of memory at ImageData creation.
```

I see this error every time I move my mouse to canvas for some time. After several closed messages I am able for some time continue my work, but after some not long time (20-30 sec) I can see this error again. I don't know is it related to fix or not, but I guess it will be interesting for you any way.

Addition Info:
For first time I saw this error after trying to open Stanner Showcase-Project in Chrome AND Firefox at the same time. So maybe it is a problem with GWT Dev mode or something like that.

Picture of error:
screenshot from 2017-03-15 13-40-56

manstis:
@romartin OK. I have it fixed! 

I now [wrap](https://github.com/manstis/kie-wb-common/blob/557b0b6ed22f97557f0838e50aa94abbfb93fe32/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/presenters/session/impl/SessionPreviewCanvasHandlerProxy.java) the ```BaseCanvasHandler``` instance created for the DefinitionSet with a proxy that does the same as your old ```CanvasHandlerProxy``` class. 

I however delegate all ```BaseCanvasHandler``` methods to the wrapped instance (other than those relating to the Graph and Rules like your old proxy). I overrode all methods as who knows what sub-classes may override in the future... At the moment it's only ```addChild(parent, child, index)``` but I thought to be future proof.

manstis:
@hasys That is worrying :-(

Could you please re-try with this revised PR (but I am afraid it won't make any difference). What browser are you using? How "long" do I have to move things around before the world crashes around me!? If this is a reproducible issue I propose a new JIRA rather than hijacking this PR.

manstis:
@hasys I've been looking into the probable memory leak. Could you try adding the following to ```CaseManagementDockingAndContainmentControlImpl``` in ```makeColorMapBackedPicker(..)``` [L#134](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/wires/CaseManagementDockingAndContainmentControlImpl.java#L134):-
```
if(m_picker!=null) {
    m_picker=null;
}
```
This explicitly declares the old instance and related image data for GC.

I fiddled around a bit and found dragging did slow down alot, although I never received the OOME.. but this could simply be that my laptop has a stonking amount of memory ;-). With the change I found the slow down seemed to disappear.

romartin:
hey @manstis !! After long comment above... and still thinking... I think you're right and that's why in fact your CM works now. So I think it's true that there exist an additional instance of canvas handler in this approach but this instance have the right type, is it? I was missing to understand the whole picture... sorry!

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/presenters/session/impl/SessionPreviewImpl.java
```diff
@@ -141,18 +149,40 @@ protected void destroyControls() {
...
-        this.shapeManager = shapeManager;
-        this.canvasCommandManager = canvasCommandManager;
+    }
+
+    BaseCanvasHandler getCanvasHandler() {
+        if (canvasHandler == null) {
+            BaseCanvasHandler handler;
+            final String defSetId = getDiagram().getMetadata().getDefinitionSetId();
+            final Annotation qualifier = definitionUtils.getQualifier(defSetId);
+            final ManagedInstance<BaseCanvasHandler> customInstances = canvasHandlerFactories.select(qualifier);
```
romartin: @manstis just a minor thing.. .I think this logic is already on the shape manager... so doing something like:
` shapeManager.getCanvasFactory(getDiagram()).newCanvasHandler()`. The same is done some lines below on this class, see method `beforeOpen`

manstis:
@romartin Thanks for the review (and more importantly; approval!)

Ideally my "proxy" (wrapper) would implement ```CanvasHandler``` and hence not contain any inherited state. I have no interest in state; but the behaviour of the ```CanvasHandler``` relating to the Diagram. Unfortunately the interface lacks many of the methods that can be overridden by subclasses; similarly for ```AbstractCanvasHandler``` (and the code you show for ```CanvasFactory```); it does not contain the methods relating to rule loading and graph building that need "null" implementations for the Preview.. So, yes, things could be improved if we moved all ```public```/```protected``` methods from the abstract classes to the interface... but that's probably a refactoring out of the scope of this PR.

manstis:
@hasys I've included the change for OOME in this PR. Did you want to re-try it before I merge?

manstis:
Jenkins please retest this.

hasys:
Hi @manstis,
hard to say. I found some new problems (as well as mentioned one) and I am able to destroy entire Case Management editor and google chrome tab now.
Sometimes (I guess it is related with subprocess movements) Preview panel is not updated, see:
screenshot from 2017-03-16 11-13-32
I tried to play more and the I saw this error:
screenshot from 2017-03-16 11-16-52
This message is can't be closed by OK button, because in this case your mouse will be over the canvas and you will see this error again. The only way is close it with X button and don't touch the canvas any more. When I close it I see this (this is a screenshot from the second time I met this error):
screenshot from 2017-03-16 11-15-24
I closed the editor and tried once more but after some time I saw this:
screenshot from 2017-03-16 11-18-41
All this takes about 5-7 minutes. I also tried with your suggested changes, but I met new issues several times and stopped my tries, so I didn't saw OutOfMemory (and it looks subjectively little bit more responsive) with clean up, but I am not sure it will help for 100%. It needs more investigations. Maybe @bsremac can help you with it.

manstis:
@hasys Thanks for the feedback. Do you know of any particular actions to reproduce (I appreciate it's probably hard to say; other than it barfed after I used it for X minutes).. I will try to replicate the setup shown in your images and fiddle around a bit :-(

manstis:
@hasys Hi, I have found an issue that seems to explain, at least in part, what you have found. If you drag a node beyond the left of the "main editing canvas" the view is updated positioning the dragged node on the far left but the Preview and Explorer remain unchanged. This leads to visual inconsistencies between "main editing canvas" and the Preview/Explorer content. That said the Preview and Explorer content are consistent - which is what I see in your screen shots. I don't suppose you can advise what happens if you always keep dragging within the "main editing canvas"? This issue appears to be caused by ```DragControlImpl``` that is setting a View elements x-coordinate to zero if it's dragged beyond the extent of the canvas.

manstis:
@hasys OK, I've included a fix for when a drag operation goes outside the bounds of the "main canvas viewer". It technically was not an issue with the "Preview Panel" (which is what this PR represents) but it did lead to inconsistent states between "View" and "Preview/Explorer" which would lead to unexpected behaviour... so, fingers crossed it's all OK now!

manstis:
Jenkins please retest this.

manstis:
@hasys I've merged, but will take the task to investigate the OOME as a new JIRA and a priority. I suspect it's the way we're using ```Context2D``` ```ImageData```, every shape on the canvas has an associated ```ImageData``` instance which is a pixel array the size of the whole editor. CM hits it particularly hard by re-creating new instances as shapes are dragged around.

manstis:
@hasys See https://issues.jboss.org/browse/JBPM-5819

PR #725, opened 2017-03-14
GUVNOR-2910: [Library] Project remains open after it is deleted

paulovmr:


paulovmr:
@Rikkola Hi, if you could take a look, please. :)

PR #734, opened 2017-03-20
Library issues fix

paulovmr:
GUVNOR-2918: Library: Add link to the import example wizard on the Library and Empty Library screens
JBPM-5637: EmptyLibraryScreen does not show option to join an existing project.
GUVNOR-2913: [Library] Project explorer is open in Library perspective
GUVNOR-2917: [Library] Library Screen should keep org. unit selected


paulovmr:
@Rikkola Hi, if you could take a look, please. :)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
```diff
@@ -244,6 +253,31 @@ public void assetSelected(@Observes final AssetDetailEvent assetDetails) {
                   assetDetails.getPath());
     }
 
+    public void projectContextChange(@Observes final ProjectContextChangeEvent projectContextChangeEvent) {
```
Rikkola: It is a lot safer to add a change listener to `ProjectContext`. The event has a bad name, it should be `ChangeProjectContextEvent`. Meaning it is used to change the context, not to follow context change. 
What you have here can work, but if there are two change events fired it might not show the latest and be out of sync with the rest of the application.

manstis: Actually, the event was originally used to _signal_ a change in the context from the navigation selections in the Project Explorer; and the current state was held in ```WorkbenchContext``` that observed these events... however that has evolved/mutated over the years and numbers of hands on the code..... still @Rikkola does know the expected current behaviour better than anybody!

Rikkola: @manstis that is true. And before I added the Handler it did both, but there were sync issues with what actually is the current context so I went with a solution that makes sure that even if event A change the context to something and after that B to some thing else. At least the client side is consistent on showing either A or B everywhere.

paulovmr: Done!

PR #738, opened 2017-03-21
JBPM-5827: Case Modeller: CM Shape should ignore underlying sizes in BPMN file

manstis:
See https://issues.jboss.org/browse/JBPM-5827

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/ActivityShape.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/ActivityShape.java
@@ -16,6 +16,7 @@
 
 package org.kie.workbench.common.stunner.cm.client.shape;
 
+import org.kie.workbench.common.stunner.bpmn.definition.BaseTask;
 import org.kie.workbench.common.stunner.cm.client.shape.view.ActivityView;
 import org.kie.workbench.common.stunner.core.client.shape.MutationContext;
 import org.kie.workbench.common.stunner.core.graph.Edge;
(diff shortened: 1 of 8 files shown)


manstis:
@romartin I dare you to find fault with this one!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/ActivityShape.java
```diff
@@ -37,9 +38,8 @@ public void applyProperties(final Node<View<W>, Edge> element,
                                 final MutationContext mutationContext) {
         super.applyProperties(element,
                               mutationContext);
-        // Apply activity attribute values for width and height.
-        final Double width = getShapeDefinition().getWidth(getDefinition(element));
-        final Double height = getShapeDefinition().getHeight(getDefinition(element));
+        final Double width = BaseTask.BaseTaskBuilder.WIDTH;
```
romartin: Hey @manstis ! All good, just share a comment for you about CM shapes. As far as I see in fact the CM Shapes never "mutate", right? I mean, once they're added into the canvas, can they be updated as from updates in the domain model?
If the answer is "CM shapes are not mutable", you can avoid "hardcoding" these kind of things, like setting same size as the initial one, by just making CM shapes implement `Shape` (or any of its sutbtypes), rather than `MutableShape` (or any of its subtypes). In that way, there is no `applyMutation` method to implement in the different shape impls, as they're supposed not be changed as from the domain model updates.
Well just a comment for further thoughts if so.. :)

manstis:
@romartin Oh, how I'd like to be more simple! I do need a mutable "title" which as far as I could tell does not exist on ```Shape``` but only when I dive deeper into the heirachy to ```ElementShape``` which extends ```MutableShape```.... if there's a more simple way I'm all ears!

romartin:
@manstis ohhh good point! yes that's true... so right now it forces to make it mutable as due to the "dynamic" title, sime improvement could be done here... anyway sorry for my comment, you're absolutely right, forgot the title arghh :)

manstis:
@krisv FYI...

PR #744, opened 2017-03-24
JBPM-5795: Stunner - make ExclusiveGateway DefaultRoute property show…

jlindop:
… node type if node name empty.


jlindop:
@manstis  In this PR, I've added a method "String getTitle();" to BPMNDefinition. This interface is implemented by BPMNDiagram and other BPMN api node type base classes such as BaseTask.  Is this change OK with you?

manstis:
@jlindop Hi, yep, fine with me.

PR #737, opened 2017-03-21
JBPM-5790: Case Modeller: Change containment rules for Diagram and St…

manstis:
…ages

See https://issues.jboss.org/browse/JBPM-5790

This PR changes the containment rules for "Diagram" to only only "Sub-processes" as direct descendants.

PR #745, opened 2017-03-24
JBPM-5790: Case Modeller: Change containment rules for Diagram and Stages

manstis:
See https://issues.jboss.org/browse/JBPM-5790

Following https://github.com/kiegroup/kie-wb-common/pull/737 @hasys reports it is no longer possible to open BPMN diagrams into Stunner/BPMN editor. I replicated the issue using the same steps as @hasys (although could not replicate using InteliJ). The problem was traced to CM's ```BPMNDiagram``` extending BPMN's ```BPMNDiagram```.


manstis:
Thanks @hasys hopefully we can get to merge https://github.com/kiegroup/kie-wb-common/pull/743 soon; and close (at least) some of the leaks (especially that relating to Explorer and CM).

PR #750, opened 2017-03-29
JBPM-5609: Address BpmnFileIndexerTest intermittent failures

jlindop:


jlindop:
@manstis Thanks for the pointer. Would be good if you'd merge this and I'll look into the change later. Thanks.

PR #753, opened 2017-03-30
Exclude not needed Hibernate deps

psiroky:
@wmedvede could you please take a look?

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml
@@ -191,12 +191,18 @@
 
     <!-- persistence -->
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.1-api</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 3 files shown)


manstis:
See https://github.com/kiegroup/kie-wb-common/pull/754 for build fix.

manstis:
Jenkins please retest this.

wmedvede:
Hello @psiroky , thanks for the explanations regarding the hibernate deps.

I think the exclusion will have to be added also here

https://github.com/wmedvede/kie-wb-common/blob/3468925b9e39346b5f00ea382248292b3219123d/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/pom.xml#L193-L193

The hibernate-envers is algo being used at:

https://github.com/wmedvede/kie-wb-common/blob/3468925b9e39346b5f00ea382248292b3219123d/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/pom.xml#L197-L197
but in this case only with test scope, so the exclusion is probably not requried here.

Thanks

psiroky:
Thanks @wmedvede, I will take a look.

psiroky:
@wmedvede I've removed the test scoped dependency as it is not used and added the second exclude.

wmedvede:
@psiroky great, thanks.
+1 in case it requires it.

PR #760, opened 2017-03-31
JBPM-5828: Case Modeller: Stages and Activities to use icons

manstis:
See https://issues.jboss.org/browse/JBPM-5828

Whilst adding icons I noticed some DOM leaks regarding glyphs (used for the icons) and fixed those too.

The leaks fixes probably relate to https://issues.jboss.org/browse/JBPM-5462

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-shapes/kie-wb-common-stunner-shapes-client/pom.xml
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-shapes/kie-wb-common-stunner-shapes-client/pom.xml
@@ -137,6 +137,18 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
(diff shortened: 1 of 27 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-shapes/kie-wb-common-stunner-shapes-client/pom.xml
```diff
@@ -137,6 +137,18 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.ahome-it</groupId>
+      <artifactId>lienzo-tests</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.eclipse.jetty</groupId>
```
romartin: should we move this exclusion to the parent KIE BOM dependency management section? or do we have to add it on each pom we need that testing artifact?

PR #765, opened 2017-04-04
JBPM-5781: Stunner - Chrome and Firefox have differences in work with Element Library

pefernan:
Modified html templates to avoid errors dragging palette elements to canvas on Firefox.

@romartin can you take a look please?


romartin:
Hey @pefernan , well done!! thanks! :)

PR #772, opened 2017-04-10
JBPM-5603: setting the rendering mode of the subform widget to read only mode, if the field is set to read only

alessioscalici:
Setting the rendering mode of the subform widget to read only mode, when the field is set to read only.
@pefernan can you review please?


pefernan:
ok to test

manstis:
Any tests?

pefernan:
+1 
@manstis that part is hard to test since it's not implementing MVP pattern and is mixing some presenter & client code (my fault here). I'm planning to refactor all this module in a middle term as a part of a related development, obviously I'll include testing for all of this components. Works for you?

manstis:
@pefernan Works for me.

PR #773, opened 2017-04-11
JBPM-5446: Validations & notifications - Validators, messages, i18n, cleans & bug fixings.

romartin:
Hey @manstis @hasys ,

Here is a good one.... sorry as it is quite big, but finally I decided to complete all missing tasks and cleans about validations the user's feedback, as was making no sense to fix that stuff but keep the messages hidden or just impossible to understand.. See [the JIRA ticket](https://issues.jboss.org/browse/JBPM-5446) as well.

The key points are:
* Some bug fixings found for the evaluation handlers
* Validations:
** Some bug fixings found on the existing graph validator
** Implemented the Validator type for Domain object beans present in the graph
** Implemented the Validator type for Diagrams - to provide complete graph structure & bean validations.
* Improved editor's notifications and the resulting messages and feedback to the user
* I18N for violations and commands messages 
* Removed canvas validation control type

PS: Once merged, will cherry-pick to `7.0.x` branch as the ticket is blocker.

Thanks!
Roger


kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/validation/impl/DiagramValidatorImpl.java
```diff
@@ -0,0 +1,116 @@
...
+        this.modelValidator = modelValidator;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public void validate(final Diagram diagram,
+                         final Consumer<Collection<DiagramElementViolation<RuleViolation>>> callback) {
+        final Graph graph = diagram.getGraph();
+        final List<DiagramElementViolation<RuleViolation>> violations = new LinkedList<>();
+        graphValidator.validate(graph,
```
manstis: Wozza... lamdas make things much easier to read ;-) 

Took a little while to follow this (but that's me being dumb!)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/CardinalityMaxRuleViolation.java
```diff
@@ -16,24 +16,33 @@
...
+
+    @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{candidate, restrictedOccurrences, currentOccurrences});
     }
 
     @Override
     public String getMessage() {
-        return "Label ['" + target + "'] can have a maximum '" + restrictedOccurrences + "' of '" + candidate + "' roles. Found '" + currentOccurrences + "'.";
+        return "The diagram requires a minimum of '" + restrictedOccurrences + "' occurrences for the candidate with roles '" + candidate + "', but only found '" + currentOccurrences + "' occurrences.";
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/CardinalityMinRuleViolation.java
```diff
@@ -16,24 +16,33 @@
...
+
+    @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{candidate, restrictedOccurrences, currentOccurrences});
     }
 
     @Override
     public String getMessage() {
-        return " Label ['" + target + "'] require a minimum '" + restrictedOccurrences + "' of '" + candidate + "' roles. Found '" + currentOccurrences + "'.";
+        return "The diagram requires a minimum of '" + restrictedOccurrences + "' occurrences for the candidate with roles '" + candidate + "', but only found '" + currentOccurrences + "' occurrences.";
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/ConnectionRuleViolation.java
```diff
@@ -16,26 +16,30 @@
...
     }
 
     @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{role, allowedConnections});
+    }
+
+    @Override
     public String getMessage() {
         return "Edge does not emanate from a GraphNode with a permitted Role nor terminate " +
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/ContainmentRuleViolation.java
```diff
@@ -33,6 +34,11 @@ public ContainmentRuleViolation(final @MapsTo("parent") String parent,
     }
 
     @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{parent, candidate});
+    }
+
+    @Override
     public String getMessage() {
         return "The '" + parent + "' does not accepts containment for the candidate with labels ['" + candidate + "'].";
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/ContextOperationNotAllowedViolation.java
```diff
@@ -16,25 +16,32 @@
...
 
     @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{context});
+    }
+
+    @Override
     public String getMessage() {
-        return "No rules for context [" + name + "]";
+        return "[" + context + "] operation not allowed";
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/DockingRuleViolation.java
```diff
@@ -33,6 +34,11 @@ public DockingRuleViolation(final @MapsTo("parent") String parent,
     }
 
     @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{parent, candidate});
+    }
+
+    @Override
     public String getMessage() {
         return "The '" + parent + "' does not accepts docking for the candidate with roles ['" + candidate + "'].";
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/EdgeCardinalityMaxRuleViolation.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{node, edge, direction, restrictedOccurrences, currentOccurrences});
+    }
+
+    @Override
+    public String getMessage() {
+        return "The node '" + node + "' can have a maximum " +
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/EdgeCardinalityMinRuleViolation.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{node, edge, direction, restrictedOccurrences, currentOccurrences});
+    }
+
+    @Override
+    public String getMessage() {
+        return "The node '" + node + "' can have a minimum " +
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/BoundsExceededViolation.java
```diff
@@ -0,0 +1,50 @@
...
+    public Optional<Object[]> getArguments() {
+        return Optional.of(new Object[]{getUUID(),
+                bounds.getLowerRight().getX(),
+                bounds.getLowerRight().getY()});
+    }
+
+    // Diagram bounds exceeded [candidate='{0}',maxX='{1}, maxY='{2}]
+    @Override
+    public String getMessage() {
+        return "Diagram bounds exceeded "
```
manstis: How is this i18n'ed?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/violations/CardinalityMaxRuleViolation.java
romartin: Hey @manstis , this message and the next ones in the comments below are not i18, theses are just for the application internal logs... So if you realize the `getMessage()` method is no longer on the interface, it's just internally used on the toString. 
The I18n for rule validations is now handled by the `StunnerTranslationService`, and the properties are based on the class name as for the key, see `StunnerCoreConstrants.properties` If you think so we can change that Strings and provide different toString results, doesn't mind for me! :) Thx

manstis: @romartin np... if the English texts are just used for logs then I'm happy :-)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/validation/impl/DiagramValidatorImpl.java
romartin: @manstis well yes, tbh I was looking at that class yesterday and realized about this too!! so we're two dumbs :)
The true is that this way, rather than focusing on if producer/consumer, there is no imperative code, and if the methods and variable names are good it really provides, when first looking, the functional operations you want to achieve, rather than focusing on structural code. 
I'm sure it can be improved again and probably make it easier to understand... I'm all ears :)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/validation/impl/DiagramValidatorImpl.java
```diff
@@ -0,0 +1,116 @@
...
+                                Optional.of((e, v) -> consumeValidations(() -> violations).accept(e,
+                                                                                                  v)),
+                                // At this point all violations have been already consumed, so no need
+                                // to use the resulting ones here.
+                                vs -> callback.accept(violations)
+        );
+    }
+
+    private BiConsumer<Element, Collection<RuleViolation>>
+    consumeValidations(final Supplier<List<DiagramElementViolation<RuleViolation>>> violations) {
```
romartin: yep agree

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/handler/impl/NodeDockingEvaluationHandler.java
```diff
@@ -63,31 +63,28 @@ public NodeDockingEvaluationHandler(final DefinitionManager definitionManager,
...
     public boolean accepts(final CanDock rule,
                            final NodeDockingContext context) {
-        final String parentId = evalUtils.getElementDefinitionId(getParent(context));
-        // As for acceptance checking, the delegated handler only needs the parent node id, no need
-        // to calculate roles for the candidate node.
-        return dockingHandler.accepts(rule,
-                                      RuleContextBuilder.DomainContexts.docking(parentId,
-                                                                                Collections.emptySet()));
+        final Set<String> parentLabels = evalUtils.getLabels(context.getParent());
+        return // As for acceptance checking, the delegated handler only needs the parent node id, no need
```
romartin: sure

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/notification/NotificationsObserver.java
```diff
@@ -0,0 +1,178 @@
...
+import org.kie.workbench.common.stunner.core.client.validation.canvas.CanvasValidationFailEvent;
+import org.kie.workbench.common.stunner.core.client.validation.canvas.CanvasValidationSuccessEvent;
+import org.kie.workbench.common.stunner.core.command.Command;
+import org.kie.workbench.common.stunner.core.command.CommandResult;
+import org.kie.workbench.common.stunner.core.rule.RuleViolation;
+import org.kie.workbench.common.stunner.core.validation.DiagramElementViolation;
+import org.uberfire.mvp.ParameterizedCommand;
+
+@Dependent
+public class NotificationsObserver {
```
romartin: well @manstis  that's a thing I'm not really 100% sure... 
I mean the `null` still exist in Java8 (an so on) and can be used!! hehe. In fact, being strict, the use of Optional have a cost, although probably negligible, but it has some cost.  So I'm not sure if from now on we should be using `Optional`'s everywhere, from my understanding it's good for a public API, where it gets clear that the argument is not mandatory, but for internal (private) use, why not keep using `null`? This is why I decided to keep some of the internal `null` references. Anyway just brainstorming, what you think?? Thx!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/resources/org/kie/workbench/common/stunner/core/StunnerClientCommon.gwt.xml
```diff
@@ -26,5 +26,6 @@
   <inherits name="org.kie.workbench.common.stunner.core.StunnerCoreCommon"/>
 
   <source path="client"/>
+  <source path="" includes="StunnerCoreEntryPoint.java"/>
```
romartin: mmm sorry not following you Mike, what you mean? I was following a supposed best practice that @pefernan told me hehe... it's about adding the entry point class here.... not sure tbh.... what you think should be better?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/pom.xml
```diff
@@ -100,6 +100,22 @@
       <artifactId>uberfire-nio2-model</artifactId>
     </dependency>
 
+    <!-- Java Bean & Errai validations. -->
```
romartin: It's nothing about AppFormer right now, but yeah, if you look at the beans used in Stunner for the different `Definitions`, they've some JSR303 annotations, right now those annotations were only being validated on the forms side, and was not difficult to achieve calling the bean validator from the Stunner main validation too, so that's it. It's for being consistent as well, if forms are validating the inputs, I should validate the bean too when validating or saving the diagram.
On the other hand, this commit provides the first part of the validations stuff - As you say, right now these JSR303 and either the rule validations on stunner are only firing on client side. Next step is to make those validations fire on backend as well, should be really easy as all that part is shared code and it's just reusing existing components.

romartin: So yeah remember that we can place JSR303 annotations on the beans, if you need so!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/GraphUtils.java
```diff
@@ -152,6 +157,33 @@ public String getElementDefinitionId(final Element<?> element) {
...
+    public List<String> getParentIds(
+            final Graph<? extends DefinitionSet, ? extends Node> graph,
+            final Element<?> element) {
+        return getParentIds(definitionManager,
+                            graph,
+                            element);
+    }
+
+    @SuppressWarnings("unchecked")
+    public static List<String> getParentIds(final DefinitionManager definitionManager,
```
romartin: yeahh agree... that part sucks... :) lemme give it a look

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/RuleManagerImpl.java
```diff
@@ -55,37 +55,38 @@ public RuleViolations evaluate(final RuleSet ruleSet,
...
-        return !empty[0] &&
-                (noRuleMatch[0] && context.isDefaultDeny()) ?
-                getDefaultViolationForContext(context) :
-                results;
+        final DefaultRuleViolations results = new DefaultRuleViolations();
+        final boolean hasRules = ruleSet.getRules().iterator().hasNext();
+        if (hasRules) {
+            final boolean[] hasEvaluations = {false};
+            ruleSet.getRules().forEach(rule -> {
+                final Optional<RuleViolations> violations = evaluate(
```
romartin: ops sorry not following you again... a language issue :) "nit-picking"??

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/resources/org/kie/workbench/common/stunner/core/StunnerClientCommon.gwt.xml
pefernan: @manstis @romartin
I know that's it's not a "best practice" but I had to do it like this when I first did the StunnerTranslationService stuff to handle the I18n for DefinitionSet's & Forms. If I remember well at that time I had to add the i18n properties files on the -api module so it would be possible having i18n on client & backend. For some reason errai TranslationService was unable to load @Bundles using absolute paths, so I had to do this trick to be able to use a relative path on the EntryPoint @Bundle. 

So the two points here are:
- Probably in this case the EntryPoint could be moved to the clent package.
- If the error on errai is already fixed (Christian was looking at it at that time, and I forgot to ask him) I can move the EntryPoints I added outside of the client package.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/GraphUtils.java
romartin: hey @manstis , just looking at that GraphUtils class now... and I think probably the more clear would be to keep all methods static and use the definitionManager instance as method arguments, if the method requires it. So it will never be injected anymore... if you need some util method here, just call it passing the necessary managed beans it requires.
What you think? If you agree, as it's an easy change, but will modify several classes, what you think about thing another PR just after this one only for this refactor?

romartin: On the other hand, another option is to split this class into an Utils and some Helper type that can be injected. I say this because some of this static methods are being called from non-managed beans, so it cannot be injected.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/notification/NotificationsObserver.java
manstis: @romartin Yeah, I can live with a difference of opinion on implementation however IMO use of ```Optional``` avoids potential missed ```null``` checks in the _internal_ implementation. IMO it's more plausible the public API may allow ```null``` values as any public API method implementation should really check the parameters are valid before anything else. My 2c. If nobody else has an opinion then happy to let it go.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/GraphUtils.java
manstis: I don't mind either way; just consistency is better than inconsistency.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/rule/RuleManagerImpl.java
manstis: oh, just the parameters on a new line after the opening bracket. https://en.oxforddictionaries.com/definition/nitpicking

manstis:
@romartin Probably just ```GraphUtils``` usage consistency to fix; then this can be merged IMO.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/notification/NotificationsObserver.java
romartin: yeah agree on both arguments you say, good points! Internally helps as well to avoid the `null` checking's and for public API that's true, arguments should be null checked, anyway I would prefer on public API using Optional rather than using javadocs to specify that semantics.. :) Thanks for the feedback, you convinced me for changing those internal fields, np!! xD

romartin:
hey @manstis , great thanks! Lemme refactor a bit those Optoinal stuff and then moving all GraphUtils methods to static - this probably will increase this PR's size, but it should not be too much problematic... so working on this!

romartin:
Hey @manstis , added a new commit which refactors the internal use of `Optional`'s and the `GraphUtils` as well. Much better now yeah, thanks for the review!!

PR #782, opened 2017-04-17
Library fixes and improvements

paulovmr:
GUVNOR-3009: [Library] Create a root screen on library (tiles) to navigate and create Organizational Units (Teams)
GUVNOR-3007: [Library] Save the project explorer state (opened/closed) by user
GUVNOR-2905: Library: Remove the "Open Project Editor" button from the project explorer.
GUVNOR-3005: Remove search textbox from the menu header
JBPM-5661: [Security management] Library perspective allows to create a project even if it is disabled


paulovmr:
Depends on: https://github.com/kiegroup/guvnor/pull/442

paulovmr:
@Rikkola Hi Toni, if you could take a look please :)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryServiceImpl.java
```diff
@@ -300,18 +314,21 @@ private OrganizationalUnit getDefaultOrganizationalUnit() {
...
 
         final List<Repository> repositories = new ArrayList<>(ou.getRepositories());
-        final Optional<Repository> primaryRepository = repositories.stream()
-                .filter(r -> r.getAlias().equalsIgnoreCase(primaryDefaultRepositoryName))
-                .findAny();
-        final Optional<Repository> secondaryRepository = repositories.stream()
-                .filter(r -> r.getAlias().equalsIgnoreCase(secondaryDefaultRepositoryName))
-                .findAny();
-
-        if (primaryRepository.isPresent()) {
```
Rikkola: Do we have tests for these `isPresent()` checks?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/resources/i18n/LibraryConstants.java
```diff
@@ -20,63 +20,108 @@
...
+    @TranslationKey(defaultValue = "Create {0}")
+    public static final String CreateOrganizationalUnit = "CreateOrganizationalUnit";
+
+    @TranslationKey(defaultValue = "Save")
+    public static final String Save = "Save";
+
+    @TranslationKey(defaultValue = "Cancel")
+    public static final String Cancel = "Cancel";
+
+    @TranslationKey(defaultValue = "Removing...")
```
Rikkola: Just checking that the "..." here and in the other locations in this file comes from UX. Some time ago I was asked to remove all the dotdotdots from the workbench. Then the policy was to not use them, but these things do change.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/OrganizationalUnitsScreen.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @PostConstruct
+    public void init() {
+        view.init(this);
+        if (organizationalUnitController.canReadOrgUnits()) {
+            libraryService.call((List<OrganizationalUnit> allOrganizationalUnits) -> {
+                organizationalUnits = allOrganizationalUnits.stream()
+                        .filter(ou -> organizationalUnitController.canReadOrgUnit(ou))
+                        .collect(Collectors.toList());
+                view.setup(organizationalUnits);
```
Rikkola: Is it possible to filter the OUs in server side? Even when the list is not visible now the OU names still leak to client.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/OrganizationalUnitsScreen.java
```diff
@@ -0,0 +1,127 @@
...
+        view.setup(filteredOrganizationalUnits);
+    }
+
+    public boolean canCreateOrganizationalUnit() {
+        return organizationalUnitController.canCreateOrgUnits();
+    }
+
+    @WorkbenchPartTitle
+    public String getTitle() {
+        return "Organizational Units Screen";
```
Rikkola: Does this need to be i18nized?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/OrganizationalUnitsView.java
```diff
@@ -0,0 +1,116 @@
...
+        if (!presenter.canCreateOrganizationalUnit()) {
+            createOrganizationalUnit.setHidden(true);
+        } else {
+            createOrganizationalUnit.setTextContent(ts.format(LibraryConstants.CreateOrganizationalUnit,
+                                                              translationUtils.getOrganizationalUnitAliasInSingular()));
+        }
+        title.setTextContent(translationUtils.getOrganizationalUnitAliasInPlural());
+        filterName.setAttribute("placeholder",
+                                ts.getTranslation(LibraryConstants.FilterByName));
+        cardByOrganizationalUnitIdentifier = new HashMap<>();
```
Rikkola: Here I would make this view, or any view, as "stupid" as possible. 

The `if` here could be in presenter and as could this variable `cardByOrganizationalUnitIdentifier`.
This way you see what the component does by looking at the presenter and the functionality can be tested with the unit test.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/popup/OrganizationalUnitPopUpPresenter.java
```diff
@@ -0,0 +1,229 @@
...
+        this.organizationalUnitService = organizationalUnitService;
+        this.afterCreateOrganizationalUnitEvent = afterCreateOrganizationalUnitEvent;
+        this.afterEditOrganizationalUnitEvent = afterEditOrganizationalUnitEvent;
+        this.notificationEvent = notificationEvent;
+        this.organizationalUnitController = organizationalUnitController;
+    }
+
+    @PostConstruct
+    public void setup() {
+        view.init(this);
```
Rikkola: Haven't used the `UberElement<Presenter>` myself. I see in `LibrarySceen` this `init` call is not needed, but here it is. When does Uberfire automatically do this for us?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/organizationalunit/OrganizationalUnitTileWidget.java
```diff
@@ -0,0 +1,171 @@
...
+        this.notificationEvent = notificationEvent;
+        this.organizationalUnitController = organizationalUnitController;
+        this.libraryInternalPreferences = libraryInternalPreferences;
+    }
+
+    public void init(final OrganizationalUnit organizationalUnit) {
+        view.init(this);
+        organizationalUnitPopUpPresenter.setup();
+        view.setup("fa-folder",
+                   "fa-folder-open",
```
Rikkola: Are the top two values always `"fa-folder"` and `"fa-folder-open"`? If yes then they could be in view.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/NewProjectScreen.java
```diff
@@ -161,7 +161,7 @@ private void hideLoadingBox() {
 
     @WorkbenchPartTitle
     public String getTitle() {
-        return ts.getTranslation(LibraryConstants.NewProjectScreen);
+        return "New Project Screen";
```
tomasdavidorg: Same question as @Rikkola has for the "Organizational Units Screen".

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/TranslationUtils.java
```diff
@@ -0,0 +1,68 @@
...
+                                    organizationalUnitAliasInSingular = loadedLibraryPreferences.getOrganizationalUnitPreferences().getAliasInSingular();
+                                    if (organizationalUnitAliasInSingular == null || organizationalUnitAliasInSingular.isEmpty()) {
+                                        organizationalUnitAliasInSingular = ts.format(LibraryConstants.OrganizationalUnitDefaultAliasInSingular);
+                                    }
+                                    organizationalUnitAliasInPlural = loadedLibraryPreferences.getOrganizationalUnitPreferences().getAliasInPlural();
+                                    if (organizationalUnitAliasInPlural == null || organizationalUnitAliasInPlural.isEmpty()) {
+                                        organizationalUnitAliasInPlural = ts.format(LibraryConstants.OrganizationalUnitDefaultAliasInPlural);
+                                    }
+                                },
+                                error -> {
```
tomasdavidorg: This repeats several times. Would it makes sense to introduce load method without error callback?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryServiceImpl.java
paulovmr: Yes, we have the `LibraryServiceImplTest#getOrganizationalUnitRepositoryInfoTest()`.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/resources/i18n/LibraryConstants.java
paulovmr: It came from me, didn't know about this. I removed it now.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/OrganizationalUnitsScreen.java
paulovmr: Actually the server side already checks this, so it's an unnecessary double check. I removed it from the client now.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/OrganizationalUnitsScreen.java
paulovmr: No, we don't show the screen title anywhere, because the panel used doesn't use Uberfire's header. But still, the method is required to be implemented.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/OrganizationalUnitsView.java
paulovmr: Fixed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/popup/OrganizationalUnitPopUpPresenter.java
paulovmr: The `init` is called only when the presenter is annotated with `@WorkbenchScreen`, which is the case of `LibraryScreen`.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/organizationalunit/OrganizationalUnitTileWidget.java
paulovmr: There is an idea to allow the user to specify icons when creating an OU in the future, so I thought it was better to leave like this to make it easier later.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/NewProjectScreen.java
paulovmr: No, we don't show the screen title anywhere, because the panel used doesn't use Uberfire's header. But still, the method is required to be implemented.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/TranslationUtils.java
paulovmr: In this case, I should have treated the error, I fixed it now. The others don't need to be treated. I think it's better to keep this API to force the developer to treat the error, or at least think about it.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryServiceImpl.java
```diff
@@ -354,11 +378,11 @@ private Repository createDefaultRepository(final OrganizationalUnit ou,
     }
 
     private String getPrimaryDefaultRepositoryName(final OrganizationalUnit ou) {
-        return getPreferences().getRepositoryAlias();
+        return getPreferences().getRepositoryPreferences().getName();
     }
 
     private String getSecondaryDefaultRepositoryName(final OrganizationalUnit ou) {
```
jhrcek: @paulovmr Could you please explain to me what's the purpose of the `secondaryDefaultRepositoryName` and when would it be used instead of primary name?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/popup/OrganizationalUnitPopUpPresenter.java
```diff
@@ -0,0 +1,228 @@
...
+    private void validateFields(final Command successCallback) {
+        final String name = view.getName();
+        if (isEmpty(name)) {
+            view.hideBusyIndicator();
+            view.showError(view.getEmptyNameValidationMessage());
+            return;
+        }
+
+        final String defaultGroupId = view.getDefaultGroupId();
+        if (isEmpty(defaultGroupId)) {
```
jhrcek: Is this the only validation for group Id? Not sure if it's needed but I figure since this is going to be used as groupId / default package name for projects of this team (am I right?) perhaps it would make more sense to restrict this some more [like this](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html). WDYT?

manstis:
There is a GAV element validation service in kie-wb-common that can
validate Group Id, Artifact Id and Version.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryServiceImpl.java
paulovmr: The repository name must be unique in the whole workbench, and not just in its organizational unit. So if the primary name was already taken by another OU, we should check for an alternative name, that uses the OU name to prevent conflict.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/popup/OrganizationalUnitPopUpPresenter.java
paulovmr: No, I also validated if this is a valid group ID in the same method, lines 192 to 293.

manstis:
@jhrcek @tomasdavidorg Any more questions before I merge (the 2 PRs)?

tomasdavidorg:
No, thanks @paulovmr and @manstis.

jhrcek:
@paulovmr Thanks for answers, @manstis +1 to merge

PR #791, opened 2017-04-20
JBPM-5718: Persistence id of the Data Object is present in the form-modeler.

pefernan:
Making properties annotated with the JPA Id annotation not available on forms.

@jsoltes @wmedvede can you take a look?

PR #792, opened 2017-04-20
JBPM-5718: Persistence id of the Data Object is present in the form-modeler.

pefernan:
Making properties annotated with the JPA Id annotation not available on forms.

@jsoltes @wmedvede can you take a look?


jsoltes:
@manstis could you please wait a little bit next time - so that i can also have a proper look - before you merge it?

pefernan:
@manstis @jsoltes do you want me to send another PR with the same changes than https://github.com/kiegroup/kie-wb-common/pull/791?

jsoltes:
@pefernan If it's only about removing the dependency on mockito-core, I can  do it as part of my next pull-request... no need to do a separate pull-request for that.

pefernan:
+1

PR #798, opened 2017-04-20
GUVNOR-3030: Resizing Project Explorer Dock throws NPE

manstis:
See https://issues.jboss.org/browse/GUVNOR-3030


manstis:
@wmedvede @ederign You're both probably best versed with "Dock" stuff.. mind reviewing?

PR #808, opened 2017-04-24
GUVNOR-2921: Importing example projects does not work on Windows

mcimbora:


mcimbora:
@manstis This is 7.0.x blocker, could you please take a look? Thanks in advance.

PR #811, opened 2017-04-24
 GUVNOR-2787 : [Library] Use Lucene to list assets in the project page

Rikkola:
Please do not merge yet.

Still updating the i18n, but since this would be good to have in today making it so that this can be reviewed early.

https://issues.jboss.org/browse/GUVNOR-2787

Extending the refactoring service for file listing. I think the name should be changed to just file search service or something similar, since it is really a query service for files and both repository and project screen in library will use the same data. Still left this out since then I would also have to rename the modules and that is a big change during code freeze. 

Related PRs

https://github.com/kiegroup/kie-wb-common/pull/811
https://github.com/kiegroup/jbpm-designer/pull/581
https://github.com/kiegroup/drools-wb/pull/465


Rikkola:
I think @paulovmr is busy today, so if you @manstis have time to review, I would appreciate it.

kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/main/java/org/kie/workbench/common/services/refactoring/backend/server/query/standard/FindRulesByProjectQuery.java
```diff
@@ -71,7 +71,9 @@ public Query toQuery( final Set<ValueIndexTerm> terms ) {
             queryBuilder.addRuleNameWildCardTerm();
         }
 
-        return queryBuilder.build();
+        Query build = queryBuilder.build();
+        String s = build.toString();
```
manstis: Debugging!?!

kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/test/java/org/kie/workbench/common/services/refactoring/backend/server/query/findresources/FindAllLibraryAssetsQueryTest.java
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
```
manstis: 2017

kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/main/java/org/kie/workbench/common/services/refactoring/backend/server/query/standard/FindRulesByProjectQuery.java
Rikkola: Yep. Well spotted.

Rikkola:
Updated, but still please hold the merge while a bug related to accessing the index with super dev mode is being fixed.

Rikkola:
@manstis and about the `Observer`. We can only use that after repository is created/imported. So it does not work for huge commits like the example import for example. 
But a better way would be to somehow know if a "huge" import has started and finished. Maybe this is possible in Uberfire.

manstis:
@Rikkola I get the following when trying to run:

library-error

manstis:
@Rikkola OK, I hacked a fix locally for the above. I notice the "Page size" selector shows "6, 15, 25..."

library-page-size

Rikkola:
@manstis Looks like the mainActionSlot is a left over from a change I left out. Going to remove it.

And 6 probable does not make sense, so going up from 15 is good enough. Going to fix that too.

manstis:
@Rikkola I'm not sure the correct index is being used... which would explain why, after a stop and restart with SDM, the index retrieval returns zero documents (as there's been a ```mvn clean``` which would delete transient indexes folders). If you create a new team, new project and add an item it appears in the Library asset list, add another; it appears. Repeat. They appear again and again. Try again with another new team but import an example. Add a new asset; it does not appear. Repeat until you're blue in the face. Move up and down the Team/Repository hierarchy. No new items appear.

manstis:
When importing an example; the repository is first cloned to ```/tmp``` (or at least it used to not sure now where, but @mcimbora would know); before being cloned into ```.niogit```. I suspect your index stuff might be using the original "clone" as opposed to the "imported clone".... still checking....

manstis:
..hmm, but stopping and restarting shows no assets for either repository.. Need to delete everything and try with a "clean" install so I know what's "new" and what isn't "old stuff"... but it'll have to wait for tomorrow.

manstis:
@Rikkola OK.. when re-running from SDM within IDEA (not tried CLI) the ```.index/repo-name``` folder is emptied and a ```write.lock``` file added (or not deleted.. not checked). When VFS starts up it checks if a ```.index/repo-name``` folder exists and if it does **NOT** exist runs ```BatchIndex```. This is why the indexes are empty on re-start.

manstis:
@Rikkola try using the System Property ```org.uberfire.metadata.index.dir``` to move the Index out of IDEA's control. For example ```-Dorg.uberfire.metadata.index.dir=/tmp/lucene```. I tried locally and re-starting IDEA no longer affects the index used by Library

jhrcek:
@manstis @Rikkola please don't merge until QE has had time to review/test this thoroughly

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java
```diff
@@ -166,39 +172,81 @@ Command detailsCommand(final Path assetPath) {
...
-        }).getProjectAssets(projectInfo.getProject());
+        }).getProjectAssets(new ProjectAssetsQuery(projectInfo.getProject(),
+                                                   view.getFilterValue(),
+                                                   firstIndex,
+                                                   view.getStep()));
     }
 
-    private void loadProject(List<AssetInfo> assets) {
-        setupAssets(assets);
+    private int getFirstIndex() {
```
jhrcek: This is non-trivial calculation. Could you please add few unit tests, mocking view and trying with various values for step and page number?

Rikkola:
@jhrcek Sure, we can wait. Going to do the small fixes @manstis picked up.

Rikkola:
@jhrcek By the way. There is a known bug on file search by name. Haven't picked up the details yet. For example LoanApplication is picked up with search term "Loan", but file "ggggg.drl" is not picked up with "gg".

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java
Rikkola: Will do.

mcimbora:
@manstis @Rikkola Content of the playground jar is first copied into [.kie-wb-playground](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-examples-screen/kie-wb-common-examples-screen-backend/src/main/java/org/kie/workbench/common/screens/examples/backend/server/ExamplesServiceImpl.java#L134) directory located in user.dir directory. A git repository is initialized in that directory so that our API can consume it. After that, the repository is [cloned](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-examples-screen/kie-wb-common-examples-screen-backend/src/main/java/org/kie/workbench/common/screens/examples/backend/server/ExamplesServiceImpl.java#L239) when fetching its projects. Skimming the code I see a potential issue - the repository is cloned every time the projects are queried, which is unnecessary for kie-wb-playground projects (should be one-timer instead).

mcimbora:
I'll take a look at it.

manstis:
@Rikkola I made a new asset "gggggg.drl" (6 g's) in "mortgages" default package and a filter of "gg" correctly returned it. I also created a new asset "ggggg.drl" (5 g's, as per your example) in "mortgages" default package and a filter "gg" correctly showed both new assets.

manstis:
@Rikkola .. and adding a new asset "ggg.drl" in the "mortgages/mortgages" package of "mortgages" correctly listed all three new assets with a filter "gg"... I conclude, without specific failing example, that filtering does work (but wait to be proven wrong by QE or you...!)

Rikkola:
@manstis Ok, going to test soon and see what I did it.

manstis:
@Rikkola I would comment that if the number of results is less than the page size then the page forward/backward buttons and page number textbox should all be disabled.

Rikkola:
@manstis Makes sense

jhrcek:
I finally built this PR locally and was playing manually with it. Here are some high-level observations:

1) Wouldn't it be better to use built-in pager component for browsing asset list (either UberfireSimplePager or com.google.gwt.user.cellview.client.SimplePager)? The current hand-rolled paging implementation looks to me like unnecessarily introducing potential for new bugs. Not to mention that it doesn't behave as usual pagers behave. For example I can navigate to pages past the last page, where it's showing the "End of file list". Standard practice is to detect that there are no more pages available (# items on current page < page size) and disabling the "next" button -> you wouldn't need this "End of file list" view with "please go back" plea then :-)

2) It seems that other project have their custom indexers - you'll need to update those as well, e.g when you create prosess using old Designer, you'll get this in server log
Caused by: java.lang.NoSuchMethodError: org.kie.workbench.common.services.refactoring.backend.server.indexing.DefaultIndexBuilder.<init>(Lorg/guvnor/common/services/project/model/Project;Lorg/guvnor/common/services/project/model/Package;)V
	at org.jbpm.designer.server.indexing.BPMN2FileIndexer.getIndexBuilder(BPMN2FileIndexer.java:216)

3) When I clone any repo from Administration perspective, it's not shown in library at all..

4) When I clone our bpms-assets repo using "Import Example" > Other in library, the indexing process there are many errors during indexing like `EJB Invocation failed on component IncrementalBuilderExecutorManager for method public void org.guvnor.common.services.builder.IncrementalBuilderExecutorManager.execute(org.guvnor.common.services.builder.AsyncIncrementalBuilder): javax.ejb.ConcurrentAccessTimeoutException: WFLYEJB0241: EJB 3.1 PFD2 4.8.5.5.1 concurrent access timeout on IncrementalBuilderExecutorManager - could not obtain lock within 5000MILLISECONDS` in error log.
In the end our main problem was not solved -> I can't still get to a point when I'm able to open guvnor project in our bpms-assets repo 😢 

Note I'm always starting my kie-wb installation with clear after installation state (I've got EAP + kie-wb setup versioned in git and resetting the git to clean installation state.

Although it might seem that I'm tearing this PR apart, I can attest that the speed improvements of asset list loading are really significant. When I actually get some assets to be displayed, stuff seems to be working fine (asset list being updated after adding / deleting etc, all this stuff is way faster than before. But we still haven't let our selenium dog loose yet :P

Rikkola:
1. I think the list comes from this wireframe http://huxhzq.axshare.com/#p=project&g=1 and the pagination comes from Pattern Fly patterns http://www.patternfly.org/pattern-library/navigation/pagination/. We can do the disable of buttons if there is less on the page than the page can fit, but if there is exactly the amount of items left that the page can fit the next page will be empty and we still need the "end of documents" to be shown. One of the issues with Lucene is that I'm not sure if we can get an actual count of the documents that it has, but if we can this can be fixed in a cleaner way.
2. Ok. Looks like I built the wrong project when I tried the jbpm indexers. These can be fixed.
3. Depends where this happens, but probably an issue with Lucene indexing or some cache not refreshing.
4. Another indexing bug. I doubt it is introduced by this.

And I did not dream that this would work 100% once it goes out ;-) Nothing ever does. I tested with the huge test repo, but the difference with that might be that there is no real content in the files and some of the indexing covers file content and to do this they even compile the files. So even if indexing takes 10 seconds the build can take an hour. We could fix this with splitting the indexing to different levels.

manstis:
@jhrcek Please remind me (via Red Hat email if you prefer) what the URL is for "bpms-assets". I doubt the errors you report for (4) have anything to do with the Lucene related changes; but plausibly the import mechanism - incremental builds are possibly coming from the copying of assets between repositories (the "examples" repository bundled with the WAR and the clone created when an example is imported).

Rikkola:
Ok. Almost everything mentioned in this PR is fixed now. 
Next checking the item #3 in @jhrcek's list.
> When I clone any repo from Administration perspective, it's not shown in library at all..

Rikkola:
@jhrcek When you say "in library at all" does that mean you can not see the repository, projects in the repository or assets in the projects?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectView.html
```diff
@@ -45,6 +45,43 @@
 
                 <div class="list-group list-view-pf list-view-pf-view" id="asset-list">
                 </div>
+                <div id="indexing-info" visible="false" class="blank-slate-pf">
+                </div>
+                <div>
+                    <form class="content-view-pf-pagination table-view-pf-pagination clearfix" id="pagination1">
+                        <div class="form-group">
+                            <select id="howManyOnOnePage" class="form-control pagination-pf-pagesize" style="width: 45px;">
```
paulovmr: I guess the UI/UX here could be improved, but it could be done in a next PR, since this is better then what we have now.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectView.java
```diff
@@ -125,9 +152,134 @@ public void addAsset(final String assetName,
...
+    public void showIndexingIncomplete() {
+        emptyState.setMessage(ts.getTranslation(LibraryConstants.IndexingHasNotFinished),
+                              ts.getTranslation(LibraryConstants.PleaseWaitWhileTheProjectContentIsBeingIndexed));
+        showEmptyState();
+    }
+
+    private void showEmptyState() {
+
+        indexingInfo.setClassName("blank-slate-pf");
+        indexingInfo.getStyle()
```
paulovmr: @Rikkola You could have used `indexingInfo.setHidden(false);`

jhrcek:
@Rikkola I don't know what I was doing exactly anymore :) I tried something along those lines and now I can see the repo fine. I'm fine with you ignoring the point #3 for now. Tomorrow I'll build this PR locally and test more thoroughly again.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/test/java/org/kie/workbench/common/screens/impl/LibraryServiceImplTest.java
```diff
@@ -68,6 +75,7 @@
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.*;
 
+@Ignore
```
Rikkola: Nope, that needs to go.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectView.html
Rikkola: Just a copy&paste from the Pattern Fly page. I think we should extract this as a widget and then do the improvements.

Rikkola: So that it is reusable.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectView.java
Rikkola: The problem I had was that in some browsers setting hidden was not enough and I had empty placeholder divs, but I have to confess I set the `visibility` property directly. Maybe `setHidden` does more.

Rikkola:
@jhrcek I think you probably ran into the same issue as @manstis. The repo change dropdown is in the corner and can be easily missed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectView.java
Rikkola: The problem I had was that in some browsers setting visibility was not enough and I had empty placeholder divs, but I have to confess I set the `visibility` property directly. Maybe `setHidden` does more.

paulovmr:
@jhrcek @Rikkola Also, the repo dropdown will only appear when there are more than one repository in the organizational unit.

Rikkola:
Ok. Tested with the repo @jhrcek provided and the examples we have that the this PR should be valid. Probably best to still wait for @jhrcek to double check tomorrow. This should be an easy merge and cherry-pick to 7.0.x.

jhrcek:
Hello @rikkola,

I built this PR locally and tested some more. 

Rant: It wasn't particularly easy process - building 4 repos & some rebasing was needed. I'm looking forward to when @psiroky implements support for triggering full downstream build for each PR wo we don't have to build multiple-repo PRs locally :)

Regarding previous issues: I can now import bpms-assets and open all the projects without any problems and the speed improvements are great

There are only few minor issues that I found, but I will report them in separate "library" jiras, so as not to postpone selenium testing of these changes:

1) Say you have multiple pages of assets and you go to 3rd page. You enter something into asset name filter and hit enter.
Now you'll see "End of list" view, which migh make the user think that there are no results for that filter. But they ARE there, alas on the 1st page, to which I have to manually return. The expected behavior (at least I'm used to that) in web apps is that after search the pager is reset to the first page so you can see the search results without having to bother with the pager

2) Sometimes when I click "next" button on the pager it reloads the whole web app and takes me to the home page. I don't yet have clear steps to reproduce this, but it happens ocassionaly and is annoying :P

3) We have some empty projects in bpms-assets (e.g. duplicate1), for which library is showing "Indexing has not finished" although I can clearly see "Completed indexing of default://master@bpms-assets/" in server.log. I would expect to see some kind of "empty project" view instead.

4) The "items per page" dropdown is too narrow - it doesn't fully show which number is selected
itemsperpagetoonarrow



I'll leave it up to you if you want to fix those in this PR or merge and fix later (hopefuly soon :) ). As far as I'm concerned +1 for merging.

I guess that once this is merged we'll have to test stuff thorougly to find and fix all kinds of issues related to indexing. Just mentioning this to make @kkufova (who is preparing library perspective test day) aware. It seems like a good idea to focus on scenarios like adding, renaming, deleting assets in various places (see [this fix](https://github.com/kiegroup/drools-wb/commit/e3d0b9fa6f3a5e887e225c55adc53d573dbfcce9) for inspiration about what might go wrong)

manstis:
@Rikkola please let me know your intentions...

Rikkola:
I think we can do the fixes in separate PRs. 

#1 There should be time to at least do the reset to first page on search.
# 2 Is something I thought only happens in Dev Mode, but looks like it is an actual bug. 
# 3 For this we need to fix the checkup that takes you to either normal library view or to the empty library view.
# 4 Is something we can also fix. The UI is also broken in Firefox, the buttons always go to 2 rows. Works in Chrome.

jhrcek:
@Rikkola FYI I created separated jiras for the outstanding issues mentioned in the comments:
https://issues.jboss.org/browse/GUVNOR-3042
https://issues.jboss.org/browse/GUVNOR-3043
https://issues.jboss.org/browse/GUVNOR-3044
https://issues.jboss.org/browse/GUVNOR-3045

Please prioritize fixing them.

PR #814, opened 2017-04-25
GUVNOR-3032: Example import: Optimize playground repository cloning

mcimbora:
Playground repository cloning is a costly operation, make sure it's not done repeatedly when querying for a list of available projects within the repository (e.g. when opening example import wizard multiple times)

@manstis Could you please review the PR? Thanks in advance


manstis:
Great @mcimbora thank-you.

PR #694, opened 2017-02-22
[6.5.x] RHBPMS-4564: Cloning managed repo as unmanaged causes menu on project editor displays different options for user using same roles in BxMS 6

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4564

PR #818, opened 2017-04-25
Revert "[6.5.x] RHBPMS-4564: Cloning managed repo as unmanaged causes menu on project editor displays different options for user using same roles in BxMS 6"

manstis:
Reverts kiegroup/kie-wb-common#694


etirelli:
@manstis Thank you!

PR #820, opened 2017-04-26
JBPM-5913_2: Fix End Event icons not showing in tree explorer widget …

jlindop:
…and toolbox

@romartin  This fixes the problems with End Event icons you told me about. Will you review - thanks.
I think this should be cherry-picked to 7.0.x (we should only cherry-pick critical fixes at this point). Do you agree or do you think this is not critical?


romartin:
hey @jlindop , just looking at this PR now.
BTW My understanding was that this commit (from this PR) was like a cherry-pick from the ones that fixed same error on 7.0.x, which are already merged, isn't it? Or does this commit do some different changes than current 7.0.x?

jlindop:
@romartin Thanks. I'll fix the icons.

jlindop:
@manstis @romartin  This needs to be updated before merging. I'll close the PR and open a new one when it's ready.

PR #821, opened 2017-04-26
[7.0.x] JBPM-5913_2: Fix End Event icons not showing in tree explorer…

jlindop:
… widget and toolbox

Agreed with krisv that this should go in 7.0.x
@romartin Will you review this one as well? Thanks.


romartin:
Hey @manstis @jlindop 

Address this one first because it's more urgent than the master one, right?

So the icons now appear on toolbox, etc. If this is urgent for a release, that' fine for me to merge.

But I think probably there are a couple of small issues, lemme know your thoughts!

1. The end none event border for the canvas shape is too big
2. The icon for the end terminate event (on palette/tree) appears, but it's completely fill using black color.. not sure if that's right one, should not be like black/white/black?

Thanks!

j-end-events

manstis:
@romartin I checked with @jlindop and he preferred this to be merged "as is".

He said he'd liaise with you about it.

jlindop:
@romartin  I agree about the icons - these are what we have from Brian. The endNoneEvent icon on the canvas is the icon and I made the border in the BaseEndEvent class more or less match the icon. The endTerminateEvent icon in the explorer is also the icon from Brian. Let's chat about these this afternoon.
In the meantime, manstis is going to merge, as it fixes the bigger problem you found.

PR #826, opened 2017-04-28
GUVNOR-2857: [Guided Decision Table] Only java.lang.Integer can be imported

manstis:
See https://issues.jboss.org/browse/GUVNOR-2857

This PR just expands the default list for new projects to include those requested in the JIRA.


kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/project/ProjectImportsServiceImpl.java
```diff
@@ -48,74 +49,82 @@ public ProjectImportsServiceImpl() {
...
+                            projectConfigurationContentHandler.toString(createProjectImports()));
         }
     }
 
     private ProjectImports createProjectImports() {
-        ProjectImports imports = new ProjectImports();
-        imports.getImports().addImport( new Import( "java.lang.Number" ) );
-        return imports;
+        ProjectImports projectImports = new ProjectImports();
+        final Imports imports = projectImports.getImports();
```
manstis: Add some more imports...

PR #827, opened 2017-04-28
GUVNOR-3038: [Library] Import examples not working

mcimbora:
Replace Files.delete by File.delete to avoid AccesDeniedExceptions if directory is marked as read-only (windows-related).


mcimbora:
@manstis That was approved even before I managed to ask for the approval :-) You're a rockstar

PR #836, opened 2017-05-03
[6.5.x] RHBPMS-4564: Cloning managed repo as unmanaged causes menu on project editor displays different options for user using same roles in BxMS 6

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4564

Before (or [see the video](https://drive.google.com/a/redhat.com/file/d/0B4qIOCgZfb2eeG4zN0c3Y19vN2s/view?usp=sharing)):
before

After:
after

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java
@@ -30,7 +30,9 @@ import javax.inject.Inject;
 import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwt.user.client.ui.Widget;
+import org.guvnor.asset.management.model.RepositoryStructureModel;
 import org.guvnor.asset.management.service.AssetManagementService;
+import org.guvnor.asset.management.service.RepositoryStructureService;
 import org.guvnor.common.services.project.builder.model.BuildResults;
 import org.guvnor.common.services.project.builder.service.BuildService;
 import org.guvnor.common.services.project.client.repositories.ConflictingRepositoriesPopup;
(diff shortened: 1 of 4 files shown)


karreiro:
@manstis Could you please take a look?

manstis:
@karreiro Are you sure you've done the right fix?

As far as I understand the JIRA (and looking at the [code](https://github.com/kiegroup/kie-wb-common/blob/6.5.x/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/editor/ProjectScreenPresenter.java#L408)) **if** the repository is imported as "unmanaged" then only the "Build & Deploy" menu item should be shown.

The problem is that (IIRC) ```system.git``` holds a setting for the repository that it is "managed" and hence your change uses the "known" state of a repository and shows "Build", "Build & Install" and "Build & Deploy" when cloning the repository as "unmanaged". In reality it should only show "Build".

If you re-cloned with a different name it _should_ not find any entries for the repository (as IIRC it uses the repository alias as key) and the "Build" menu would only show "Build & Deploy". Things are a little worse still as settings (from ```system.git```) are cached in ```ConfigurationServiceImpl``` and resulting ```Repository``` objects in ```ConfiguredRepositories```.

I think you are right to remove ```isRepositoryManaged()``` as it relies upon client-side state that may be stale; but I also think some of the caches mentioned above need clearing (or check they are cleared when the original repository is deleted) and the re-cloned repository's environment map has the "managed" flag not set.

Perhaps check with @hasys that my understanding of the JIRA is correct too :-)

karreiro:
Hi @manstis, thanks for the comment.

Regarding this:
> As far as I understand the JIRA (and looking at the code) if the repository is imported as "unmanaged" then only the "Build & Deploy" menu item should be shown.

I think that we don't have the concept of an unmanaged repository, which is multimodule, see:
https://github.com/kiegroup/guvnor/blob/6.5.x/guvnor-asset-mgmt/guvnor-asset-mgmt-backend/src/main/java/org/guvnor/asset/management/backend/service/RepositoryStructureServiceImpl.java#L286-L291

The load method keeps the flag (managed) consistent. So my fix was to call it.

Am I right about this?
cc @hasys

manstis:
@karreiro Great. Thanks for the explanation.

OK, let's state some facts:-

1) You can have a managed single module repository.
2) You can have an unmanaged single module repository.
3) If you have a multi-module repository it has to be managed.

So, if the repository you're importing/deleting/re-cloning is a multi-module repository (I forget it it was) then your fix for those is spot on :-) It would be good to check the other two scenarios; a single module repository that was first managed and re-cloned as unmanaged. The "Build" menu should then show only "Build & Deploy". 

If you can confirm creating/deleting/re-cloning the three different scenarios work as expected then I'll approve this PR and get it merged :-) You should be able to create a single-module unmanaged repository; reclone it as managed and the "Build" menu show the 3 options. Conversely you should be able to create a single-module managed repository; reclone it as unmanaged and the "Build" menu show the sinlge option.

karreiro:
@manstis see https://www.youtube.com/watch?v=16pUHyHtl14&feature=youtu.be :-)

PR #837, opened 2017-05-03
Stunner showcase - Added missing dashbuilder dependencies.

romartin:
Hey @manstis @hasys ,

The stunner's showcase was not starting due to some missing class dependencies to Dashbuilder, here is a fix for it.

Thanks!!


manstis:
Same as https://github.com/kiegroup/drools-wb/pull/471 (for ```drools-wb```)

PR #851, opened 2017-05-09
RHBPMS-4375 - Container is not started when being created

mswiderski:
depends on https://github.com/kiegroup/droolsjbpm-integration/pull/948

this PR introduces new checkbox on the new container wizard to allow directly start the container when created
screen shot 2017-05-09 at 19 42 19

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/wizard/container/NewContainerFormPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/wizard/container/NewContainerFormPresenter.java
@@ -66,6 +66,8 @@ public class NewContainerFormPresenter implements WizardPage {
 
         void setContainerName( final String containerName );
 
+        void setStartContainer( boolean startContainer );
+
         String getContainerName();
 
         String getContainerAlias();
@@ -75,6 +77,8 @@ public class NewContainerFormPresenter implements WizardPage {
         String getArtifactId();
 
         String getVersion();
+
+        boolean isStartContainer();
 
         void errorOnContainerName();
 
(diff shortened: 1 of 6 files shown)


kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/wizard/container/NewContainerFormPresenter.java
```diff
@@ -76,6 +78,8 @@
 
         String getVersion();
 
+        Boolean isStartContainer();
```
manstis: Not that it probably matters a great deal; but does this need to be the ```Boolean``` class as opposed to ```boolean``` primitive? (Objects introduce the possibility of ```null``` values)...

mswiderski: nothing in particular, but get used to work with Objects so it's more of a habit. If this is best to be primitive so it's consistent with other types used in kie-wb-common will update, just say a word :)

manstis: pfffff.. I hate telling people what to do. I'll leave it to your conscience... IMO ```Boolean``` suggests a possible third state (```null```) so if I picked up your code in the future I'd be looking for use of it being ```null```.. then I'd ask why was it not an ```Optional```. All of which could be resolve by using a primitive.

mswiderski: updated to use primitive, thanks @manstis

manstis:
Thanks @mswiderski

PR #852, opened 2017-05-10
JBPM-5718: Persistence id of the Data Object is present in the form-modeler

pefernan:
Removed wrong properties from property selector on create TableColumnMeta

@wmedvede @jsoltes can you take a look?


pefernan:
jenkins retest this.

PR #855, opened 2017-05-11
[7.0.x] GUVNOR-3066: GUVNOR-3043: Workbench reloading on User interactions.

manstis:
See https://issues.jboss.org/browse/GUVNOR-3043 and https://issues.jboss.org/browse/GUVNOR-3066

Use of ```<button>``` inside ```<form>``` elements leads to the form being resubmitted when either the ```<enter>``` key is pressed or the User clicks the ```<button>``` element. This is expected behaviour (and correctly implemented by Chrome).

This is the corollary of https://github.com/kiegroup/kie-wb-common/pull/854 for 7.0.x

(cherry picked from commit 888ea4130accb13298389365da0a2f528bd6af26)

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/EmptyProjectView.html
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/EmptyProjectView.html
@@ -22,9 +22,9 @@
         <div class="col-md-8 col-lg-9">
             <div class="main-content-kie">
                 <div class="toolbar-pf">
-                    <form data-field="assets-toolbar" class="toolbar-pf-actions">
+                    <div data-field="assets-toolbar" class="toolbar-pf-actions">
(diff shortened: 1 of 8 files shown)


Rikkola:
Wondering if the PatternFly team can fix the examples they have. Do they really need to use forms for some weird reason. Personally I think we should try and keep our HTML as close to the Pattern Fly spec as possible, but not sure how possible that is.

Rikkola:
And now that I'll try to find the example with form from Pattern Fly I can't. Well going to report it to them if I find it.

manstis:
@Rikkola I had a quick look (I was about to forward your proposal to the PatternFly team) however their ```form``` [examples](http://www.patternfly.org/pattern-library/forms-and-controls/forms/#/code) show ```buttton``` outside the ```form``` itself or include the ```type``` attribute. Did you have an example?

Rikkola:
@manstis Ok so looks like I had to do some changes to the examples. I remembered I only had to change the styles ( that is one thing I need to report ). Turns out I also had to change the links to buttons to catch the click events.

So Pattern Fly is clean, but I need to investigate what the best way to fix this is. We should just be able to use Pattern Fly code 1:1 in our app.

PR #861, opened 2017-05-12
Unify Plugin Management and Artifact Repository names for drools-wb

tomasdavidorg:
There are two more occurrences of the names which wasn't changed to the new form.

This is followup on https://github.com/kiegroup/kie-wb-distributions/pull/514.


tomasdavidorg:
@psiroky could you please have a look?

PR #862, opened 2017-05-12
GUVNOR-3042 :[Library] Filtering by asset name should take you to first page with results

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3042

Also while there:
- Cleaning the buttons out. I think one of my iterations captured GWTEvents so I changed to buttons and then did not change back
- Disable/enable for forward backward was also broken. Could be because of the change from button to link. Anyway fixed that.


Rikkola:
@jhrcek Ok, I think I need to drop the PatternFly styling and just go with my own setup that looks the same. Feels like it is buggy, works in one browser and then breaks in the next.

Rikkola:
Ok, splitting the list and paging logic to a separate component so they are easier to unit test. Also we can then reuse this in the project list.

ederign:
@paulovmr can you please take a look on this?

paulovmr:
@Rikkola Will you do the changes you mentioned on this PR or another one? Just asking to check if I should review it as is or wait.

Rikkola:
@paulovmr Please wait. Going quite a lot of changes, best to have it in this one.

Rikkola:
Updated with changes. Split the navigation and list logic to AssetList component. Making it easier to test and reuse.

Going to take a quick look at the UI issue that happens in Firefox I believe, worked in Chrome.

Rikkola:
Ok could not reproduce the UI issue @jhrcek reported in either Chrome or FF. Could be that my html changes fixed it.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/test/java/org/kie/workbench/common/screens/impl/LibraryServiceImplTest.java
```diff
@@ -76,6 +76,7 @@
 import static org.mockito.Mockito.*;
 
 @RunWith(MockitoJUnitRunner.class)
+@Ignore
```
Rikkola: Whoops. Removing, does this fail locally for anyone else?

Rikkola:
@paulovmr Ok, this is now ready for a review.

Rikkola:
Noticed the UI bug is there. 
Strange, but UI works in SDM and fails with the war.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/AssetListView.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    @Override
+    public void range(int from,
+                      int to) {
+        fromToRange.setInnerHTML(from + " - " + to);
+    }
+
+    private String getDisabledClassName(final boolean disabled) {
+        // TODO: test and move to presenter
```
paulovmr: @Rikkola Missed one TODO here :)

jhrcek:
Please don't merge yet, I'd like to verify this one locally first. I'll provide more feedback in few hours..

jhrcek:
My bad, It was enough just to rebase this PR onto latest master, building again..

jhrcek:
@Rikkola  Finally I tested this locally with the following results
### Fixed problems
1. The issue this had was supposed to fix is gone, working fine
2. The "Next" pager button disabling now works properly

### Newly introduced problems

1. Asset list is "blinking" every time I type a letter into the filter (probably caused by this piece of code 

```
public void onFilterChange() {
        view.resetList();
       loadProjectInfo();
     }
```
Can't this fix be done without resetting the view on every letter I type? The change introduced by @manstis (postponing backend requests for few hundret ms) made things look really smooth and this resetting makes the UI really jerky :( Compare the difference yourself: 
[videoBeforeAfter.zip](https://github.com/kiegroup/kie-wb-common/files/1007736/videoBeforeAfter.zip)

2. The visual glitch of pager from https://github.com/kiegroup/kie-wb-common/pull/862#pullrequestreview-38031672 is still there :(

Rikkola:
Going to take a look at the blinking.

@paulovmr was kind enough to take a look at the visual glitch. It works in drools-wb so I believe it is a UI setting issue in kie-wb-distributions. So it will be done in a separate PR. I don't think the bug was introduced by this PR since there was a similar issue before.

Rikkola:
Yep. I broke the previous fix for blinking. Fix will soon be in this PR.

Rikkola:
I don't think I managed to fix it, but we are running out of time. But the flickering is a minor bug, while this ticket fixes the more serious one.

Rikkola:
@paulovmr Ok, this is ready for a review.

manstis:
IMO we should not merge stuff that introduces more (known) issues; even if it fixes others.. 

If something is incomplete we push back on it's inclusion.. or time-frames are extended.

Rikkola:
True, here the deadline made me want to merge fast. Since we have testing for these branches we can and should go for "no known bugs go in" policy. This way we are forced to make the master/7.0x/any-branch as good as possible.

PR #870, opened 2017-05-16
JBPM-5980 & JBPM-5981: Bug fixings for shapes docking & delete node commands.

romartin:
Hey @manstis @hasys 

This commit contains the fixes for these two JIRA tickets: [JBPM-5980 ](https://issues.jboss.org/browse/JBPM-5980)and [JBPM-5981](https://issues.jboss.org/browse/JBPM-5981).

Sorry, it contains quite some changed files but it's because had to do a package and a few public API refactors. But basically it contains:
- Bug fixing on the canvas draw, and add child/node commands (JBPM-5980)
- Bug fixings on the delete node commands for both canvas and graph structures. (JBPM-5981)
- Added an improvement in the delete node that it shortcuts the incoming/outgoing connectors for a node, if applies, instead of keeping the connectors present but not connected.
- @manstis Refactored the containment/docking acceptor implementations for both default and CM implementations. There was some bug, no tests and the code was old and was not clear, so just doing clean implementations with new tests.
- Finally found a bug while testing the app that does not appear on the tickets, but using the keyboard to undo and redo actions, was causing issues when several editors were open.

Thanks guys!


krisv:
Since the PR is big and introducing a few unrelated changes as well, this will no longer target 7.0.x but master (7.1) instead.

PR #876, opened 2017-05-17
GUVNOR-3075: [Library] Perspective doesn't show files from Other section

manstis:
See https://issues.jboss.org/browse/GUVNOR-3075

This moves "Library" specific index information into a new generic ```Indexer```. It also moves **all** the "Library" related (index/query) code from the ```refactoring-backend``` module to the ```library-backend``` module.

I've tested with ```drools-wb``` and WID's and uploaded files; and it appears OK. I've also tested with ```kie-wb``` and the auto-generated files from Designer (when exporting the PNG) appears in the Library too.


Rikkola:
@manstis @paulovmr 
Do we dare to close this one now?
 https://issues.jboss.org/browse/GUVNOR-3044

paulovmr:
@Rikkola Yes, tested it locally and it is fixed. :) Thanks, @manstis !

@ederign Ready to merge.

PR #879, opened 2017-05-18
GUVNOR-3080: Project page filter no longer has a search delay

manstis:
See https://issues.jboss.org/browse/GUVNOR-3080

(See https://github.com/kiegroup/kie-wb-common/pull/862#issuecomment-302079268 for background)


Rikkola:
Testing locally.

PR #881, opened 2017-05-18
[DO NOT MERGE] GUVNOR-3082: Asset Search: The ability to search by meta-data has disappeared

manstis:
See https://issues.jboss.org/browse/GUVNOR-3082

@paulovmr The feature used to be available on the [Authoring Perspectives](https://github.com/kiegroup/kie-wb-distributions/blob/master/kie-drools-wb-parent/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/perspectives/DroolsAuthoringPerspective.java#L110). IDK if this is the correct place to add it back; but it makes sense to me (unless UX had different ideas?!)

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/perspective/LibraryPerspective.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/perspective/LibraryPerspective.java
@@ -20,20 +20,30 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
 import javax.inject.Inject;
 
+import org.jboss.errai.ui.client.local.spi.TranslationService;
 import org.kie.workbench.common.screens.library.api.search.FilterUpdateEvent;
+import org.kie.workbench.common.screens.library.client.resources.i18n.LibraryConstants;
 import org.kie.workbench.common.screens.library.client.util.LibraryPlaces;
 import org.kie.workbench.common.widgets.client.search.ContextualSearch;
 import org.kie.workbench.common.widgets.client.search.SearchBehavior;
 import org.uberfire.client.annotations.Perspective;
+import org.uberfire.client.annotations.WorkbenchMenu;
 import org.uberfire.client.annotations.WorkbenchPerspective;
+import org.uberfire.client.mvp.PlaceManager;
 import org.uberfire.client.workbench.panels.impl.MultiListWorkbenchPanelPresenter;
 import org.uberfire.lifecycle.OnOpen;
 import org.uberfire.lifecycle.OnStartup;
 import org.uberfire.mvp.Command;
 import org.uberfire.mvp.PlaceRequest;
+import org.uberfire.mvp.impl.DefaultPlaceRequest;
 import org.uberfire.workbench.model.PanelDefinition;
+import org.uberfire.workbench.model.PartDefinition;
 import org.uberfire.workbench.model.PerspectiveDefinition;
+import org.uberfire.workbench.model.impl.PartDefinitionImpl;
 import org.uberfire.workbench.model.impl.PerspectiveDefinitionImpl;
+import org.uberfire.workbench.model.menu.MenuFactory;
+import org.uberfire.workbench.model.menu.MenuPosition;
+import org.uberfire.workbench.model.menu.Menus;
 
 @ApplicationScoped
 @WorkbenchPerspective(identifier = "LibraryPerspective")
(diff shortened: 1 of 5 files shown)


manstis:
Added @jhrcek and @jomarko for a QE opinion (I assume we have no tests for this).

manstis:
After change (see "Asset search" top right)
guvnor-3082

jhrcek:
@manstis We had asset search tests but it was using old Authoring and hasn't been migrated to Library, as the feature wasn't present in the Library :) Let Jenkins execute full downstream build and then let me check tomorrow that it works properly..

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/perspective/LibraryPerspective.java
```diff
@@ -95,6 +110,15 @@ public void onOpen() {
         libraryPlaces.refresh(callback);
     }
 
+    @WorkbenchMenu
+    @SuppressWarnings("unused")
+    public Menus buildMenuBar() {
+        return MenuFactory
+                .newTopLevelMenu(translationService.getTranslation(LibraryConstants.AssetSearch)).position(MenuPosition.RIGHT).respondsWith(() -> placeManager.goTo("FindForm"))
```
jomarko: Just minor thing but I think for it would be more readable after splitting this line.

manstis:
@jomarko Updated.

paulovmr:
@manstis Hi, I think that is good, but needs some improvements to keep the general navigation experience:

* Include breadcrumb when you go to this screen (see `LibraryPlaces.java`);
* Remove the header of the Uberfire Part. This can be done like [this](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java#L468-L472). This should be done to keep this screen view consistent to the teams, library and project ones.


Also, we must make sure that going to this screen (and the assets that it lists) won't break the navigation (window and breadcrumb) independently of which screen (context) was opened before that.

manstis:
Hi @paulovmr thanks for looking.

I've made the change (locally) to remove the header as requested; however am thinking about the breadcrumbs. The "Asset search" feature spans **_all_** repositories (to which the User has access) and not just the current project. I think adding it to the breadcrumbs gives the impression that it is scoped to the current project (which it is not). WDYT? Happy to add if you so desire - you're the king of Library!

paulovmr:
@manstis I see your point... WDYT if the breadcrumb was "Teams > Asset search", independent of what screen was open before? Then when an asset is opened, it would go back to the current format (e.g.: Teams > myteam > myproject > myasset).

Considering the rate that you are opening PRs to Library, you will take mine and Toni's crown. :)

jhrcek:
@manstis we now have Asset search tests for library in Selenium. Tried it manually and it works. The only think I don't like is the additional "stripe" at the top that's added just to make the "Asset search" link available (is that what you fixed in https://github.com/kiegroup/kie-wb-common/pull/881#issuecomment-302427394 ?). Can't this be added to existing library's [BreadcrumbToolbar](https://github.com/AppFormer/uberfire/blob/master/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/breadcrumbs/UberfireBreadcrumbsView.java#L42) instead?

screenshot from 2017-05-19 09-37-01

manstis:
@jhrcek Hi.. soooo, this PR will be updated (when I finish with the crazy stuff for 7.0.0.Final). @paulovmr and I were chatting about this on IRC; the "Asset search" link will move from the "Perspective menu bar" to the breadcrumbs (like "Project settings"). ATM I'm keeping this PR open to remind me to finish it!

Rikkola:
I believe the long term goal is to move asset search options into the filtering options in the asset list, but since that will probably happen for 8.0.x this is a good option for 7.0.x branch.

porcelli:
do we really need this back? I'd prefer invest time to get the new filter more powerful =/

manstis:
@porcelli Well... I guess it missed the 7.0.0.Final boat so we could just as well implement the alternative for 7.1.0.Final (assuming there's resource somewhere to do something more specific than add back a menu item). Having *something* in 7.1.0 is, IMO, better than nothing.

porcelli:
@manstis i still think that improve the current filter would be better :)

manstis:
Closing since the current proposal is unsuitable for inclusion in 7.1.0.Final.

porcelli:
@manstis +1 for close.. sometimes less is more :)

jhrcek:
Just a generic note about Asset search: Even after we've developed some tests for that I was never actually sure what to enter in those search fields (I know it comes from data entere in asset metadata tab, but never found this feature user friendly). I'd vote for expanding existing library filtering capabilities (or rather its asset name filtering capabilit*y*) to include more metadata fields.

PR #885, opened 2017-05-18
[7.0.x] GUVNOR-3083: Library: Indexer should not index folders

manstis:
See https://issues.jboss.org/browse/GUVNOR-3083

This is the corollary of https://github.com/kiegroup/kie-wb-common/pull/883 for 7.0.x

(cherry picked from commit 41fe528630fe7bd876c27489044e725cd6e7dbea)


Rikkola:
+1 if we re-tag.

PR #886, opened 2017-05-19
Fix small bugs found by Findbugs

hasys:


kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors/src/main/java/org/kie/workbench/common/forms/adf/processors/util/FormGenerationUtils.java
```diff
@@ -27,7 +27,7 @@
     public static void sort(String firstName,
                             List<Map<String, String>> elements) {
 
-        if (elements == null && elements.isEmpty()) {
+        if (elements == null || elements.isEmpty()) {
```
triceo: This is the kind of problem that is very obviously a bug. It can not be anything but a bug.
Can we have Findbugs fail the build on those?

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-common-rendering/kie-wb-common-forms-common-rendering-client/src/main/java/org/kie/workbench/common/forms/common/rendering/client/widgets/decimalBox/DecimalBox.java
```diff
@@ -55,7 +55,7 @@ public void setValue(Double value) {
     public void setValue(Double value,
                          boolean fireEvents) {
 
-        if (this.getValue() != value) {
+        if (this.getValue().equals(value)) {
```
hasys: So, if I correct this is by design, but is it have to be in this way?

pefernan: @hasys good point! the idea is checking that the new value is different before setting it, I'd do something like:
if (this.getValue() == null || !this.getValue().equals(value))

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors/src/main/java/org/kie/workbench/common/forms/adf/processors/util/FormGenerationUtils.java
psiroky: I definitely agree and we can enable this. But first someone needs to fix all the current occurrences, otherwise the build would start to fail imminently, which is probably something we don't want.

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-common-rendering/kie-wb-common-forms-common-rendering-client/src/main/java/org/kie/workbench/common/forms/common/rendering/client/widgets/integerBox/IntegerBox.java
```diff
@@ -54,7 +54,7 @@ public void setValue(Long value) {
     @Override
     public void setValue(Long value,
                          boolean fireEvents) {
-        if (this.getValue() != value) {
+        if (!this.getValue().equals(value)) {
```
pefernan: @hasys same here

hasys:
Thank you, @pefernan, amended.

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors/src/main/java/org/kie/workbench/common/forms/adf/processors/util/FormGenerationUtils.java
hasys: Hi @triceo, it looks like we are not able to enable just one rule. We played with it for some time to be sure that Critical level have no false positives. So, for now it looks like we can fix remaining issues and enable findbugs on this level only, WDYT @pefernan ?

triceo: I believe @rsynek had a setting somewhere (was it jbpm?) that was pretty fine-tuned.

hasys: @triceo yes, but kie-wb-common repository had some GWT&CDI related false positives

rsynek: @hasys It's possible to exclude individual features Findbugs looks for, please see https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/kie-build-tools/src/main/resources/findbugs-excludes.xml

PR #888, opened 2017-05-19
GUVNOR-3087: Library Project Page: List should double check the file exists

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3087

A tricky one to reproduce, but this should fix it.

Fix from @manstis test from me.


Rikkola:
Test actually found an area that is missing test coverage. Something I'll improve when I next dive to these parts.

PR #890, opened 2017-05-19
GUVNOR-3089 : Workbench: Asset list is not complete after importing a project

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3089

Power level 2000!

@mcimbora or anyone please take a look.


mcimbora:
@Rikkola +1, however let's make sure there's a follow-up and the list is correctly loaded under all circumstances.

Rikkola:
Made a ticket just in case I forget.
https://issues.jboss.org/browse/GUVNOR-3098

PR #891, opened 2017-05-19
[7.0.x] GUVNOR-3089 : Workbench: Asset list is not complete after importing a project

Rikkola:
(cherry picked from commit ac2baf838210b76c4d297206df6f34df1b288c82)

Related to #890 

And another fix that should still go in.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java
@@ -291,7 +291,7 @@ public class ProjectScreen {
                                                    return false;
                                                }
                                            },
-                                           1000);
+                                           2000);
     }
 
     public void onFilterChange() {


Rikkola:
jenkins please retest this

Rikkola:
Base on the IRC spam the 7.0.x branch is broken ATM.

ederign:
jenkins please retest this

Rikkola:
@ederign so this is the ticket that just came up in the weekly meeting. Let's hope the build passes.

Rikkola:
Thank you

manstis:
@Rikkola Might be worth thinking about changing the "indexing complete" message from ```BatchIndex``` (make some explicit actions to monitor; rather than just text messages) and have Library continue to poll until the message is received. The "poll" interval will be just that then and not used to determine if indexing is complete (i.e. if there are no changes to the result set after 2 successive polls).

Rikkola:
@manstis We would have to change when the event is fired. Could be a bug, but at the moment the event is fired once the repository is created. This happens before the content is cloned in. So the repository is empty at that point.

manstis:
@Rikkola what event? 

ATM ```BatchIndex``` calls into a generic ```Observer``` several generic "message" for when indexing starts and completes. I was proposing we (remove/enhance) this to report a specific "indexing complete" message that then fires an event to Library to inform it indexing has completed and hence stop polling. So change [this](https://github.com/AppFormer/uberfire/blob/master/uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/BatchIndex.java#L203) to call ```observer.indexingComplete(repository)``` and the [incremental indexing](https://github.com/AppFormer/uberfire/blob/master/uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/IOServiceIndexedImpl.java#L251). This can be then caught by [this](https://github.com/kiegroup/guvnor/blob/master/guvnor-message-console/guvnor-message-console-backend/src/main/java/org/guvnor/messageconsole/backend/DefaultIndexEngineObserver.java) to raise a client-side event used by Library to signal indexing has completed..

Might be more effort than it's worth; but it'd remove the notion of "no changes in results == indexing complete" which is subject to timing problems as we've found.

Rikkola:
@manstis I was referring to event that causes the log message in UI that says "indexing complete". So what the `observer.indexComplete(repository)` most likely fires.
 
The problem is the reporting panel in the UI says "indexing complete" and still the search finds nothing. So either it reports "indexing complete" too soon or indexing is actually complete, but the index is still being written when the query is first ran. Lucene might not write it with the same speed we pass the data in.

I'll take a another look at the event when I work on [GUVNOR-3098](https://issues.jboss.org/browse/GUVNOR-3098). Using the event is indeed the better way to do this, but didn't use it due to the reasons above and being afraid I'll run out of time if it does not work for in time.

PR #892, opened 2017-05-22
Set Findbugs maven plugin to fail on violations for Stunner project

hasys:


romartin:
Ah well you're updating both plugins yeah, just seen now the complete change, sorry :)

hasys:
Hi @romartin,
yes, sorry. There are also minor update for checkstyle. From now it will print checkstyle violations to console, not just in a file.

And yes, it will run findbugs also as checkstyle only with `full` profile.

Thank you for review :)

romartin:
Great! thanks for clarifying @hasys  :)

PR #893, opened 2017-05-22
Enable Findbugs to fail on violations for Forms project

hasys:
Pleas don't merge for now, need to test this PR locally after Full-downstream build will be done.


hasys:

hasys:
This Commit contain fix for Findbugs violation *Class defines field that masks a superclass field* [1] and enable Fail on Violations for Forms project. I tried full downstream build locally and everything is fine, ready to merge. @manstis, @pefernan, can you review also?

[1]
> This class defines a field with the same name as a visible instance field in a superclass. This is
> confusing, and may indicate an error if methods update or access one of the fields when they wanted
> the other.

PR #899, opened 2017-05-24
[7.0.x] GUVNOR-3108: RuleSelector: Refactor to use Bootstraps Select widget

manstis:
See https://issues.jboss.org/browse/GUVNOR-3108

This is the corollary of https://github.com/kiegroup/kie-wb-common/pull/898 for 7.0.x

(cherry picked from commit d01a4d38087fa61fcb157ba46eb9a9d80ddd2a5f)


manstis:
@krisv This is the fix for the "Rule inheritance widget" issue @Rikkola reported.

IDK whether it should be considered a "blocker" (nobody noticed - I only found when doing an unrelated JIRA).

krisv:
+1, Edson recommended we would try to pick up this fix

PR #904, opened 2017-05-25
GUVNOR-3112: Package list box should initialize with "groupId.artifactId" package selected

paulovmr:


kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/handlers/PackageListBox.java
```diff
@@ -53,112 +56,133 @@
...
+        }).resolvePackages(context.getActiveProject());
+    }
+
+    Package getGroupIdProjectPackage(final Project project,
+                                     final List<Package> packages) {
+        final GAV projectGav = project.getPom().getGav();
+        final String projectGroupId = projectGav.getGroupId();
+        final String projectArtifactId = projectGav.getArtifactId();
+
+        final String activePackageCaption = projectGroupId + "." + projectArtifactId;
```
manstis: Please note the "default" package created for new projects is not simply ```GroupID```+"."+```ArtifactId``` but [this](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/project/ProjectSaver.java#L220). However for the purpose of this PR (for 7.0.0.Final) and the edge cases this won't work this is probably good enough. Perhaps raise a new JIRA to ensure the ```activePackageCaption``` is constructed the same as the backend for 7.1.0.Final.

paulovmr: @manstis You are absolutely right, I will create this ticket. Thanks!!

PR #906, opened 2017-05-25
JBPM-5950 : Add a new Error Management view.

nmirasch:
Adds a new Perspective id for the Error management view


nmirasch:
I have been added that in https://github.com/kiegroup/kie-wb-distributions/pull/531

PR #913, opened 2017-05-31
JBPM-5978: Improve marshalling tests in BPMNDiagramMarshallerTest

jlindop:
@romartin , @hasys   This is incomplete but I think worth committing at this point. I've changed some of the 'testMarshall...' tests so that checks are made against the Eclipse BPMN2 API objects which are created by the marshaller, rather than testing against strings in the output XML. The advantage is that in many cases, the tests are no longer testing for matching UUID's, so it's much easier to update existing bpmn test files and test cases when necessary. When the tests were for matching UUID's, all UUID's for that test case would need to be updated if a change to the test process was made.


jlindop:
jenkins please retest this

romartin:
@manstis @hasys @krisv See comments [in the ticket](https://issues.jboss.org/browse/JBPM-5978), I would close this ticket for now.. (plus is not completed yet). WDYT? Thx!

manstis:
@romartin I've commented the JIRA.

hasys:
@romartin also commented the Jira issue

romartin:
Hey @hasys , please do not forget this one, @manstis  requested review from your side. Notice that  @mdproctor  and @krisv agreed on merging this.

PR #914, opened 2017-06-01
GUVNOR-3132: [Library] Increase separation of Indexer tests

manstis:
See https://issues.jboss.org/browse/GUVNOR-3132


manstis:
Different tests were using the same "Project root" and hence in a parallel build system (Jenkins) it was plausible the state of one test could be visible to another test. This PR increases the separation between tests by ensuring each uses different "Project roots".

PR #921, opened 2017-06-06
PLANNER-763: Workbench: Import example polishing

mcimbora:
Add support for Filter as you type in the project list of Import example wizard.


manstis:
@mcimbora Do consider (for a future improvement) that filtering the view on _every_ key press may not be such a wise idea; unless you know the list is always small and can be handled entirely synchronously client-side.

Here's an example of a similar scenario in [Library](https://github.com/kiegroup/kie-wb-common/blob/14aeb4c54559899775a510be5d9832d4f6d14225/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java#L103). ```onUpdateAssets()``` is called by a ```KeyPressHandler``` and only when the timer elapses do we get the filter (tag) value from the UI and use it.

PR #923, opened 2017-06-07
GUVNOR-3134: Test scenarios fail after upgrade to 7.1 because rule expectation in 6.x doesn't use FQN

manstis:
See https://issues.jboss.org/browse/GUVNOR-3134

### DIFF
--- a/kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/test/resources/org/kie/workbench/common/services/refactoring/backend/server/query/findrules/drl5.drl
+++ b/kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/test/resources/org/kie/workbench/common/services/refactoring/backend/server/query/findrules/drl5.drl
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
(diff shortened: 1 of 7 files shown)


manstis:
Part of an ensemble:

- https://github.com/kiegroup/drools/pull/1299
- https://github.com/kiegroup/kie-wb-common/pull/923
- https://github.com/kiegroup/drools-wb/pull/522

manstis:
Jenkins please retest this.

manstis:
@yurloc @jomarko OK.. all fixed. Rules with "```.```" in their name can be used; and I only show/use "Simple Rule Names" so 6.x legacy Test Scenarios should continue to function unaffected. Please re-review the ensemble.

kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/test/resources/org/kie/workbench/common/services/refactoring/backend/server/query/findrules/drl5.drl
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.workbench.mock.package;
+
+import org.kie.workbench.common.services.refactoring.backend.server.drl.classes.Applicant;
+
+rule "my.Rule4"
```
jomarko: Is this name intentional? I would expect `rule "my.Rule5"`. But I may be wrong.

manstis: It's intentional and based on the rule names in the files. File ```drl4.drl``` has a rule called ```noPackage```; so file ```drl5.drl``` follows on the rule names from ```drl3.drl```... but if it's confusing I can change.

jomarko: Thanks for fast reply. I think I can live with the current filenames and rulenames inside. I got idea that filenames could stay as are and just rulenames could be: "parentRule", "childRule", "standaloneRule", "noPackage", "rule.with.dots.in.name"

PR #925, opened 2017-06-07
[JBPM-6069] Missing CRUD buttons in asset editors

ederign:
Reverting https://github.com/kiegroup/kie-wb-common/commit/a18fe0c3a2b27a480a5d166966c432e70877bbb5#diff-7987251173d3d91720dcb2a2b380e509 .

### DIFF
--- a/kie-wb-common-widgets/kie-wb-metadata-widget/src/main/java/org/kie/workbench/common/widgets/metadata/client/KieEditor.java
+++ b/kie-wb-common-widgets/kie-wb-metadata-widget/src/main/java/org/kie/workbench/common/widgets/metadata/client/KieEditor.java
@@ -44,6 +44,7 @@ import org.uberfire.mvp.Command;
 import org.uberfire.mvp.ParameterizedCommand;
 import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.workbench.model.menu.MenuItem;
+import org.uberfire.workbench.model.menu.Menus;
 
 public abstract class KieEditor
         extends BaseEditor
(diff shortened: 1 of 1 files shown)


ederign:
This is a quick fix. I don't know if we should invest time in a proper refactoring/test of this.

ederign:
jenkins retest this

jhrcek:
It seems the bug was introduced by my change [here](https://github.com/ederign/kie-wb-common/commit/a18fe0c3a2b27a480a5d166966c432e70877bbb5#diff-7987251173d3d91720dcb2a2b380e509L53)

In that commit I removed 3 fields based on 3 of [these 4](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-common-pullrequests/1267/findbugsResult/) findbugs warnings. However this bug shows that proper functionality of al least one of those classes depended on the field being declared also in the subclass.

@ederign @manstis can we refactor those classes in some other way to avoid this findbugs failing the build? Note that I don't thing reverting my "Enable findbugs failing the build" is a good solution. Instead we should somehow avoid the bad practice of needing same-named protected field at two places in the inheritance hierarchy.

manstis:
@jhrcek @ederign Yes, it does point to a more serious issue.. fixing at the moment.

Thanks @ederign for your investigation I, as much as anyone I guess, was _really_ surprised by the cause.

ederign:
I didn't intend to revert all the commit, just the bug introduced by it.  Are we sure that removing fields from KieTextEditorPresenter and ProjectImportsServiceImpl.java also doesn't break anything?

PR #930, opened 2017-06-13
JBPM-5861: Stunner - Client side export features.

romartin:
Hey @manstis @hasys ,

This commit provides export to JPG, PNG and PDF capabilities. From the user perspective, just added three new buttons on the toolbar, so feel free to give it a try :)

It DEPENDS on [this other PR for AppFormer](https://github.com/AppFormer/uberfire/pull/744).

[This is the JIRA ticket](https://issues.jboss.org/browse/JBPM-5861) for it.

Thanks!


romartin:
Hey @manstis @hasys 
I have updated this commit due to some changes in the uberfire side.  Remember it depends on:
https://github.com/AppFormer/uberfire/pull/744
Thanks!

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/LienzoLayer.java
```diff
@@ -103,15 +103,12 @@ public String toDataURL() {
...
-                                                x,
-                                                y,
-                                                width,
-                                                height);
+    public String toDataURL(final URLDataType type) {
+        switch (type) {
+            case JPG:
+                return layer.toDataURL(DataURLType.JPG);
+        }
+        return layer.toDataURL(DataURLType.PNG);
```
hasys: I guess we need FailFast technique here as it is don in CanvasFileExport:
```java
switch (type) {
    case JPG:
        return layer.toDataURL(DataURLType.JPG);
    case PNG:
        return layer.toDataURL(DataURLType.PNG);
}
throw new UnsupportedOperationException("No mimeType supported for " + type);
```

romartin:
Hey @hasys 

Yeah thanks for testing this. Added a new commit. Lemme share my comments back:

- Added the "caret" in the export drop-down menu (as in "Last Version")

- About auto-deselect before export -> yeah it's something I noticed too, but in fact it's not just doing deselection, is disabling the toolbox and any other action/controls present on the canvas at export time. So reported [this bug](https://issues.jboss.org/browse/JBPM-6111) to keep track of it, as it's not just a quick fix...

- Storing images on server -> Not sure if the exported image must be stored on server side as well. In fact the idea of this feature was considering it a client-side-only feature (although it finally goes to backend to check if there are different pdf preferences), so the content export and file stream generation only happen on client side. Probably @tsurdilo or @krisv  can give us some feedback about this, anyway is not a problem sending the image to server, but does really the user expect that either?

- About export options -> Actually there exist a few PDF export options, available from the workbench's Preferences (as this feature has been included in UF and can be reused by any other component, is generic). On the other hand I'm agree that we could provide concrete export options for Stunner, like the background color (now is always white), or the grid, or the file name, etc, probably using same Preferences or using some popup export options screen once clicking. A new (enhancement/task) ticket can be open once this gets merged and available for this, if you think so.

- About HTTP requests - yes, it's true that for PDF export, it generates two requests. In fact I cannot say right now why two, because I'm just using the errai bus, so I understand it's errai that generates the request and payload. In addition, I did not found any event for preferences being updated, so each time you click on export it checks the preferences, which are stored on server. Lemme know please if I'm wrong :)

Thanks!

romartin:

krisv:
@romartin the console currently depends on an SVG file of the process to be saved next to the process itself.  This would be similar to what you worked on, but since this does not yet support SVG, I think it's fine to ignore this requirement for now.  I don't think it makes sense to store the pdf or png/jpg on file system.

hasys:
@krisv, thank you! 
Another issues for the future: [JBPM-6110](https://issues.jboss.org/browse/JBPM-6110)

PR #931, opened 2017-06-13
JBPM-6088 - Not possible to run Simulation of process with Send Task

tsurdilo:
@manstis i would really like your opinion on how to write a meaningful test for this small change if there is one. 

The actual test is in jbpm-designer https://github.com/kiegroup/jbpm-designer/pull/617 but when designer runs in workbench (with stunner) BpmnProcessDataEventListener also gets serialized and afaik stunner does not yet support send task to write a test against)


tsurdilo:
@manstis during simulation we get a KieSessionFluent from ExecutableBuilder. This session will include all listeners including this 
BpmnProcessDataEventListener. Due to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6750650 however "messages" cannot be serialized on this listener and the simulation process fails. 

After looking some more it looks I can write a similar tests for send task in stunner afterall. Will do that and add.

hasys:
Hi @tsurdilo, thank you for quick fix! Can we also add tests from https://github.com/kiegroup/jbpm-designer/pull/617 to this PR as well?

manstis:
@tsurdilo Thanks for the explanation.

tsurdilo:
per @hasys request I will close this pr and create a new one so that branch names match the associated jbpm-designer pr (sorry i didnt do that tobegin with)

PR #948, opened 2017-06-27
JBPM-6156: Stunner -  Use same stencil icons for canvas and for all widgets.

romartin:
Hey @manstis  @hasys 

Sorry for this huge PR, I know it results difficult to review, but once trying to fix some BPMN shapes and icons realized that the changes only for changing a few images were too much, in fact there was some duplicity on how to define the "glyphs" (icons) for each element, and this was causing that the icons displayed were different between components and the canvas.  So finally decided to spend a few days doing some cleans and improving the glyph and shapes integration and support. All the stuff is related, so it was really difficult to split in different commits.

On the other hand, I think that from this point is much easier and intuitive to define the shapes and the icons to use, so I expect that this commit will save future coding time :)

Please see the detailed changes in [the JIRA ticket](https://issues.jboss.org/browse/JBPM-6156).

@manstis This will cause compilation issues for your actual DMN stuff, but I don't expect any big issues tbh. The conflicts should appear on client side, and being more specific in your DMN `ShapeDefs` and `ShapeFactories`,  This commit includes refactors for BPMN and CM, so there are a few examples as well.  If you can move to some other task for just an hour, we can do as we did on previous times, I will checkout your stuff and do the refactor (should be just minutes for me!), so you can continue from that point. Just lemme know what's better for you!!! And sorry for the inconveniences.

PS: Added lots of tests that were missing in addition to the tests for new stuff , we're not yet in a high coverage rate probably, but getting there soon!! :)

Thanks guys!!

PR #952, opened 2017-06-29
GUVNOR-3220: [DMN Editor] FlowActionsToolboxControlProvider always uses a 'default' connector

manstis:
See https://issues.jboss.org/browse/GUVNOR-3220

@romartin This moves the fix for always using "default connector" in ```AddNodeCommand``` to ```master``` which also includes the fix to ```CommonLookups``` that did not check if an "allowed connection" (to a target node) originated from the "source node". 

It _might_ interfere with https://github.com/kiegroup/kie-wb-common/pull/948 as we;'ve both dabbled with ```AddNodeCommand```.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/lookup/util/CommonLookups.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/lookup/util/CommonLookups.java
@@ -60,7 +60,7 @@ import static org.uberfire.commons.validation.PortablePreconditions.checkNotNull
 /**
  * An utils class that provides common used look-ups and other logic for querying the domain model and the rules model,
  * that is used along the application.
- * <p>
+ * <p/>
  * // TODO: Some kind of cache to avoid frequently used lookups? Consider performance and memory, this class
  * is shared on both server and client sides.
  */
@@ -151,7 +151,7 @@ public class CommonLookups {
      * the given edge (connector) identifier.
      * This method only returns the definition identifiers that are considered the default types for its morph type,
      * it does NOT return all the identifiers for all the allowed target definitions.
-     * <p>
+     * <p/>
      * TODO: Handle several result pages.
      */
     public <T> Set<String> getAllowedMorphDefaultDefinitions(final String defSetId,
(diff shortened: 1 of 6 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/lookup/util/CommonLookups.java
```diff
@@ -377,12 +378,17 @@ private RuleSet getRuleSet(final String defSetId) {
                                                        pageSize);
         if (null != rules && !rules.isEmpty()) {
             final Set<String> result = new LinkedHashSet<>();
+            final Optional<Set<String>> sourceDefLabels = Optional.ofNullable(getDefinitionLabels(sourceDefinition));
```
hasys: Probably, in this case `Optional` is overkill.Code is less readable with this construction here and bring no gain.

To eliminate further regressions about `NullPointerException` in this piece of code I guess we can add just a test with mocked `getDefinitionLabels` method: 
```java
when(getDefinitionLabels(anyObject())).thenReturn(null);
```

manstis:
Hi @hasys the issue was only evident in my DMN Editor [work](https://github.com/manstis/kie-wb-common/tree/GUVNOR-3164) that has different types of connector (BPMN's use of Stunner only has one type of connector and therefore the issues this PR fixes are not evident). @jomarko has been looking at my fork (for DMN Editor related work) perhaps he can verify the issue does not exist (on my fork)? This PR simply replicates the fix (that is on my fork) into ```master```.

manstis:
@hasys Removed the use of ```Optional```. 

I looked into mocking ```getDefinitionLabels(anyObject()``` however if a ```Node``` lacks any "labels" (```null``` or _empty_) then the code I changed does not get hit as prior checks for "allowed targets" are not satisfied.

romartin:
yeah @manstis , this PR conflicts with the other big one, no problem for me, if this one gets merged before I'll rebase the other one. I'm only trying to get the big one merged as soon as possible anyway for avoiding further issues with any other PRs... but don't worry for my big PR, if necessary will rebase again  :) BTW doing a try to this now. thanks!!

jomarko:
@hasys I believe it is enough to install this PR locally  and then package the [manstis's referenced fork|https://github.com/manstis/kie-wb-common/tree/GUVNOR-3164/kie-wb-common-dmn].

manstis:
@jomarko @hasys The problem was that with the DMN Editor you could not add new nodes from the "Toolbox" controls (select a node, then pick "Click to create a XYZ"). The pre-PR Stunner code always used a "default connector" but for DMN different "target nodes" needed different types of connector.

jomarko:
I can confirm that using given Toolbox controls different connectors are used.

screenshot from 2017-06-30 12-47-43

manstis:
Thanks @hasys I have fixed the URL in the pom on my [fork](https://github.com/manstis/kie-wb-common/commit/50e1508e4ac6c3829209ed6f0782a53b9fe612bf).

PR #963, opened 2017-07-06
Library improvements

paulovmr:
GUVNOR-3150: [Library] Disabling different user role permissions does not work
GUVNOR-3152: [Library] Read/create permissions have unexpected behavior
GUVNOR-3153: [Library] Errors after creating teams with invalid name
GUVNOR-3154: [Library] Incorrect project description
GUVNOR-3160: [Library] Search field does not work as expected when an asset is opened


paulovmr:
Part of an ensemble:
* https://github.com/kiegroup/guvnor/pull/472
* https://github.com/kiegroup/kie-wb-common/pull/963

paulovmr:

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/perspective/LibraryPerspective.java
```diff
@@ -101,6 +104,24 @@ public void onClose() {
...
+            final ProjectInfo projectInfo = libraryPlaces.getProjectInfo();
+            final Command callback = () -> filterUpdateEvent.fire(new FilterUpdateEvent(searchFilter));
+
+            if (projectInfo != null && projectInfo.getProject() != null) {
+                if (PlaceStatus.CLOSE.equals(placeManager.getStatus(LibraryPlaces.PROJECT_SCREEN))) {
+                    libraryPlaces.goToProject(projectInfo,
+                                              false,
+                                              callback);
+                } else {
+                    placeManager.goTo(LibraryPlaces.PROJECT_SCREEN);
```
manstis: Doesn't ```PlaceManager``` have an overloaded ```goTo(..)``` method that accepts a command to execute after the place has been navigated to? Would use of that be better to ensure the "Project Screen" was actually loaded before firing the "Search Filter" event?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/popup/OrganizationalUnitPopUpPresenter.java
```diff
@@ -165,10 +166,14 @@ void saveCreation(final String name,
                                                                                             NotificationEvent.NotificationType.SUCCESS));
                                                view.hide();
                                            },
-                                           new HasBusyIndicatorDefaultErrorCallback(view)).createOrganizationalUnit(name,
-                                                                                                                    owner,
-                                                                                                                    defaultGroupId,
-                                                                                                                    repositories);
+                                           (o, throwable) -> {
+                                               view.hideBusyIndicator();
```
manstis: What if an exception other than ```InvalidOrganizationalUnitNameException``` is thrown? Would use of ```CommandDrivenErrorCallback``` not be better, so you can report invalid OU _file_ names separately to, for example, a NPE or other unexpected exception?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
```diff
@@ -90,20 +90,22 @@ public LibraryToolbarPresenter(final View view,
 
     public void init(final Command callback) {
         libraryService.call((OrganizationalUnitRepositoryInfo info) -> {
-            LibraryToolbarPresenter.this.info = info;
-            view.init(LibraryToolbarPresenter.this);
+            if (info != null) {
```
manstis: ..and if it does equal ```null``` do we want to put the view into a state that prevents erroneous manipulation (and further errors)?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/perspective/LibraryPerspective.java
paulovmr: There is no overloaded method for that, but there is a way to register a callback for when one place is opened. However, that is not necessary in this case, because the else clause only will be executed if the place is already opened. The `goTo` is executed only to focus the already opened place.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/organizationalunit/popup/OrganizationalUnitPopUpPresenter.java
paulovmr: I agree, this way is better. Done.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
paulovmr: In this case the screen will be empty. But in any case, I've added an else clause to show a message to the user telling that more permissions are needed to use authoring, since `info` is null when no OUs could be fetched or created.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/perspective/LibraryPerspective.java
manstis: @paulovmr OK, np.. I was thinking of the ```private goTo(place, panel, doWhenFinished)```.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
manstis: Better, but is it good enough? ```LibraryToolbarPresenter``` is ```@ApplicationScoped```; so if as a User I _had_ access to an OU the screen would be populated. Now let's assume for some reason my access is revoked (and assuming I don't log out, or go home and come in the next day etc) when I revisit this screen it will appear that I still have access (```info == null``` but the UI would hold the old state). Admittedly I'll now get a message saying "no permission" but the UI would not reflect that.

paulovmr:
@manstis And I thank you for that! :)

This method `LibraryToolbarPresenter#init` is called every time the perspective is opened, so the screen would be blank and the error message would appear as soon as the user changed perspectives. If the user doesn't leave the perspective, he/she would get lots of errors due to lack of permissions, but I guess this could happen in any screen of the workbench when some permission changes. Do you think that is enough?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
manstis: @paulovmr What makes you think "...so the screen would be blank"? The ```Presenter``` is ```@ApplicationScoped``` so we have a single re-used instance. The ```View``` (whilst ```@Dependent``` by default) will only be created an injected once (into the ```Presenter```). So, if the User first had access to an OU the ```View``` would be initialised when the Perspective was navigated to and the OU shown (```info != null```). If the User's access to _any_ OU was then removed and the Perspective re-visited ( ```info == null```) the ```View``` would not be cleared (as we only instantiate it once when the ```@ApplicationScoped``` ```Presenter``` was first created) and would appear to still show the original OU.

tomasdavidorg:
Hi @paulovmr,
I tried to disable some permission for specific project (GUVNOR-3150) and for example I am still able to delete the project even the permission is disabled.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
paulovmr: @manstis You are right, I was testing an OU with just one repository, so the screen already had the right behavior. Now I fixed the code to clear the view properly when necessary. Thanks!

paulovmr:
Hi @tomasdavidorg , I couldn't reproduce this problem. Could you provide the exact steps that you did?

paulovmr:
jenkins retest this

paulovmr:

tomasdavidorg:
@paulovmr I tried the new build and the problem I mentioned is gone. But:
GUVNOR-3150: It is still possible to change and save the Deployment and Persistence descriptor. 
GUVNOR-3152: The 'Create' permission is still independent to the read permission. So if user has only Create permission he can create a project but then after he clone the project he cannot access the project anymore.

paulovmr:
@tomasdavidorg Thanks, I've now fixed the update permission issue for all editors. As we talked about, the create permission being independent from the read one is a generic behavior of our security. If this is to change, it should be dealt with in another JIRA.

@manstis A lot of changes since the first review, if you please could take another look. Thanks! :)

paulovmr:

paulovmr:
jenkins retest this

paulovmr:

paulovmr:

tomasdavidorg:
Thanks @paulovmr, I created new jira for the create permission improvement - https://issues.jboss.org/browse/AF-567.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
```diff
@@ -78,32 +82,40 @@ public LibraryToolbarPresenter(final View view,
...
+                projectContextChangeEvent.fire(event);
 
-            callback.execute();
+                callback.execute();
+            } else {
+                notificationEvent.fire(new NotificationEvent(view.getNotEnoughPermissionsToAccessLibraryMessage(),
+                                                             NotificationEvent.NotificationType.ERROR));
+            }
+
+            setBranchSelectorVisibility();
```
manstis: @paulovmr Can you confirm the "Branch Selector" and "Repository Selector" are the only preserved "state" in the ```View```? (If yes.. so a leading question for you.... then I am happy that _should_ the User's permissions change then the ```View``` will not maintain old state) :-)

manstis:

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/library/LibraryToolbarPresenter.java
paulovmr: @manstis Yes, these are the only two components of the view. :)

paulovmr:
@manstis Hi, I've opened new PRs for this other repositories, I included the PR list [here](https://github.com/kiegroup/kie-wb-common/pull/963#issuecomment-313513085)

PR #978, opened 2017-07-17
JBPM-6119: Error after generating process/task form containing data object with more forms.

pefernan:
@jsoltes @wmedvede could you take a look?


jsoltes:

PR #979, opened 2017-07-17
GUVNOR-3284: [DMN Editor] Two same connections between nodes

manstis:
See https://issues.jboss.org/browse/GUVNOR-3284

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/rules/SingleConnectorPerTypeGraphRule.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/rules/SingleConnectorPerTypeGraphRule.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.dmn.api.rules;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+import javax.enterprise.context.ApplicationScoped;
+
+import org.kie.workbench.common.stunner.core.graph.Edge;
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.content.definition.Definition;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+import org.kie.workbench.common.stunner.core.rule.RuleViolations;
+import org.kie.workbench.common.stunner.core.rule.context.GraphConnectionContext;
+import org.kie.workbench.common.stunner.core.rule.ext.RuleExtension;
+import org.kie.workbench.common.stunner.core.rule.ext.RuleExtensionHandler;
+import org.kie.workbench.common.stunner.core.rule.violations.DefaultRuleViolations;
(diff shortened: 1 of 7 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/rules/SingleConnectorPerTypeGraphRule.java
```diff
@@ -0,0 +1,121 @@
...
+    protected boolean isConnectionAlreadyFormed(final Node<? extends View<?>, ? extends Edge> source,
+                                                final Node<? extends View<?>, ? extends Edge> target,
+                                                final Edge<? extends View<?>, ? extends Node> connector) {
+        //Check for existing outgoing connections of the same type
+        final Set<Edge<? extends View<?>, ? extends Node>> existingEdges = new HashSet<>();
+        final Class<?> newConnectorClass = connector
+                .getContent()
+                .getDefinition()
+                .getClass();
+        source.getOutEdges()
```
triceo: As an exercise for myself, I tried to write this snippet of code without the use of a collection inbetween. I also got rid of the null-check and the final `Optional`. Feel free to use this code if you like it.

    return source.getOutEdges()
        .stream()
        .filter(e -> !Objects.equals(e, connector))
        .flatMap(e -> {
            final Object content = e.getContent();
            if (content instanceof Definition) {
                final Class<?> existingConnectorClass = ((Definition) content).getDefinition().getClass();
                if (newConnectorClass.equals(existingConnectorClass)) {
                    return Stream.of(e);
                }
            }
            return Stream.empty();
        })
        .distinct()
        .map(Edge::getTargetNode)
        .anyMatch(n -> Objects.equals(n, target));

manstis: Sweet.. let my brain chew on this "more readable" form (I too _tried_ to avoid the temporary collection!)

triceo: Good use of the quotes there - I was actually also wondering if my change makes the code nicer or not. :-)

triceo: One more note - the `distinct()` call there was only added to mimic the behavior of the original `Set`. If this call were removed - that is, if it were acceptable that the elements in the stream would not be unique - then the stream would become lazy and `anyMatch()` would return when encountering the first element that matches the final `Predicate`.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/rules/SingleConnectorPerTypeGraphRule.java
```diff
@@ -0,0 +1,121 @@
...
+                                      target,
+                                      connector)) {
+            result.addViolation(new RuleViolationImpl(ERROR_MESSAGE));
+        }
+
+        return result;
+    }
+
+    @SuppressWarnings("unchecked")
+    protected boolean isConnectionAlreadyFormed(final Node<? extends View<?>, ? extends Edge> source,
```
romartin: hey guys, here it is what I should do by doing some initial casts (not great code I know, had to review at some point...), but  this way the stream and functions to apply seams more clear... up to you!

` 
protected boolean isConnectionAlreadyFormed(final Node<? extends View<? extends Definition>, ? extends Edge> source,
                                                final Node<? extends View<? extends Definition>, ? extends Edge> target,
                                                final Edge<? extends View<? extends Definition>, ? extends Node> connector) {

        final String connectorDefId = getDefinitionId(connector.getContent());
        final List<Edge<View<? extends Definition>, Node>> outEdges =
                (List<Edge<View<? extends Definition>, Node>>) source.getOutEdges();
        return outEdges
                .stream()
                .filter(e -> connectorDefId.equals(getDefinitionId(e.getContent())))
                .map(Edge::getTargetNode)
                .filter(n -> n.equals(target))
                .findFirst()
                .isPresent();
    }

    private String getDefinitionId(final Definition definition) {
        return definitionManager.adapters().forDefinition().getId(definition);
    }

    private String getDefinitionId(final View view) {
        return definitionManager.adapters().forDefinition().getId(view.getDefinition());
    }
`

PR #990, opened 2017-07-24
Teams/Project metrics dashboards fixes

dgutierr:
Some fixes related to the Teams/Project metrics dashboards:

GUVNOR-3315: I removed the assets counter from the metric as is not possible to know (with existing uberfire's NIO2 API) what are the files a commit is related to. So instead of assets I'm just showing the number of commits per author in the project dashboard whereas the a projects counter is shown in the Teams dashboard.

GUVNOR-3310: The refresh mechanism has been replaced by an async update which occurs when the contributions dataset in the backend changes. That way we get rid of all the unnecessary polling requests the metric is sending to the backend.

@jhrcek @manstis Can you guys review please?


jhrcek:
@dgutierr ok, I'm going to test this locally today and provide feedback asap

dgutierr:
@jhrcek I understand your point. The reason I stuck with the bubble chart is because if the number of authors is high then either a bar or pie chart sucks. A bubble chart scales better.

jhrcek:
@dgutierr ok, now I understand your motivation for sticking with bubble chart. I'm not really sure bubble chart scales better in this case. Since both X and Y are based on # commits, all the bubbles are shown on the diagonal. Since all bubbles are big, they inevitably overlap and are not readable. See example from 2 projects from our testing repo:
bubble

In general case there's usually no more than 10-20 developers working on even relatively big projects. And we can always set x axis label angle to make even larger amount of x axis values readable. See example bar charts I configured to display similar data for our testing repo:


bar1
bar2


If you still think bubble chart is better, is there no other dimension we can find in the data to avoid repeating # of commits on both X and Y axes? Unfortunately looking at the code (all data being extracted from list of HistoryRecords) this would probably add additional complexity :(

If you still think the current implementation is ok and not worth the effort to improve, I'm willing to lower my high expectations and approve this for merging :)

dgutierr:
@jhrcek The bubble chart is ok imho. Notice, the chart itself only gives an overview of the contributions. For full details, the dropdown can be used, as every entry has a tooltip with the number of commits.

PR #997, opened 2017-07-27
[Guided Rule Editor] Unable to construct From Collect

jomarko:


kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/AsyncPackageDataModelOracleImplTest.java
```diff
@@ -162,212 +166,261 @@ private PackageDataModelOracleIncrementalPayload createDefaultPayload() {
...
+        collectionTypes.put("org.CustomCollection",
+                            true);
+        collectionTypes.put("com.CustomCollection",
+                            true);
+
+        return collectionTypes;
+    }
+
+    @Test
+    public void testGetAvailableCollectionTypes() throws Exception {
```
manstis: @jomarko Just an opinion.. it might be worth adding a comment as to _why_ it's these 3 (oracle's package is ```org``` and both ```List``` and ```HashSet``` are imported). It took me a while to figure it out; and a simple comment might save some head scratching in the future.

manstis: @jomarko You might also like a test for when there's an entry in "Project Collection Types" with value of ```false```

PR #998, opened 2017-07-27
Fix for random fails in Indexer tests.

hasys:
Hi @pefernan, @romartin,

I used IntelliJ _rerun until fail_ feature to test my fix. And this test passed about 100 times in a row without any problem. Without fix it failed in first 5 execution. So, I hope, it will help.

### DIFF
--- a/kie-wb-common-forms/kie-wb-common-forms-editor/kie-wb-common-forms-editor-backend/src/test/java/org/kie/workbench/common/forms/editor/backend/indexing/FormDefinitionIndexerTest.java
+++ b/kie-wb-common-forms/kie-wb-common-forms-editor/kie-wb-common-forms-editor-backend/src/test/java/org/kie/workbench/common/forms/editor/backend/indexing/FormDefinitionIndexerTest.java
@@ -97,6 +97,7 @@ public class FormDefinitionIndexerTest extends BaseIndexingTest<FormResourceType
     public void testFormsIndexing() throws Exception {
         List<Path> pathList = new ArrayList<>();
 
+        ioService().startBatch(ioService().getFileSystem(basePath.toUri()));
         for (String formFile : FORMS) {
             Path path = basePath.resolve(formFile);
             pathList.add(path);
@@ -104,6 +105,7 @@ public class FormDefinitionIndexerTest extends BaseIndexingTest<FormResourceType
             ioService().write(path,
                               formContent);
         }
+        ioService().endBatch();
 
         Path[] paths = pathList.toArray(new Path[pathList.size()]);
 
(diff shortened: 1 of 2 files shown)


hasys:
@manstis, should I create PR for 7.X branch as well?

manstis:
@hasys Firstly thank-you for looking into the failures. Secondly a PR for 7.2.x probably won't hurt.

hasys:
Hi @jsoltes,
can you review it as well, because it affect new forms as well.

@manstis, ok, thank you. Will do.

Thank you.

hasys:
Hi @manstis, @pefernan, @romartin, I have classic situation there. One good news and one bad news.

Good news: our's tests are stabilized.
Bad news: it looks like Uberfire indexer has two bags:
- Sometimes (one from 10 or 20 times) indexer shows the same file twice and we have "more" files than expected. It can occurs on Forms and on Stunner test.
- Sometimes Indexer lose "brokenSignal.bpmn" file. Indexer can found it on first check (all files are there) and lose it on the second check (filter "brokenSignal.bpmn" file only).

I need more time to investigations what's going on, because I am not sure we want a workaround for this kind of issues. WDYT?

manstis:
@hasys Yes, there are indeed some weird shenanigans regarding indexing tests when ran in parallel (i.e. on Jenkins). Each test class is meant to be isolated by using a different repository name but it clearly is insufficient. It could be as simple as [this](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/test/java/org/kie/workbench/common/services/refactoring/backend/server/BaseIndexingTest.java#L56) needing synchronising.

You should be aware (if you're treading the indexer tests path) that AppFormer received a PoC change thanks to @porcelli to [remove](https://github.com/AppFormer/uberfire/commit/0ea8a9e5063dfd0ef51c295ebac37c4aba2e9fdd) the need to have indexer tests wait. This not only means tests are much faster but should also minimise the false negatives we keep seeing.

In an ideal world **ALL** indexer tests in **ALL** repositories would be changed to remove the wait.. but as with most things occasional false negatives and a marvellous CI system to simply re-test to correct failures means any effort to fix the tests takes a very low priority.

Volunteers welcome :-)

hasys:

hasys:
Hi @manstis, @porcelli @ederign 

I reported two issues to Uberfire: 
- [AF-576](https://issues.jboss.org/browse/AF-576) - The same file can be indexed twice
- [AF-577](https://issues.jboss.org/browse/AF-577) - Random ArrayIndexOutOfBoundsException in Indexer test

First issue is first point from my previous message. And it looks like the second issue from my message is very rear combination of two reported issues (failed indexing of incorrect business process and one correct file is indexed twice).

PR #1007, opened 2017-07-28
JBPM-6049: Allow user to put spaces in the name of processes

Christopher-Chianelli:
To do this I need to URL-encode the name when creating a new resource, and URL-decode the name when displaying the resource name. What are all the widgets that read the title names? I did the main ones, but there are a few I missed (ex: Case Management).

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonUnmarshaller.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonUnmarshaller.java
@@ -1035,6 +1035,12 @@ public class Bpmn2JsonUnmarshaller {
         for (RootElement root : rootElements) {
             if (root instanceof Process) {
                 Process process = (Process) root;
+                if (process.getId() != null) {
+                    String processId = process.getId().trim();
+                    processId = processId.replaceAll("\\s",
+                                           "");
+                    process.setId(processId);
+                }
                 List<FlowElement> flowElements = process.getFlowElements();
                 for (FlowElement fe : flowElements) {
                     if (fe instanceof DataObject) {
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/test/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonUnmarshallerTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/test/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonUnmarshallerTest.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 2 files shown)


Christopher-Chianelli:
@manstis Both Business Process and Case Management need to be encoded; Packages and Data objects aren't allowed to have spaces or percent sign in their name, so the encoding does nothing for them. The only exception is Project, but I think that has a different new resource handler. The most ideal solution would be to only encode/decode it in whatever is making/receiving requests (`AbstractVFSDiagramService`), but doing it only there causes it to not find the diagram. There IS a very odd workaround though. If you create a title with no spaces, then rename it to have spaces, no issues occur. So a trick I can do is create the file using an UUID, and then rename it to the original file name. Unfortunately, I have no idea how I would go about that.

manstis:
@Christopher-Chianelli Sorry, but I still do not understand why we need to escape the file name in the GWT code. Designer's [NewProcessHandler](https://github.com/kiegroup/jbpm-designer/blob/master/jbpm-designer-client/src/main/java/org/jbpm/designer/client/handlers/NewProcessHandler.java) and Case Management's [NewCaseDefinitionHandler](NewCaseDefinitionHandler) will both happily pass a file name with a space to ```DesignerAssetService``` so where is the encoding actually _needed_? I still think adding the encoding/decoding in the client is masking where the problem actually is.

manstis:
Jenkins please retest this.

manstis:
@Christopher-Chianelli You really need to see some of these ```kie-wb-common``` PRs through to completion!

Christopher-Chianelli:
@manstis This one is a bit tricky; I am still looking for the root cause of the bug to avoid encoding the spaces, no idea where to begin looking to be honest.

romartin:
Hey @Christopher-Chianelli , is this commit related to the `https://issues.jboss.org/browse/JBPM-6036` too? That ticket is critical for next release, and I see that its state is on "pull request sent", but not sure which is the right PR for it. Please try to move this forward and close the critical issue asap, deadline is next week.
Thanks!

Christopher-Chianelli:
@romartin Yes; I have a work around for the issue (url-encoding the names), but not a fix for the root cause (which I still don't know). Is there another PR for this? If so, it probably that one.

romartin:
@Christopher-Chianelli  well that's why I was asking hehe... IDK if there is some other PR or only this one will fix the ticket. If you look at the ticket it says PR sent....
Anyway please, move this forward, identify the commits missing to merge etc and fix those, if necessary, this error is identified as a critical issue for next release, so we need this merged into `master` and `7.3.x` asap, deadline is next week.
Thanks!

Christopher-Chianelli:
@romartin I have pushed a work around which basically creates a new file with a UUID then renames it to the actual file name (since file renaming with spaces is allowed, but file creating with spaces is not).

Christopher-Chianelli:

Christopher-Chianelli:
@romartin Yeah; hence a "work-around". I am trying to find the root issue; appears it is in the unmarshaller somewhere...

Christopher-Chianelli:
@romartin After a lot of searching through the JsonUnmarshaller, I found the root issue: in `updateIDs` they only removed the whitespace in the id's of the process flow elements, forgetting to remove the whitespace in the process id as well.

Christopher-Chianelli:

romartin:
great @Christopher-Chianelli ! lemme do it a try :)

romartin:
Hey @Christopher-Chianelli , tested and everything seems good, just please add the missing tests. Thanks!

Christopher-Chianelli:
@romartin Writting tests for this is somewhat awkward; I cannot define a new bpmn file with a space in the id in the BPMNMarshallerTest, since that is illegal and will cause an exception when unmarshalling it in the test (the fix doesn't prevent the exception from reading bad XML files, but rather prevent writting a bad XML file). Nor could I easily create a diagram with a space in the name (the JSONParser giving a NPE when I use the factory mocked in the test). Thus I need to mock a bunch of things just to create the test! I am not sure if I can get the test ready by tommorow...

romartin:
hey @Christopher-Chianelli , well do not have right now all the concrete detail for this, another option could be creating a package protected method for your concrete logic and just testing, at least, the method does what's expected to do. Anyway AFAIK the deadline is next Friday (Oct 6th), not tomorrow, so no worries for today.

Christopher-Chianelli:
@romartin Added a test on the updateIDs method (which fixes the IDs)

romartin:
Jenkins please retest this

hasys:

romartin:
@manstis  are you ok with merging this one? cannot see your approval yet :) Thanks!

PR #986, opened 2017-07-19
GUVNOR-3168 GUVNOR-3167 [DMN Editor] Write unmarshaller marshaller

tarilabs:
Early to be merged, just sharing to validate the current ongoing approach... @manstis can you kindly try it and give me some feedback, please?

PR #1016, opened 2017-08-01
GUVNOR-3362: Unit test for AssociationConverter

jomarko:
This PR is dependent on #986 

@manstis @tarilabs question for you, during writing this test and playing with DMN editor, I realized that the associations can be created just in direction to TextAnnotations. So I think TextAnnotaion can't have outgoing edges in our DMN editor, am I missing something? Is there really need for the code processing `outgoing edges` in the `AssociationConverter` class? If there is need, I should extend my test.Thanks for your comments.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/AssociationConverterTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/AssociationConverterTest.java
@@ -1,0 +1,270 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.backend.definition.v1_1;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.junit.runner.RunWith;
+import org.kie.workbench.common.dmn.api.definition.v1_1.Association;
+import org.kie.workbench.common.dmn.api.definition.v1_1.BusinessKnowledgeModel;
+import org.kie.workbench.common.dmn.api.definition.v1_1.Decision;
+import org.kie.workbench.common.dmn.api.definition.v1_1.DRGElement;
+import org.kie.workbench.common.dmn.api.definition.v1_1.InputData;
+import org.kie.workbench.common.dmn.api.definition.v1_1.KnowledgeSource;
+import org.kie.workbench.common.dmn.api.definition.v1_1.TextAnnotation;
+import org.kie.workbench.common.dmn.api.property.dmn.Description;
+import org.kie.workbench.common.dmn.api.property.dmn.Id;
+import org.kie.workbench.common.stunner.core.graph.Edge;
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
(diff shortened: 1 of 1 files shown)


tarilabs:
According to DMN spec Association can make a link between a Artifact and a DMNElement in any "direction". cc'd @etirelli who may confirm or correct me accordingly.

At today we are using between a TextAnnotation (the only standard "node" artifact) and the DRGElements which is the only combination I can achieve; dmn test file `potpourri.dmn` in the original PR demonstrates this, in either of "directions".

Finally please be aware the need is arising because Stunner graph cannot be queried for "Edges", so currently the only possible implementation was to "lock" from a TextAnnotation node, and process the edges from that perspective.

etirelli:
@jomarko @tarilabs not sure I understand the question?

tarilabs:
@etirelli in other words, an Association could be drawn in either of the following "direction", sourceRef Vs targetRef:
```
TextAnnodation --(Association)--> DRGElement
DRGElement --(Association)-->  TextAnnodation
```

So if the above is correct, it would imply TextAnnotaion **_can_** have outgoing edges

etirelli:
@tarilabs @jomarko page 43 of the spec details the Association element. In particular it uses the associationDirection attribute to define if the link should have no arrow, or arrow on one side, or arrow on both sides of the link. So I believe not only both of the scenarios @tarilabs described on the previous comment are possible, but also the arrows may or may not be present, depending on that attribute.

manstis:
I'd be interested in understanding what tooling supports defining different directions of Association. At the moment our editor only supports one direction.

jomarko:
@tarilabs @manstis @etirelli Thank you all for your comments and explanations. I decided to cover also outgoing edges in this PR however not so properly as incoming edges. Please have a look and feel free to approve/decline/any action.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/AssociationConverterTest.java
```diff
@@ -0,0 +1,254 @@
...
+        when(knowledgeSourceAssociation.getId()).thenReturn(new Id("knowledgeSourceAssociationId"));
+        when(inputDataAssociation.getId()).thenReturn(new Id("inputDataAssociationId"));
+        when(drgElementAssociation.getId()).thenReturn(new Id("drgElementAssociationId"));
+
+        when(textAnnotationNode.getInEdges()).thenReturn(Arrays.asList(edgeFromDecision,
+                                                                       edgeFromBkm,
+                                                                       edgeFromKnowledgeSource,
+                                                                       edgeFromInputData));
+        when(textAnnotationNode.getOutEdges()).thenReturn(Arrays.asList(outgoingEdge));
+    }
```
manstis: You have to love this :-)

~210 lines to setup the test and 40 lines of tests!

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/AssociationConverterTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    @Test
+    public void testDmnFromWBNoEdges() throws Exception {
+        when(textAnnotationNode.getInEdges()).thenReturn(new ArrayList<Edge<View<Association>, ?>>());
```
manstis: Personal preference... use of ```Collections.emptyList()```

manstis:
@tarilabs Please can you review too; re: sufficient coverage of your converter.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/AssociationConverterTest.java
jomarko: I agree, I will try to enhance/simplify the setup.

tarilabs:
Hi, putting aside my bias for Mocking, I'm a bit worried the current might be duplicating code logic and if we change the real code we might need to realign in two sides. That is why currently the test in the original PR try to leverage a round-trip mechanism, so to try maximize code coverage in both directions. I just wonder if what would benefit the most currently would actually be a programmatic test, building a Stunner graph and roundtrip from the UI perspective, so to obtain the dual as well. What do you think? 

Anyway I understand this is test style for kie-wb so besides my suggestion above which I hope is helpful, I have no further comments. Thank you!

jomarko:
Well, I think I can't simplify this too much. Leaving this as it is. Regarding other converters, I will try to find another way (no so much mocking) how to test them. Please merge this when #986 is in.

PR #1039, opened 2017-08-04
Index test stabilization

hasys:
Hi @manstis,

There are tree changes:

- Increase wait time (will fix ~ 4% of random fails)
- Use FS Batch operation (will fix ~ 80% of random fails)
- shadow the Uberfire bug. I will send a PR directly to Uberfire to test it. After that I will make a same PR (with Set instead of List) for Forms and Stunner. (have to fix the rest ~ 15%) of fails

P.S. But there are still small chance for random fail in case of sleep time won't be enough.


manstis:
@hasys Do you know why the use of "batch" and ```Set``` should fix anything?!?

Also, are you aware that Uberfire has support for removal of all "waits" completely? But we've never found time to make the necessary changes to all the "indexing" related tests in ```kie-wb-common```, ```drools-wb``` and probably Form Modeller, OptaPlanner, jBPM Console etc..

hasys:
@manstis, 

- `Set` will eliminate file duplication caused by [AF-576](https://issues.jboss.org/browse/AF-576)
- Batch force FS indexation, if I correct
- I am not sure that Byteman alternative is better than simple Wait, I wish one day we will have better way (some events or something similar) how to get notification about finished indexation.

hasys:

manstis:

PR #1044, opened 2017-08-07
GUVNOR-3369: Test Marshalling of basic Text Annotation properties

jomarko:
@tarilabs @manstis again, something for your eyes.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
@@ -27,6 +27,7 @@ import org.apache.tools.ant.filters.StringInputStream;
 import org.jboss.errai.marshalling.server.MappingContextSingleton;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.kie.api.KieServices;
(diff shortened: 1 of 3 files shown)


jomarko:
@tarilabs very good point. I will check CDATA too.

jomarko:
@tarilabs please check the PR now. I added the test&file that uses `CDATA` for all properties (text, text format and description).  The problem is that the added test fails because of `CDATA` in text format and description. If there is `CDATA` just for text then marshalling works well. Could you please download my PR locally, unignore the test a help me with investigating this?

tarilabs:
I checked.
Concerning the attribute, AFAIK you can't place CDATA inside the attribute value in xml document because it's already considered a CDATA.
Concerning the line with ideogram, I don't know what is the problem tho my XML editor did not correctly recognized the end of the CDATA with the double `]]`. I believe you might have copy pasted by mistake some spurious unicode characters in there. I solved locally by revwriting by hand the element and just pasting the 2 ideograms inside the CDATA once I completed it.
Or you can try with the below
```
  <textAnnotation id="60915990-9E1D-42DF-B7F6-0D28383BE9D1" textFormat="text/html">
    <description><![CDATA[描述]]></description>
```
which I copied from my editor and solved the problem for me.

For instance, before it was:
image

Then I erased the complete element,
I rewrote the decrirption element open close
I rewrote the CDATA open close
and then inside I pasted the 2 ideograms got:
image

Although it "looks" the same, clearly something in the underlying byte representation is not. Unicode encoding is a funny world =)

jomarko:
@tarilabs thanks. I amended the commit. Lets see the test result.

jomarko:

jomarko:

jomarko:
@manstis it is green, finally :)

PR #1045, opened 2017-08-07
Guvnor 3373

desmax74:
Fix for kieAllBuild

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/test/projects/dummy/pom.xml
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/test/projects/dummy/pom.xml
@@ -39,4 +39,27 @@
             </plugin>
         </plugins>
     </build>
+     <repositories>
+        <repository>
+            <id>kie-group</id>
+            <url>https://repository.jboss.org/nexus/content/groups/kie-group/</url>
+        </repository>
+        <repository>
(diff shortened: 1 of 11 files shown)


kie-wb-common-services/kie-wb-common-services-backend/src/test/projects/dummy/pom.xml
```diff
@@ -39,4 +39,27 @@
...
+            </releases>
+            <snapshots>
+                <enabled>true</enabled>
+                <updatePolicy>daily</updatePolicy>
+            </snapshots>
+            <id>jboss-public-repository-group</id>
+            <name>JBoss Public Repository Group</name>
+            <url>https://repository.jboss.org/nexus/content/groups/public/</url>
+        </repository>
+    </repositories>
```
psiroky: This should IMO be configurable externally, probably via settings.xml.

desmax74: Yep, but for this we need to add a class for this in the external339 package, this is a short term fix

desmax74: @psiroky  I'll close this PR and open deserved only for this problem to document better

PR #1046, opened 2017-08-07
GUVNOR-3389 fix for test fail on kieAllBuild

desmax74:
Short term fix until we add and external339 class to manage correctly settings.xml


desmax74:
@psiroky @porcelli

manstis:
@psiroky @desmax74 IIRC I saw some comments on a similar (identical?) PR about parameterizing the repository URLs? Can this be merged, or should I wait for something?

porcelli:
This is not the solution, we should work on real solution

PR #1054, opened 2017-08-09
JBPM-6103 Fix removing unnecessary depencency from showcase standalone

tiagodolphine:
@romartin just removing the shared dependency that is not necessary on the showcase standalone module in response to the comment on: https://github.com/kiegroup/kie-wb-common/commit/a33699b8e0d6b64ec6a946449e80af960124a1c8#commitcomment-23545859


romartin:
oh it was already merged, great I'm always on time... :)

PR #1062, opened 2017-08-09
GUVNOR-3354: [DMN Editor] Toolbar: Add support for using different SessionCommandFactory

manstis:
See https://issues.jboss.org/browse/GUVNOR-3354

This PR adds a new ```SessionCommandFactory``` for the DMN Editor; with a ```ClearSessionCommand``` that does not respond to DMN-centric events that would lead to the Toolbar button becoming enabled.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/DMNCommand.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/DMNCommand.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


manstis:
Jenkins please retest this.

manstis:
Hi @jomarko feel free to propose alternative interface names :-) 

It is only additional ```@Observers``` that veto the event and not the first execution/undo of the event. So for this PR, for example, it is only the ```@DMNEdtior``` version of ```ClearSessionCommand``` that is vetoing commands.

```NavigateToDRGEditorCommand``` only implements ```VetoUndoCommand``` as an undo "means" navigating _to_ the Expression editor (_from_ the DRG editor). In this scenario we do not want the ```ClearSessionCommand``` to re-enable the "Clear" toolbar button. Likewise ```NavigateToExpressionEditorCommand``` only implements ```VetoExecutionCommand``` as, when navigating _to_ the Expression editor (_from_ the DRG Editor) we do not want the ```ClearSessionCommand``` to re-enable the "Clear" toolbar button.

jomarko:
@manstis thanks a lot for this answer, I appreciate your patience. I think there is no need for alternative interface names. There was just problem in my understanding. It is much better now, Thanks.

PR #1063, opened 2017-08-09
[JBPM-5544] Text displayed in the model now support auto-wrap

Christopher-Chianelli:


kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExt.java
```diff
@@ -176,6 +176,7 @@ public T setFillGradient(final Type type,
         this.fillGradientEndColor = endColor;
         if (null != getShape()) {
             final BoundingBox bb = getShape().getBoundingBox();
+            textViewDecorator.setTextBoundaries(bb);
```
manstis: Why the need to set text bounds when setting the fill gradient? And.. if this is for good reason, then why not do it [here](https://github.com/Christopher-Chianelli/kie-wb-common/blob/23c7dfd45dfff9d7840b2f9e3f08d26228dbfe5c/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExt.java#L189) so that potential use of either public methods sets the text bounds?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExt.java
```diff
@@ -357,14 +358,23 @@ private void addTextAsChild() {
...
+            removeChild(textViewDecorator.getView());
+            addTextAsChild();
+            textViewDecorator.setTextBoundaries(getShape().getBoundingBox());
         });
         HandlerRegistration r2 = addWiresResizeEndHandler(wiresResizeEndEvent -> {
             final ResizeEvent event = buildResizeEvent(wiresResizeEndEvent);
             resizeHandler.end(event);
+            removeChild(textViewDecorator.getView());
+            addTextAsChild();
+            textViewDecorator.setTextBoundaries(getShape().getBoundingBox());
```
manstis: These changes will ensure the text bounds are updated when a shape's size is changed with DnD. However I do not see any changes relating to when a shapes' size is changed via the Properties Panel. Should you not also be adding support for the text bounds size change in ```AbstractCanvasHandler.applyElementMutation(..)```  or related methods too?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExt.java
Christopher-Chianelli: @manstis This is to make sure the text is position correctly when the `ShapeView` is constructed. For some reason doing the call in the constructor results in the text being incorrectly positioned, but doing it here results in it being correctly positioned. I moved it to `updateFillGradient`.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExt.java
Christopher-Chianelli: @manstis Added an override to `refresh`, so mutations should cause the text bounds to be updated.

Christopher-Chianelli:
@manstis Tests are here, since the majority of the patch is lienzo side: https://github.com/ahome-it/lienzo-tests/blob/master/src/test/java/com/ait/lienzo/client/core/shape/wires/WiresTextTest.java

manstis:
@Christopher-Chianelli Thanks. Those tests check text formatting works; but not that ```WiresShapeViewExt``` adjusts the text's bounding box when resized etc so I still think there are tests missing.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeView.java
```diff
@@ -103,6 +103,7 @@ public T setSize(final double width,
                              0d,
                              width,
                              height);
+        refresh();
```
manstis: What about when the radius of a circle is changed?

romartin:
going to review it just right now, sorry for the delay guys

manstis:
@romartin https://www.youtube.com/watch?v=MXXRHpVed3M

romartin:
@Christopher-Chianelli I'm getting a NPE in client side when trying to open the evaluation process, you?? Creating a new process for now works...

romartin:
@Christopher-Chianelli also the text for the line has been broken :( It's rotated, and now it appears on the wrong location:

lane-text-issue

manstis:
4 @romartin Happy PTO: https://www.youtube.com/watch?v=rQtcpDGWx0w

romartin:
@manstis hahahaha I though that videos couldn't exist!! :P

romartin:
@manstis yeah can see myself like the ones in the videos... same structural body as mine xD

Christopher-Chianelli:
@romartin Found out the cause of the NPE: `String.split()` returns an empty array if the String consists of only spaces, and I assumed `String.split()` always return at least one element. The bug is on lienzo side; I can walk around the issue by trimming the whitespace on the end of the strings (an empty string returns an array of one element, since the String wasn't split).

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeView.java
Christopher-Chianelli: @manstis I couldn't find any method for when the radius of a circle changes. However, wouldn't said method call `refresh()` or `resize(...)`, both of which update the Text's `BoundingBox`?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExt.java
romartin: yes, it was missing the text update as @manstis pointed. So it's fine @Christopher-Chianelli  now overriding the `refresh`, but why `refresh` just updates the text boundaries, but on the other hand,  the resize handlers also remove/add the text primitive? So why not same logic in both cases, by calling some common method for it?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeView.java
romartin: well Stunner actually doesn't care about kind of shape, it only considers bounding boxes for elements/shapes, so circles, diamonds, rectangles, paths.. whetever.. once are being resized from the UI, the model element for it is updated as for the new bounding box side, and then this setSize method is called too. So this method is called for whatever kind of shape, but actually, bpmn does not allows resizing circles or gateways.
PS: @Christopher-Chianelli  can be a good test, if you have time, to just enable the resize for events and gateways locally, by changing [this flag,](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/shape/def/StartEventShapeDef.java#L141) and check the text wrapping works as expected too.

romartin:
@Christopher-Chianelli  thanks, it works fine now. Just added a couple of minor comments in this PR, but rather than that, all good for me!

romartin:
Hey @Christopher-Chianelli , 

Just found another trivial bug, it appears once cliking for opening the preview/explorer dock panel, see screenshot:

screenshot from 2017-08-13 01-00-02

Pleeease, you've to try to get used to spend more time on testing the commits... I spent 2 minutes testing this and found 2 really basic features broken (also was doing a great video trying to show your stuff.. and suddenly zaaas, crashed!! haha).  Also the downside is that this consumes times for everyone who's testing your stuff, and re-testing it again... We all produce bugs, me included as the first candidate always!! So let's try to reduce the probability by getting used to do some smoke quick tests for each delivery. Thanks!! :)

Christopher-Chianelli:
Fixed it; didn't know that `setTitle` could be called with a null argument. Did some smoke testing this time in both bpmn and case-management; didn't notice any bugs caused by this pull request, but noticed sometimes text is positioned incorrectly when undoing. Adding unit tests.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeView.java
Christopher-Chianelli: Just tested changing the size of the start event using @romartin suggestion. `refresh()` is being called, but the radius attribute isn't updated in `com.ait.lienzo.client.core.shape.Circle`, so it is using the old radius. What do you think @manstis?

Christopher-Chianelli:
@romartin @manstis Added unit tests

manstis:
Jenkins please retest this.

romartin:
Please remember to cherry-pick this into 7.X as well, once it gets merged! Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
```diff
@@ -0,0 +1,92 @@
...
+                                                                height);
+        WiresShapeViewExtTest.setPrivateField(WiresShapeViewExt.class,
+                                              tested,
+                                              "textViewDecorator",
+                                              textDecorator);
+    }
+
+    @Test
+    public void testTitle() {
+        tested.setTitle(null);
```
manstis: Missing assertion/verification.. what should be/should not be happening when title is ```null```?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void testTextWrapBoundariesUpdatesOnRefresh() {
+        tested.refresh();
+        verify(textDecorator).setTextBoundaries(any());
```
manstis: Can we not verify that the text boundaries are correct? Verifying that we simply set the boundaries on the text decorator is better than nothing but can we not ascertain the values that should be being set?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void testTextWrapBoundariesUpdatesOnResize() {
+        tested.resize(0,
+                      0,
+                      10,
+                      10,
+                      true);
+        verify(textDecorator).setTextBoundaries(any());
```
manstis: Can we not verify that the text boundaries are correct? Verifying that we simply set the boundaries on the text decorator is better than nothing but can we not ascertain the values that should be being set?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void testTextWrapBoundariesUpdatesOnSetSize() {
+        tested.setSize(10,
+                       10);
+        verify(textDecorator).setTextBoundaries(any());
```
manstis: Can we not verify that the text boundaries are correct? Verifying that we simply set the boundaries on the text decorator is better than nothing but can we not ascertain the values that should be being set?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExtTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test
+    public void testTitle() {
+        tested.setTitle(null);
+    }
+
+    @Test
+    public void testTextWrapBoundariesUpdates() {
+        tested.refresh();
+        verify(textDecorator).setTextBoundaries(any());
```
manstis: Can we not verify that the text boundaries are correct? Verifying that we simply set the boundaries on the text decorator is better than nothing but can we not ascertain the values that should be being set?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExtTest.java
```diff
@@ -0,0 +1,80 @@
...
+                                              PATH);
+        setPrivateField(WiresShapeViewExt.class,
+                        tested,
+                        "textViewDecorator",
+                        textDecorator);
+    }
+
+    @Test
+    public void testTitle() {
+        tested.setTitle(null);
```
manstis: Missing assertion/verification.. what should be/should not be happening when title is ```null```?

romartin:
oh good points @manstis !! forgot to take a look at the new tests on my previous review!... sorry :(

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
Christopher-Chianelli: This test is asserting an exception is not thrown; it happened often enough that `setTitle()` threw an exception when called with a null argument I decided to add a test for it. Unfortunately, there isn't a "assert no exception is thrown" assertion that I know of, I can probably make it more explicit using
```
try {
    tested.setTitle(null);
}
catch (Exception e) {
    fail("The follow exception occurred: " + e.getStackTrace());
}
```

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
Christopher-Chianelli: It is possible, it just require a lot of mocking...

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExtTest.java
Christopher-Chianelli: See above

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresShapeViewExtTest.java
Christopher-Chianelli: See above

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
romartin: Just adding a quick javadoc in the test method could be fine for me

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/DecoratedShapeViewTest.java
romartin: IMO I'm agree with @manstis that makes sense testing that the wrap boundaries set to the text instance are the right ones! At least test it once in a separate method (with the necessary mocks..). Then is fine keeping these other checks just using `any()` for verifying the method is called on refresh...

Christopher-Chianelli:
@romartin @manstis Added javadoc and assertions for tests.

PR #1073, opened 2017-08-14
GUVNOR-3374: [Stunner] Add support to enable/disable KeyboardEvent handling

manstis:
See https://issues.jboss.org/browse/GUVNOR-3374

This PR introduces a new ```CanvasControl``` for ```KeyboardControl``` that is instantiated when a session is created - as with all other controls. All existing keyboard (shortcut) code has been moved to the new control. A new ```SessionAware``` interface has been added and used on controls and commands that need to be bound to a session. These changes allow the ```KeyboardControl``` to be enabled/disabled (which is used to have a different keyboard handler used by the DMN Editor when in "Expression editing" mode of operation).

@jomarko The only "testable" Use Case is to navigate to a (Literal) Expression select the table and press ```DELETE```. Before this change Stunner's ```DeleteSelectionSessionCommand``` would execute and the selected Node on the graph deleted (you'd see it disappear in the preview panel and navigating back to the DRG Editor would show the node had been deleted). The ```DELETE``` key in the Expression Editor will clear a cell (but that'll be implemented with https://issues.jboss.org/browse/GUVNOR-3266).


manstis:
@hasys I also added you in case you want to look at the Stunner changes (that are probably more extensive than those for the DMN Editor side of things).

manstis:
@jomarko negative checks added to keyboard shortcut tests and requested.

I also refactored the tests to extend a base class to minimise repetition.

PR #1081, opened 2017-08-17
JBPM-6052: Stunner - Component Library is hidden after scrolling canvas

tiagodolphine:
Changing palette and presenter css to fix palette position on canvas scrolling.

@manstis 
@hasys 
@ibek


manstis:
I'll review this on the basis of the changed code and not the reported issue!

Personally, I believe the issue is that the ```LienzoPanel``` should be sized to fit into the available Workbench perspective real-estate (i.e. browser window width minus palette width minus width of other panels etc) and adjusted as other items change. We do this for the Guided Decision Table Editor and the "canvas" never has a scroll bar (Stunner sort of just picks an arbitrary width for ```LienzoPanel```). Anyway... I digress ;-)

hasys:

manstis:
@hasys Please let me know whether your waiting on the full downstream build.

hasys:
@manstis Full downstream is green here. One test failed in normal build. I guess we can merge it, because Full Downstream build was successful.

manstis:
@hasys Ah yes, I just saw one red and one green....

PR #1088, opened 2017-08-18
Reduce amount of random test fails

hasys:


hasys:
Fix this random fails: https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-common-pullrequests/1795/testReport/junit/org.kie.workbench.common.screens.impl/FindAllLibraryAssetsSortedQueryTest/listAllInProjectSortedPaged/

manstis:
Jenkins please retest this.

hasys:
Hi @manstis, it is green now.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/test/java/org/kie/workbench/common/screens/impl/FindAllLibraryAssetsSortedQueryTest.java
```diff
@@ -58,25 +61,34 @@ public ResponseBuilder getResponseBuilder() {
...
+    private void setUp(String path) throws IOException, InterruptedException {
+        ioService.startBatch(basePath.getFileSystem());
+        //Add test files
+        for (String fileName : FILE_NAMES) {
+            addTestFile(path,
+                        fileName);
+        }
+        ioService.endBatch();
+
+        Thread.sleep(5000); //wait for events to be consumed from jgit -> (notify changes -> watcher -> index) -> lucene index
```
rsynek: @hasys too late, it has been already merged. Still - the test will run every time for 5+ seconds. I wonder if the Thread.sleep could be replaced by polling on certain condition with timeout?

hasys: Hi @rsynek, unfortunately for now - there are no condition to pull. But we have to alternatives:

- try assertion -> if failed, wait some time -> try assertion again. It can reduce waiting for 1-2 seconds but will artificially complicate test code, I am not sure it is much better.
- Use Byteman, but I am not sure it is much better way as well.

manstis: @rsynek @hasys There is support to use Byteman in AppFormer (the _source_ of the "indexing" code and tests). We simply have not had the time to move all indexing related tests (guess at ~200) to use the new provision.

rsynek: @hasys If you prepare a utility method providing such a functionality, you can reuse it from multiple tests as a one-liner and it won't be so much complicated.

@manstis thanks for update, I would rather avoid using Byteman, but it's probably just a matter of taste. 

One solution or another - your call guys, I wanted only to point out that we should avoid simple Thread.sleep() where ever possible.

PR #1096, opened 2017-08-21
JBPM-6304 Stunner - gateway doesn't show properties

tiagodolphine:
Now it is possible to open the Exclusive Gateway properties panel.
The problem was that the **DefaultRouteFormProvider** was using a **TreeMap** implementation to provide SelectorData to the Combobox of the DefultRoute property.
The methods **get(key)** and **containsKey(key)** of the **TreeMap** class throws NullPointerException whether the key parameter is null. The HashMap doesn't throw Exception in case key is null.

Reference:
https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html
https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html

@hasys 
@manstis

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/gateway/DefaultRouteFormProvider.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/gateway/DefaultRouteFormProvider.java
@@ -16,6 +16,7 @@
 
 package org.kie.workbench.common.stunner.bpmn.client.forms.fields.gateway;
 
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


tiagodolphine:
Hi @manstis  thanks for the feeback, I agree with you and maybe that was the reason TreeMap was used, but since this is a data provider, I think it would be better that this role of sorting and the sorting rule is handled by the component renderer that knows how to display data to the user, in this case the renderer is on the forms module [1] and it makes use of _null_ values, I think it should be more elegant to refactory this renderer do support empty values without using null and the sorting rules as well, if so, do you think is this the scope of this task? or should we opened other task to the forms module? anyway, using HashMap will solve this specific problem but I agree that we should solve the cause that is on the renderer. 

[1]
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/main/java/org/kie/workbench/common/forms/dynamic/client/rendering/renderers/selectors/listBox/AbstractListBoxFieldRenderer.java#L47

hasys:

hasys:
Hi @tiagodolphine, @manstis I agree with you, this issue is not related to this PR. I reported Jira issue to handl it later on Forms side: [JBPM-6372](https://issues.jboss.org/browse/JBPM-6372).

PR #1122, opened 2017-08-31
JBPM-6104: Stunner - not possible to remove container with not empty container inside

alepintus:
Hi @hasys ,  @manstis , @romartin ,
Fixed a bug that did not allow to remove a container with a non-empty container inside.

For more information see [this ticket](https://issues.jboss.org/browse/JBPM-6104).

Could you please review this PR?

Thanks.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/SafeDeleteNodeProcessor.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/SafeDeleteNodeProcessor.java
@@ -16,10 +16,11 @@
 
 package org.kie.workbench.common.stunner.core.graph.util;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.Stack;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Stream;
@@ -70,7 +71,7 @@ public class SafeDeleteNodeProcessor {
 
     @SuppressWarnings("unchecked")
     public void run(final Callback callback) {
-        final Stack<Node<View, Edge>> nodes = new Stack<Node<View, Edge>>();
+        final Deque<Node<View, Edge>> nodes = new ArrayDeque();
         processedConnectors.clear();
         childrenTraverseProcessor
                 .setRootUUID(candidate.getUUID())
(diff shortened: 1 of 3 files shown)


manstis:
@alepintus A unit test would be nice :-)

manstis:
ok to test

hasys:

alepintus:
Hi @manstis , @hasys , @romartin  
Added missing test.

Could you review please?

Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/SafeDeleteNodeProcessor.java
```diff
@@ -92,10 +92,16 @@ public boolean startNodeTraversal(final List<Node<View, Edge>> parents,
                                   return true;
                               }
                           });
+
         // Process delete for children nodes.
-        nodes.forEach(node -> processNode(node,
-                                          callback,
-                                          false));
+        int nodesSize = nodes.size();
```
romartin: ok so it seems the fix is about reversing the collection order, why not keeping the `nodes.forEach` and instead just changing the type for `nodes`?  Actually it's a Stack, so what about just changing it to a `Queue`, like an `ArrayDequeue`? This way no need to add logic for controlling the iteration, you can just use the internal iterator `forEach`.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/util/SafeDeleteNodeProcessorTest.java
```diff
@@ -139,4 +144,43 @@ public void testDeleteParentNode() {
         verify(callback,
                times(1)).deleteCandidateNode(eq(graphHolder.parentNode));
     }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testDeleteParentWithNonEmptyContainerInside() {
```
romartin: that's fine, you're asserting all commands are being executed with the right arguments, but one of the key points to test is the iteration order, as your fix does (by reverting the actual order), so it seems to be missing here this 2nd part - ensuring the methods are being called in the right order.

alepintus:
Hi @romartin @hasys ,
code modified as suggested and enforced check on the invocation order of the called methods.

Could you please review this PR?

Thanks!

romartin:
@alepintus  once this gets merged, please remember to cherry-pick it into `7.3.x` branch as well. Thanks!

PR #1124, opened 2017-09-01
JBPM-6220: Unable to complete your request after setting the Show Time property to false.

pefernan:
Copy of https://github.com/kiegroup/kie-wb-common/pull/1123

There was two issues here:

a JS error going on when the component floating div was added to the DOM because the container for it wasn't initialized.
a CSS issue when displaying the DatePicker floating div appear on scroll bars instead of overflow the layout element container.
This PR initializes the DatePicker with the container for the floating div, now it's being added to the body (as the DateTimePicker does by default) to avoid overflow issues.

@jsoltes can you review?


jsoltes:

manstis:
@pefernan If you want this in **todays** 7.3.0.Final release you'll need to let @mbiarnes know the commit.

PR #1133, opened 2017-09-05
AF-593: Decouple kie-wb-common-datamodel from Drools

mbarkley:


mbarkley:
Related:
https://github.com/AppFormer/uberfire/pull/822
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/541
https://github.com/kiegroup/droolsjbpm-knowledge/pull/263
https://github.com/kiegroup/drools/pull/1444
https://github.com/kiegroup/jbpm/pull/973
https://github.com/kiegroup/droolsjbpm-integration/pull/1163
https://github.com/kiegroup/drools-wb/pull/596
https://github.com/kiegroup/jbpm-designer/pull/654
https://github.com/kiegroup/jbpm-wb/pull/864
https://github.com/kiegroup/optaplanner-wb/pull/218

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-api/src/main/java/org/kie/workbench/common/services/datamodel/spi/DataModelExtension.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.kie.workbench.common.services.datamodel.spi;
+
+import java.util.List;
+
+import org.appformer.project.datamodel.oracle.ExtensionKind;
+import org.uberfire.java.nio.file.DirectoryStream.Filter;
+import org.uberfire.java.nio.file.Path;
+
+public interface DataModelExtension {
```
manstis: Would this be better suited in AppFormer?

mbarkley:
Jenkins, please retest this.

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-api/src/main/java/org/kie/workbench/common/services/datamodel/spi/DataModelExtension.java
mbarkley: This will be moved to AppFormer eventually, but for now I have kept it in the same module as the DMO.

There will be another round of PRs next fixing Guvnor dependencies in the DMO, after which (maybe in the same PR) the entire kie-wb-datamodel module will move to AppFormer.

mbarkley:
@manstis Yes the extensions were added to decouple the DMO from Drools, but the motivation was more pragmatic than design-driven. Drools depends on uberfire-maven-utils. It makes sense to keep Maven utilities in AppFormer so I needed to break the DMO dependency on Drools.

I don't know enough about how enumerations are used to have an opinion one way or the other. But I would prefer to not include such as change as part of the AppFormer migration so I can keep the change-set as small as possible. If you feel strongly that globals and workbench enumerations don't belong in AppFormer I could open JIRA for this.

manstis:
@mbarkley "Globals" are a Drools feature so probably have little use in AppFormer. 

"Workbench" enumerations could have a use in AppFormer (as they provide a means to define lists for _other_ data-types; e.g. I could have an ```int``` field and define discrete values for it; and they provide the only means to have dependent enumerations - city/state for example).

So a JIRA for "Globals" would probably be of use. Enumerations.: It's your or @ederign call!

mbarkley:
@manstis Okay. Thanks for expanding on that. I will open a JIRA for migrating Globals.

In the mean time, it appears that I missed droolsjbpm-tools in this migration, so I have some more work to do here.

mbarkley:
Jenkins, please retest this.

mbarkley:
@manstis I've created a [JIRA issue](https://issues.jboss.org/browse/AF-612) for migrating the Globals.

mbarkley:
Jenkins, please retest this.

mbarkley:
Jenkins, please retest this.

mbarkley:
Jenkins, please retest this.

PR #1135, opened 2017-09-06
JBPM-6399 - Execution Server - Process config is not aligned with dep…

mswiderski:
…loyment descriptor

depends on https://github.com/kiegroup/guvnor/pull/509 and complete explanation why this is needed in that PR too.


mswiderski:
added tests for building process config.

mswiderski:
Jenkins retest this

PR #1139, opened 2017-09-07
JBPM-6389 : Upload file in task form is not working for eap7

nmirasch:


nmirasch:
@jsoltes, @pefernan fix cherrypicked to master. Could you take a look? Thanks!

PR #1147, opened 2017-09-08
JBPM-6354, GUVNOR-3457, GUVNOR-3176

paulovmr:
JBPM-6354: Cryptic error message when trying to create a project with name that already exists in Maven repo
GUVNOR-3457: [Library] edge case when filtering projects by name leads to javascript error
GUVNOR-3176: [Library] Filtering assets and projects is not unified


paulovmr:

PR #1156, opened 2017-09-12
JBPM-5745 Unclear May not be null validation error in the MultipleSubform.

alessioscalici:
**Duplicate** of https://github.com/kiegroup/kie-wb-common/pull/1157, pointing to master

The message will change as "Field is required".

@pefernan @jsoltes can you please review?


manstis:
@alessioscalici You really need to see these PRs through to completion!

PR #1164, opened 2017-09-15
GUVNOR-3479 : [GSS-RFE] Add support for missing keywords like "not matches" condition in Guided Rule

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3479

Looks like there were tests after all.


manstis:
Jenkins please retest this.

Rikkola:
There is at least one more test in drools-wb with the same root cause. And I was worried that there was no tests. Fixing ASAP.

PR #1166, opened 2017-09-15
Stunner - Fix README's wrong versions

romartin:
Hey @manstis ,
Just realized some old versions are specified in the README's. 
Thanks!


romartin:
oh so fast thanks! you're a like a bot! :P

PR #1171, opened 2017-09-19
[6.5.x] RHBPMS-4856: [GSS](6.4.z) Calendar is not visible on screen if guided decision table data grows BPM Suite 6.4

manstis:
See https://issues.jboss.org/browse/RHBPMS-4856

### DIFF
--- a/kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/main/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCell.java
+++ b/kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/main/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCell.java
@@ -17,12 +17,18 @@ package org.kie.workbench.common.widgets.decoratedgrid.client.widget.cells;
 
 import java.util.Date;
 
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.event.logical.shared.ValueChangeEvent;
 import com.google.gwt.event.logical.shared.ValueChangeHandler;
 import com.google.gwt.i18n.client.DateTimeFormat;
 import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
 import com.google.gwt.user.client.ui.PopupPanel.PositionCallback;
+import org.gwtbootstrap3.client.shared.event.HideEvent;
+import org.gwtbootstrap3.client.shared.event.HideHandler;
+import org.gwtbootstrap3.client.shared.event.ShowEvent;
+import org.gwtbootstrap3.client.shared.event.ShowHandler;
+import org.gwtbootstrap3.extras.datepicker.client.ui.base.constants.DatePickerPosition;
 import org.uberfire.ext.widgets.common.client.common.DatePicker;
 
 /**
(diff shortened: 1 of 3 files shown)


kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/main/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCell.java
```diff
@@ -33,47 +39,60 @@
...
+        this.datePicker = GWT.create(DatePicker.class);
+        datePicker.setFormat(format.getPattern());
+        datePicker.setPosition(DatePickerPosition.AUTO);
 
         // See https://issues.jboss.org/browse/GUVNOR-2322
-        // The DatePicker was being closed, before the ValueChangeHandler invoked, in response to the
-        // containing PopupPanel being automatically hidden when another Element received events.
-        datePicker.setContainer( vPanel );
-        panel.addAutoHidePartner( datePicker.getElement() );
+        // Register the DatePicker DOM components as "auto hide partners"
```
manstis: And here lays the crux of the fix.... When the date picker "popup/down selector" is shown it is added to the DOM as the first sibling of the ```TextBox``` widget; however the parent ```PopupPanel``` closes itself (hence removing the ``DatePicker```) if a mouse event occurs on DOM elements other than itself or registered "auto hide partners". Therefore I register the newly created element containing the "popup/down selector" when it is added to the DOM ("shown") as an "auto hide partner" (and remove on hide). Setting the ```DatePicker``` container to ```vPanel``` (now with hindsight) was a hack that did not work well with ```DatePickerPosition```.

kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/test/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCellTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private PopupDateEditCell cell;
+
+    private boolean autoHidePartnerAdded = false;
+
+    private boolean autoHidePartnerRemoved = false;
+
+    @Before
+    public void setup() {
+        cell = new PopupDateEditCell(DateTimeFormat.getFormat(DateTimeFormat.PredefinedFormat.DATE_LONG),
```
jomarko: I think this could be just `spy(new PopupDateEditCell(DateTimeFormat.getFormat(DateTimeFormat.PredefinedFormat.DATE_LONG), false));` without overriding `addDatePickerAsAutoHidePartner` and `removeDatePickerAsAutoHidePartner`. Then you could remove fields `autoHidePartnerAdded` and `autoHidePartnerRemoved` and last change would be change from `assertTrue(autoHidePartnerAdded);` to `verify(cell).addDatePickerAsAutoHidePartner()`. Similar for `autoHidePartnerAdded`. Or am I missing something?

manstis:
Hi @jomarko you had me worried! Your video shows it working perfectly:-) Compare your video to that in the JIRA. Your video shows the date picker popping "up" and not partially hidden. Compare that to the pop "down" and partial clipping in the JIRA video. Also I tried "spying" first but since the methods are called in the constructor they are not "spied" until after the constructor completes and the static "spy" method wraps the object.

jomarko:
@manstis thanks for quick answer
- partially hidden, yeah, that was the difference, you are right, now I am not able to get picker partially hidden
- spying, yes, you are right, I had also this constructor troubles in past, there would be need to move initialization form constructor to some post construct method. This is for 6.5.x. lets keep this as it is.

PR #1185, opened 2017-09-27
[DO NOT MERGE] GUVNOR-2642: Display of Data Objects in guided rule tab is limited to…

markcoble:
… 15 items (cannot remove items beyond this limit)
@manstis this is just the temp PR we discussed.  Still not showing data on the first load.


manstis:
Closed as incomplete. JIRA set to "open" and reassigned to me.

PR #1188, opened 2017-09-28
AF-622: [Library] Advanced Import dialog doesn't show up due to JavaScript error

paulovmr:


jhrcek:

manstis:
@paulovmr I'd like to understand why this fixes the problem. The ```ProjectContextChangeEvent``` _was_ used to control the selected state (OU, repository, project, package) of the Workbench and _was_ used by the Project Explorer (and Package Selector). Therefore IMO removing the triggering of the event may be simply masking the issue (and possibly adding more problems to Project Explorer?)

jhrcek:
@paulovmr You might also want to check analysis of this issue I did in https://issues.jboss.org/browse/RHBRMS-2915 I thought the issue would be fixed by adding project context change event to the if branch as I mentined, not by removing it from somewhere else.

paulovmr:
@manstis @jhrcek Hi, so basically the call I removed should have been removed when that optimization was included. 

The idea is to keep the user on the context of the project that was selected lastly, until another project/team is selected. The call I removed was executed when the user went back to the Teams screen. In that step, the user could select the same team and the same project as before, so no need to change the context yet. If a different team is selected, the project context will be updated accordingly.

PR #1203, opened 2017-10-04
JBPM-5568 Cannot open Diagrams that contain "unsupported" Definitions 

tiagodolphine:
When loading a diagram that contains unsupported definitions on Stunner, than an error message is displayed.
It was created `DiagramClientErrorHandler ` to handle the error generating a message based on the `ClientTranslationService` using the bundle properties that contains the message template on `StunnerCoreConstants.properties`.

@romartin 
@hasys 
@manstis

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/exception/DefinitionNotFoundException.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/exception/DefinitionNotFoundException.java
@@ -14,17 +14,24 @@
  * limitations under the License.
  */
 
-package org.kie.workbench.common.stunner.core.service;
+package org.kie.workbench.common.stunner.core.definition.exception;
 
-import org.jboss.errai.bus.server.annotations.Remote;
-import org.kie.workbench.common.stunner.core.diagram.Diagram;
-import org.kie.workbench.common.stunner.core.diagram.Metadata;
-import org.kie.workbench.common.stunner.core.graph.Graph;
+import org.jboss.errai.common.client.api.annotations.Portable;
 
-/**
- * Default Stunner's diagram service implementation.
- */
-@Remote
-public interface DiagramService extends BaseDiagramService<Metadata, Diagram<Graph, Metadata>> {
-    void registryDiagrams();
+@Portable
+public class DefinitionNotFoundException extends RuntimeException {
+
+    private String definitionId;
+
+    public DefinitionNotFoundException() {
+    }
+
+    public DefinitionNotFoundException(String message, String definitionId) {
+        super(message);
+        this.definitionId = definitionId;
+    }
+
+    public String getDefinitionId() {
+        return definitionId;
+    }
 }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/error/ClientErrorHandler.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/error/ClientErrorHandler.java
@@ -14,17 +14,21 @@
  * limitations under the License.
  */
 
-package org.kie.workbench.common.stunner.core.service;
+package org.kie.workbench.common.stunner.core.client.error;
 
-import org.jboss.errai.bus.server.annotations.Remote;
-import org.kie.workbench.common.stunner.core.diagram.Diagram;
-import org.kie.workbench.common.stunner.core.diagram.Metadata;
-import org.kie.workbench.common.stunner.core.graph.Graph;
+import java.util.function.Consumer;
 
-/**
- * Default Stunner's diagram service implementation.
- */
-@Remote
-public interface DiagramService extends BaseDiagramService<Metadata, Diagram<Graph, Metadata>> {
-    void registryDiagrams();
+import org.kie.workbench.common.stunner.core.client.i18n.ClientTranslationService;
+import org.kie.workbench.common.stunner.core.client.service.ClientRuntimeError;
+
+public abstract class ClientErrorHandler {
+
+    protected ClientTranslationService translationService;
+
+    public ClientErrorHandler(ClientTranslationService translationService) {
+        this.translationService = translationService;
+    }
+
+    public abstract void handleError(ClientRuntimeError error, Consumer<String> showError);
+
 }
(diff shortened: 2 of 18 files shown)


hasys:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/exception/DefinitionNotFoundException.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.kie.workbench.common.stunner.core.definition.exception;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+@Portable
+public class DefinitionNotFoundException extends RuntimeException {
+
+    private String definitionId;
```
romartin: I would encourage immutability as much as possible, and this field seems a good candidate for it, for being `final`, isn't it? :)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/exception/DefinitionNotFoundException.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public DefinitionNotFoundException() {
+    }
+
+    public DefinitionNotFoundException(String message, String definitionId) {
```
romartin: In case you finally use `final` fields, remember to use `@MapsTo` annotation in the constructors, to let errai know how to build the instances.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/error/ClientErrorHandler.java
```diff
@@ -0,0 +1,34 @@
...
+package org.kie.workbench.common.stunner.core.client.error;
+
+import java.util.function.Consumer;
+
+import org.kie.workbench.common.stunner.core.client.i18n.ClientTranslationService;
+import org.kie.workbench.common.stunner.core.client.service.ClientRuntimeError;
+
+public abstract class ClientErrorHandler {
+
+    protected ClientTranslationService translationService;
```
romartin: Same as for `DefinitionNotFoundException`, I would vote for immutability here too...

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/service/DiagramService.java
```diff
@@ -26,5 +26,5 @@
  */
 @Remote
 public interface DiagramService extends BaseDiagramService<Metadata, Diagram<Graph, Metadata>> {
-
+    void registryDiagrams();
```
romartin: Not really sure if makes sense adding the `registryDiagrams` method here...  IMO:

-  Looking at the AP there exist a `DiagramService` and a `DiagramRegistry`, so the idea is that both are "decoupled" from each other, so the behavior about how/when to register (caching) diagrams is given by top level components or by the service/registry implementations. 
- IMO adding this method creates additional coupling, as an operation about the service's internal registry instance is being made public to use.

As far as I remember the actual `DiagramService` implementation populates its internal registry once the managed bean instance is being created (probably on the `@PostConstruct`), why was this causing issues? why cannot remove this `registryDiagrams` and keep using that "on-demand load" approach? Probably is not possible or doesn't makes sense what I'm saying sorry... just trying to see if we can keep service and registry as less coupled as possible. WDYT??

Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/exception/DefinitionNotFoundException.java
tiagodolphine: Agreed, I'll do that.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/error/ClientErrorHandler.java
tiagodolphine: Agreed, I'll do that.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/service/DiagramService.java
tiagodolphine: Hey @romartin  the problem is exactly  that on the `@PostConstruct` implies that the instance is not being created by the CDI, because the marshaling process that throws the `DefinitionNotFoundException` is thrown on `diagrams.forEach(diagram -> getRegistry().register(diagram));`. So, the Exception is not gotten on the proper way on the client side to display the error message, in fact we got a `RuntimeException` not the `DefinitionNotFoundException`.
On the implementation I've decided to move the logic of registering all diagrams out of the `@PostConstruct` because it seems to make sense do perform that operation only when it was required, in this case, on the `lookup` method on the `DiagramLookupServiceImpl`, but I agree that we are coupling the registry process on the `DiagramService` that is not good, may be we can perform that operation on other service...  maybe a DiagramRegistryService ? any more ideas ?
Well, in summary, if the exception is thrown on the instantiation of the class, we could not get it on the proper way on the client side, that was what happened when testing it.

Thanks !

tiagodolphine: @romartin I will close this PR and later open a new one with the complete solution with the registering  refactory with the metadata-only marshaling as we discussed today.
Thanks

tiagodolphine:
@romartin I will close this PR and later open a new one with the complete solution with the registering refactory with the metadata-only marshaling as we discussed today.
Thanks

PR #1216, opened 2017-10-05
JBPM-5543: Changed inline editor widget to support new lines

Christopher-Chianelli:


romartin:

romartin:
@manstis assigned to you because I remember you did something about the text too... so just to double check it works for you. Thanks!

manstis:
Roger, I've not ran this code so I'm not entirely sure what's changed but
my contributions were to render Lienzo Text with line breaks whereas this
PR seems to use a TextArea so think Lienzo has no part to play... but
correct me if I'm wrong.

Christopher-Chianelli:
@manstis This PR is to fix the issue where line breaks were removed if you activated the inline name editor by double-clicking the name. The issue occurs since a Text Box automatically removes all new lines (since text boxes don't support new lines). Changing it to a Text Area preserves the new lines. (So this PR fixes inputing/saving text with new lines)

Christopher-Chianelli:
@hasys Pushed a fix to allow the inline editor to receive new lines via enter and updated the properties panel to allow new lines. Unfortunately, due to class encapsulation in lienzo, I am unable to create a TextWrapper that both wraps to a bounding box and respects new lines. (So we could either have text wrap or new lines, not both, until next lienzo release). How do you want enter to behave on the inline widget? Save changes on enter if no modifier keys are pressed, add a new line otherwise?

romartin:
Hey @Christopher-Chianelli ,

Rather than removing comments in code, have tested it. Actually the new lines are working, but two things seem broken:

1. Text does not auto-wrap now. We just added this a few weeks ago, but it seems disabled, is it? I understand the text should auto-wrap for white-spaces but keep new-lines, if possible. I would say both text wrapping and new-line characters support are compatible.

2. As now the canvas input is a `TextArea`, it seems it's capturing the `enter` key press event, so the keyboard shortcut for saving the text and closing the input is no longer working, as it was before. Probably it's as easy as using another shortcut for it, so instead of only the `enter` key... using  `shift/control + enter`? WDYT? @manstis @hasys  any ideas?

Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/TextEditorBoxImpl.java
```diff
@@ -115,9 +115,9 @@ private void processKey(final int keyCode,
                             final String value) {
         this.value = value;
         // Enter key produces save.
-        if (13 == keyCode) {
-            onSave();
```
romartin: at which point is the call to `onSave` done now?

Christopher-Chianelli:
@romartin See my above comment to @hasys. Basically, lienzo provides an interface, `ITextWrapper`, to control wrapping. Lienzo provides three implementations of it: `TextNoWrap` (no wrapping), `TextBoundsWrap` (wrap on bounds, new lines are ignored) and `TextLineBreakWrap` (wrap on new lines, ignore bounds). Ideally, we want a combination of `TextBoundsWrap` and `TextLineBreakWrap`. Issue is, you cannot create another implementation of the interface due to how it is defined: one of it methods (`drawString`) uses `Text.DrawString` as a parameter, which is package private (default), so you cannot import the interface to use it. The things that were commented out were for the `TextBoundsWrap`, which would need to be re-added once we have a wrapper that wraps on both new lines and bounds. `setTextBoundaries` does indeed do nothing since it currently uses `TextLineBreakWrap`, which doesn't support setting wrap boundaries. The  commented out`if (13 == keycode)` statement is responsible for the save shortcut, which I removed so users can enter new lines without needing to copy and paste. I agree that enter without modifier keys pressed should be save; what modifier keys should be used to cause it to input a new line?

romartin:
Hey @Christopher-Chianelli ,

Well IMO I think we should refactor/fix whatever you think necessary to make this work, I mean text wrapping is a mandatory feature we cannot remove at this point, so we need to make the text wrapping work with new lines as well. I'll ask for a new lienzo release soon, probably this week, so if you do changes on lienzo this week we can try to do the right solution once new release is uploaded. WDYT??

About the key to use.... mmm tbh not sure hehe, probably the best option is to ask UX guys about it, anyway I'm happy for example by using `Shift/Ctrl+Enter` for new lines, but just my opinion.

Christopher-Chianelli:
Depends on https://github.com/ahome-it/lienzo-core/pull/195

romartin:
Hey @Christopher-Chianelli ,

Finally found a fix for the issue about text wrapping and the resulting wrong shape's bounding box. It works fine for me but would prefer to double check with you as well.

I think the root cause of the issues was that you always rely on updating the text wrap boundaries by using the shape's bounding box, but on the other hand the shape itself includes the text instance as well, so that's why you was removing/adding child at some points too. So I changed the approach and moved to update the text wrap boundaries but given a certain size, this size is well known from the resize event handlers and from the `DecoratedShapeView#setSize` method, so I'm just calling to rebuild the boundaries from that locations.  Removed as well some other calls to this method that I think were not necessary.

Please take it a quick look at lemme know what you think, if you think is good, I'll provide it in my next PR np. Here is the commit:
https://github.com/romartin/kie-wb-common/commit/adb0595c028ce544657728191d4ce691b16d6268

Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/TextEditorBoxView.java
```diff
@@ -24,10 +25,15 @@
 public interface TextEditorBoxView extends UberElement<TextEditorBoxView.Presenter> {
 
     interface Presenter extends TextEditorBox<AbstractCanvasHandler, Element> {
+        final int CTRL = 1 << 0;
+        final int SHIFT = 1 << 1;
+        final int ALT = 1 << 2;
+        final int META = 1 << 3;
```
jomarko: These final integers seems to be unused. Please remove them if that is the case.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/TextEditorBoxImplTest.java
```diff
@@ -196,19 +224,29 @@ protected void verifyNameNotSaved() {
...
         verify(closeCallback).execute();
     }
+    
+    private Event getEventFromKeyCode(int keyCode) {
+        Event e = mock(Event.class);
+        when(e.getKeyCode()).thenReturn(keyCode);
+        when(e.getCtrlKey()).thenReturn(false);
+        when(e.getShiftKey()).thenReturn(false);
+        when(e.getAltKey()).thenReturn(false);
+        when(e.getMetaKey()).thenReturn(false);
```
jomarko: Is here really needed to mock behavior for `getAltKey()` and `getMetaKey()` ?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/TextEditorBoxImpl.java
```diff
@@ -111,11 +112,11 @@ public void onClose() {
...
     }
 
-    private void processKey(final int keyCode,
+    private void processKey(final Event event,
                             final String value) {
         this.value = value;
-        // Enter key produces save.
-        if (13 == keyCode) {
+        //Enter key w/o shift/ctrl produces save.
+        if (13 == event.getKeyCode() && !(event.getCtrlKey() || event.getShiftKey())) {
```
jomarko: I miss a test that would assert save operation is not done if CTRL or SHIFT were pressed during Enter key press. I think it could be easy to add such test next to `testSaveWithNewlinesByPressingEnter`.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/TextEditorBoxImplTest.java
Christopher-Chianelli: Due to the current implementation, no, but if the implementation was changed to use `getAltKey()` and `getMetaKey()`, their return value would be unknown (I think currently Mockito defaults to returning false for boolean, but I prefer for it to be independent from Mockito implementation). Thus, I rather have their return value known rather than unknown in case the implementation change (so the test is independent from the implementation).

Christopher-Chianelli:
@jomarko I have added your suggested changes. @romartin That commit looks good! The `updateTextBoundaries()` call in refresh is for when `setSize()` is called programmatically if I recall correctly, but you seemed to move the call there which also works. I think the one in `updateFillGradiant()` was put there as a precaution. If text wrapping is working as expected, then I am good with the commit!

hasys:

hasys:

romartin:
hey @hasys , see you're asking for jenkins build - as far as I know this depends on the next new lienzo-core `2.0.293-RELEASE`, so the build will keep failing at this point.. will let you know once we can test it! thanks anyway

romartin:
Closing as this job is being done here https://github.com/kiegroup/kie-wb-common/pull/1549

PR #1218, opened 2017-10-06
JBPM-6419 - When opening more than one Diagram the Palette is not rendered on all of them

tiagodolphine:
Now if more than one diagram is opened the palette is displayed on all of them.
The problem was basically the `AbstractPaletteDefinitionFactory.newPalette` was returning always the same instance instead of creating a new instance on each method call, in this way there was only one instance of the palette. Now, every method call `AbstractPaletteDefinitionFactory.newPalette` a new instance is returned.
This issue was affecting **BPMN** and **DMN**.

**Important**: there was a bug inserted on the master branch on this line https://github.com/kiegroup/kie-wb-common/blob/2b2ff376755db5d918fbc7e032c160e5be91f34d/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidget.java#L258 probably on this commit https://github.com/kiegroup/kie-wb-common/commit/cdf85d73afb77946d87d20f273a37cc795c6b4c5, the main problem is that the form is trying to be validated without any element/diagram previwous selected, on the project showcase this not happens because the `ProjectDiagramPropertiesScreen` is loaded after the diagram is selected, and on the standalone showcase and dmn the `SessionPropertiesScreen` is loaded before any selection, and then try to validate. To solve this I've inserted a check on the `DynamicFormRenderer` to check whether it is initialized AND valid, avoiding the NullPointerException, this solves the bug returning false in case it is not initialized.

@romartin 
@manstis 
@hasys

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/components/palette/factory/DMNPaletteDefinitionFactory.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/components/palette/factory/DMNPaletteDefinitionFactory.java
@@ -18,21 +18,22 @@ package org.kie.workbench.common.dmn.client.components.palette.factory;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
+import org.jboss.errai.ioc.client.api.ManagedInstance;
 import org.kie.workbench.common.dmn.api.DMNDefinitionSet;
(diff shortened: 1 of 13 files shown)


romartin:

romartin:
@hasys pleeease give a look at this one, let's see if we can get it merged and cherry-picked to `7.3.x` for next Mon EOF... thanks!!!

PR #1236, opened 2017-10-12
[RHBPMS-4972] avoid double logout

psiroky:
session.logout() already happens as part of logout.jsp


psiroky:

PR #1276, opened 2017-11-02
JBPM-5645: Stunner - Add Throwing Intermediate Signal Event

wmedvede:


wmedvede:
early pushed by mistake

PR #1282, opened 2017-11-08
AF-822: Fix null in error message when uploading non-existing file

tomasdavidorg:
https://github.com/AppFormer/uberfire/pull/901

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-default-editor/kie-wb-common-default-editor-client/src/main/java/org/kie/workbench/common/screens/defaulteditor/client/editor/NewFileUploader.java
+++ b/kie-wb-common-screens/kie-wb-common-default-editor/kie-wb-common-default-editor-client/src/main/java/org/kie/workbench/common/screens/defaulteditor/client/editor/NewFileUploader.java
@@ -55,11 +55,11 @@ public class NewFileUploader
     }
 
     @Inject
-    public NewFileUploader( final PlaceManager placeManager,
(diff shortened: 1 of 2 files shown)


manstis:
@tomasdavidorg Approved, but I'd be interested in understanding why [this](https://github.com/AppFormer/uberfire/blob/f8f3737a854dc64f1075b076514c8630bf3da5f2/uberfire-extensions/uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/editors/defaulteditor/DefaultEditorFileUploadBase.java#L90) is not called when a file is uploaded and the validation performed in the super-class.

jomarko:

jomarko:
@manstis I had a look and I am not sure what is the problem with the place you pointed. The method `isValid` seems to me as initializing the `FileUpload` widget, that seems to me appropriate for handling empty / not empty file name. Maybe the problem is that `FileUpload` combine both Presenter and View principles?

manstis:
@jomarko Hi.. my issue was that ```DefaultEditorFileUploadBase.isValid()``` performs the ```null```/empty check on the filename and reports that a file has not been selected. This validation happens when the ```FORM``` is submitted. However this is not called until after ```NewFileUploader``` tries to extract the file extension from a ```null``` filename - which caused the NPE. So.... the fix seems OK now on reflection.

PR #1285, opened 2017-11-09
[AF-894] Delete old cluster infrastructure

ederign:


ederign:
Depends on https://github.com/AppFormer/uberfire/pull/899

PR #1317, opened 2017-12-12
[AF-913]: Added new Project View and Assets List

adrielparedes:
Project View and Assets View UX changed to improve navigation and redability.

Issues related:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/686
https://github.com/kiegroup/appformer/pull/116
https://github.com/kiegroup/kie-wb-common/pull/1317
https://github.com/kiegroup/drools-wb/pull/727
https://github.com/kiegroup/jbpm-designer/pull/732
https://github.com/kiegroup/jbpm-wb/pull/966
https://github.com/kiegroup/kie-wb-distributions/pull/664


adrielparedes:
jenkins retest this

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @EventHandler("prev-page")
+    public void prevPage(final ClickEvent event) {
+        this.presenter.prevPage();
+    }
+
+    @EventHandler("filter-text")
+    public void filterTextChange(final KeyUpEvent event) {
+        presenter.setFilter(filterText.value);
```
manstis: Yikes.. so for **every** key up event we're updating the filter and calling ```getProjectAssets(..)``` which makes a network round trip, invokes an index search, returns some assets, updates the view.. There have been other places where we collect the typed text and only invoke the foregoing process every 250ms (or so). See the use of ```Timer``` [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java).

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.java
```diff
@@ -0,0 +1,212 @@
...
+    @Override
+    public void setCurrentPage(int currentPage) {
+        this.currentPage.value = String.valueOf(currentPage);
+    }
+
+    @Override
+    public void setPageIndicator(int from,
+                                 int to,
+                                 int total) {
+        this.pageIndicator.textContent = from + "-" + to + " of " + total;
```
manstis: i18n of " of "

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetScreen.java
```diff
@@ -0,0 +1,187 @@
...
+
+    private void update() {
+        this.view.clear();
+        List<NewResourceHandler> filteredHandlers = this.filterAndSortHandlers(this.newResourceHandlers,
+                                                                               this.filter,
+                                                                               this.getCategory(filterType));
+        filteredHandlers.forEach(handler -> addAssetItem(handler));
+    }
+
+    private Category getCategory(String filterType) {
```
manstis: I'm sure this is duplicated code!? Can it not be moved to the enum itself (assuming we keep the enum!).

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetScreen.java
```diff
@@ -0,0 +1,187 @@
...
+        try {
+            return Category.valueOf(filterType.toUpperCase());
+        } catch (Exception e) {
+            return null;
```
manstis: Would returning some default (IDK, like ```Category.UNDEFINED```) be better?

Is, for example, ```filterAndSortHandlers(...)``` null-safe? (indeed ALL uses of ```Category``` in this whole PR?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetView.java
```diff
@@ -0,0 +1,73 @@
...
+        this.assetsList.appendChild(view);
+    }
+
+    @Override
+    public void clear() {
+        this.assetsList.textContent = "";
+    }
+
+    @Override
+    public void setCategories(SelectComponent categories) {
```
manstis: Same comment as before (re: ```SelectComponent```).

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetView.java
```diff
@@ -0,0 +1,73 @@
...
+        this.filterType.appendChild(categories.getView().getElement());
+    }
+
+    @EventHandler("filter-text")
+    public void onFilterText(KeyUpEvent keyUpEvent) {
+        this.presenter.setFilter(this.filterText.value);
```
manstis: Same comment as before (re: possible use of a ```Timer```)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectScreen.java
```diff
@@ -0,0 +1,259 @@
...
+        this.view.init(this);
+        this.buildExecutor.init(this.view);
+        this.view.setTitle(libraryPlaces.getProjectInfo().getProject().getProjectName());
+        this.view.setContributorsCount(this.contributorsListScreen.getContributorsCount());
+        this.resolveAssetsCount();
+        this.showAssets();
+        newFileUploader.acceptContext(new Callback<Boolean, Void>() {
+            @Override
+            public void onFailure(Void reason) {
+
```
manstis: You probably want to disable some view components in the event of a failure!?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectView.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Override
+    public void enableImportButton() {
+        isImportEnabled = true;
+    }
+
+    @Override
+    public void init(ProjectScreen presenter) {
+        this.presenter = presenter;
+        if (!this.presenter.canEditContributors()) {
```
manstis: Personally I'd like to see these moved to the presenter (as a prior comment) when ```view.init( this )``` is called.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/delete/DeleteProjectPopUpScreen.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public void delete() {
+        final String confirmedName = view.getConfirmedName();
+        if (!project.getProjectName().equals(confirmedName)) {
+            view.showError(view.getWrongConfirmedNameValidationMessage());
+            return;
+        }
+
+        view.showBusyIndicator(view.getDeletingMessage());
+        projectService.call(v -> {
```
manstis: ```ErrorCallback``` to close "busy indicator."

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpScreen.java
```diff
@@ -0,0 +1,64 @@
...
+        view.init(this);
+    }
+
+    public void show(final ProjectInfo projectInfo) {
+        this.projectInfo = projectInfo;
+        view.buildModal(projectInfo);
+        view.show();
+    }
+
+    public void fork() {
```
manstis: We should not be committing incomplete work.. or at least hide the buttons; so Users don't start asking why it does not work.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/common/select/SelectComponent.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+
+import org.jboss.errai.ioc.client.api.ManagedInstance;
+import org.uberfire.client.mvp.UberElemental;
+
+public class SelectComponent {
```
manstis: My only comment is that IF these "select" related classes represent a generic "list box" then these classes should better live in [AppFormer](https://github.com/kiegroup/appformer/tree/master/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/dropdown) (or somewhere similar) and not in the ```kie-wb-common``` "Library" module. I think we have (at least) 3 "list box" implementations kicking around now (@dgutierr @mbarkley @ederign @Rikkola)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/test/java/org/kie/workbench/common/screens/library/client/screens/assets/AssetsScreenTest.java
```diff
@@ -0,0 +1,82 @@
+package org.kie.workbench.common.screens.library.client.screens.assets;
```
manstis: License please.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-api/src/main/java/org/kie/workbench/common/stunner/cm/resource/CaseManagementDefinitionSetResourceType.java
```diff
@@ -47,6 +48,11 @@ public int getPriority() {
         return 0;
     }
 
+    @Override
+    public Category getCategory() {
+        return Category.MODEL;
```
manstis: I'd have thought "Case Management" was more akin to ```PROCESS``` than ```MODEL```

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-project-client/src/main/java/org/kie/workbench/common/stunner/cm/project/client/type/CaseManagementDiagramResourceType.java
```diff
@@ -39,6 +40,11 @@ public CaseManagementDiagramResourceType(final CaseManagementDefinitionSetResour
         super(definitionSetResourceType);
     }
 
+    @Override
+    public Category getCategory() {
+        return Category.MODEL;
```
manstis: I'd have thought "Case Management" was more akin to ```PROCESS``` than ```MODEL```

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.java
paulovmr: I'll go further and say that in this case (network round trip needed) we should have a button to perform the search.

manstis: IDK what UX's design is for the "Search" boxes... but I agree (however would we also need to make all searches within Library consistent?)

paulovmr: The UX design is this:
image

I haven't included the button on the project list because it filters only at the client. But I agree that it is best to keep the UI consistent. @adrielparedes I can "copy" your impl to the project list screen in another PR if you think it is best.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsScreen.java
```diff
@@ -0,0 +1,529 @@
...
+
+        return new ProjectAssetsQuery(libraryPlaces.getProjectInfo().getProject(),
+                                      filter,
+                                      startIndex,
+                                      amount,
+                                      suffixes);
+    }
+
+    protected List<String> getSuffixes(Category category) {
+        return this.syncBeanUtils.getResourceTypeDefinitions(category)
```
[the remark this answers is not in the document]
manstis: @jomarko ```SyncBeanUtils.getResourceTypeDefinitions( category )``` does some extra processing - it defensively checks if a ```ResourceTypeDefinition``` has a category.. but ```SyncBeanUtils``` _could_ use ```@Inject Instance<ResourceTypeDefinition>``` instead of ```SyncBeanManager``` - and arguably would be better using that; as we'd not need to worry about a possible memory leak **if** a ```ResourceTypeDefinition``` is not ```@ApplicationScoped``` (at the moment ```SyncBeanUtils``` instantiates instances of ```ResourceTypeDefinition``` but does not dispose of them. Using ```Instance<ResourceTypeDefinition>``` would eliminate the need to manually maintain their life-cycle.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/common/select/SelectComponent.java
mbarkley: +1 to this. I'd suggest putting them somewhere in uberfire-widgets.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
```
paulovmr: I don't think this is needed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.java
```diff
@@ -0,0 +1,212 @@
...
+        String pageNumber = currentPage.value;
+        if (pageNumber.matches("\\d+")) {
+            presenter.setCurrentPage(Integer.valueOf(pageNumber));
+        }
+    }
+
+    @Override
+    public void showEmptyState(EmptyState emptyState) {
+
+        indexingInfo.className = "blank-slate-pf";
```
paulovmr: Probably better if we move these CSS properties to the CSS/LESS file and just handle (add/remove) CSS classes here.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetView.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
```
paulovmr: I don't think this is needed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetView.html
```diff
@@ -0,0 +1,55 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<div class="ac library">
+    <div class="page-content-kie">
+        <div class="main-container col-md-12 col-lg-12">
+            <!--<div class="populated-library">-->
```
paulovmr: Could we remove this line?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetView.html
```diff
@@ -0,0 +1,55 @@
...
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <div class="container-cards-pf">
+                <div class="row row-cards-pf" data-field="assets-list">
+                </div>
+            </div>
+            <!--</div>-->
```
paulovmr: Could we remove this line?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectView.java
```diff
@@ -0,0 +1,271 @@
...
+            this.editContributors.hidden = true;
+        }
+        if (!this.presenter.userCanUpdateProject()) {
+            this.addAsset.hidden = true;
+            this.importAsset.hidden = true;
+        }
+        if (!this.presenter.userCanBuildProject()) {
+            this.build.disabled = true;
+            this.deploy.disabled = true;
```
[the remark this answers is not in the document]
paulovmr: Personally I think it would be best to hide things when the user doesn't have permission and disable them when they have permission but the action can't be made in the current state.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpView.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
paulovmr: I don't think this line is needed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/rename/RenameProjectPopUpView.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
paulovmr: I don't think this line is needed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
```diff
@@ -666,6 +667,14 @@ public void goToAsset(final ProjectInfo projectInfo,
         }
     }
 
+    public void goToAddAsset(final ProjectInfo projectInfo) {
+
+        final PlaceRequest addAssetScreen = new DefaultPlaceRequest(LibraryPlaces.ADD_ASSET_SCREEN);
+        final PartDefinitionImpl part = new PartDefinitionImpl(addAssetScreen);
+        placeManager.goTo(part,
```
paulovmr: Probably missing a `part.setSelectable(false);` to hide the screen header and keep the UI consistent.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/common/select/SelectOptionView.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
paulovmr: I don't think this line is needed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/project/NewAssetHandlerCardView.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
```
paulovmr: I don't think this line is needed.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/test/java/org/kie/workbench/common/screens/library/client/screens/ProjectMetricsScreenTest.java
```diff
@@ -34,6 +34,7 @@
 import org.kie.workbench.common.screens.contributors.model.ContributorsDataSets;
 import org.kie.workbench.common.screens.library.api.ProjectInfo;
 import org.kie.workbench.common.screens.library.client.events.ProjectMetricsEvent;
+import org.kie.workbench.common.screens.library.client.screens.project.ProjectMetricsScreen;
```
paulovmr: Better to move the test to the same package as the class it tests.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/add/AddAssetView.java
adrielparedes: @manstis I agree with the comment before, but I think here is ok because the search is not being done in the backend but in the view. it is just a small filter in a collection of assets. But in this situation it never goes to the backend to refresh the list.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/delete/DeleteProjectPopUpScreen.java
adrielparedes: @manstis I don't get this. the `projectService` call has as second parameter a `HasBusyIndicatorDefaultErrorCallback`, is that what you were talking about?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpScreen.java
adrielparedes: @manstis this button is hidden, people can't see it in the UI. I can remove it, but this is going to be done in the next weeks.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-project-client/src/main/java/org/kie/workbench/common/stunner/cm/project/client/type/CaseManagementDiagramResourceType.java
adrielparedes: Sent you an email with information about how I make this decisions, we can continue discussion here :)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/FindAllLibraryAssetsQuery.java
```diff
@@ -77,13 +78,15 @@ public void validateTerms(final Set<ValueIndexTerm> queryTerms)
                                      NAME,
                                      new String[]{
```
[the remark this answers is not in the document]
adrielparedes: @jomarko actually that class test has the same behaviour: https://github.com/adrielparedes/kie-wb-common/blob/ef41890fa0a559bd05371f5095d2ccc4c4087330/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/test/java/org/kie/workbench/common/screens/impl/ListAssetsTest.java#L107

May be I'm missing something.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsScreen.java
```diff
@@ -0,0 +1,529 @@
...
+
+    protected int totalPages(int elements,
+                             int size) {
+        return Integer.valueOf((int) Math.ceil(elements / size)) + 1;
+    }
+
+    protected void update() {
+        this.view.clear();
+        this.hideEmptyState();
+        int offset = getOffset();
```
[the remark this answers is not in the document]
adrielparedes: Agree :)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/delete/DeleteProjectPopUpScreen.java
manstis: @adrielparedes You're correct :-)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpScreen.java
manstis: @adrielparedes Please point me to where it is hidden!?! I cannot see.

"Next few weeks.." can often slip/extend and *if* this work is incomplete by the time we do 7.6.0.Final *and* you later forget to remove it we'd have a release with a button that does nothing and at best confuse Users.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-project-client/src/main/java/org/kie/workbench/common/stunner/cm/project/client/type/CaseManagementDiagramResourceType.java
manstis: @adrielparedes Thanks for the email.. All OK with me (looks like it was voted on!)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpScreen.java
adrielparedes: @manstis is hidden in the html, but I can hide it into code too. But I think you are right, I don't want to have a button that does nothing, I will remove it

manstis: @adrielparedes Ah, OK, I see.. I had a closer look and there is no HTML for a button for "fork" (nor a ```@DataField``` in the ```View``` implementation).. All good now. Thanks for putting up with me!

adrielparedes:
jenkins retest this

adrielparedes:
@jomarko I fixed thing you commented. The scenario where morgates is imported twice after a delete is a know bug with Library that is already being fixed. The others scenarios where fixed. Setting is intended to be like that because it is going to change very very soon (work is already started)

paulovmr:
Hi @adrielparedes , I've tested the war with your modifications. Here are my comments:

1) "Add Asset" screen: the card subtitle should be the category of the asset type (eg.: Decision, Model, etc); also the filter by category is not working.
2) The "Add project" button has a little CSS problem (observed on kie-wb-webapp), the two buttons (Add Project and the down arrow) are a little separated (on both empty and populated library).
3) When listing projects, the project card count is wrongly positioned, should be on the right.
4) When going back from an empty project to the list of projects (through the breadcrumbs) I get an error popup: `Unable to complete your request. The following exception occurred: Parameter named 'project' should be not null!.`.

Also, I get an error popup when trying to add a project: `Unable to complete your request. The following exception occurred: org.apache.maven.project.ProjectBuildingException: 1 problem was encountered while building the effective model for com.myteam:fdfdfd:1.0.0 [FATAL] Non-resolvable parent POM for com.myteam:fdfdfd:1.0.0: Failure to transfer com.myteam:aa:pom:1.0.0 from http://localhost:8080/kie-wb-7.6.0-SNAPSHOT-wildfly10/maven2/ was cached in the local repository, resolution will not be reattempted until the update interval of guvnor-m2-repo has elapsed or updates are forced. Original error: Could not transfer artifact com.myteam:aa:pom:1.0.0 from/to guvnor-m2-repo (http://localhost:8080/kie-wb-7.6.0-SNAPSHOT-wildfly10/maven2/): Unauthorized (401) @ line 5, column 11 for project com.myteam:fdfdfd:1.0.0.` (maybe this is a problem with my environment...).

adrielparedes:
hi! @paulovmr All issues you pointed me are resolved. I'm uploading a new kie-wb in case you need to check it again. The only thing I could't reproduce is your last comment. Thanks!

adrielparedes:
jenkins retest this

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsScreen.java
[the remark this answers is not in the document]
porcelli: @adrielparedes could you clarify about your option here?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpScreen.java
porcelli: +1 to remove any reference for something that we don't implement. could you please remove the "fork" reference?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/fork/ForkProjectPopUpView.java
```diff
@@ -0,0 +1,107 @@
...
+import org.jboss.errai.ui.shared.api.annotations.Templated;
+import org.kie.workbench.common.screens.library.api.ProjectInfo;
+import org.kie.workbench.common.screens.library.client.resources.i18n.LibraryConstants;
+import org.kie.workbench.common.screens.library.client.widgets.common.PopUpUtils;
+import org.uberfire.ext.editor.commons.client.file.popups.CommonModalBuilder;
+import org.uberfire.ext.widgets.common.client.common.BusyPopup;
+import org.uberfire.ext.widgets.common.client.common.popups.BaseModal;
+
+@Templated
+public class ForkProjectPopUpView implements ForkProjectPopUpScreen.View,
```
porcelli: No fork should be added to this PR

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/test/java/org/kie/workbench/common/screens/library/client/screens/ProjectMetricsScreenTest.java
porcelli: +1 for paulo comment

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-api/src/main/java/org/kie/workbench/common/stunner/cm/resource/CaseManagementDefinitionSetResourceType.java
porcelli: + for process

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/SyncBeanUtils.java
```diff
@@ -0,0 +1,81 @@
...
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Instance;
+import javax.inject.Inject;
+
+import org.kie.workbench.common.widgets.client.handlers.NewResourceHandler;
+import org.uberfire.workbench.type.Category;
+import org.uberfire.workbench.type.ResourceTypeDefinition;
+
+@ApplicationScoped
+public class SyncBeanUtils {
```
porcelli: Tests?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-api/src/main/java/org/kie/workbench/common/screens/library/api/index/LibraryValueFileExtensionIndexTerm.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.kie.soup.commons.validation.PortablePreconditions;
+import org.kie.workbench.common.services.refactoring.model.index.terms.valueterms.ValueIndexTerm;
+
+@Portable
+public class LibraryValueFileExtensionIndexTerm extends LibraryFileExtensionIndexTerm implements ValueIndexTerm {
```
porcelli: Tests?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/common/PopUpUtils.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Arrays;
+
+import com.google.gwt.user.client.ui.Widget;
+import org.gwtbootstrap3.client.ui.Button;
+import org.gwtbootstrap3.client.ui.ModalFooter;
+import org.gwtbootstrap3.client.ui.constants.ButtonType;
+import org.uberfire.ext.widgets.common.client.common.popups.footers.GenericModalFooter;
+import org.uberfire.mvp.Command;
+
+public class PopUpUtils {
```
porcelli: Test?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsScreen.java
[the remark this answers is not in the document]
adrielparedes: @porcelli I've changed this part like @manstis recommended, because I need to get every suffix from the category selected to the the search, because indexed information doesn't contain a "category" but a suffix. That's why this class exist.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/test/java/org/kie/workbench/common/screens/library/client/screens/ProjectMetricsScreenTest.java
adrielparedes: @porcelli this is already done, I moved it to project. They have the same package

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/widgets/common/PopUpUtils.java
adrielparedes: @porcelli this is pure GWT, it is just an auxiliar to create those buttons because I didn't want to repeat code. Can this be tested?

adrielparedes:
jenkins retest this

adrielparedes:
jenkins full downstream build

adrielparedes:

adrielparedes:

adrielparedes:
@jomarko thanks! I will fix them after is merged!

ederign:
@manstis when you have a chance can you please take a look at this?

adrielparedes:
jenkins retest this

adrielparedes:
jenkins retest this

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsScreen.java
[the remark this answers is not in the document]
ederign: @adrielparedes  I wonder if others category will  appear here (due to manually instance creation)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/SyncBeanUtils.java
```diff
@@ -0,0 +1,115 @@
...
+
+import org.kie.workbench.common.widgets.client.handlers.NewResourceHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.uberfire.workbench.category.Category;
+import org.uberfire.workbench.category.Undefined;
+import org.uberfire.workbench.type.ResourceTypeDefinition;
+
+@ApplicationScoped
+public class SyncBeanUtils {
```
ederign: What is the reason for this class?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/SyncBeanUtils.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public Set<Category> getAvailableCategories() {
+        Comparator<Category> comparator = Comparator.comparing(Category::getName);
+        return this.getCategories().stream().filter(category -> {
+            List<ResourceTypeDefinition> resourceDefinition = this.getResourceTypeDefinitions(category);
+            return !resourceDefinition.isEmpty();
+        }).collect(Collectors.toCollection(() -> new TreeSet<>(comparator)));
+    }
+
+    public List<ResourceTypeDefinition> getResourceTypeDefinitions(Category category) {
```
ederign: As I told you before, I would use ActivityBeansCache for ResourceType lookup and for category lookup. 

This is important in order to keep your sanity and consistency.

We have dynamic resource types (like sample_resource_type.js) and this should be also considered. Another thing is to prevent the leak and work with dynamic plugins (a poc feature).

Looking at dynamic resource types code, this reminds me that the should probably support categories as well. Is up to you and @porcelli to decide if they will be fixed right now or later. Without this support, dynamic editor will be broken after this PR.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/SyncBeanUtils.java
ederign: I would strongly advise to remove this class and concentrate everything in ActivityBeansCache in order to centralize this for next endeavours (dynamic/modular workspaces)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/SyncBeanUtils.java
```diff
@@ -0,0 +1,115 @@
...
+        return newResourceHandlers;
+    }
+
+    protected List<Category> getCategories() {
+        return this.categories;
+    }
+
+    public Category getCategory(String filterType) {
+        if (filterType == null) {
+            return new Undefined();
```
ederign: This is a CDI BEAN.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-project-client/src/main/java/org/kie/workbench/common/stunner/bpmn/project/client/type/BPMNDiagramResourceType.java
```diff
@@ -38,6 +40,11 @@ public BPMNDiagramResourceType(final BPMNDefinitionSetResourceType definitionSet
         super(definitionSetResourceType);
     }
 
+    @Override
+    public Category getCategory() {
+        return new Process();
```
ederign: CDI BEAN.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-project-client/src/main/java/org/kie/workbench/common/stunner/cm/project/client/type/CaseManagementDiagramResourceType.java
```diff
@@ -39,6 +41,11 @@ public CaseManagementDiagramResourceType(final CaseManagementDefinitionSetResour
         super(definitionSetResourceType);
     }
 
+    @Override
+    public Category getCategory() {
+        return new Model();
```
ederign: This is a CDI Bean.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/SyncBeanUtils.java
manstis: @ederign One word of caution using ```ActivityBeansCache``` to get meta-data about editors: https://github.com/kiegroup/appformer/blob/master/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/authz/EditorTreeProvider.java#L115 (there can be "difficult to find" problems - as I discovered! - if the editor instance is not destroyed after meta-data extracted...)

ederign: Maybe the best option for Adriel, is to take the scan for metadada to also extract all the categories in runtime

manstis: @ederign ```getAvailableCategories()``` returns **only** the categories that are _used_ by ```ResourceTypeDefinition``` and not all categories on the classpath. To be honest I don't have a problem with what he's done here (I just had a vested interest in not creating instances of a Editor to retrieve its metadata!) Happy to chat on IRC (after lunch) if it's easier; I could just have misinterpreted things.

adrielparedes:
jenkins retest this

jhrcek:

jomarko:
I saw DOMUtil in some review comment, so wanted to point to this class [Elemental2DomUtil](https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/AssetsView.java
```diff
@@ -0,0 +1,49 @@
...
+    HTMLDivElement content;
+
+    @Override
+    public void init(AssetsScreen presenter) {
+        this.presenter = presenter;
+    }
+
+    @Override
+    public void setContent(HTMLElement element) {
+        this.content.textContent = "";
```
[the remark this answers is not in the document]
adrielparedes: @jhrcek actually neither I. I saw that pattern a few time, that's why I did that. But I need that line to remove all children from that asset view.

adrielparedes:

adrielparedes:

adrielparedes:

adrielparedes:

manstis:
@adrielparedes Wozza.. does this PR really have 1,086 files changed!?

manstis:
@adrielparedes Having a quick scroll it looks like you've not rebased with ```master```?

adrielparedes:
@manstis yes yes, sorry my bad, I was trying to solve it

adrielparedes:

adrielparedes:

adrielparedes:
jenkins retest this

adrielparedes:

adrielparedes:
jenkins retest this

adrielparedes:

tiagobento:
@adrielparedes Settings screen tested approved :) Just so I don't say anything, the settings screen does look a bit off because of the gray nav-bar. 

As a suggestion, I'd remove the `<div class="col-xs-12">` surrounding the `<ul class="library nav-tabs">` (in both `LibraryView.html` and `ProjectView.html`). 

Then, I'd add a new CSS class to the `<ul>` (maybe `nav-tabs-library` is a good name) with the following style:

```css
.nav-tabs.nav-tabs-library {
	background-color: white;
	padding: 0 20px 0 20px;
}
```

Before:
screen shot 2018-02-21 at 18 15 33

After:
screen shot 2018-02-21 at 18 16 06


Thanks again for taking the time to merge it into your PR!

jhrcek:
@tiagobento @adrielparedes today I went thoroughly though the changes introduced in this PR in the UR and first of all, good job, I like the improved consistency of the UI as well as the visual side.

Here are some issues I discovered. Even though the list seems long, I'm impressed about how few issues I was able to find :-) Please let me know which of these issues you agree with (just numbers are enough), which you'd like to fix in this PR and which I should file in separate jiras.

1. In Project's "General Settings" there's an URL field
- whatever I enter into this field stays present only in the editor while i'm in LibraryPerspective. But when I leave the perspective and return back the URL is not there. Looking at projects's git history in .niogit I don't see any commit when I edit URL and change it which confirms my suspicion that it does nothing.
- also there's some button on the right of the URL field, but it doesn't have any description, and does nothing (no network activity in the browser dev tools tab after I click it) - what is it for?

1generalsettings_urlfield

2. General Settings > "Disable GAV conflict check"
- MINOR This setting can also be changed globally in PreferencesCentralPerspective, where it has nice  tooltip explaining what it does. It would be nice to have the same tooltip in Project specific setting screen, because the meaning of this checkbox is rather unclear from the label itself

2disablegavcheckingmissingtooltip

- MINOR The "Double negation" makes the meaning of this checkbox less intuitive than it could be - how about inverting the semantics of the checkbox and using label like "Enable GAV conflick check"? It would be more intuitive like checked == enabled, unchecked == disabled (whereas now it's checked ==disabled, unchecked == enabled)
- Enhancement - This setting is tightly coupled to the list of repositories in "Validation" section. Wouldn't it make more sense to have it in the "Validation" section? As it basically says "when validating gav I'm looking into these repositories"



3. "Allow child GAV edition" - has no effect. In previous implementation when you had a project which had parent there were these 2 sections in cases when you opened a project which had a parent declared. Now there seems to be no way to see/edit parent GAV. (@Rikkola do we still support projects with pom inheritance / parent child relationships?).

3notpossibletoeditparentgav



4. Inconsistent highlighting of "Currently selected section" in Settings.
- When you enter Library (afresh without doing anything before), open project and click its Settings tab, no section is highlighted, even though General settings section is opened.
- When you click one of the section it gets higlighted with a blue stripe
-> Expected would be to ALWAYS have a higlight for the currently selected section (as in the 2nd half of screenshot)
4nothinghighlighted



5. Settings > Dependencies - there is a backend bug when uploading some artifacts (I tested this with a few popular packages from maven central https://mvnrepository.com/popular)
- When I add dependency with GAV dom4j:dom4j:1.6.1 to the project (artifact available in maven central) there's a useles error displayed obviously propagated from backend.
Taking advantage of the new [sourcemaps debugging in kie-wb](https://github.com/kiegroup/kie-wb-distributions/pull/678) which makes sourcemaps available in kie-wb wars for each full downstream pull request I was able to track down the root cause

5backendbugwhenuploadingdom4j




6. In Dependencies there are 2 radio buttons "Package white list". No matter what I do, "All" is always selected and I'm unable to select "NONE" for some artifacts, which is very confusing. Why is this possible for some artifacts, but not for others?

6whitelistallalwaysselected



7. Knowledge Bases when you add a knowledge base and within it you add Knowledge session, the headers of the table are not aligned with the contents below
7headersnotaligned

tiagobento:
@jhrcek Thanks for looking into this.

1)
    - @adrielparedes is already looking into this issue.
    - This button should probably be removed. @adrielparedes can you remove along with your URL fixes?


2. - Good catch. I'll add the tooltip.
   - I agree with the double negation but maybe people are already used to having this as it is, so changing might be even more confusing.
	- I think I'd rather keep it in sync with the prototype. See https://rawgit.com/srambach/BxMS-frontend/setting-tabbed/spaces-settings.html


3. - I could really use @Rikkola's help on this one :)


4. - I'll fix that.


5. - I'm guessing it can have something to do with the merge.. @adrielparedes @paulovmr @ederign do you have any idea what could be causing this?


6. - That's because the dependency `a:jansproj:2.0.0` has no packages I'm assuming, so `All` and `None` have the same logical meaning.


7. - Yeah.. so the thing is this 'table' is not a regular PatternFly component nor a stable widget on the Workbench. @srambach and I came to the conclusion that it would be best to leave it unalligned than to have it with fixed widths.

ederign:
@tiagobento number 5 looks like a project oriented issue. Could you please file a jira with [Project Oriented] on the name?

tiagobento:
@ederign I'm working alongside with @adrielparedes to solve that already.

Rikkola:
@jhrcek @tiagobento Ok so about 3.

At the moment projects with multiple modules are broken. Not by this PR they are a project orientated bug. This is since we removed the asset-mgmt features and are adding in a replacement. And when I say we are adding in a replacement. I mean after all the currently known Project Orientated bugs are fixed.

adrielparedes:

adrielparedes:

adrielparedes:
jenkins retest this

adrielparedes:
jenkins retest this

adrielparedes:

Rikkola:
@jhrcek Sure, there is no Jira. The feature will be called "Multimodule support", but maybe it is worth making a bug Jira about the broken parent support. Just to be safe.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.java
```diff
@@ -0,0 +1,223 @@
...
+    private TranslationService ts;
+
+    @Inject
+    private SelectComponent selectComponent;
+
+    private PopulatedAssetsScreen presenter;
+    private final Elemental2DomUtil domUtil;
+
+    public PopulatedAssetsView() {
+        domUtil = new Elemental2DomUtil();
```
mbarkley: Why not inject this?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/delete/DeleteProjectPopUpView.java
```diff
@@ -0,0 +1,159 @@
...
+    HTMLLabelElement confirmedNameLabel;
+
+    @Inject
+    @DataField("confirmed-name")
+    HTMLInputElement confirmedName;
+    private Elemental2DomUtil domUtil;
+
+    @Override
+    public void init(final DeleteProjectPopUpScreen presenter) {
+        domUtil = new Elemental2DomUtil();
```
mbarkley: Why not inject this?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/assets/PopulatedAssetsView.java
adrielparedes: you are right, fixing

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/delete/DeleteProjectPopUpView.java
adrielparedes: actually here is never used, removing

PR #1330, opened 2017-12-21
JBPM-6629: Stunner - Wrong location when adding new nodes from toolbox

alepintus:
Hi @romartin @hasys ,
fixed a bug on the canvas positioning algorithm when adding new nodes from toolbox.

Could you please take a look?

Thanks!


romartin:
ok to test

romartin:

romartin:
Hey @alepintus , thanks for this fix! I've tested it and works much better than before. 
Found a minor issue - when if the parent node is "full", it adds the node "outside" the parent but it's added as a child as well, which causes a bit of confusion as the shape is outside the parent but if you drag the parent you realize the shape is also moved.. so it's a child.
Anyway I'll merge as it works much better than before and would like to have this fix present before the xmas shutdown. Let's try to improve it a bit once we're back ;)
Thanks!!

PR #1336, opened 2018-01-05
JBPM-6730: Stunner - BPMN marshallers re-implementation

evacchi:
This is a **work in progress**. I'm opening this PR for getting early feedback.

Most work happens under 
`kie-wb-common/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend` 
involved packages are mostly
- `org.kie.workbench.common.stunner.bpmn.backend`
- `org.kie.workbench.common.stunner.bpmn.backend.converters`

I have defined a new `BPMNDirectDiagramMarshaller` (for lack of a better name) which skips the JSON marshalling process and goes straight to the new data model, e.g.
- `org.kie.workbench.common.stunner.bpmn.backend.converters.ProcessConverter`
- `org.kie.workbench.common.stunner.bpmn.backend.converters.FlowElementConverter`

I have added a few utilities (`o.k.w.c.s.b.b.c.Match`, `o.k.w.c.s.b.b.c.TypedFactoryManager`) to minimize casts/better readability

Any feedback/suggestions is much appreciated at this stage.


evacchi:
@romartin @tsurdilo please chime in for review :-)

tsurdilo:
@evacchi is there a Jira for these changes that describes in detail what you are trying to fix? I looked at the code don't fully understand what the benefits/updates are. Is the concern for the JSON generation performance related? If it's not then just seems weird to me that if we do spend time on the bpmn2 generation in Stunner, it's not related to rewriting the marshallers but to further "patch" the code that was already taken from jbpm-designer as-is and make it even more confusing? Please don't take this as a rant but just as a real honest question I have :)

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/Misc.java
```diff
@@ -0,0 +1,117 @@
...
+import java.net.URLEncoder;
+import java.util.List;
+import java.util.Map;
+
+public class Misc {
+    private static final Logger _logger = LoggerFactory.getLogger(Misc.class);
+
+    public static void writeAssociations(Process process,
+                                   String elementId,
+                                   JsonGenerator generator) throws IOException {
```
tsurdilo: if goal is to remove json why are we still writing it? Don't understand

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/Misc.java
```diff
@@ -0,0 +1,117 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.List;
+import java.util.Map;
+
+public class Misc {
```
tsurdilo: why is important bpmn2 marshalling operations in a class called "misc". I think if you are extracting utility methods from marshallers thats fine but maybe name them appropriately

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/Misc.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        try {
+            return URLEncoder.encode(s,
+                    "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            return s;
+        }
+    }
+
+    public static DiagramElement findDiagramElement(
```
tsurdilo: this does nothing

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/Properties.java
```diff
@@ -0,0 +1,204 @@
...
+import org.jboss.drools.ImportType;
+import org.jboss.drools.MetaDataType;
+import org.kie.workbench.common.stunner.bpmn.backend.legacy.util.Utils;
+
+import java.io.IOException;
+import java.util.*;
+
+import static org.kie.workbench.common.stunner.bpmn.backend.legacy.Bpmn2JsonPropertyIds.*;
+
+public class Properties {
```
tsurdilo: I think this will be super hard to maintain. Every time we add a new property or extension element it will be hard imo to maintain all the places it has to be checked. Also this still includes the parsing logic that we did to put in the JSON but my understanding here is that we are trying to get rid of JSON?

tsurdilo:
@evacchi added some review comments as much as I could not knowing the code that much. Hope it helps. Overall my feel is that tackling the actual marshallers to not write json at all but the graph information instead would probably be better long-term thing to do. Thanks.

evacchi:
@tsurdilo thanks for your feedback.

w.r.t. the Jira, I'm pretty sure there must be one, but I'd ask @romartin to chime in.

In short -- quoting a mail from Roger, -- right now we have that an XML BPMN2 file is translated into a Stunner BPMN model by going through the following process:

    | XML | <--(1)--> | Eclipse BPMN2 | <--(2)--> | Oryx model (JSON) | <--(3)--> | Stunner BPMN model |

the JSON conversion is lossy and not needed any more (AFAIU it was used *before* Stunner was introduced). The goal of this PR is to drop the JSON conversion and going straight from Eclipse BPMN2 to Stunner BPMN via model transformation.

As for Misc and Properties, you are correct in pointing out that that code should be dropped; in fact that's actually part of the original  `org.kie.workbench.common.stunner.bpmn.backend.legacy.Bpmn2JsonMarshaller`.
The reason those are there, is that my first attempt at tackling the problem (and learning about the code) was refactoring that class.
They should be removed in the final version of this patch, and to be fair, I can probably delete those already.

tsurdilo:
@evacchi thanks for the explanation!

romartin:
Hey @evacchi @tsurdilo 
Thanks for all the comments! Yeah just created the ticket [JBPM-6730](https://issues.jboss.org/browse/JBPM-6730) for this.
@tsurdilo yep, as you know the idea at some point was to re-implement the actual BPMN marshallers and avoid the "intermediate" use of Oryx in which Stunner relies on... so this is just the task @evacchi  has started to work on.  It's really important your collaboration here @tsurdilo , you have infinite knowledge on what we support, we do not support and what we are customizing, so yeah please, whatever feedback you can provide can safe us lot of timee :)
@evacchi BTW review in progress ;)

romartin:
Hey @evacchi @tsurdilo 

Had been looking at this. Yep that's the right direction to follow. Not much to say at code level because it sounds good :) 

Anyway previously to start coding probably I think it worth considering the different solutions we can take in order to address this task and their implications. I mean, rather than the implementation itself, now we're assuming that: 
* We'll keep (and improve) the actual Stunner BPMN domain model for one side
* On the other side, for the marshalling stuff, we'll rely on the Eclipse BPMN model (Eclipse marshallers as well)
* Finally we'll rely on some kind of model transformation to achieve process marhsalling

In fact that's the solution that fits better for our needs, but we can take another alternatives, and so each one with some implications on the further development task:
* First important thing is to ask ourselves if the Eclipse model is really the best option we can take. Which are the alternatives? Rather than the domain model API, do Eclipse marshallers provide the excepted support for our custom needs? 
* About model conversion/transofrm -> Instead of spending time on do the model transforms, we can consider extending the actual Stunner model, either by inheriting/implementing the Eclipse model objects in Stunner or even either by trying to reuse the Eclipse domain model as it is in Stunner (by adding metadata on top of it), although that sounds not feasible for our customizations..
* If we're going to use third party libraries for this (as expected), do they provide client & server impls? Can we add missing client support on top if so? Remember that the idea is to do as much as possible shared stuff between client & server, and just create enterprise beans for each of the components which implementation differs from both sides, this way can be injected where the side specific code is required

Well brainstorming with you guys and giving ideas, just before start coding like geeks :-P

@evacchi now just a concrete comment about this commit - I see you crated the `ProcessConverter` etc, that's fine! But probably we can do this paert a bit more generic and accessible from other marshallers like the DMN or CM ones too (not just BPNM)..   mean WDYT about creating some kind on ModelTransformer/Converter interface type, which the base Stunner marhsallers could handle, and this way create a set of enterprise "tranformer" beans , which can be handled and injected via the IOC/CDI container.  This way we can (probalby) provide shared tranformer implementations at this point, being able this way to perform the marshalling on pure client side (although not sure if it's completely possible...). Another option that Mark said was about relying on XSLT, you think makes sense here? Although I understand we'll need custom Java code at some point for sure...

@evacchi Well I think the best we can do is to have some talk tomorrow and get started with this... will ping you. Thanks for this!! :)

Thanks guys!!

tsurdilo:
@romartin regarding "if the Eclipse model is really the best option we can take." - I don't know of any better alternative (excepting writing your own object model which I think would in the end look pretty much same as whats being used now. 

One idea that you might could look at if there is time is using/reusing the eclipse bpmn2 modeler parsers code as that modeler supports the entire bpmn2 spec transformation (still uses the eclipse bpmn2 api).

evacchi:
> But probably we can do this part a bit more generic and accessible from other marshallers like the DMN or CM ones too (not just BPNM).. I mean WDYT about creating some kind on ModelTransformer/Converter interface type, which the base Stunner marhsallers could handle, and this way create a set of enterprise "tranformer" beans , which can be handled and injected via the IOC/CDI container.

I can certainly see the emerging pattern; I mean, at the end of the day a converter *is* just a function A -> B, i.e., it could implement a `java.util.function.Function[A,B]`.

Can you give an example of where IOC/CDI would come in handy here?

romartin:
@tsurdilo hehe yeah I don't know other BPMN domain models in Java at this point either... so no worries, asking is free :P About the Eclipse parsers yeah that's the idea AFAIK, @evacchi can confirm, but yep this is one of the big upsides of this refactor, trying to delegate to other parsers as much as possible, keeping our codebase for this lightweight and easy to maintain... what I asked @evacchi to confirm that  in case we require to apply additional parsing logic on top of Eclipse parsers (for whatever customization our engine does), we need to be sure that that parsers provide extension points so we can extend if necessary. I hope so but better looking before start coding ... Anyway thanks for your feedback again! :+1: 

@evacchi Yeah well in fact the converter can be considered just a Function yep, but by creating an interface for it and making it visible to the IoC container, we are defining a bean type in our enterprise context, this way we can use CDI in a way like `@Inject Instance<T> instance` to produce our converter beans. Well we already had been discussing this in a call before... but does it makes sense for you?

romartin:
@evacchi As commented before too, I would try to build the converter implementations in a shared package, but before being able to do that, we should check if we can make the Eclipse BPMN domain model visible on client side as well. This is not for now, I mean there is no need at this point to export the Eclipse model to client side, neither the shared converters in fact, but it I think it worth taking a look at the Eclipse beans and probably even doing a quick test trying to export that model to client side (via the `errai.properties`), this way we can have an idea of where and how to implement the converters... ping me if you have Qs, I can help :)

evacchi:
> I would try to build the converter implementations in a shared package, but before being able to do that, we should check if we can make the Eclipse BPMN domain model visible on client side as well. 

so, the problem here is that the Eclipse model makes extensive use of the EMF framework; e.g., all root objects are EObjects. I don't think it would be easy to do away with this dependency, so one might be tempted to see if ecore has been ported to GWT; but now the question may be: do we really need to bring in such a heavy dependency on the client side? OTOH if at some point we really need to bring this to the client side, we might want to generate our own, which, as @tsurdilo noted, would probably look pretty much the same

> Yeah well in fact the converter can be considered just a Function yep, but by creating an interface for it (eg: Converter) and making it visible to the IoC container, we are defining a bean type in our enterprise context, this way we can use CDI in a way like @Inject Instance<Converter> instances to produce our converter beans

so I'll have to think a bit about this one; I'm trying to think of a use case for which bringing in CDI now will give us a concrete benefit *at this time*. Anyway, we can always factor this in at a later stage, if it makes sense :)

romartin:
Hey @evacchi 
- oh good point , the Eclipse model was tight to EMF... it sounds difficult to achieve the client side support for now yep. No worries, I just wanted to have this clear that's fine!!
- About the converter type and CDI well yes, another option is just to make the ConvertersFactory an enterprise bean, it can internally use CDI or just the `new` operation to build the different instances, it makes sense for me too. But agree with you, that's just a minor thing at this point we can do a quick refactor for this later on, np!
Thanks!

romartin:
@evacchi Just another comment trying to convince you for the use of some kind of `Converter` bean types hehe.. So IMO one of the upsides for creating the base interface type `Converter` and making it visible to the IoC container, it's possible to easily append new converters on third party modules (f.i. some customer that needs some extensions)  by just implementing the interface Converter and adding some bean scope annotation (like `@Dependant`, because the injection points you created will resolve it too. It's useful too for example if some bug appears on a concrete converter, as from a third party module you can just specialize the bean (`Specializes`).
Summarizing - IMO doing this help to keep our type decoupled and easily extensible, with no "cost" from our side, we just take advantage of the enterprise capabilities.
Well just brainstorming, feel free to add your thoughts, as i know you're a coding geek as well hehe , it's  always great to discuss and learn something new, right? :)

romartin:
Hey @evacchi , FYI - did some cleans on the domain model side. See https://github.com/kiegroup/kie-wb-common/pull/1362

evacchi:
I've noticed that the `org.eclipse.bpmn2` dependency is at version `0.7.6-jboss`, which implies it is kind of vendored. I believe it might be outdated. In particular, I have noticed that 
- `MessageEventDefinition#getEventRef` returns a Message
- `ErrorEventDefinition#getErrorRef` returns an Error

however:

- `SignalEventDefinition#getSignalRef` returns a String (the ID that is being ref'd) instead of a Signal!

At first I thought this might be intended, but the code at http://git.eclipse.org/c/bpmn2/org.eclipse.bpmn2.git/tree/org.eclipse.bpmn2/src/org/eclipse/bpmn2/SignalEventDefinition.java#n49 shows otherwise.

Can we upgrade to the latest version of this dependency, at least for the marshallers?

romartin:
Hey @evacchi , yep I'm agree on this... in that case, which are the other solutions we can take? We are not forced to check the new marshallers work by using this way... so I'm open to ideas :)

evacchi:
superseded by #1408

PR #1338, opened 2018-01-08
DROOLS-1959 [DMN Editor] Extend marshaller support for Expressions

tarilabs:
add support for FunctionDefinition

/cc @manstis @jomarko


manstis:
Jenkins please retest this.

tarilabs:
@jomarko thanks for reporting it. You are correct.

The problem happens only when the QName of the FunctionDefinition is created from scratch from the WB side. 

In turn, it lies in the management of the WB's QName. There is certainly some limited modification I can do to align and make it work also for this case, but I'm wondering more in general how the QName is expected to work.
QName per se is a qualified XML name, and normally contains a reference-name, plus some URI and optionally some prefix. I'm just wondering if we really expect the end user will fill-in all of these 3 attributes, or actually the WB's QName will be handled behind the scene by the workbench itself. Personally, I think is the latter, but I'm not too much involved in this kind of details.

Before I go and I try to unify the string representation of a QName as a String for these 3 attributes, or anything else, can we clarify this point please @manstis ?
Thanks

manstis:
@tarilabs Hi.. I know handling of (workbench) ```QName``` is in progress.. I'd just like to add at the moment all ```QNames``` instantiated in the UI have a value "My QName" which doesn't really match any specification ;-) If there's a better real value we can use please say!

tarilabs:
@jomarko latest rebase and commits should solve the problem you reported. thanks again.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/DMNModelInstrumentedBase.java
```diff
@@ -20,11 +20,17 @@
 
 import org.jboss.errai.common.client.api.annotations.NonPortable;
 import org.kie.workbench.common.dmn.api.definition.DMNDefinition;
+import org.kie.workbench.common.dmn.api.property.dmn.QName;
 import org.kie.workbench.common.stunner.core.definition.builder.Builder;
 
 public abstract class DMNModelInstrumentedBase implements DMNDefinition {
 
+    public static final String URI_FEEL = org.kie.dmn.model.v1_1.DMNModelInstrumentedBase.URI_FEEL;
```
manstis: @tarilabs Have you tried running the webapp with this change? (I will, so don't worry too much).

I fear, since (this) ```DMNModelInstrumentedBase``` can be used "client-side" (in JavaScript in the browser) that we will not be able to reference ```org.kie.dmn.model.v1_1.DMNModelInstrumentedBase``` and may need to duplicate the constants literally here instead.

tarilabs:
@manstis I taken the chance to `feel:string` as a default for WB's QName.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/DMNModelInstrumentedBase.java
tarilabs: @manstis but the API has visibility on `kie-dmn-model`, hence why I set it up here. If it's a problem we can dup the constant, I leave the call up to you. Thanks

tarilabs:
```
Test Result (1 failure / +1)org.kie.workbench.common.forms.editor.backend.indexing.FormDefinitionIndexerTest.testFormsIndexing
```

Jenkins, test again

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/FunctionDefinition.java
```diff
@@ -28,7 +28,7 @@
 @Portable
 public class FunctionDefinition extends Expression implements HasExpression {
 
-    public static final QName KIND_QNAME = new QName("http://www.drools.org/kie/dmn/1.1:kind");
+    public static final QName KIND_QNAME = new QName("{" + DMNModelInstrumentedBase.URI_KIE + "}kind");
```
jomarko: If I compare to previous version, it seems to me that "{" and "}" are added, is this intended please?

tarilabs: Yes.
This is to align with standard JDK behaviour of QName.toString() method.
ref: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/javax/xml/namespace/QName.java#418

manstis: Good find @tarilabs

jomarko: thanks

PR #1346, opened 2018-01-11
DROOLS-2232: Increase coverage of the Literal Expression model mapper

jomarko:
@manstis please have  look when you have a moment.


jomarko:
@manstis also ready for merge.

PR #1356, opened 2018-01-12
JBPM-5815 - Improve the keyboard integration support

Josephblt:
https://issues.jboss.org/browse/JBPM-5815

Stunner - Improve the keyboard integration support

@romartin 
@tiagodolphine


Josephblt:

porcelli:
ok to test

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
```diff
@@ -180,8 +256,19 @@ public void handle(MouseExitEvent event) {
...
         }
-        return getCommandManager().execute(canvasHandler,
-                                           command);
+
+        final CommandResult<CanvasViolation> result = getCommandManager().execute(canvasHandler,
+                                                                                  command);
+
+        if (!CommandUtils.isError(result) && shapeLocationsChangedEvent != null) {
+            Collection<String> uuids = new ArrayList<>();
+            for (final Element element : elements) {
```
tiagodolphine: @Josephblt in cases like this you can use a more functional code like 

`List<String> uuids = elements.stream().map(Element::getUUID).collect(Collectors.toList());`
 
I think it is cleaner and easier to understand, but just as a suggestion.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LienzoMultipleSelectionControlTest.java
```diff
@@ -225,6 +246,60 @@ public void testOnSelectEvent() {
...
+                                                             eq(100d),
+                                                             eq(100d),
+                                                             eq(overLayer),
+                                                             eq(10.0));
+
+        //Same canvas with selected items
+
+        when(shapeLocationsChangedEvent.getCanvasHandler()).thenReturn(canvasHandler);
+
+        ArrayList<String> selectedUUIds = new ArrayList<String>();
```
tiagodolphine: @Josephblt  you can use an interface on the declaration

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LienzoMultipleSelectionControlTest.java
```diff
@@ -225,6 +246,60 @@ public void testOnSelectEvent() {
...
     }
 
+    @Test
+    public void testOnShapeLocationsChanged() {
+        //Different canvas
+        when(shapeLocationsChangedEvent.getCanvasHandler()).thenReturn(mock(CanvasHandler.class));
+        tested.onShapeLocationsChanged(shapeLocationsChangedEvent);
+
+        verify(selectedItems, never()).rebuildBoundingBox();
+        verify(selectionManager, never()).drawSelectionShape(eq(0d),
```
tiagodolphine: @Josephblt just as a suggestion to use constants for the values 0d, 100d, ...

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/event/ShapeLocationsChangedEvent.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+package org.kie.workbench.common.stunner.core.client.canvas.event;
+
+import java.util.Collection;
+
+import org.kie.workbench.common.stunner.core.client.canvas.CanvasHandler;
+
+public class ShapeLocationsChangedEvent {
```
tiagodolphine: @Josephblt you can extend `AbstractCanvasHandlerEvent`

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/event/keyboard/KeyboardEvent.java
```diff
@@ -23,10 +23,10 @@
         CONTROL(17),
         SHIFT(16),
         DELETE(46),
-        ARROW_UP(24),
-        ARROW_DOWN(25),
-        ARROW_LEFT(27),
-        ARROW_RIGHT(26),
+        ARROW_UP(38),
```
tiagodolphine: @Josephblt just curious were the codes wrong?

Josephblt:
Yeah @jomarko. That's a bug will work to fix that.
@tiagodolphine Will try and make those adjustments too.

Josephblt:

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
```diff
@@ -77,6 +80,10 @@
         extends AbstractCanvasHandlerRegistrationControl<AbstractCanvasHandler>
         implements LocationControl<AbstractCanvasHandler, Element> {
 
+    private final double LARGE_DISTANCE = 25d;
```
tiagodolphine: @Josephblt it is just missing the `static` on the declaration.

jomarko:
@Josephblt please let me know knew PR is ready for retest.

Josephblt:
@tiagodolphine - fixed the missing static modifiers.
@jomarko - Yes. I'ts ready for testing.
Tahnk you guys

Josephblt:

Josephblt:

romartin:

romartin:
Hey @Josephblt , I see you pushed new commits here, but not sure if this is ready to test.. please lemme know. Thanks!!

Josephblt:
@romartin
@jomarko 
@tiagodolphine 

 Ready for tests and review;

Josephblt:

Josephblt:

romartin:
Hey @Josephblt , the build is failing due to some checkstyle violations, look at the console log... please fix those and re-fire a build when you have a chance, so we can consume the generated wars (actually not being generated). Thanks!

manstis:
@jomarko It looks to be caused by the ```Command``` stack exceeding 200 entries.

The stack is used for undo/redo operations, however I'd expect Stunner to handle the limit better. IDK keep only the last X entries and discard older ones so you can only undo the last X changes.

@romartin Do you have a JIRA for this already?

Josephblt:

tiagodolphine:
@manstis some time ago I was talking to @romartin and I think all stunner registries should have an eviction policy not only for commands maybe we could have common implementation (more robust) for all of them, I mean registry/caches. Then we can provide some approaches for eviction like FIFO, LRU, LFU... let's track that on a JIRA. 
WDYT @romartin @manstis ? 
but anyway, this is not related to this PR.

manstis:
@tiagobento Sounds good to me :-)

Josephblt:
@romartin All good now. Ready to test, thanks.

romartin:
@jomarko @manstis good catch guys, that registry implementation is really basic and this must be updated - just created a new ticket for it -> [JBPM-6776](https://issues.jboss.org/browse/JBPM-6776)
@Josephblt thanks! gonna take a look at this now :+1:

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
```diff
@@ -174,14 +255,23 @@ public void handle(MouseExitEvent event) {
...
+                                          parentConstrained);
                 builder.addCommand(c);
                 i++;
             }
             command = builder.build();
         }
-        return getCommandManager().execute(canvasHandler,
-                                           command);
+
+        final CommandResult<CanvasViolation> result = getCommandManager().execute(canvasHandler,
```
romartin: Actually when using the arrow keys you try to move a shape out of canvas or current parent's bounds, it does not allows you, that's ok, but on the other hand it appears an error popup on each arrow key pressed... which IMO makes no sense from the end user perspective. 
This is due you're trying to execute the command but it fails, so there exist a client component that listen for failed executions and displays an error popup.... In order to avoid this, let's split the command execution here by first calling the `allow` method, which will check if bounds are correct and no error will be thrown yet, and if the result is success, then let's call the `execute` method, as it is right now. This way no error popups will appear, as if the update position command is not allowed, you will never perform the execution. I have tested this locally and works.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/command/UpdateElementPositionCommand.java
```diff
@@ -28,18 +28,22 @@
 
     protected final Node<View<?>, Edge> element;
     protected final Point2D location;
+    protected final boolean parentConstrained;
 
     public UpdateElementPositionCommand(final Node<View<?>, Edge> element,
-                                        final Point2D location) {
+                                        final Point2D location,
+                                        final boolean parentConstrained) {
```
romartin: Well IMO there is no need of the `parentContrained` flag due to:
1. When updating some position is expected just passing the element and the new location as arguments, isn't it? So this flag seems internal command stuff, probably even better creating two different commands if necessary, one that takes the parent's node into account as well. 

2.- Anyway I think in our case is better always doing the parent check, no need for flag. We don't want shapes "outside" the parent bounds in any case, right? So better if by default our command implementation just does the check, isn't it?

Locally I have set the `parentConstrained` boolean to `true` always, and did some tests and it seems it works fine. So IMO I would just remove this flag from everywhere - from the command class and this constructor, from the command factory too, and the different refactored calls. So you can start by setting it always to `true` locally, as I did, and run your tests too.. if you think it's fine, yeah let's remove it. Anyway lemme know if that makes sense for you or any issues appear.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/keyboard/KeysMatcher.java
```diff
@@ -40,4 +40,19 @@ public static boolean doKeysMatch(final KeyboardEvent.Key[] actualKeys,
 
         return matches.size() == expectedKeys.length;
     }
+
+    public static boolean isKeyMatch(final KeyboardEvent.Key[] actualKeys,
+                                      final KeyboardEvent.Key... expectedKey) {
+        if (actualKeys == null) {
```
romartin: mmm once we moved to java 8 we can probably do this in a shorter and better way just using streams, although just a minor comment....

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/UpdateElementPositionCommand.java
```diff
@@ -49,25 +49,32 @@
...
+                                        final @MapsTo("previousLocation") Point2D previousLocation,
+                                        final @MapsTo("parentConstrained") boolean parentConstrained) {
         this.uuid = PortablePreconditions.checkNotNull("uuid",
                                                        uuid);
         this.location = PortablePreconditions.checkNotNull("location",
                                                            location);
         this.previousLocation = PortablePreconditions.checkNotNull("previousLocation",
                                                                    previousLocation);
+
+        this.parentConstrained = parentConstrained;
```
romartin: as commented before.. .I would try to remove this `parentConstrained` flag from everywhere.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/UpdateElementPositionCommand.java
```diff
@@ -109,16 +116,31 @@ public String getUuid() {
...
         final Element<? extends View<?>> element = getNodeNotNull(context);
         final Graph<DefinitionSet, Node> graph = (Graph<DefinitionSet, Node>) getGraph(context);
+
         final BoundsImpl newBounds = getTargetBounds(element);
+
         final GraphCommandResultBuilder result = new GraphCommandResultBuilder();
-        if (checkBoundsExceeded(graph,
-                                newBounds)) {
-            ((View) element.getContent()).setBounds(newBounds);
+        if (parentConstrained) {
```
romartin: well in general that if sentences are a bit confusing at first look, I see duplicated code in both if/else clauses... so probably this can be improved and make it more clear and easy to maintain. Anyway if removing the `parentConstrained` flag (as we can consider it `true` always), you'll simplify this...

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/GraphUtils.java
```diff
@@ -207,24 +191,39 @@ public static Point2D getPosition(final View element) {
...
         return new double[]{Math.abs(w), Math.abs(h)};
     }
 
+    @SuppressWarnings("unchecked")
+    public static Bounds getBounds(final Graph<DefinitionSet, ? extends Node> graph) {
+        return graph.getContent().getBounds();
+    }
+
+    public static boolean isRootNode(Element<? extends View<?>> element, final Graph<DefinitionSet, Node> graph) {
+        if (element instanceof Node) {
```
romartin: can be simplified in a single line

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/UpdateElementPositionCommand.java
```diff
@@ -115,24 +115,30 @@ public String getUuid() {
...
+                } else {
+                    result.addViolation(new BoundsExceededViolation(parentBounds)
+                                                .setUUID(element.getUUID()));
+                }
+            }
         } else {
-            result.addViolation(new BoundsExceededViolation(parentBounds)
-                                        .setUUID(element.getUUID()));
+
+            if (GraphUtils.checkBoundsExceeded(graph, newBounds)) {
```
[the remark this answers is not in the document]
Josephblt: Yeah this duplicated code will be deleted after the requested changes. As for the key improvement I think it's a great idea. If @romartin agrees we could try adding that.

romartin:
hey @Josephblt @tiagodolphine 
I also realized about holding arrow keys does not keeps moving the shapes.... so would be nice yeah if we can provide this, on the other hand not sure if it's just as easy... I mean by holding the key pressed lots of events will be received, so actually lots of commands will be executed as well, so both command registry (in session) and the the undo will be affected - the registry will grow fast and the undo will only execute one command undo at time, which not sure if makes sense from the user perspective.
So probably to achieve this in a good way we should schedule the keypress event handler and avoid it firing so many times... well just an idea.
Summarizing - that's not as  easy as it seems, so I will not complain for not having this for now if you think guys that will take long... we have lots of other priorities. But if you think the solution is feasible yeah, go ahead :)

jomarko:
Hi @romartin @tiagodolphine I think moving elements while the key pressed is nice idea, however I also think it is worth to address in completely separate jira. WDYT?

Josephblt:
As discussed with @romartin I opened 3 new Jiras regarding bugs and enhancements we found. Tthe The links for those tickets are as follows: 

Stunner - Child elements should not be larger than their parent container
https://issues.jboss.org/browse/JBPM-6790
Stunner - Canvas and Graph are desynchronizing after some operations.
https://issues.jboss.org/browse/JBPM-6791
Stunner - Improve Arrow Keys support
https://issues.jboss.org/browse/JBPM-6789

These first two are bugs and should not be considered for this review.
The third is for the continuous movement with arrow keys wich we will address at a more appropriate time. Thanks guys.

@romartin 
@tiagodolphine 
@jomarko

Josephblt:

Josephblt:

Josephblt:

Josephblt:

romartin:
Tested the whole kie-wb locally again and it works fine. Merging.

PR #1369, opened 2018-01-18
DROOLS-2155: Table editor should use full canvas area available

manstis:
See https://issues.jboss.org/browse/DROOLS-2155

@romartin This introduces some small changes to Stunner. A continuous "inheritance path" is required from the (Lienzo) ```Layer``` used to hold the DMN grids to the outer most Stunner ```SessionPresenterView``` that supports ```RequiresResize``` so the DMN grid layer can resize and adjust its virtual scrollbars. Therefore I've had to make a few ```FlowPanel``` into ```ResizableFlowPanel``` in Stunner.

I also added the ability for the ```SessionPresenterView``` to have set whether it's main ```DIV``` controls scrollbars on the content (as you have now, a large (Lienzo) ```Panel``` in a small ```DIV``` with ```overflow: auto``` or whether the content provides its own scrollbars (as is the case with the DMN grid layer).

I've tested with DMN standalone, Stunner standalone and ```kie-drools-wb```.. The "new Process designer" looks to work entirely as it used to before this PR.. so all should be good.

The changes in here will be of use when you move Stunner to use the "virtual scrollbar" stuff @karreiro did originally for Decision Tables, now re-used by the DMN grid editor and to be used for Stunner as we've previously discussed (I think there is a JIRA for it somewhere).

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/pom.xml
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/pom.xml
@@ -165,6 +165,11 @@
 
     <dependency>
       <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-workbench-client</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.uberfire</groupId>
       <artifactId>uberfire-workbench-client-views-patternfly</artifactId>
     </dependency>
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/general/BaseNavigateCommand.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/general/BaseNavigateCommand.java
@@ -18,6 +18,9 @@ package org.kie.workbench.common.dmn.client.commands.general;
 
 import java.util.Optional;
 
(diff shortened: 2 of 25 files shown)


manstis:
Jenkins please retest this.

jomarko:
screenshot from 2018-01-19 13-39-52
screenshot from 2018-01-19 13-39-20

manstis:
@jomarko Thanks for the review.

I've added an extra commit renaming use of ```TranslationService```

The other issue appears to be a bug with how Workbench "maximise" and "minimise" is interacting with the Workbench Panel. When you "maximise" the "Palette" reappears on the left edge that causes the "grid" to overflow - the vertical scroll bar is there; just not visible. "Minimising" the editor keeps the "Palette" visible. It should not be shown when the editor is in "grid" mode.

I'd prefer a separate JIRA for it, if you don't mind, as the problem may not be in the DMN Editor but AppFormer and it's implementation of "maximise" and "minimise"... but I will investigate (and probably be the one who fixes it anyway!)

manstis:
@jomarko It looks like Stunner is showing the "Palette" when the panel is maximised. I will work on a fix for the DMN Editor as it needs to behave differently. A JIRA in DROOLS for the DMN Editor would therefore be perfect :-)

jomarko:
https://issues.jboss.org/browse/DROOLS-2256 thanks in advance for looking into it.

PR #1373, opened 2018-01-19
DROOLS-2256: [DMN Designer] Scrollbars in full screen mode

manstis:
See https://issues.jboss.org/browse/DROOLS-2256

@tiagobento the changes you made for https://issues.jboss.org/browse/GUVNOR-3357 forces the Palette to show (for a ```@DiagramEditor``` widget) when a workbench Panel is maximised or minimised. This breaks the DMN Editor that has two modes of operation: (1) The normal "Stunner" graph grid display, (2) A sub-view when the User chooses to edit the detail of a node (the sub-view is still part of the ```@DiagramEditor``` "main" view). In the second scenario we hide the Palette as it is not needed; however maximising/minimising a Panel caused the Palette to reappear.

The change in this PR allows me to prevent this undesirable behaviour by only showing the Palette if it has not been hidden "on purpose" by the User editing the node detail.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/palette/BS3PaletteWidgetImplTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/palette/BS3PaletteWidgetImplTest.java
@@ -23,10 +23,13 @@ import org.kie.workbench.common.stunner.client.widgets.palette.categories.Defini
 import org.kie.workbench.common.stunner.client.widgets.palette.factory.BS3PaletteViewFactory;
 import org.kie.workbench.common.stunner.core.client.api.ShapeManager;
 import org.kie.workbench.common.stunner.core.client.components.glyph.ShapeGlyphDragHandler;
+import org.kie.workbench.common.stunner.core.client.event.screen.ScreenMaximizedEvent;
+import org.kie.workbench.common.stunner.core.client.event.screen.ScreenMinimizedEvent;
 import org.kie.workbench.common.stunner.core.client.service.ClientFactoryService;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 
+import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.verify;
 
 @RunWith(MockitoJUnitRunner.class)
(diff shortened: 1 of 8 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/palette/BS3PaletteWidgetImplTest.java
```diff
@@ -71,4 +74,39 @@ public void checkDestructionReleasesResources() {
...
+
+        verify(view).showEmptyView(false);
+    }
+
+    @Test
+    public void checkOnScreenMinimisedNotDiagramEditor() {
+        final ScreenMinimizedEvent event = new ScreenMinimizedEvent(false);
+        palette.onScreenMinimized(event);
+
+        verify(view).showEmptyView(false);
```
jomarko: Hi @manstis shouldn't be this like [1], as we are not minimizing the diagram screen, but the screen for properties of a decision node?

[1]
```java
verify(view).showEmptyView(true);
```

manstis: @jomarko When the (for example) "Properties panel" is minimised (restored to its original size/location) we _show_ the Palette; so I believe verifying the call to _not_ show an _empty_ view is correct (i.e. verify that we are _showing_ the Palette).

manstis: @jomarko Or, more succinctly, it does not matter what _type_ of screen _was_ shown; when we minimise the screen we always show the Palette. We only hide the Paletter _if_ it was not a "Diagram" editor.

manstis:
@jomarko Your proposed tidy-up does not work! 

The reason I use a _different_ ```PaletteVisibility``` implementation in ```BaseNavigationCommand``` is to be able to differentiate between the palette being hidden by the navigating to the Expression Editor vs hidden by, for example, maximising the "Properties Panel". I tried your code and if, in the "graph" editor you maximise the "Properties Panel" the Palette is correctly hidden. However if you then minimise it the Palette is not shown. This is because in ```DMNPaletteWidget``` we forgo calling ```setVisible(VISIBLE)``` **if** it is hidden.

manstis:
Jenkins please retest this.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/palette/BS3PaletteWidgetImplTest.java
jomarko: Sorry Michael, I am still little bit confused. I had look on `ScreenMinimizedEvent` constructor. It takes one `boolean` argument `diagramScreen`. That is why I thought:
- `new ScreenMinimizedEvent(true)` represents minimizing:
  - 
before-save

- `new ScreenMinimizedEvent(false)` represents minimizing:
  - 
screenshot from 2018-01-22 14-02-13

And if we minimize the socnd one, we don't want to show palette, or? I am probably still missing something.

PR #1381, opened 2018-01-23
JBPM-5912: Lists of basic data types are handled in the same way as if f they were not list.

pefernan:
Now the forms engine supports lists of basic types (Lists of values of basic types as supported on the Data Modeler). There are two new form widgets to handle them:
- LOVSelector: lets the user select values from an existing list
- LOVInput: lets the user create / edit a List of values

This PR isn't finished yet there are still some minor issues to solve on LOVInput:
- the add new value button refreshes the screen the first time it is pressed
- the table for List<Date> doesn't work quite well yet.

@jsoltes could you please start testing it? you can use the lists-example project on my playground (https://github.com/pefernan/jbpm-playground.git)


pefernan:

pefernan:

pefernan:

kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/test/java/org/kie/workbench/common/forms/dynamic/client/rendering/renderers/lov/creator/input/widget/LOVCreationComponentTest.java
```diff
@@ -0,0 +1,394 @@
...
+
+        Assertions.assertThat(component.getSelectedValues())
+                .isNotNull()
+                .isNotEmpty()
+                .contains(entry);
+    }
+
+    @Test
+    public void testPromoteThirdElement() {
+        testSelectSecondElement();
```
jsoltes: I believe here should be "testSelectThirdElement()" instead...

kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/test/java/org/kie/workbench/common/forms/dynamic/client/rendering/renderers/lov/creator/input/widget/LOVCreationComponentTest.java
```diff
@@ -0,0 +1,394 @@
...
+
+        TableEntry entry = component.getSelectedValues().get(0);
```
jsoltes: These same lines are repeated 7 times in this test - maybe they can be extracted to a helping method?

kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/test/java/org/kie/workbench/common/forms/dynamic/client/rendering/renderers/lov/creator/input/widget/impl/AbstractEditableColumnGeneratorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+        assertNotNull(column);
+        assertNotNull(column.getFieldUpdater());
+
+    }
+
+    @Test
+    public void testGetReadOnlyColumn() {
+
+        generator.registerColumn(pagedTable, callback,true);
```
jsoltes: here is also a lot of code repeating - also can be extracted to a method..

jsoltes:
@pefernan I also have some comments regarding the UI testing I performed. It looks like a lot, but most of them aren't criticals, just a way how to make the user's life better (I tried to put the most important first).

**Issues common for both widgets:**
1. When you try to start the CreationListsProcess or SelectListsProcess, you get an exception: "Unable to complete your request. The following exception occurred: Cannot convert between java.lang.String and java.util.ArrayList for property [listOfStrings] in java.util.HashMap."
2. Converters. I get it that the widget doesn't support all types of variable, but I think there should be some UI warning when adding a model field with unsupported binding (e.g. "This widget does not support BigDecimal values, the value will be converted to Double) and not change it without the knowledge of the user.
3. List of Bytes allows to put bigger value than Byte (also for Short), maybe this can be restricted programatically. If not I would provide the warning as in 1.
4. When you input too big Long, you get the warning: "The value should be an Integer" even though Long values aren't converted to Integer.
5. When you add some inputs/select some values and then remove and add a widget, it remembers its state. This also feels weird and it is also not consistent with the other widgets.
6. When you create some inputs/select something from list and then change any field property, the widget resets to default (you loose the values you previously selected/added). Feels unnatural.
7. Removing of the lines from the table feels kind of like hard work. If it was possible to select the lines just by clicking on them would make it easier. Or alternatively having the remove button activated always (when at least one line is present) and by clicking on it removing the last line.
8. When you add 6 lines to the table, a scroll appears on the right side with no actual use.

**MultipleInput issues:**
1. When you run clean kie-wb, import the list-example2 project, open "LIstCreationAll" form and click on + button in the first widget, the kie-wb restarts.
2. The input field in the table has fixed size, but could be stretched for the whole width of the cell (could be useful for bigger values).
3. It is not possible to get the insert mouse just by clicking into the input field (it always selects the whole field).

**MultipleSelect issues**
1. When changing the field binding to nothing and back, the values in the table disappear (for every type of binding except of String).
2. The widget is not wide enough for the dates - when selecting them, the layuts gets broken.
3. I believe the field "Elements visible on the dropdown" is not needed. This is simply not different than putting less fields into the table.
4. String values in the selector are always put to uppercase (user might want to have it in lowercase).
5. It would be better if the search filter was case sensitive.

kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/test/java/org/kie/workbench/common/forms/dynamic/client/rendering/renderers/lov/creator/input/widget/LOVCreationComponentTest.java
pefernan: well seen! you caught me on a copy&paste

pefernan:
@jsoltes working fixing this issues. Just some questions:

**Issues common for both widgets:**
1.- fixed there was a converters issue.
2.- the widgets support all data types, I was using errai converters because that's the strategy I've been following since now to keep things simple on the widget. I removed all converters for this widget (left a default one to avoid errai autocalculate the binding converter) to fix this issue and also 1.
3 & 4.- I'd like to keep user away of knowing the internals of the data types that are being used behind the form because they might be non technical. So I changed the error messages to look like "_Wrong value: the value should be a integer/decimal number between {0} and {1}_"
5 & 6.- Is that error happening on runtime or authoring?
7.- I cannot control the click on row to select sorry, but if you want to select a given row you can click the select checkbosx. I'm not sure if the remove-last would work here, user could press the remove button and remove elements from other pages without knowing... we might ask UX.
8.- need to see who puts that scroll.

**MultipleInput issues:**
1.- working on this
2&3 .- That's a problem on the GWT TextEditCell, I don't know if I'll be able to fix.

**MultipleSelect issues**
1.- I'll be taking a look but not sure how to fix
2.- Ok
3.- That's part of the LiveSearch Component and I think it has some use. Imagine a dropdown with a thousand options with this setting you can control the number of visible items and navigate throught them on the filter.
4.- Yes I capitalize the first letter, I will remove that.
5.- The LiveSearch Component is not Case Sensitive (https://silviomoreto.github.io/bootstrap-select/examples/#live-search_1) but I can add an extra setting to make it case sensitive.

jsoltes:
Hello @pefernan, just to answer your questions...

**Issues common for both widgets:**
5. & 6. The issue is in editor (remember, runtime was not working :)).
7. This is not as important, I am ok with the way it is.

**MultipleInput issues**
2.&3. Just if there is an easy fix, don't have very high priority.

**MultipleSelect issues**
1. This is definitely a bug. The problem is that the values in the lines disappear and the lines don't. Otherwise I would understand that removing the binding would make loose the state. Although, the preferred behavior would be keeping the state when you go back.
3. What you probably mean is "Elements visible on the dropdown title", what i meant is "Elements visible on the dropdown". Which can be simply defined by putting values in the table.
5. On the second thought you don't have to. It is standard behavior of many searches (even the browser one is case-insensitive).

For all the others, thank you for fixing them!

pefernan:
@jsoltes sent PR fixing all issues.

kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-backend/src/main/java/org/kie/workbench/common/forms/dynamic/backend/server/context/generation/dynamic/impl/fieldProcessors/time/AbstractDateMultipleieldValueProcessor.java
```diff
@@ -25,11 +25,11 @@
 import org.kie.workbench.common.forms.dynamic.service.context.generation.dynamic.FieldValueProcessor;
 import org.kie.workbench.common.forms.model.FieldDefinition;
 
-public abstract class AbstractDateLOVFieldValueProcessor<F extends FieldDefinition> implements FieldValueProcessor<F, List<?>, List<Date>> {
+public abstract class AbstractDateMultipleieldValueProcessor<F extends FieldDefinition> implements FieldValueProcessor<F, List<?>, List<Date>> {
```
jsoltes: Just a typo here > missing F...

jsoltes:
jenkins please execute full downstream build

jsoltes:
Good job, Pere! Looks like everything we agreed on is fixed except of 6. in common issues and 3. in MultipleSelect issues...

pefernan:
@jsoltes fixed tye typo.

As we agreed in the chat: we can leave the canges required for #6 to be done until we can use the Preview functionality on layout editor so we can have widgets disabled on the editor and use the preview to test the form.
For #3 let's keep it (for now if there are no complains from QE) since with that option we can use the filter as a query too.

pefernan:
jenkins please execute full downstream build

pefernan:
jenkins please execute full downstream build

pefernan:
jenkins retest this

pefernan:
jenkins retest this

PR #1362, opened 2018-01-16
JBPM-6748: Stunner - Improve domain model declaration by making some annotations not mandatory.

romartin:
Hey @manstis @jomarko @hasys 

As this week we're doing kinda workshop, I'd been showcasing the Stunner domain beans and did some cleans that were missing from time ago, this way you will see the code for beans and properties has been reduced.

This commit provides no fixes neither new stuff, it's just a code improvement, so nothing special to test here rather than it works as usually I guess... :) Added some missing tests too.

See [JBPM-6748](https://issues.jboss.org/browse/JBPM-6748)

It **depends on** 
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/672
- https://github.com/kiegroup/kie-wb-distributions/pull/670
- https://github.com/kiegroup/jbpm-wb/pull/971/

Thanks!

PR #1382, opened 2018-01-23
DROOLS-2260: [DMN Designer] Background Color change throws an Exception

manstis:
See https://issues.jboss.org/browse/DROOLS-2260

This PR is probably impacted by #1362 however I'm happy to keep this PR pending (merge of #1362) and fix accordingly (the mentioned PR removes ```defaultValue``` but I'll move to static literals instead).


manstis:
@jomarko This PR updated following merge of #1362

manstis:
@jomarko The font colour can be changed; but the text has a black border. If you increase the font size to something big, really big, you'll see the font has changed colour.. Also, as for changing the node name, it does appear a little "sensitive" to where you double click however it works OK - it appears updated in the "graph" however due to [this](https://issues.jboss.org/browse/DROOLS-1937) the wrong property in the Property Panel is updated. I wasn't sure what you meant by "can't be changed"?

PR #1385, opened 2018-01-24
DROOLS-2266: [DMN Designer] Override newSizeHandler for DMNSVGShapeDefImpl

manstis:
See https://issues.jboss.org/browse/DROOLS-2266


manstis:
@jomarko Hi, yes the "Properties Panel" will be included when this integrates into the workbench.

What you report is a Stunner issue; could you please raise a JIRA against that project?

PR #1386, opened 2018-01-25
DROOLS-2270: [DMN Editor] Replace 'Return to DRG' button with Link

manstis:
See https://issues.jboss.org/browse/DROOLS-2270

@jomarko Due to [this](https://issues.jboss.org/browse/DROOLS-1937) the name of the node shown in the "back link" in the UI is always "My name" however you can see from the Unit Test that it does use the value extracted from the ```HasName``` instance. I also stepped through the code in the browser and the ```HasName``` passed is identical to the node selected in the graph.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
@@ -23,6 +23,7 @@ import javax.inject.Inject;
 import com.ait.lienzo.client.core.types.Transform;
 import com.google.gwt.event.dom.client.ClickEvent;
 import org.jboss.errai.common.client.api.IsElement;
+import org.jboss.errai.common.client.dom.Anchor;
 import org.jboss.errai.common.client.dom.DOMUtil;
 import org.jboss.errai.common.client.dom.Div;
 import org.jboss.errai.common.client.dom.Document;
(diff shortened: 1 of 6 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
```diff
@@ -57,8 +59,8 @@
 
     private ExpressionEditorView.Presenter presenter;
 
-    @DataField("exitButton")
-    private Div exitButton;
+    @DataField("returnToDRG")
+    private Anchor returnToDRG;
```
jomarko: I see the `org.jboss.errai.common.client.dom.Anchor` is deprecated. I tried to replace with `elemental2.dom.HTMLAnchorElement;` what was quite straightforward. Just replace `anchor.setTextContent()` with `anchor.textContent =`, however this is not testable. SO please consider this Deprecated class but easily testable vs. Non deprecated class but not so easily testable. Let me know your preference.

manstis: Yes, I have "move to Elemental2" on my TODO list: https://issues.jboss.org/browse/DROOLS-1963

jomarko: You don't stop surprising me. Thanks.

manstis:
@jomarko Test updated following your recommendation.

PR #1390, opened 2018-01-29
fix equals, hashcode in core

evacchi:
This is a preliminary PR after checking with @hasys that equals is not implemented in a few core classes.
Open issues:
- NodeImpl should probably be compared against their nodes as well, but the `Node` interface returns a List (e.g. `getInEdges`). I think it might be better to return a Set, otherwise it might be more cumbersome to deep equal those. e.g.: `forall e in that.getInEdges() this.inEdges.contains(e)` would be quadratic
- I'm not sure about GraphImpl. It compares against hashes, I guess they are expected to equal if the nodes are equal, but are they? At a first glance I can't say  EDIT: see review below

/cc @romartin

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
@@ -41,13 +41,14 @@ public class DefinitionImpl<T> implements Definition<T> {
 
     @Override
     public int hashCode() {
-        return HashUtil.combineHashCodes(definition.hashCode());
+        return HashUtil.combineHashCodes(definition == null ? 0 : definition.hashCode());
     }
 
     @Override
     public boolean equals(Object obj) {
         if (obj instanceof Definition) {
-            return ((Definition) obj).getDefinition().equals(definition);
+            Object d = ((Definition) obj).getDefinition();
+            return this.definition == null ? d == null : this.definition.equals(d);
         }
         return false;
     }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
@@ -41,13 +41,13 @@ public class DefinitionImpl<T> implements Definition<T> {
 
     @Override
     public int hashCode() {
-        return definition.hashCode();
+        return HashUtil.combineHashCodes(definition.hashCode());
     }
 
     @Override
     public boolean equals(Object obj) {
         if (obj instanceof Definition) {
-            return ((Definition)obj).getDefinition().equals(definition);
+            return ((Definition) obj).getDefinition().equals(definition);
         }
         return false;
     }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
@@ -18,6 +18,7 @@ package org.kie.workbench.common.stunner.core.graph.content.definition;
 
 import org.jboss.errai.common.client.api.annotations.MapsTo;
 import org.jboss.errai.common.client.api.annotations.Portable;
+import org.kie.workbench.common.stunner.core.util.HashUtil;
 
 @Portable
 public class DefinitionImpl<T> implements Definition<T> {
@@ -36,5 +37,18 @@ public class DefinitionImpl<T> implements Definition<T> {
     @Override
     public void setDefinition(final T definition) {
         this.definition = definition;
+    }
+
+    @Override
+    public int hashCode() {
+        return definition.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof Definition) {
+            return ((Definition)obj).getDefinition().equals(definition);
+        }
+        return false;
     }
 }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
@@ -63,8 +63,8 @@ public final class ViewImpl<W> implements View<W> {
     public boolean equals(Object o) {
         if (o instanceof View) {
             View other = (View) o;
-            return definition.equals(other.getDefinition()) &&
-                    bounds.equals(other.getBounds());
+            return (definition == null ? other.getDefinition() == null : definition.equals(other.getDefinition())) &&
+                    (bounds == null ? other.getBounds() == null : bounds.equals(other.getBounds()));
         }
         return false;
     }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
@@ -19,6 +19,7 @@ package org.kie.workbench.common.stunner.core.graph.content.view;
 import org.jboss.errai.common.client.api.annotations.MapsTo;
 import org.jboss.errai.common.client.api.annotations.Portable;
 import org.kie.workbench.common.stunner.core.graph.content.Bounds;
+import org.kie.workbench.common.stunner.core.util.HashUtil;
 
 @Portable
 public final class ViewImpl<W> implements View<W> {
@@ -50,5 +51,21 @@ public final class ViewImpl<W> implements View<W> {
     @Override
     public void setBounds(final Bounds bounds) {
         this.bounds = bounds;
+    }
+
+    @Override
+    public int hashCode() {
+        return HashUtil.combineHashCodes(definition.hashCode(),
+                                         bounds.hashCode());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof View) {
+            View other = (View) o;
+            return definition.equals(other.getDefinition()) &&
+                    bounds.equals(other.getBounds());
+        }
+        return false;
     }
 }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
@@ -55,8 +55,8 @@ public final class ViewImpl<W> implements View<W> {
 
     @Override
     public int hashCode() {
-        return HashUtil.combineHashCodes(definition.hashCode(),
-                                         bounds.hashCode());
+        return HashUtil.combineHashCodes(definition == null ? 0 : definition.hashCode(),
+                                         bounds == null ? 0 : bounds.hashCode());
     }
 
     @Override
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElement.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElement.java
@@ -21,6 +21,7 @@ import java.util.Set;
 
 import org.kie.soup.commons.validation.PortablePreconditions;
 import org.kie.workbench.common.stunner.core.graph.Element;
+import org.kie.workbench.common.stunner.core.util.HashUtil;
 
 public abstract class AbstractElement<C>
         implements Element<C> {
@@ -63,12 +64,15 @@ public abstract class AbstractElement<C>
             return false;
         }
         AbstractElement that = (AbstractElement) o;
-        return uuid.equals(that.uuid);
+        return uuid.equals(that.uuid) &&
+                labels.equals(that.labels) &&
+                content.equals(that.content);
     }
 
     @Override
     public int hashCode() {
-        return uuid == null ? 0 : ~~uuid.hashCode();
+        return HashUtil.combineHashCodes(
+                uuid.hashCode(), labels.hashCode(), content.hashCode());
     }
 
     @Override
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElement.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElement.java
@@ -64,15 +64,17 @@ public abstract class AbstractElement<C>
             return false;
         }
         AbstractElement that = (AbstractElement) o;
-        return uuid.equals(that.uuid) &&
-                labels.equals(that.labels) &&
-                content.equals(that.content);
+        return uuid.equals(that.getUUID()) &&
+                labels.equals(that.getLabels()) &&
+                content == null ? that.getContent() == null : content.equals(that.getContent());
     }
 
     @Override
     public int hashCode() {
         return HashUtil.combineHashCodes(
-                uuid.hashCode(), labels.hashCode(), content.hashCode());
+                uuid.hashCode(),
+                labels.hashCode(),
+                content == null ? 0 : content.hashCode());
     }
 
     @Override
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/GraphImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/GraphImpl.java
@@ -22,11 +22,6 @@ import org.kie.soup.commons.validation.PortablePreconditions;
 import org.kie.workbench.common.stunner.core.graph.Edge;
 import org.kie.workbench.common.stunner.core.graph.Graph;
 import org.kie.workbench.common.stunner.core.graph.Node;
-import org.kie.workbench.common.stunner.core.graph.content.definition.Definition;
-import org.kie.workbench.common.stunner.core.graph.content.definition.DefinitionSet;
-import org.kie.workbench.common.stunner.core.graph.processing.traverse.tree.AbstractTreeTraverseCallback;
-import org.kie.workbench.common.stunner.core.graph.processing.traverse.tree.TreeWalkTraverseProcessor;
-import org.kie.workbench.common.stunner.core.graph.processing.traverse.tree.TreeWalkTraverseProcessorImpl;
 import org.kie.workbench.common.stunner.core.graph.store.GraphNodeStore;
 import org.kie.workbench.common.stunner.core.util.HashUtil;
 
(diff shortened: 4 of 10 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
```diff
@@ -37,4 +38,17 @@ public T getDefinition() {
     public void setDefinition(final T definition) {
         this.definition = definition;
     }
+
+    @Override
+    public int hashCode() {
+        return definition.hashCode();
```
jomarko: I think we should use `HashUtil.combineHashCodes(definition.hashCode())`, WDYT?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/GraphImpl.java
```diff
@@ -79,45 +74,16 @@ public void clear() {
...
-                                      Object def = ((Definition) (node.getContent())).getDefinition();
-                                      myHashArr[0] = HashUtil.combineHashCodes(myHashArr[0],
-                                                                               def.hashCode());
-                                  }
-                                  return true;
-                              }
-                          });
-        return hashArr[0];//Get the hash from the graph traversal
+        return HashUtil.combineHashCodes(
+                getUUID().hashCode(), this.nodeStore.hashCode());
```
evacchi: So, I am assuming here we can delegate to the node store, since it is implemented via Map

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionImpl.java
evacchi: I assumed they would be equivalent (when there is only 1 arg) but I should check

evacchi: hah. they don't; this assertion fails:

        assertEquals("foobar".hashCode(), HashUtil.combineHashCodes("foobar".hashCode()));

evacchi: addressed in my last commit

jomarko:
@evacchi sorry, one more post-review question, looking on this [class](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/diagram/DiagramImpl.java) shouldn't we care before `combineHashCodes()` is called that the hash codes we combine are not null?

evacchi:
@jomarko "damn thou, mutability!" (*shakes fist*)
BTW does it look ok now? 
also, waiting for @hasys

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
```diff
@@ -51,4 +52,20 @@ public Bounds getBounds() {
...
+
+    @Override
+    public int hashCode() {
+        return HashUtil.combineHashCodes(definition.hashCode(),
+                                         bounds.hashCode());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof View) {
```
hasys: Sometimes you test `instanceof Interface`, sometimes `instanceof Implementation`. Is it expected?

evacchi:
@hasys can you point me to the test suites for these classes so I can add these cases as well?

jomarko:
@evacchi I discussed this a little bit with @hasys , he explained me that `HashUtil` usage is most important in the "gwt client" classes. I agree with him, that some tests of the added methods would be fine, please let me know once his comments are incorporated, I will give quick check if the DMN designer was not affected by this PR.

hasys:
@evacchi,

tests can be placed on the same directory but under Test directory instead of Java directory. Also you can use this class as start point.
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/HashCodeAndEqualityTest.java

Also as inspiration can be used: 
https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals%28java.lang.Object%29
https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode%28%29

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
hasys: I see.. It make sense for GraphNodeStoreImpl, but Graph#nodes() method like usable for us here.
What do you think?

evacchi: right. fixed in my last commit

evacchi:
hey @hasys: what do you think about the tests I have added? Do they make sense?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElement.java
```diff
@@ -63,12 +64,17 @@ public boolean equals(final Object o) {
             return false;
         }
         AbstractElement that = (AbstractElement) o;
-        return uuid.equals(that.uuid);
+        return uuid.equals(that.getUUID()) &&
```
evacchi: So I figured it might not make sense to compare UUIDs, because the UUID *identifies* an element, yet two elements might have different UUIDs, but matching contents. I'm not sure what the policy should be in this case. 

For instance, Parent/Child edges have necessarily different UUIDs, yet they might be equal

i.e. I would expected (a,b) equals (x,y) rwhen a equals x, b equals y, regardless UUIDs. 

Specifically, the reason I'm doing this PR is that @hasys suggested to do migration tests between new/old impl of the BPMN marshallers via `assertEquals(graphOld, graphNew)`. This all falls over if equality relies on UUIDs, which might be automatically generated because they are not present in the BPMN serialization to XML. We might also think of a different Edge implementation that ignores UUIDs when it's a relationship. I don't know what's the better way. WDYT?

/cc @romartin

hasys: From me it's OK do not compare UUIDs. I assume that if element B has the same UUID as A, than B IS A. But if they have different UUIDs but same content -> B EQUALS A.

@romartin, your turn :)

CC @tsurdilo

romartin:
Hey @evacchi @hasys @jomarko 
First of all thanks for all the job you're doing guys :+1: 
Lemme brainstorm and give my 5 cents too:
About changing the collection instances for the node in/out connectors to `Set` instead of actual `List`
I don't see any issues on this, probably can be used `Collection` or even `Stream` in the `Node` interface method return types too, and the `NodeImpl` will just rely on Sets... well just brainstorming, but anyway I don't see any problem on changing this
About instance equality stuff - tbh started writing feedback about your comments but it's getting too long... I think better brainstorming this together in a quick call tomorrow (Tue 30)... what about after lunch time? 3-4pm? Will ping you or reach me out on IRC. I think we should discuss this in more detail as it has some implications...
Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertEquals(a.hashCode(),
+                     b.hashCode());
+
+        b.setDefinition("b");
+        assertNotEquals(a.hashCode(),
+                        b.hashCode());
+
+        b.setDefinition("a");
+        assertEquals(a.hashCode(), b.hashCode());
+
```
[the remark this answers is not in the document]
evacchi: I have added null assertions in all of the test cases

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,71 @@
...
+    @Test
+    public void testDefinitionEquals() {
+        DefinitionImpl<String> a = new DefinitionImpl<>("a");
+        DefinitionImpl<String> b = new DefinitionImpl<>("a");
+        assertEquals(a,
+                     b);
+
+        b.setDefinition(null);
+        assertNotEquals(a,
+                        b);
```
hasys: Please, add
```java
assertNotEquals(b, a);

Object c = new Object();
assertNotEquals(a, c);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,71 @@
...
+        assertEquals(a.hashCode(),
+                     b.hashCode());
+
+        b.setDefinition("b");
+        assertNotEquals(a.hashCode(),
+                        b.hashCode());
+
+        b.setDefinition(null);
+        assertNotEquals(a,
+                        b);
```
hasys: Here should be `hashCode` method
```java
assertNotEquals(a.hashCode(), b.hashCode());
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionSetHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,93 @@
...
+        b.setDefinition("a");
+        b.setBounds(new BoundsImpl(new BoundImpl(0.0,
+                                                 0.0),
+                                   new BoundImpl(5.0,
+                                                 5.0)));
+        assertNotEquals(a,
+                        b);
+
+        a.setBounds(null);
+        assertNotEquals(a,
```
hasys: Same here:
```java
assertNotEquals(b, a);

[...]

a.setDefinition(null);
assertNotEquals(a, b);
assertNotEquals(b, a);

Object c = new Object();
assertNotEquals(a, c);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionSetHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,93 @@
...
+        a.setBounds(null);
+        assertNotEquals(a.hashCode(),
+                        b.hashCode());
+
+        a.setBounds(new BoundsImpl(new BoundImpl(0.0,
+                                                 0.0),
+                                   new BoundImpl(5.0,
+                                                 5.0)));
+
+        assertEquals(a.hashCode(), b.hashCode());
```
hasys: b.setDefinition(null);
        assertNotEquals(a.hashCode(), b.hashCode());

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,106 @@
...
+                                                 0.0),
+                                   new BoundImpl(5.0,
+                                                 5.0)));
+        assertNotEquals(a,
+                        b);
+
+        b.setBounds(null);
+        assertNotEquals(a,
+                        b);
+
```
hasys:
```java
assertNotEquals(b, a);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,106 @@
...
+        b.setBounds(null);
+        assertNotEquals(a.hashCode(),
+                        b.hashCode());
+
+        b.setBounds(new BoundsImpl(new BoundImpl(0.0,
+                                                 0.0),
+                                   new BoundImpl(1.0,
+                                                 1.0)));
+
+        assertEquals(a.hashCode(), b.hashCode());
```
hasys:
```java
        b.setDefinition(null);
        assertNotEquals(a.hashCode(), b.hashCode());
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,106 @@
...
+        b.setBounds(null);
+        assertNotEquals(a,
+                        b);
+
+        b.setBounds(new BoundsImpl(new BoundImpl(0.0,
+                                                 0.0),
+                                   new BoundImpl(1.0,
+                                                 1.0)));
+
+        assertEquals(a, b);
```
hasys:
```java
Object c = new Object();
assertNotEquals(a, c);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,71 @@
...
+@RunWith(JUnit4.class)
+public class DefinitionHashCodeAndEqualityTest {
+
+    @Test
+    public void testDefinitionEquals() {
+        DefinitionImpl<String> a = new DefinitionImpl<>("a");
+        DefinitionImpl<String> b = new DefinitionImpl<>("a");
+        assertEquals(a,
+                     b);
+
```
hasys:
```java
assertEquals(a, a);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/definition/DefinitionSetHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,93 @@
...
+@RunWith(JUnit4.class)
+public class DefinitionSetHashCodeAndEqualityTest {
+
+    @Test
+    public void testDefinitionSetEquals() {
+        DefinitionSet a = new DefinitionSetImpl("a");
+        DefinitionSetImpl b = new DefinitionSetImpl("a");
+        assertEquals(a,
+                     b);
+
```
hasys:
```java
assertEquals(a, a);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,106 @@
...
+                                                           new BoundImpl(1.0,
+                                                                         1.0)));
+        ViewImpl<String> b = new ViewImpl<>("a",
+                                            new BoundsImpl(new BoundImpl(0.0,
+                                                                         0.0),
+                                                           new BoundImpl(1.0,
+                                                                         1.0)));
+        assertEquals(a,
+                     b);
+
```
hasys:
```java
assertEquals(a, a);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElementHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    @Test
+    public void testAbstractElementEquals() {
+        AbstractElement<String> a = new ConcreteTestElement("a");
+        AbstractElement<String> b = new ConcreteTestElement("a");
+        assertEquals(a,
+                     b);
+
```
hasys:
```java
assertEquals(a, a);

Object c = new Object();
assertNotEqulas(a, c);
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/impl/AbstractElementHashCodeAndEqualityTest.java
```diff
@@ -0,0 +1,92 @@
...
+        a.setContent("x");
+        b.setContent("z");
+        assertNotEquals(a,
+                        b);
+
+        a.setContent(null);
+        assertNotEquals(a,
+                        b);
+
```
hasys:
```java
assertEquals(b, a);
```

hasys:
All comments above should be written as "Please, add this lines here" :)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/ViewImpl.java
```diff
@@ -51,4 +52,20 @@ public Bounds getBounds() {
...
+    public int hashCode() {
+        return HashUtil.combineHashCodes(definition == null ? 0 : definition.hashCode(),
+                                         bounds == null ? 0 : bounds.hashCode());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof View) {
+            View other = (View) o;
+            return (definition == null ? other.getDefinition() == null : definition.equals(other.getDefinition())) &&
```
hasys: It can be simplified:
```java
Objects.equals(definiton, other.getDefinition());
```

evacchi: omg love this

evacchi:
So, I spoke to @romartin and, w.r.t. **edge** equality I think we can proceed as follows:

- let be e1 = (id1, content1, src1, tgt1), e2 = (id1, content2, src2, tgt2) where:
   - id is an identifier
   - content is the edge content
   - src is the source node
   - tgt is the target node 

(I am omitting labels for simplicity here)

then **generally** it should be:

      e1=e2 iff id1=id2, content1=content2, src1=src2, tgt1=gt2

BUT, in the special case when we have relationship edges (e.g. parent/child) ids are not stable across session, because these edges **are not serialized** (*)

(*) in BPMN

So, the right thing ™ would probably be to subclass Edges and override equals accordingly; however, it is probably easier to special case `equals` when `content instanceof Child` or Parent or Docker. In this case we relax the constraint as follows:

     e1=e2 iff src1=src2, tgt1=gt2

romartin:
Hey @evacchi @hasys 
Yes, what @evacchi says sounds good for me. 
There is probably another remark that makes sense brainstorming now - performance. So at runtime, while authoring from the UI and Stunner is mutating the graph behind the scenes, it's performing several operations that probably are checking graph instance equality. Actually the equality for nodes/edges etc just relies on the `UUID`'s, so those should be quite cheap evaluations. 
If at that point we're going to add additional logic, which is only being used for that situation about "comparing" diagrams, probably has an additional cost at runtime as well. 
Not sure if I'm explaining... anyway just brainstorming too :)

romartin:
Hey @evacchi , not sure if this PR still makes sense - you want it to be merged? Otherwise please feel free to close it. Thanks!!

evacchi:
let's close it for now

PR #1395, opened 2018-02-01
AF-980: Indexing: Improve on false negatives during parallel Unit Test execution

manstis:
See https://issues.jboss.org/browse/AF-980

Reviewers just those people with an interest in this being fixed (or at least worked around)..

I have a fear (if this works) that it simply masks the problem and any underlying issue never fixed.

### DIFF
--- a/kie-wb-common-forms/kie-wb-common-forms-editor/kie-wb-common-forms-editor-backend/src/test/java/org/kie/workbench/common/forms/editor/backend/indexing/FormDefinitionIndexerTest.java
+++ b/kie-wb-common-forms/kie-wb-common-forms-editor/kie-wb-common-forms-editor-backend/src/test/java/org/kie/workbench/common/forms/editor/backend/indexing/FormDefinitionIndexerTest.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.annotation.concurrent.NotThreadSafe;
 import javax.enterprise.inject.Instance;
 
 import org.junit.Test;
@@ -60,6 +61,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.when;
 
+@NotThreadSafe
 @RunWith(MockitoJUnitRunner.class)
 public class FormDefinitionIndexerTest extends BaseIndexingTest<FormResourceTypeDefinition> {
 
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/test/java/org/kie/workbench/common/screens/datamodeller/backend/server/indexing/ImpactAnalysisJavaFileTest.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/test/java/org/kie/workbench/common/screens/datamodeller/backend/server/indexing/ImpactAnalysisJavaFileTest.java
@@ -26,6 +26,8 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Matcher;
 
+import javax.annotation.concurrent.NotThreadSafe;
+
 import org.apache.lucene.analysis.Analyzer;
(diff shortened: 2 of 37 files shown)


manstis:

jhrcek:
@manstis I'm for running full downstream build 10 times to confirm it reduces indexing test failures. But I doubt this will help.

From my experiments playing with those tests (executing `mvn clean test -Dtest=SpecificRandomlyFailing#testMethod` using bash script 100 times serially and getting 4-6 failures every time) i doubt this issue has anything to do with test modules being executed in parallel. What I understood from what @porcelli was saying is that it's the parallelism at the level of indexing threads that's causing the problem..

manstis:
@jhrcek I completely agree. 100%

.. and I mentioned my fear that this will just mask the issue to make our CI world more stable/usable.

I'll leave the decision as to whether having this (possible) workaround is better (or worse) than having the intermittent false-negatives for general discussion hoping others on the review list comment in due course.

mareknovotny:
basically I am not against this changes, but as @jhrcek wrote above the root cause should be probably discovered first as parallel build can mean different things.

porcelli:
@mareknovotny @jhrcek @manstis the reality is that this has lower priority over scalale workspace solution... until we get `scalale workspace` done, I won't be able to go deep on this issue and provide a definitive fix.

manstis:
@mareknovotny ... agreed.. and then it comes back to the discussion on BSIG again.. we either live with the continued false negatives for indexing tests - as nobody has time to investigate and fix properly at the moment - or we put in place a quick workaround to make our CI lives easier...

jhrcek:

jhrcek:
Ok, I'm willing to accept this workaround, but let's make sure it actually removes those random failures by running full downstream build 5-10 times.

So far 2 executions without errors.
3rd try: Jenkins execute full downstream build

manstis:

Failure was due to version changes for ```7.7.0-SNAPSHOT```

manstis:
Indexing issue in ```drools-wb``` (I have not made the necessary changes in that repository).

```
org.drools.workbench.screens.guided.rule.backend.server.indexing.IndexRuleMultipleTypesTest.testIndexDrlRuleMultipleTypes

java.lang.AssertionError: Number of docs fulfilling the given query criteria expected:<2> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.kie.workbench.common.services.refactoring.backend.server.IndexingTest.searchFor(IndexingTest.java:242)
	at org.drools.workbench.screens.guided.rule.backend.server.indexing.IndexRuleMultipleTypesTest.testIndexDrlRuleMultipleTypes(IndexRuleMultipleTypesTest.java:55)
```

manstis:

manstis:

jhrcek:
@manstis  Looks like we have 5 executions without errors in modified classes. Please rebase and runfull build for the last time and I'll approve.

manstis:

jomarko:
Looks like index fail appeared.

manstis:
@jomarko Yes and no.. No as the only tests to have had the workaround applied are in ```kie-wb-common```; but yes... there is a failing test in ```drools-wb``` (that has not had any workaround applied).. If the general consensus is to merge this PR I'll make similar changes in ```drools-wb```. So I guess, ATM, we're only looking for failing tests in ```kie-wb-common```.

manstis:
@jhrcek Your understanding is spot on - and indeed identical to my original thinking (as I first added the annotations to just the top-level base indexing classes and thought "yah, it'll be inherited by our whole test suite"). However I soon found that ```@NotThreadSafe``` is not an _inherited_ annotation and hence will need adding to every (indexing) test class. I'll update ```drools-wb``` after this PR is merged and then pick up others as and when we see failures (I suspect we'd then be limited to Designer and possibly OptaPlanner).

manstis:
Merging as full downstream failures are in ```drools-wb``` (additional PR to follow for that repository),

PR #1400, opened 2018-02-02
RHDM-360: Increase coverage of EnumLoading

jomarko:
@manstis please have a look


jomarko:
@manstis this is ready for merge I believe

PR #1335, opened 2018-01-05
AF-930: [Layout Editor] Use dock panels to increase the editor content area

dgutierr:
Changes required by https://github.com/kiegroup/appformer/pull/125

@pefernan Still pending to integrate the component palette dock panel into the authoring perspective. Can you please help me with that? :-)

PR #1407, opened 2018-02-06
Prevent some random failures of ValidatorTest

jhrcek:
@manstis this should  prevent some random failures of ValidatorTest which accounts for quite a number of failures in the past few months:
screenshot from 2018-02-06 08-25-01

Fix is based on [your comment](https://github.com/kiegroup/kie-wb-common/pull/1335#issuecomment-362204190). Also cleaned up the class (no need to use mockito for this class, and some java 8 boilerplate fixes).


manstis:
Thanks @jhrcek

PR #1408, opened 2018-02-06
JBPM-6730: Stunner - New unmarshaller implementation

evacchi:
Enable with flag:
    -Dbpmn.marshaller.experimental=true

details in `kie-wb-common-stunner-bpmn-backend/README.md`

supersedes #1336 

@romartin @hasys


romartin:
ok to test

romartin:
Hey @evacchi just quick note for further PRs - although is not mandatory, we're using a naming convention for the github integration with the CI, the idea is to use as branch name for the pull requests the ticket id. This way your tickets' status will be automatically updated and if you have different related PRs from different repos, Jenkins does the magic too and will build the concrete branches.

romartin:
Jenkins please retest this

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/Definition.java
```diff
@@ -16,7 +16,7 @@
 
 package org.kie.workbench.common.stunner.core.graph.content.definition;
 
-public interface Definition<T> {
+public interface Defihttps://github.com/kiegroup/kie-wb-common/pull/1390nition<T> {
```
romartin: ooops - I think this is the compilation issue ::)

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/background/BackgroundSet.java
```diff
@@ -17,6 +17,8 @@
 
 import javax.validation.Valid;
 
+import com.google.common.base.MoreObjects;
```
romartin: oh that's new for me! :) is this shared code? xO

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/graph/content/definition/Definition.java
evacchi: whoops

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/background/BackgroundSet.java
evacchi: it's Guava's `Objects#toStringHelper`, Guava is in your dependencies so i figured it could be used. We can revert that, though. Reason for `MoreObjects` instead of  `Objects` is that  `Objects#toStringHelper` it's deprecated

evacchi:
> Thanks for the readme! It helps 👍 IMO Would be useful here a small section of what to do when adding support fore more nodes, I mean, if adding a new activity type, it requires creating/updating converters and propertyReaders, etc. Does that makes sense for you?

it certainly does! I'm starting to work on the marshallers and trying to figure out if I can come up with something symmetrical to the unmarshalling side. It's complicated, because the two representations are not very similar.

> Not necessary yet but we must think as well on providing some CDI extension point/s that make sense, I mean for possible specializations (customer/community fixes/improvements) or future product patches

The reason why I haven't added CDI extension points (yet) is that I don't want to make the code too "rigid" at this point, by making it implement some interface. It's still very much in flux, so I'm changing API signatures very often. Once that has stabilized, I think the common design will come up naturally, and we can abstract those details more easily. That should make it simpler to add new components; right now it's a bit "manual".

evacchi:
so it is building fine, all tests are passing **but** migration tests are broken (ignored for now)

unfortunately... I fear that they're broken because there was some regression *not* on my end. 

- package name on the Diagram used to defaut to "org.jbpm". Now it is null when missing (I have "fixed" this on my end, not sure it is the intended behavior)
- scripts are missing sometimes, e.g. see userTaskAssignees.bpmn and startTimerEvent.bpmn (my impl is parsing it as I'd expect). An empty script is returned, instead of the body
- also, in a task, there is **always** at least one empty script with language = "java", even when there is no script at all in the BPMN file

HTH

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/background/BackgroundSet.java
romartin: okk thanks for the feedback! Well this is shared code, so if google does not complains at compile time it's fine hehe, but I would say guava is just server side stuff. Anyway np for me, was just curious :)

evacchi: oh I didn't understand that you meant "shared between server and client". btw if it's a problem I can drop the dependency (but I think that Guava might exist for gwt)

romartin:
Hey @evacchi 

Yep sorry - wer're "on fire" right now adding and adding stuff, so code changes quite fast... please be patient,  you're task is more long term so... :)

About the migration issues a few notes:
* About the process id, which it was `org.jbpm` before - I'm not aware of any change on Stunner side about this, @wmedvede are you? Anyway we can check this together in next call , no worries.
* About the scripts missing and empty default script to user task... well @wmedvede was just working on this ticket, which has been merged a few days ago. Can you please @wmedvede  point @evacchi on this? Should he some marshalling be affected for this?

I'm trying to figure the issue on Jenkins too, I'm facing the same one - the maven enforcer plugin complains, but not sure if it's due to our changes... just looking at this, will let you know

romartin:
oh BTW @evacchi tip for free :P - when you get different build error locally rather than from Jenkins, you can try to run the maven build locally using `-Dfull` -> this enables all plugin and checks as Jenkins does
So `mvn clean install -Dfull` is the full equip build, but will take long!

romartin:
yes @evacchi please, it seems the Jenkins build error is due to maven enforcer rules failed, just run the maven build using the `full` profile (`-Dfull`) and I'm pretty sure you'll get the build error, lemme know!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/background/BgColor.java
```diff
@@ -73,4 +73,9 @@ public boolean equals(Object o) {
         }
         return false;
     }
+
+    @Override
+    public String toString() {
+        return String.format("BgColor(%s)", String.valueOf(value));
```
jhrcek: Calling `String.valueOf()` on a String seems redundant. Why not just ` String.format("BgColor(%s)", value);`

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/background/BorderColor.java
```diff
@@ -73,4 +73,9 @@ public boolean equals(Object o) {
         }
         return false;
     }
+
+    @Override
+    public String toString() {
+        return String.format("BorderColor(%s)", String.valueOf(value));
```
jhrcek: Same as above (search for all occurrences of this throughout this PR)

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/DefinitionResolver.java
```diff
@@ -0,0 +1,84 @@
...
+    private final Map<String, Signal> signals = new HashMap<>();
+    private final Map<String, ElementParameters> elementParameters = new HashMap<>();
+
+    public DefinitionResolver(Definitions definitions) {
+        for (RootElement el : definitions.getRootElements()) {
+            if (el instanceof Signal) {
+                signals.put(el.getId(), (Signal) el);
+            }
+        }
+        FeatureMap value = definitions.getRelationships().get(0).getExtensionValues().get(0).getValue();
```
jhrcek: Can we guarrantee that getRelationships will always be nonempty (similarly for ExtensionValues), or are we creating potential for index out of bounds exceptions?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/background/BgColor.java
evacchi: uhm I recalled that String.valueOf(null) printed null, whereas just printing null would print an empty string. But I was wrong, and this actually throws NPE... I will revert this.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/DefinitionResolver.java
evacchi: I am assuming those wrote this knew what they were doing:
https://github.com/evacchi/kie-wb-common/blob/158b6b473d27f79358e025c26c62ba56b013dfe8/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonMarshaller.java#L301

in other words, this code mimics what the old marshallers were doing

wmedvede:
Hi @romartin , @evacchi 
In the Stunner models the scripts are defaulted in general with an empty script = "" and language = "java", or with script = "autocomplete" and language = "mvel".
At parsing time the scripts are populated with the values that comes from the process bpmn2 and the language is defaulted (only when missing) with the same criteria as let's say the old marshalling, i.e. normally to java.
But see that as we spoke there has been some changes and now the scripts are mapped into a ScriptTypeValue object, instead of having separate properties as we had before.

Here is an example about how the OnEntryScriptType in the bpmn2 extended model is converted into a ScriptTypeValue in the Stunner model. 

https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonMarshaller.java#L960

And here is an example that shows the other direction

https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonUnmarshaller.java#L4418

wmedvede:
Hi @romartin @evacchi regarding he process id, this is something that was left broken in the middle of some commits by @romartin  and the new WorkspaceProjects work recently merged to the master. And I had to do some repairs because the new process was created wrong.

Here is how it works now. (and this is also how it worked before the errors I mentioned above)

If we have a project "MyProject" and we create a new process HelloWorld in a package "myPackage" this are the values the process will have.

Process Name = HelloWorld
ID = MyProject.HelloWorld
package = myPackage

Here's the fix:

https://github.com/kiegroup/kie-wb-common/commit/423c664200d73fdc803e0e0bdfc552d39730ae30#diff-36fe92ebe1da31be931dcff8fd83d941

evacchi:
>  In the Stunner models the scripts are defaulted in general with an empty script = "" and language = "java", or with script = "autocomplete" and language = "mvel".
> [...]
>  But see that as we spoke there has been some changes and now the scripts are mapped into a ScriptTypeValue object, instead of having separate properties as we had before.

I am aware of that and aligned my code accordingly, but it looks like scripts are parsed as empty in userTaskAssignees.bpmn and startTimerEvent.bpmn (possibly others) although they are non-empty in the file. Have you checked it now works as expected? Then I'll rebase again my code.

evacchi:
@romartin 

> yes @evacchi please, it seems the Jenkins build error is due to maven enforcer rules failed, just run the maven build using the full profile (-Dfull) and I'm pretty sure you'll get the build error, lemme know!

so the build fails because of "duplicates", but it seems to me it's all unrelated code: `common/stunner/code` :S any idea?

https://gist.github.com/evacchi/1dbcb984482bb1cde1bce2e38b71f6e0

wmedvede:
Hi @evacchi 

In the userTaskAssignees.bpmn:

The UserTask has no onEntryAction or onExitAction scripts, so no scripts are parsed for this task.
BUT this bpmn file has two additional scripts tasks, and the script and language is parsed well for both.

In the startTimerEvent.bpmn:

The same, the start event has no scripts. But instead in the file there are a 
script task. And the script and language is parsed well too.

wmedvede:
Hmm, no I see @evacchi. Even when the UserTask has no onEntryAction in the bpmn file, a sort of "by default on onEntryAction" with no script is created. This is because of the fact that in the Stunner model, when the UserTask is created it's poulated with the by default values we want to show in the UI. In the past the by default values were managed separated, i.e. it was as special @Default annotation for indicating the by default value for a field, but @romartin did some simplifications and now the only way we have let's say to set by default values is by setting by default values in the Stunner model, directly in the java class. So when a new UserTask is created in the Stunner model, it'll be automatically populated with the by default values.

wmedvede:
@romartin @evacchi I believe the "problem" of by default values in the java class might affect all the nodes. See this.
When the node in the Stunner model is created, since the by default value is "programmed" in the java class, a by default value is set.
When the unmarshaller needs to populate the node, if there's a value in bpmn2 file, then alight, the value from the bpmn is set. BUT if not value is present in the bpmn2 file, then the Stunner node is left as is. And it will carry the by default value even when no value exists in the bpmn2 file.

evacchi:
yeah @wmedvede I have understood the thing about the default value, but in my migration tests I've found that 

userTaskAssignees:

     scriptTask id="_BC95EB85-9025-4D32-823D-6AECA2F89F26"

has this script body:

         <bpmn2:script>
              <![CDATA[System.out.println("s1  s1 = " + s1);]]>
        </bpmn2:script>

but the unmarshaller returns the empty string (i.e. the default value). Can you confirm that, or is there a problem in my own build? Thanks!

wmedvede:
Hi @evacchi,
I confirm that the script for the task you are mentioning scriptTask id="_BC95EB85-9025-4D32-823D-6AECA2F89F26" is properly unmarshalled.
i.e. if you run the BPMNDiagramMarshallerTest.testUnmarshallUserTaskAssignees() test, the diagram returned by this test contains the script task with the expected script loaded.

evacchi:
@wmedvede ok I'll check again, then! Thanks a lot

romartin:
Jenkins please retest this

romartin:
Hey @evacchi 

Had been now looking at the build issue about the maven enforcer plugin. 

Well tbh cannot connect to VPN now but doing the `full` maven build locally I got a maven enforcer error on the `bpmn-api` module as well, but it's not due ban duplicated classes as the one you pointed me before. So the error I face is:
`[WARNING] Rule 0: de.is24.maven.enforcer.rules.IllegalTransitiveDependencyCheck failed with message:
Found 2 illegal transitive type dependencies in artifact 'org.kie.workbench.stunner:kie-wb-common-stunner-bpmn-api:jar:7.7.0-SNAPSHOT':
1.) com.google.common.base.MoreObjects
2.) com.google.common.base.MoreObjects$ToStringHelper`
I just removed the use of that classes locally to check if this works, you can fix the pom by including the dependency declaration if so.

On the other hand I cannot see any error about ban duplicated classes, which is strange as it was there for you, anyway if this persists what you can do for now is just try adding an exclusion to `kie-wb-common-stunner-core-common:jar:compile` in the declaration of the `kie-wb-common-stunner-core-common:jar:test` (`bpmn-api` module), as it seems both jars contain the same stuff.... anyway I think that would be not necessary, only fixing the above "illegal transitive" checks the build works for me..

Good luck! :+1:

romartin:
@evacchi sorry well the build fails due to [this error](https://gist.github.com/romartin/2c8e7f3efd5ff2e0991305f8f66d4849), which seems related to your new stuff, but at least it's building now the `bpmn-backend` module, which means the issue for the `bpmn-api` one is fine... lemme know if it's the same for you.

evacchi:
@romartin ok that was a botched commit when I checked out the other branch, let's try again

evacchi:
jenkins please retest this

hasys:
Hi @evacchi 
there are still a compilation issue, mentioned by @romartin:

```
[INFO] Analyze dependencies of artifact 'org.kie.workbench.stunner:kie-wb-common-stunner-bpmn-api:jar:7.7.0-SNAPSHOT'.
[WARNING] Rule 0: de.is24.maven.enforcer.rules.IllegalTransitiveDependencyCheck failed with message:
Found 2 illegal transitive type dependencies in artifact 'org.kie.workbench.stunner:kie-wb-common-stunner-bpmn-api:jar:7.7.0-SNAPSHOT':
1.) com.google.common.base.MoreObjects
2.) com.google.common.base.MoreObjects$ToStringHelper

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------

```

you can reproduce it executing `mvn clean install -Dfull` under `kie-wb-common/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api`

evacchi:
weird thing is that I'm getting a different error

evacchi:
ok, I think the transitive dependency issue has been fixed, @hasys is taking another look

hasys:
Something strange with build not related to the PR. 
Jenkins retest please

hasys:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/test/java/org/kie/workbench/common/stunner/bpmn/backend/service/diagram/MigrationDiagramMarshallerTest.java
```diff
@@ -0,0 +1,629 @@
...
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import org.junit.Before;
+import org.junit.Ignore;
```
hasys: It can be removed since Migration tests work now.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/test/java/org/kie/workbench/common/stunner/bpmn/backend/service/diagram/MigrationDiagramMarshallerTest.java
```diff
@@ -0,0 +1,629 @@
...
+                BPMN_INTERMEDIATE_ERROR_EVENTCATCHING,
+                BPMN_INTERMEDIATE_SIGNAL_EVENTTHROWING,
+                BPMN_INTERMEDIATE_MESSAGE_EVENTCATCHING,
+                BPMN_INTERMEDIATE_MESSAGE_EVENTTHROWING,
+                BPMN_INTERMEDIATE_TIMER_EVENT,
+                BPMN_ENDSIGNALEVENT,
+                BPMN_ENDMESSAGEEVENT,
+                BPMN_ENDNONEEVENT,
+                BPMN_ENDTERMINATEEVENT,
+                //BPMN_PROCESSPROPERTIES,
```
hasys: This and two following processes can be uncomented and related methods removed since all of them are works now.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+    -Dbpmn.marshaller.experimental=true
+    
+### Design
+
+- Entry point: `BPMNDirectDiagramMarshaller` which implements `BPMNDirectDiagramMarshaller` 
+- Actual unmarshalling from XML is delegated to Eclipse BPMN2 library
+- Mapping between Eclipse BPMN2 data model to Stunner BPMN data model is performed through **Converter** classes
+
+#### Converters
```
hasys: What if some parse functionality move to concrete nodes classes? In this case converters method will be splitted to smaller peaces and some of the functionality can be moved to  base classes. I mean, it can be like this:
```java
class GatewayConverter {
  Node converter(Gateway gateway) {
    [...]
    gateway.parse();
    [...]
  }
}

class BaseGateway {
  convert() {
    [...]
    return this;
  }
}

class ParallelGateway {
  convert() {
    super.convert();
    [...]
    return this;
  }
}
```
Also there are no Unit tests, for all converters, as I can see. Am I correct?

WDYT is it possible and is it worth of efforts?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/marshall/json/builder/BPMNGraphGenerator.java
```diff
@@ -143,8 +143,11 @@ public boolean isClosed() {
     @SuppressWarnings("unchecked")
     public void close() throws IOException {
         logBuilders();
```
hasys: I guess this method should be removed entirely here and below. (It was one time and it flush process to server log twice).

As an option it can be optional using some system (maven) property for debugging purpose.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonMarshaller.java
```diff
@@ -326,7 +326,7 @@ public void marshall(JsonGenerator generator,
         marshallDefinitions(def,
                             generator,
                             preProcessingData);
-        generator.close();
+        //generator.close(); FIXME: can we remove this from here?
```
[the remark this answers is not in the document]
evacchi: most probably. Please keep in mind that all changes under backend/legacy and backend/json are minor refactorings intended for me to better understand the old marshallers. Those will be likely dropped as soon as we move to the new infrastructure

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/test/java/org/kie/workbench/common/stunner/bpmn/backend/service/diagram/MigrationDiagramMarshallerTest.java
evacchi: cool!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/marshall/json/builder/BPMNGraphGenerator.java
evacchi: see prev comment abount backend/json

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/README.md
evacchi: at this point I'm still not 100% sure about the structure. I am working on the marshallers now (different branch) and I'm getting close to a structure that its kind of symmetrical. Method signatures might change in the meantime, though. In any case, I wouldn't advise merging concrete classes with their converters, I'd rather do a Decorator-style class (see `*PropertyReader(s)`) 

Right now we have the following two components
- Converters, describing the association between properties in our data model and properties of the eclipse model (I'm using the term 'property' here in a loose way, because sometimes the properties are deeply nested)
- PropertyReaders, exposing accessors to the Eclipse model in a more convenient way, mapping more closely to the stunner model. PropertyReaders are not strictly necessary for conversion, but the code would be much more verbose if they weren't there
- the marshallers (converting from stunner model to eclipse) will have PropertyWriters which will be dual to PropertyReaders.

As for testing, I agree that we should strive for better coverage, but at this point -- when things are still in flux -- it might be a lot of work for little gain. At least we have the huge unit for (un)marshalling and the migration test.

evacchi:
just to let you know I'm working on **marshallers** (this PR is about **un**marshalling) on this other branch https://github.com/evacchi/kie-wb-common/tree/bpmn-marshaller

wmedvede:
Hi @evacchi , just to let you know that there has been a small change in the Bpmn2JsonMarshaller as part of https://github.com/kiegroup/kie-wb-common/commit/8bdc30b1439b3bdcc1ad46c50437bd39be977f38#diff-72489e3fe7c7f614a2f2f73dd6daff2dR2585

It's really silly, but you might probably have include it in the new marshallers.

Thanks.

romartin:
Hey @evacchi , yeah that works sounds good for me in general, so please rebase it (again sorry), keep working with @hasys to obtain the approval and let's try to merge it asap, as it's quite big and it implies continuous refactorings.

evacchi:
@romartin rebased and fixed

@wmedvede thanks Walter, that should be aligned now

@hasys I have addressed most of your concerns, as for code structuring, I'd wait to decide upon this up until we have covered marshalling and unmarshalling, but I wouldn't put code right inside of the Stunner data model, better keep it separated. As for test, i think the best way to proceed right now is to continue with migration tests that we have already, so that we aren't misaligned.

Again, this code is still WIP, and the only reason we want it to be merged is so that we can get early feedback on the progress so far; marshallers and further refactoring are in the pipeline

hasys:

evacchi:
So I'm all for merging this ASAP but I don't want this to get in the way of the release. My code is of course linking to current Stunner datamodel, and of course it had to be fixed when other PRs affected such a model. Is this going to happen again before the release? My fear is that these bits might get in your way as you merge other, more urgent PRs ! 

ping @romartin

romartin:
Hey @evacchi 

* There is one failed test - `org.kie.workbench.common.stunner.bpmn.backend.service.diagram.MigrationDiagramMarshallerTest.testUnmarshallOldStuff`. I think it's related to your work, isn't it? Cannot merge until it's being fixed.

* About this new code - well at some point we have to integrate this and start maintaining the new stuff. At this point not so much new nodes are going to be included for next GA, so I don't expect too much changes related to this, anyway yeah @evacchi you'll have to support @wmedvede or @alepintus once they are going to update any bean or property and let them know how to proceed. I understand that if finally this results really uncomfortable, which doesn't expect so, we can always revert the new packages and re-add this stuff later again, but hope that will be not the situation. So I'm agree on merging this.. but let's ask the team as well - WDYT @hasys @wmedvede @alepintus @tiagodolphine ?

Thanks!

evacchi:
so the test were failing because I restored *.bpmn tests cases that were in master
apparently:

if `drools:boundaryca=""`:

- BoundaryEvent => `CancelActivity` == false
- Not BoundaryEvent => `CancelActivity` == true

I'm not sure that's the expected behavior (ping  @wmedvede @alepintus @tiagodolphine @Josephblt) but now it's aligned.

hasys:
Hi guys,
at Friday, I provided Mutation testing report to @evacchi, now you also can download it here [1].

It is only for (un)marshaller staff, so if you want to do something similar to your code, you can use PIT maven plugin added by @yurloc some time ago. To invoke it just execute `mvn org.pitest:pitest-maven:mutationCoverage` in folder with maven KIE Group project. Just one caveat: only marshallers tests took more than 2 hours on my machine, so don't execute it against entire Stunner, or be ready to get results only in the next morning at the best case :)

About merging. I guess we can do it, but in this case, I guess, it is better DO NOT change current implementation of (un)Marshaller. So, we will need to support all it's "interesting" solutions in new (un)marshaller until we will be 100% sure that they are compatible and refactor old staff only after full migration when we will have 100% test coverage. For now we can put all such findings to TODO and FIXME comments to appropriate places or report them to separate Jira.  WDYT @romartin @evacchi @wmedvede @alepintus  @tiagodolphine @Josephblt ?

@evacchi About `CancelActivity`: if node is not a boundary event `CancelActivity` property is ignored, so it is make no effect the process, but if event is a boundary event - it does. I guess it is the reason why it behaves little bit different. Not, sure is it a good reason to do so, just a notion. And yes, for now, I guess, it is better to follow this behavior until merge and refactoring as I mentioned above.

[1] [pit-reports.zip](https://github.com/kiegroup/kie-wb-common/files/1736365/pit-reports.zip)

romartin:
Hey @evacchi ,

As commented on today's meeting, we're all agree on merging this, so let's go ahead! :+1: 

I cannot merge until the Jenkins builds are ok, please I need you to keep an eye on this PR  and fire the builds again when the drools issues are fixed, once finally it's ok lemme know and I'll merge. Thanks!!

romartin:
@hasys @jhrcek please need your approvals too, when you think it's ready... Thanks!

evacchi:
alright, I'll check again tomorrow. BTW the other branch (rebased on this) now passes 75 out of 76 tests :O

evacchi:
jenkins please retest this

evacchi:
So, apparently, this unrelated project is failing the full downstream build:

      [INFO] KIE (Drools) Workbench Tests :: GUI ................ FAILURE [ 12.318 s]

romartin:
ok @evacchi , re-firing the build let's see... anyway I'm still missing the @hasys and @jhrcek approvals...

romartin:

romartin:

romartin:
Hey @evacchi ! Now the builds are fine, although the downstream one is failed due to some tests, those are not realted to this, so we're ready for merging!! I'm looking at this PR every day, to check if it's ready to merge, but you need to obtain the approvals from @hasys  or @jhrcek please, we need somebody else to double check and approve the PRs for being merged. So please keep chasing them, or send them some IRC messages, or some mail... just not to forget to review this PR, once done lemme know and I'll merge!!

Thanks!! And again, great job ! :+1:

evacchi:
paging @hasys (or @jhrcek) for approval

romartin:
Please @hasys or @jhrcek let's do not forget this, as long it keeps as a PR more work will be necessary to keep it up-to-date. Thanks!

hasys:
Hi @evacchi,

thank you looks good, just some clean up provided by me in this PR: https://github.com/evacchi/kie-wb-common/pull/2

there you can find some suggestion about some minor clean up. I guess, all code you, probably, will need in the future will still preserve in your private branch so for now we can bravely remove all commented/unused new code. WDYT?

Also there are removed duplicated checks and some refactoring to streams.

Just **one question** there are an invocation without assert in the tests: https://github.com/evacchi/kie-wb-common/compare/bpmn-unmarshaller...hasys:1408CleanUp?expand=1#diff-5772128498f7023e3ecae4a4ca73646bL1123

When I add an assertion instead of Equals method invocation I see that check is failed. Is it a wrong method invocation or it is a bug?

evacchi:
thanks for the PR, I have replied there

evacchi:

romartin:
Great job @evacchi !! Thanks!

PR #1419, opened 2018-02-08
JBPM-6798: Stunner & Forms - Field updates are not fired when selecting a canvas shape

pefernan:
This PR fixes two issues:
- changes on firn fields on stunner are lost when selecting a canvas element. Value change event is not triggered when clicking on a canvas element. So for now before rendering a properties form, if already there's a properties form loaded and the status is valid FormPropertiesWidget will synchronize the model with the form values.

- Fixed validation issues: due to the old static forms for LiveSpark project there was two binding strategies static & dynamic. That was making that in some cases for dynamic forms the validations where executed over the BindableProxy instead of the actual model binded on the form. Now this binding strategies have been removed and DefaultModelValidator will make sure that the actual model is validated instead of the BindableProxy.

@jsoltes @romartin could you please take a look?

This is part of an ensemble merge with:
https://github.com/kiegroup/lienzo-core/pull/7
https://github.com/kiegroup/kie-wb-common/pull/1419


pefernan:
jenkins please execute full downstream build

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/DefaultModelValidator.java
```diff
@@ -49,6 +50,10 @@ public boolean validate(Collection<FormField> fields,
                             MODEL model) {
         boolean isValid = true;
 
+        if (model instanceof BindableProxy) {
```
romartin: hehe...

pefernan: @romartin hehehe ;)

pefernan:
jenkins please execute full downstream build

pefernan:
jenkins please execute full downstream build

pefernan:
@hasys weird! let me check

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/test/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImplTest.java
```diff
@@ -358,17 +361,15 @@ protected void runWrongValidationTest(boolean skipGetModel) {
         assertTrue(formHandler.validate(USER_BIRTHDAY_FIELD));
         assertTrue(formHandler.validate(USER_MARRIED_FIELD));
         assertFalse(formHandler.validate(USER_ADDRESS_FIELD));
+
+        verify(proxy, times(7)).deepUnwrap();
```
jsoltes: @pefernan I noticed that the same method is called in the last 6 calls of validate. Wouldn't it be enough to test one field or is a different behavior expected out different fields? Also it can be hard to identify where the problem is, if the deepUnwrap() is called 6 times and not 7 times. So if it has to be really tested for each of the fields, I would call verify after each, or if not - I would just test it with one field... This is just a nitpick, but the unusually high number in "times" dragged my attention. :)

pefernan:
@hasys @romartin @jsoltes 

After cheking the errors that @hasys provided I found out that it was impossible to run the forceModelSynchronization method on formRenderer with custom widgets (timer settings, scripts...) because they use complex logic to generate/validate their values.

The root of the problem is that clicking anywhere on the canvas is forcing the FormPropertiesWidget to render a form and that happens before the change events from the form widgets are triggered. So after the new form is rendered the change events from the previous form are triggered and nobody was listening for them to update the model...

The only solution I could find is delaying the removal of the previous formhandler so it has time to catch the previous change events and it can update the model.

Please take a look and let me know if you find anything.

hasys:

hasys:
Hi @pefernan,

there is one failing test in this PR [1], looks like it is related to your changes. Otherwise code looks good, thank you. Waiting for a build and than will provide final review :)

[1] org.kie.workbench.common.stunner.forms.client.widgets.FormPropertiesWidgetTest#testFormContextIsPathAware

pefernan:

pefernan:
@hasys fixed, let's see if we have the double green

hasys:
Hi @pefernan,

thank you, now it's work. But I found 3 more issues, sorry :)

1. If you put some code to Script property of Script Task and click on the same Script Task the Script property will be cleared. If you activate another node and than activate Script Task again - Script property will contain entered earlier value.
2. If I select Node and than will select any property value by Mouse and click Delete button on my keyboard - node will be deleted instead of property value. Some times node will be removed even when I select a node, put cursor to any value and press Delete button to remove one symbol.
3. User Task and Sequence Flows priority property doesn’t notify about incorrect value, just clear it. When we click on another node and show warning when we click on another property. Probably, it should raise validation of the value "on the fly" to prevent it.

pefernan:
@hasys @romartin 
1.- I think it has an easy fix, I'll update the PR soon.
2.- I think this issue needs a separate JIRA since it seems that it is not related to this PR. 
3.- This is caused by how events are executed by the browser (the origin of this issue), click event on canvas is triggered always before the change event on form widget so the new form is being rendered before old form catches the field change and evaluates if the values are valid or not. In the end if the value is wrong and the model isn't updated looks like an acceptable solution.

pefernan:
@hasys @romartin fixed issue #1 (rendering form when clicking for second time the selected element). The other issues are not going to be solved on this PR

pefernan:

pefernan:
same tests failing...

hasys:
Hi Pere,

sorry, but found two new issues:

1. Field are still not updates it's values when selecting a canvas shape in Firefox browser.
2. When I morph node to another one it's form remains for the previous one. For example if I have a User Task and morph it to Script Task form on the right side remains with properties for the User task.

pefernan:
@hasys fixed all issues, please let me know if you find anything else.

pefernan:

hasys:
Hi @pefernan,

I sent you small fixes there: https://github.com/pefernan/kie-wb-common/pull/2

Also I tested it with Chrome, FF, and IE11. In Chrome and IE11 and found two new issues:

1. (Firefox only) When you change node name and click on another node name is not changed on the canvas, but if you click back on the node with changed name it's new value will be preserved. You need to click somewhere else (outside of the canvas) to apply your new value. In Chrome and IE 11 it works as expected - new name value will be applied at once you click somewhere on the canvas.

2. Back to the priority property we discussed before. It looks like now when you put any incorrect value and click on the another node - it will keep value in the form. But if you save and reopen process this value will be lost. I guess we should be consistent and keep priority property clean of incorrect values or preserve incorrect value even for save and reopen of process.

pefernan:
@hasys try again, I did some minor changes to fix the issues, but requires some changes from lienzo side (https://github.com/kiegroup/lienzo-core/pull/7)

pefernan:
jenkins please execute full downstream build

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/Form.java
```diff
@@ -0,0 +1,45 @@
...
+        if(formField == null) {
+            throw new IllegalArgumentException("FormField cannot be null");
+        }
+
+        fields.add(formField);
+    }
+
+    public FormField findFormField(String fieldName) {
+        return fields.stream()
+                .filter(field -> field.getFieldName().equals(fieldName) || fieldName.equals(field.getFieldBinding()))
```
jomarko: To avoid NPE, could we use:
```java
Objects.equals(field.getFieldName(), fieldName) || Objects.euqals(fieldName, field.getFieldBinding())
```

pefernan: @jomarko well seen, it will be changed

pefernan:
jenkins please execute full downstream build

pefernan:
jenkins please execute full downstream build

hasys:
Hi @pefernan, @romartin 

Full downstream issues are not related to this changes. We are ready to go with this PR :)

PR #1421, opened 2018-02-09
[DROOLS-2310]: Fixed redirection issue after import projects

adrielparedes:
@manstis could you review? May be we should add project oriented people to this conversation. I've tested it in several cases and it works.


adrielparedes:
@manstis actually exists a test and the module needs to be null (I have just realized that). I'm trying to find project oriented guys to understand why

adrielparedes:
@manstis I've been talking with @mbarkley and we decided to remove the null module check because it looks that it was not necessary and to keep the behaviour it has now with the new changes.

manstis:
@adrielparedes If there is already a test then how/why did this regression come into existence!?! There may not be a test for the specific criteria needed to trigger the issue. Might I suggest one (or two?) for ``` onChange(WorkspaceProjectContextChangeEvent previous, WorkspaceProjectContextChangeEvent current)```?

adrielparedes:
@manstis you were right there was one condition that it was not tests. I added a test to check that if the module is not null it still can go to project view. The other tests were already there

mbarkley:
@manstis The `onChange` method was added with the project oriented changes. I think this is a consequence of having mostly mockito tests: our tests have trouble identifying changes between components that cause integration issues.

After a second thought, I think we should get input from @Rikkola as I believe he added the `onChange` method originally.

mbarkley:
The build failure is from an indexing test so I am considering this ready to merge.

PR #1424, opened 2018-02-14
DROOLS-2272: Increase coverage

jomarko:
Dependent on #1422 

@manstis please have a look, I would like to merge this when your PR is in.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/undefined/UndefinedExpressionGrid.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/undefined/UndefinedExpressionGrid.java
@@ -167,18 +167,20 @@ public class UndefinedExpressionGrid extends BaseExpressionGrid<Expression, Unde
 
     @Override
     public void onItemSelected(final ListSelectorItem item) {
-        final ListSelectorExpressionTypeItem ei = (ListSelectorExpressionTypeItem) item;
-        expressionEditorDefinitionsSupplier
-                .get()
-                .stream()
-                .filter(definition -> definition.getModelClass().isPresent())
-                .map(ExpressionEditorDefinition::getType)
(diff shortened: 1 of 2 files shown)


manstis:
@jomarko You can rebase now.. changes look OK though.

jomarko:
Dear jenkins ping me or michael when you are done with tests :coffee:

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/undefined/UndefinedExpressionGrid.java
```diff
@@ -167,18 +167,20 @@ public boolean isEnabled() {
...
-                .stream()
-                .filter(definition -> definition.getModelClass().isPresent())
-                .map(ExpressionEditorDefinition::getType)
-                .filter(type -> type.equals(ei.getExpressionType()))
-                .findFirst()
-                .ifPresent(type -> {
-                    cellEditorControls.hide();
-                    onExpressionTypeChanged(type);
-                });
+        if (item instanceof ListSelectorExpressionTypeItem) {
```
manstis: This type check isn't strictly necessary as **I know** all instances will be of type ```ListSelectorExpressionTypeItem``` from [here](https://github.com/jomarko/kie-wb-common/blob/fb8e14c28c017374d18d1277fa63e5b62fb50ba5/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/undefined/UndefinedExpressionGrid.java#L149) but then again.. the check does not harm either!

jomarko: Thanks for pointing this method, actually I was thinking if the instanceof check is needed or not, however I didn't find the method you pointed me to, so I decided to add the check, as there are multiple implementations of the `ListSelectorItem` in our codebase.

jomarko:
@manstis green

PR #1427, opened 2018-02-14
AF-1010: [Project Oriented] Library tries to close all places when cr…

mbarkley:
…eating new DRL file (showing unsaved popup)


mbarkley:
TBH I'm not sure why we have to close places at all, but it seems clear that this condition was inverted based on the behaviour in `goToProject(WorkspaceProject)` in the same class.

ederign:
jenkins retest this

mbarkley:
@paulovmr Thank you for the explanation 👍

PR #1436, opened 2018-02-19
Cover parent column width update

jomarko:
@manstis please have a look. Some coverage increase.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGridGeneralTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGridGeneralTest.java
@@ -17,11 +17,15 @@
 package org.kie.workbench.common.dmn.client.widgets.grid;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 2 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGridGeneralTest.java
```diff
@@ -298,6 +303,63 @@ public void testPaddingWithNoParent() {
...
+                .mapToObj(width -> mockColumn(width, grid))
+                .collect(Collectors.toList());
+        grid.getModel().appendRow(new DMNGridRow());
+        columns.stream().forEach(column -> grid.getModel().appendColumn(column));
+
+        // force the peers width update
+        columns.get(indexOfColumnToUpdate).updateWidthOfPeers();
+
+        // assert parent width is equal to sum of nested columns widths
+        Assertions.assertThat(parentColumn.getWidth()).isEqualTo(Arrays.stream(widthsOfNestedColumns).sum());
```
manstis: I think if a nested grid has padding the parent column width would be wider than the sum of the nested columns.

Is it worth doing a test for this scenario?

jomarko: You are completely right, unfortunately I was not able to mock in that way the paddings was not zero, but I will give one more try, I think it is worth to test in the way you described.

jomarko:
@manstis please, would you mind to have a look on the new version of the test?

PR #1438, opened 2018-02-20
JBPM-6809: Stunner - Make the default canvas size bigger

alepintus:
Hi @romartin @hasys ,
changed the default size of the canvas to 2800 x 1400.
Could you please take a look?

Thanks.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/factory/BPMNGraphFactory.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/factory/BPMNGraphFactory.java
@@ -25,6 +25,6 @@ import org.kie.workbench.common.stunner.core.factory.graph.GraphFactory;
 public interface BPMNGraphFactory extends BPMNFactory,
                                           GraphFactory {
 
-    double GRAPH_DEFAULT_WIDTH = 1400d;
-    double GRAPH_DEFAULT_HEIGHT = 700d;
+    double GRAPH_DEFAULT_WIDTH = 2800d;
+    double GRAPH_DEFAULT_HEIGHT = 1400d;
 }


romartin:
ok to test

romartin:

hasys:
Hi @alepintus, @romartin, 

I guess it is a question to @ibek is this size OK or it's better to make it bigger.

ibek:
Could we make the resize dynamic so that at the border there would be arrow buttons which would enlarge the canvas size? First click would make it 2800x1400 second one 3600x2000, and similarly?

romartin:
Hey @ibek 
Well this ticket is considered just a quick win for the next release until we do the right updates. The 
 approach we already discussed is about having an "infinite" canvas, like the decision tables are doing, but that 's something to address after the GA, it already exist the [JBPM-5536](https://issues.jboss.org/browse/JBPM-5536) for it.  So yeah I know that's not the ideal solution but for now that's exactly what we're expecting... I mean not spending time on this (yet).

ibek:
ok in that case please make it 4 times bigger than what we have now: 5600x2800 or just 2 times bigger 2800x1400 but with 2 times smaller nodes which could be zoomed in if that's an option.

romartin:
**@ibek**  please can you add a comment in the ticket about this ([JBPM-6809](https://issues.jboss.org/browse/JBPM-6809))? I mean this way UX can confirm the right values as well, in fact the values were proposed by them, so probably better double checking.

**@alepintus** once the final decision is taken (in the ticket comments) and UX is agree, please just update the commit.
Thanks guys!!

manstis:
Check browser memory usage... I could be completely wrong but I suspect the browser will reserve memory for the whole canvas.. 4x bigger, potentially 4x memory use.

jomarko:
I tried locally and unfortunately can confirm Michael's concern, the size 5600x2800 cause designer is extremely slow.

romartin:
oh @manstis really good point!! I forgot this for a while, thanks for letting us know!! :+1:

jomarko:
Please let us know which size do we agreed on and we should try before merging.

romartin:
hey @jomarko @hasys @alepintus what's the status on this? I think that finally the size actually proposed in this commit fine, due to comments on the ticket, so it's just missing the QE review? or are you guys discussing yet the sizes?

romartin:
hey @alepintus @hasys @jomarko 
I forgot that the idea is using the workbench preferences in order to specify the canvas size, with some default values, thanks for the reminder @alepintus ! 
So let's wait a few more to check this PR and see if we can do it via preferences...

romartin:
hey @alepintus @hasys @jomarko 
After reviewing this with @krisv it really does not makes sense at this point implementing the canvas size configurable via preferences, as it will require a few time to be implemented and also we're going to use an "infinite" canvas approach soon, once we have time for it. 
So IMO let's double the canvas size as @alepintus has done here, I've tested it and works fine at performance level, and this exactly what we're expecting.. a quick win, just updating the values to obtain acceptable results on the UI, but not producing costs on the implementation side.
I'm agree on merging this commit as it is, please let us know.
Thanks!

PR #1446, opened 2018-02-21
Do not Merge [AF-1022]: Fixed double delete message issue

adrielparedes:
Backend fires the Deletion event notification, that is why it appeared two times. Fixed.

@Rikkola can you review the PR please?


adrielparedes:
@Rikkola I solved this issue in AF-913 with the new Project View I think we should check some of this issues before continue

PR #1439, opened 2018-02-20
DROOLS-2332: [DMN Editor] Performance poor for lots of nested tables

manstis:
See https://issues.jboss.org/browse/DROOLS-2332

This PR removes the need to "walk" the (nested) data model of parent-child grids when looking for a "parent grid".

PR #1448, opened 2018-02-22
DROOLS-2332: Increase coverage and introduce util method to reduce codebase

jomarko:
Dependent on #1439 

@manstis please have a look, I will rebase once your PR is in.


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/util/CommandUtils.java
```diff
@@ -93,4 +94,10 @@ public static void updateParentInformation(final GridData uiModel) {
             }
         });
     }
+
+    public static GridCellValue<?> extractGridCellValue(final GridCellTuple cellTuple) {
```
manstis: Great improvement thanks... I'll ask however... any reason you didn't choose to return ```Optional<GridCellValue<?>>``` and remove the ```Optional.ofNullable(..)``` in the commands? Only a question.. happy "as is".. it's a nice improvement.

jomarko: Because I was not thinking enough :D Will apply definitely good idea.

jomarko:
@manstis updates done.

jomarko:
Rebased, lets wait for tests results.

jomarko:
@manstis seems we can merge

PR #1457, opened 2018-02-26
DROOLS-2330: Cover ghost render for Literal and Undefined expression grids

jomarko:
Dependent on #1449 

@manstis please have a look, I will rebase once your PR is in.


jomarko:
@manstis I think you can re review now. Thanks in advance.

PR #1472, opened 2018-03-02
JBPM-6873: Stunner - None Start Event has dashed border instead of solid one

wmedvede:


wmedvede:
Hello @romartin @hasys would you mind review this PR?
Thanks

hasys:

hasys:
Hi @wmedvede,

I found an issue in our backlog [JBPM-6813](https://issues.jboss.org/browse/JBPM-6813) and have a question now: 

Does it make sense to make IsInterrupting property always visible for now within the current PR?

hasys:
@wmedvede, @romartin,

I mean if I correct in this PR @wmedvede added IsInterrupting property, but it is not visible now even inside of Event Subprocess.
image

For now we can make this property visible always as it's done in current designer and use alredy reported jira as improvements for the future. WDYT, @wmedvede, @romartin?

romartin:
Hey @hasys ,
Yeah well in fact next task for @alepintus will be working on the is interrupting property for events. It behaves a bit different, so that's already identified and will be soon there, feel free to add comments in the ticket so @alepintus can address them. 
So if that's good for you, I'll merge this, lemme know.
Thanks guys!

wmedvede:
@romartin @hasys 
yes, the isInterrupting property is not visible for the StartNoneEvent. Let's say that it's more for fixing the issue but also for compliance with the org.eclipse.bpmn2.StartEvent. 
For the particular case of the EventSubprocess I believe that what needs to be done here is to adjust the Rules, because AFAIK the EventSubprocess can only be started with events. i.e., I believe that it shouldn't be possible to put StartNoneEvent inside the EventSubprocess

PR #1474, opened 2018-03-05
JBPM-5509: Stunner - Error message in Server Log after Business Process creation

wmedvede:


wmedvede:
Hello @romartin @hasys, would you mind review this PR?
Thanks.

romartin:

romartin:
@manstis @jomarko adding you to review just to double check plz. I just copy-pasted this `UUID` generator class when starting in Stunner from Wires, not sure if we can feel free to change this or there is some point we're missing... 
The goal is to avoid the BPMN parsers to show warnings about the ids generated, it seems there are some constraints there.. so adding this change here will update the ids for DMN as well, hope no implications... is that good for you guys?
thanks!

hasys:
Hi @wmedvede,

I am still can see IDs like "_qR-_PCLCEeiPz5TLmOC29w". It looks like Stunner using at least two different ID generators. Is it possible to use just one instead of two?

romartin:
Hey @wmedvede , this PR still needs love, isn't it? :)

wmedvede:
@romartin 
Well, after wasting considerably time on this, here are my findings.
When the process is created, Stunner creates it's Graph model. This is fine, and the Process elements identifiers are created by Stunner.
BUT Sunner just creates the Grahp model, and not ALL the xml elements that will be finally saved in the resulting bpm2 file.

All the stuff, tipically located at the end of the bpmn2 file (see bellow) is created at the parser/marshaller. Of course some information is
taken from the nodes in the Graph, but not all. And in particular the identifiers for this "additional" xml elements are automatically generated
by Eclipse classes when the xml with the bpmn2 model is saved. This Eclipse classes also manages and waranty the cross reference integrity between this "additional" xml elements. 
Unfortunatelly identifiers generated by eclipse has a slightly different format from the ones generated by Stunner. BUT, both identifiers are valid.
This has been this way also in the jBPM designer since the begininig of time. i.e., identifiers generated by jBPM had a different format from the Eclipse generated. (and probalby was a bit worst than Stunner too)

  <bpmndi:BPMNDiagram id="_hJw9QUR-EeiFdZOB6a5cgQ">
	//digarm related stuff
  </bpmndi:BPMNDiagram>
  <bpmn2:relationship id="_hJw9RER-EeiFdZOB6a5cgQ" type="BPSimData">
	//simulation information stuff
  </bpmn2:relationship>

So to summarize my suggestion is to not waste more time on this. Note that this identifiers are completelly internal, and we have also millions of higher priorities tasks to do.
We could of course try to copy paste the Eclipse UUID generator class code, and try to make it GWT client side compatible code, and try to use this in Stunner, but I don't recomend it

ps. was rebased with master

romartin:

romartin:
I would vote for cherry-picking this to `7.7.x` too!

hasys:
@romartin agreed, and we already have a RHPAM issue for it: https://issues.jboss.org/browse/RHPAM-436

PR #1479, opened 2018-03-06
AF-985+: Fix duplicate loading of assets when page is updated rapidly.

mbarkley:
Note that because of other changes in master, the loading popup prevents you from easily reproducing the original issue.

The easiest way I found to test this is open the chrome inspector while the asset list is loading and hide the overlay. Then you can rapidly click the project breadcrumb.


mbarkley:
Jenkins, please retest this.

PR #1481, opened 2018-03-07
DROOLS-2369: Refactor use of class ListSelector to use interface ListSelectorView.Presenter

manstis:
See https://issues.jboss.org/browse/DROOLS-2369


manstis:
@jomarko This one is ready to review... it's a search/replace exercise.. no other changes.

PR #1486, opened 2018-03-08
DROOLS-2374: [DMN Designer] Literal Expression Cell can not be updated once erased

manstis:
See https://issues.jboss.org/browse/DROOLS-2374

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/relation/RelationColumn.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/relation/RelationColumn.java
@@ -29,7 +29,6 @@ import org.uberfire.ext.wires.core.grids.client.model.GridCell;
 import org.uberfire.ext.wires.core.grids.client.model.GridCellValue;
 import org.uberfire.ext.wires.core.grids.client.model.GridData;
 import org.uberfire.ext.wires.core.grids.client.model.GridRow;
-import org.uberfire.ext.wires.core.grids.client.model.impl.BaseGridCell;
 import org.uberfire.ext.wires.core.grids.client.model.impl.BaseGridCellValue;
 import org.uberfire.ext.wires.core.grids.client.widget.context.GridBodyCellRenderContext;
 import org.uberfire.ext.wires.core.grids.client.widget.dom.HasDOMElementResources;
@@ -95,15 +94,13 @@ public class RelationColumn extends DMNGridColumn<RelationGrid, String> implemen
                      final GridBodyCellRenderContext context,
                      final Callback<GridCellValue<String>> callback) {
         factory.attachDomElement(context,
-                                 (e) -> e.getWidget().setValue(assertCell(cell).getValue().getValue()),
+                                 (e) -> e.getWidget().setValue(assertCellValue(assertCell(cell).getValue()).getValue()),
                                  (e) -> e.getWidget().setFocus(true));
     }
 
-    private GridCell<String> assertCell(final GridCell<String> cell) {
-        if (cell != null && cell.getValue() != null && cell.getValue().getValue() != null) {
-            return cell;
-        }
-        return new BaseGridCell<>(new BaseGridCellValue<>(""));
+    @Override
+    protected GridCellValue<String> makeDefaultCellValue() {
+        return new BaseGridCellValue<>("");
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/relation/RelationColumn.java
```diff
@@ -95,15 +94,13 @@ public void edit(final GridCell<String> cell,
                      final GridBodyCellRenderContext context,
                      final Callback<GridCellValue<String>> callback) {
         factory.attachDomElement(context,
-                                 (e) -> e.getWidget().setValue(assertCell(cell).getValue().getValue()),
+                                 (e) -> e.getWidget().setValue(assertCellValue(assertCell(cell).getValue()).getValue()),
                                  (e) -> e.getWidget().setFocus(true));
     }
 
-    private GridCell<String> assertCell(final GridCell<String> cell) {
-        if (cell != null && cell.getValue() != null && cell.getValue().getValue() != null) {
```
jomarko: We have replaced this line with:
```java
(e) -> e.getWidget().setValue(assertCellValue(assertCell(cell).getValue()).getValue())
```
The new implementation doesn't check if `cell.getValue().getValue() != null`. Is that correct?

In other words, if I add the test below into `BaseDOMElementSingletonColumnTest` the test will fail.
```java
@Test
    public void checkEditNullCellValueValue() {
        final GridCell<String> cell = new BaseGridCell<>(new BaseGridCellValue<>(null));

        column.edit(cell,
                    context,
                    result -> {/*Nothing*/});

        assertCellEdit(DEFAULT_VALUE);
    }
```

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/model/DMNGridColumn.java
```diff
@@ -51,6 +54,24 @@ public DMNGridColumn(final List<HeaderMetaData> headerMetaData,
...
+
+    protected GridCellValue<T> assertCellValue(final GridCellValue<T> cellValue) {
+        if (cellValue != null) {
+            return cellValue;
+        }
+        return makeDefaultCellValue();
+    }
+
+    protected GridCellValue<T> makeDefaultCellValue() {
+        return null;
```
jomarko: Question here, wouldn't be better to throw an exception or make the method abstract to force user override appropriately in `DMNGridColumn` implemetations?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/relation/RelationColumn.java
manstis: Good spot! Added check and your test. Thanks.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/model/DMNGridColumn.java
manstis: I've moved the assertions to a new sub-class ```DMNSimpleGridColumn``` that is the base for all "simple" columns that use ```TextBox``` or ```TextArea```. The only columns not extending this are ```ExpressionEditorColumn```, ```UndefinedExpressionColumn``` (both of which provide a custom widget for editing the cell value) and ```function.supplementary.NameColumn``` that is not editable. There is now no need to either have all columns implement ```makeDefaultCellValue()``` or throw an exception.

manstis:
Jenkins please retest this.

manstis:
```
org.kie.workbench.common.services.backend.compiler.plugin.KieMetadataTest.compileAndLoadKieJarMetadataAllResourcesPackagedJar

Failing for the past 11 builds (Since Unstable#3113 )
Took 51 sec.
Stacktrace
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.kie.workbench.common.services.backend.compiler.plugin.KieMetadataTest.compileAndLoadKieJarMetadataAllResourcesPackagedJar(KieMetadataTest.java:109)
```

manstis:
@ederign Test ```KieMetadataTest.compileAndLoadKieJarMetadataAllResourcesPackagedJar()``` has been failing for the last 11 builds. Could you please take a look (you made the last change to the class by the look of it).

ederign:
@manstis I just touched this in project oriented refactoring! :-)

But anyway: https://github.com/kiegroup/kie-wb-common/pull/1498

manstis:
Oh sorry @ederign but it did look like you knew why it was failing; so I thank-you :-)

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

PR #1487, opened 2018-03-08
Fail creating/validating server template

ederign:
@cristianonicolai I don't know if that is the best solution because I basically use a catch as if 😰, but I don't know if it's ok to change this method:
https://github.com/kiegroup/droolsjbpm-integration/blob/c0f7b6a3318140d6c022b0e1756224af5f252c66/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/service/SpecManagementServiceImpl.java#L148-L150


cristianonicolai:
@ederign yeah but I think this is the best way to go.

PR #1490, opened 2018-03-09
DROOLS-2377 More tests for various DMN elements marshalling

kurobako:
https://issues.jboss.org/browse/DROOLS-2377

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ImportConverter.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ImportConverter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.dmn.backend.definition.v1_1;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.kie.workbench.common.dmn.api.definition.v1_1.Import;
+import org.kie.workbench.common.dmn.api.property.dmn.LocationURI;
+import org.kie.workbench.common.dmn.api.property.dmn.QName;
+
+public final class ImportConverter {
+
+  public static Import nodeFromDMN(org.kie.dmn.model.v1_1.Import source) {
+    LocationURI locationURI = new LocationURI(source.getLocationURI());
+    Import result = new Import(source.getNamespace(), locationURI, source.getImportType());
+    Map<QName, String> additionalAttributes = new HashMap<>();
+    for (Map.Entry<javax.xml.namespace.QName, String> entry : source.getAdditionalAttributes().entrySet()) {
+      additionalAttributes.put(QNamePropertyConverter.wbFromDMN(entry.getKey()), entry.getValue());
+    }
+    result.setAdditionalAttributes(additionalAttributes);
+    return result;
+  }
+
+  public static org.kie.dmn.model.v1_1.Import dmnFromNode(Import source) {
+    org.kie.dmn.model.v1_1.Import result = new org.kie.dmn.model.v1_1.Import();
+    result.setImportType(source.getImportType());
+    result.setLocationURI(source.getLocationURI().getValue());
+    result.setNamespace(source.getNamespace());
+    Map<javax.xml.namespace.QName, String> additionalAttributes = new HashMap<>();
+    for (Map.Entry<QName, String> entry : source.getAdditionalAttributes().entrySet()) {
+      additionalAttributes.put(QNamePropertyConverter.dmnFromWB(entry.getKey()).get(), entry.getValue());
+    }
+    result.setAdditionalAttributes(additionalAttributes);
+    return result;
+  }
+}
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
@@ -18,12 +18,17 @@ package org.kie.workbench.common.dmn.backend;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.StringReader;
 import java.math.BigDecimal;
+import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Scanner;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.function.Consumer;
@@ -31,6 +36,11 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import javax.enterprise.inject.spi.BeanManager;
+import javax.xml.namespace.NamespaceContext;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
 
 import org.apache.tools.ant.filters.StringInputStream;
 import org.jboss.errai.marshalling.server.MappingContextSingleton;
@@ -105,6 +115,7 @@ import org.kie.workbench.common.stunner.core.util.DefinitionUtils;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 import org.uberfire.commons.uuid.UUID;
+import org.xml.sax.InputSource;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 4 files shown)


manstis:
Jenkins please retest this.

manstis:
@tarilabs Would you mind reviewing? You know the limitations of the marshaller and what other DMN constructs we need to test for (and possibly implement for). Thanks.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<semantic:definitions xmlns:semantic="http://www.omg.org/spec/DMN/20151101/dmn.xsd" xmlns="http://www.trisotech.com/dmn/definitions/_9cd3704d-34c3-44cd-887a-f7be50b997af" xmlns:feel="http://www.omg.org/spec/FEEL/20140401" xmlns:tc="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:triso="http://www.trisotech.com/2015/triso/modeling" xmlns:trisofeed="http://trisotech.com/feed" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" exporter="DMN Modeler" exporterVersion="6.0.3.2" id="_9cd3704d-34c3-44cd-887a-f7be50b997af" name="dummy" namespace="http://www.trisotech.com/dmn/definitions/_9cd3704d-34c3-44cd-887a-f7be50b997af" triso:logoChoice="Default">
+    <semantic:extensionElements/>
+    <semantic:import namespace="http://example.com" importType="http://example.com"/>
```
tarilabs: I would refrain from using `import` as it has semantics --it would try to import the other file.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
```diff
@@ -0,0 +1,23 @@
...
+    <semantic:elementCollection name="alpha" id="alpha">
+        <semantic:drgElement href="http://example.com"/>
```
tarilabs: I believe this has the semantic for the DRD, so unless it is implemented for the editor, I believe is better to avoid to introduce beforehand the `elementCollection`.
Btw this `elementCollection` element has no influence for runtime.
/cc @etirelli to confirm

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
```diff
@@ -0,0 +1,23 @@
...
+    </semantic:elementCollection>
+    <semantic:performanceIndicator name="beta" id="beta">
+        <semantic:impactingDecision href="http://example.com"/>
```
tarilabs: From a runtime PoV, this `performanceIndicator` has no execution semantic, but the actual content does not link to a Decision. I don't know if this will influence the editor later.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
```diff
@@ -0,0 +1,23 @@
...
+    </semantic:performanceIndicator>
+    <semantic:organizationUnit name="gamma" id="gamma">
+        <semantic:decisionMade href="http://example.com" />
+        <semantic:decisionOwned href="http://example.com" />
```
tarilabs: analogous as above.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
etirelli: it depends, if this is a marshalling-only test, I think it would be good to add import in there to make sure it is working, but we need the additional "name" attribute in there. E.g.:
```
  <semantic:import xmlns:drools="http://www.drools.org/kie/dmn/1.1"
        drools:name="Imported Model" importType="http://www.omg.org/spec/DMN1-2Alpha/20160929/MODEL"
        namespace="http://www.trisotech.com/dmn/definitions/_f27bb64b-6fc7-4e1f-9848-11ba35e0df36"/>
```

 The runtime semantics are not implemented yet though.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
etirelli: I am not sure about this one... on one side we have no use for them, but if we are importing a model created by someone else, we might need to keep consistency of this element. If it is easy to add support for it, then go ahead and do it. Otherwise postpone.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
etirelli: If it is easy to add support for it, do it, otherwise postpone.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
etirelli: If it is easy to add support for it, do it, otherwise postpone.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn
tarilabs: Edson, usually for the `DMNMarshallerTest` which this PR is about, we have a marshalling roundtrip and then we then compile and execute the model, to be sure the result out of the marshalling roundtrip did non wrongly impacted the semantics of the model.

kurobako:
@etirelli @tarilabs @baldimir updated my PR with detailed tests

Commented out lines are for elements which are currently removed by marshaller. I understand that marshaller does not need to preserve 100% of the document's structure (and that is the reason why I found assertJ capabilities unpractical after all) but let's discuss this and modify this PR once again based on that.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -1005,4 +1016,92 @@ public void test_wrong_context() throws IOException {
...
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:performanceIndicator/semantic:impactingDecision/@href"), original, roundtripped);
+
+        //assertXPathEquals(xpath.compile("boolean(/semantic:definitions/semantic:organizationUnit)"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:organizationUnit/@name"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:organizationUnit/@id"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("boolean(/semantic:definitions/semantic:organizationUnit/semantic:decisionMade)"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:organizationUnit/semantic:decisionMade/@href"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("boolean(/semantic:definitions/semantic:organizationUnit/semantic:decisionOwned)"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:organizationUnit/semantic:decisionOwned/@href"), original, roundtripped);
```
jomarko: @kurobako hello, do we need to keep these commented lines?

kurobako: this is what I asked in the last comment

tarilabs: if @manstis or @jomarko confirms these elements will never be required by the authoring editor, I believe we can simply skip.
If these elements will be managed eventually by the authoring editor, I believe we should implement the converter hence these need to result in being effective and not commented.
If we don't know yet about these elements, I believe we can leave as-commented.

kurobako: @tarilabs totally agree. Let's see what Michael and Jozef say.

manstis: @etirelli This is the PR I mentioned in the meeting.

IMO all elements will eventually be supported... note the "eventually" bit ;-)

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -1005,4 +1016,92 @@ public void test_wrong_context() throws IOException {
...
+        diagram.setGraph(marshaller.unmarshall(null, getClass().getResourceAsStream("/dummy.dmn")));
+        String roundtripped = marshaller.marshall(diagram);
+        XPath xpath = namespaceAwareXPath("semantic", "http://www.omg.org/spec/DMN/20151101/dmn.xsd");
+        assertXPathEquals(xpath.compile("boolean(/semantic:definitions/semantic:extensionElements)"), original, roundtripped);
+
+        //assertXPathEquals(xpath.compile("boolean(/semantic:definitions/semantic:import)"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:import/@namespace"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:import/@importType"), original, roundtripped);
+        //assertXPathEquals(xpath.compile("/semantic:definitions/semantic:import/@locationURI"), original, roundtripped);
+
```
etirelli: /definitions/import is something we need to support from the first version, so this needs to be working.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
etirelli: Support for "import" is necessary. The others, we can postpone if it requires too much extra work. If it is just a matter of copy/paste for now, then we can just support it in the marshaller and ignore it in the UI for the first version.

manstis:
@etirelli I'd like to understand how "import" will work at the UI level. From the DMN Specification I understand it allows the import of types, decisions, BKM etc from external sources. Presumably the backend (marshaller) would populate a UI DMN object graph including "imported" items; however I also assume such items would be read-only in the UI. If I understand correctly we'd need some form of classifier/flag on an "imported" item to drive the UI or some other way of aascertaining whether an item is "locked".

etirelli:
Good points Michael, I don't have a definitive answer for you, so we need UX involvement on how to present this in the UI. But you are correct in:
* "imported" items would be read-only in the "importing" model
* they need to be available to use in the "importing" model. E.g., if a decision invokes and "imported" BKM, that BKM node has to be added as a knowledge requirement of the decision, and show up in the DRD. How? I don't know: a palete "library" where you can drag imported nodes from? a right click menu item that opens a dialog for you to select which imported node do you need? some other way?
* The standard itself is very poor on the "import" definition, so it is up to us to define how to graphically represent a "read-only imported node". An import icon? a lock icon?
* It would be great if the user would be able to navigate from the importing model directly to the imported model. Maybe a context menu (right click) option?

kurobako:
@tarilabs updated my PR, please take a look when you have a moment
Since drools:name is not guaranteed to be present by the XSD, I employed a pair of namespace and URI as fallback strategy for generating node uuid if name is unavailable, though it seems to not matter really
Also added tests to make sure additional attributes are preserved

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ImportConverter.java
```diff
@@ -0,0 +1,48 @@
...
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+import org.kie.workbench.common.stunner.core.graph.content.view.ViewImpl;
+import org.kie.workbench.common.stunner.core.graph.impl.NodeImpl;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+public final class ImportConverter implements NodeConverter<org.kie.dmn.model.v1_1.Import, org.kie.workbench.common.dmn.api.definition.v1_1.Import>{
```
tarilabs: I believe here is the culprit: I see you are implementing `NodeConverter` but is this required?
Wouldn't be easier to do something just as a plain converter, for example https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ItemDefinitionPropertyConverter.java#L26 ?
Does it need to convert to a Node? afaik currently import is used just as a plain Pojo from the UI/authoring

tarilabs:
@kurobako concerning:
> Since drools:name is not guaranteed to be present by the XSD

If you need a QName reference you can use here: https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-model/src/main/java/org/kie/dmn/model/v1_1/Import.java#L22
as we are doing analogously for other things which are not originally covered by the XSD.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ImportConverter.java
tarilabs: /cc @manstis could you kindly advise, following this PR discussion, on the UI/authoring functionality if the DMN `Import` element needs to be converted to _its own node_, or is just sufficient for it be available inside the Definitions' "main" node graph, please?

tarilabs: (my understanding was defaulting for it NOT to need its own node)

kurobako: @tarilabs regarding node/plain converter: well, I do not know. I followed the example and it worked. If we are gonna create edges for imported files, though, we might have to deal with nodes anyway.
Are you concerned with simplicity only or there is any other downside of implementing NodeConverter?

kurobako:
[ERROR] src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ImportConverter.java:[11] (imports) CustomImportOrder: Import statement for 'java.util.HashMap' is in the wrong order. Should be in the 'STANDARD_JAVA_PACKAGE' group, expecting not assigned imports on this line.
[ERROR] src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ImportConverter.java:[12] (imports) CustomImportOrder: Import statement for 'java.util.Map' is in the wrong order. Should be in the 'STANDARD_JAVA_PACKAGE' group, expecting not assigned imports on this line.

you must be kidding me...

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -296,11 +299,14 @@ public Graph unmarshall(final Metadata metadata,
...
         }
 
+        List<Node<View<org.kie.workbench.common.dmn.api.definition.v1_1.Import>, ?>> imports = dmnXml.getImport().stream().map(importConverter::nodeFromDMN).collect(Collectors.toList());
+
         Graph graph = factoryManager.newDiagram("prova",
                                                 BindableAdapterUtils.getDefinitionSetId(DMNDefinitionSet.class),
                                                 metadata).getGraph();
         elems.values().stream().map(kv -> kv.getValue()).forEach(graph::addNode);
         textAnnotations.values().forEach(graph::addNode);
+        imports.forEach(graph::addNode);
```
manstis: @kurobako @tarilabs @etirelli I am happy to be educated; but is there a difference between an "import" being something that **could** be added to the DRD vs an "import" being something that **is** added to the DRD? The analogy being a Java import. Just because I've imported a class does not mean it is used.

From an authoring/UI stance; I'd need to know what items **could** be imported so I can offer them to the User to select from and then add **to** the DRD. In my mind a DMN "import" was that something **could** be used; not **is** used. 

If this assumption is true then I do not think we need to add all imports to the UI graph; only when we detect an import is **used** in the DRD do we need to add it. If my assumption is completely wrong; then does anybody know how we can show to Users what **could** be added but as yet has not **been** added?

Depending on the answer(s) to this question do I think we need ```ImportConverter``` to implement ```NodeConverter```. Otherwise, as @tarilabs says, it could simply contain some static methods to handle conversion and add the DMN/kie ```Import`` instance to the UI model.

Looking at https://github.com/kurobako/kie-wb-common/blob/4d6ff02656ae927dac58d554f09cad104f3bcdc3/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/resources/dummy.dmn I did not see how the "import" is effectively consumed.

Sorry if these questions are blindingly obvious to those more knowledgeable about the DMN specification that I.

jomarko:
@kurobako I believe the import error will gone if you apply kiegroup codestyle. 

https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration

etirelli:
@manstis your assumption is correct that a DMN import works similar to Java import a.b.c.* in the sense that importing a namespace will give access to the nodes within that namespace, but they don't need to be used.
Also, imports don't have a UI representation (they are not a node or anything), but I believe the UI needs that information to access nodes in the imported namespace.
Finally, imports in DMN define an alias (the "name" attribute) to the namespace and in the model that alias is used to refer to nodes in the imported namespace. E.g.:

<import namespace="http://some.name.space/foo" drools:name="bar"/>

For the rest of the model, all types defined in that imported namespace will be prefixed by "bar.". All invocations to BKMs in that namespace will also use "bar." prefix.

So, while there is no UI representation of the import itself, I believe you need the import information in the UI to allow the user to view/edit it.

Let me know if this is unclear and we can have a quick chat to discuss.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
kurobako: @manstis it seems there is enough confusion about this case already, so I will keep my two cents and let Edson or Matteo respond; just one thing: if you need to access content that is being imported (like give users a preview of what's inside) but not necessarily used this PR needs to be modified.
For the time being only information present in the graph nodes is URI of the file imported and such; what can be done *if needed* is to build a graph of the imported file's contents without linking it to the outside nodes but let's first specify what to implement exactly.

manstis:
@etirelli Clear as mud ;-) Perhaps we can chat at sometime.

@kurobako Thank-you for your comments. As my understanding is cloudy; I'm happy to leave things as they are. We may need to change things in the future, when I am back on DMN authoring. At the moment, if a Node is added to the graph it will be visible to Users as a node in the DRD - even if not connected to another concrete Node that was not imported.

kurobako:
@manstis thanks for the reply, Michael. So I guess we need another way to pass additional information about imports in metadata instead of graph to not force import nodes being shows as part of it.
I checked Metadata and see no suitable methods there. @etirelli @tarilabs do you have a suggestion here maybe? Should I add something like addImport() to Metadata? Michael, would that work for you? 
I would like to avoid rewriting my PR more times than necessary, so let's discuss API first and then do the implementation.

tarilabs:
> do you have a suggestion here maybe? 
I thought based on latest feedback from @manstis and confirmation from @etirelli about the semantic, this could have been done indeed without the need of implementing the NodeConverter interface and simply use the static-method based converters instead?

tarilabs:
i.e.: the static method converter would be called inside the `Definitions` converter, similarly to other elements. This would ensure data is converted to UI-based pojos, without the need for it being converted to a Node.

If the example is not clear I'm always available for a call if preferred.

kurobako:
@tarilabs that's a given. the problem is I do not see where could I put marshalled import node. we pass basically two things to UI (based on code I see): graph and metadata.graph won't work because eveything there is automatically shown. metadata has no suitable methods and that's why I asked in the previous comment if you are okay with me adding methods there to save import information.

tarilabs:
> that's why I asked in the previous comment if you are okay with me adding methods there to save import information.

I am sorry but I think we are still talking of 2 separate things.

With my suggestion there is no need to add methods, unless I missed something.

Let's work another example, by analogy.

DMN `Description` has no UI Node conversion.
It is converted inside the Definitions converters here:
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DefinitionsConverter.java#L36

Analogous things happen for ItemDefinitions, here:
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DefinitionsConverter.java#L45

in both cases there is no need for node conversion, and is simply added to their expected parent.

similarly, a
DMN Import would need its own ImportPropertyConverter
this is called again inside Definitions converter
The result of Import conversion to UI based pojo is then added to Definitions.getImport().add(...);
in line similarly to linked example

If this suggestion is still not clear I strongly suggest a call and a screenshare of your IDE, please?

kurobako:
@tarilabs ah, I think I know what you mean. I will try it in a while. I have to set up my project from scratch again because I updated IDEA and I thus I have to reimport modules from kie-wb-common since whole project just wrecks its brain. I will probably update this PR tomorrow morning. Thanks.

manstis:
@tarilabs @kurobako OK, so IMO we should just use static converter methods for imports; so the UI model has the same imports as the KIE model. This would mean whatever is in the DMN model is the same (i.e. the User has imported something and added it to the DRD). 

How we get "what is available for import" to the UI model is one of the things us UI guys have to do ;-)

The graph is in fact part of the content we send to the UI. What we'll need to do is also include objects relating to "what can be imported" and then use this to construct whatever UI components that the User will interact with to add imported things to the DRD.

For now (for the purpose of merging this PR) I think we need not worry about what I need to do to allow the User to select from "what can be imported" and then add imports and nodes to the DRD; that'd be marshalled back from the UI DMN model to the KIE DMN model to DMN XML.

kurobako:
@tarilabs updated my PR
@manstis so now import elements are simply added to Definitions. I expect this PR to be merged now but please let me if you'll need to update marshalling in the future; I invested quite a bit of time understanding this part of a codebase and would be glad to work on it more if such need arises

kurobako:
12:57:22 Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
12:57:22 /home/jenkins/workspace/kie-wb-common-pullrequests/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/service/SpecManagementServiceCDI.java:[24,37] cannot find symbol
12:57:22   symbol:   class KieServerControllerNotFoundException

Doesn't look like it has anything to do with my change. Did someone ruin upstream?

mvn clean install for kie-wb-common-dmn works fine for me locally.

kurobako:
jenkins retest this

kurobako:
jenkins retest this

kurobako:
jenkins retest this

kurobako:
so, yep, it wasn't this PR fault after all. 
@etirelli can we merge this now?

etirelli:
@kurobako if you and @tarilabs  are in agreement, go ahead.

tarilabs:
Yes I ticked it https://github.com/kiegroup/kie-wb-common/pull/1490#pullrequestreview-112352250
but I have no merge rights, especially on UI stuff ;)
@manstis can you kindly review and merge eventually please?

PR #1492, opened 2018-03-09
BAPL-764-migration-tool: Provide migration path from old Form Modeler to the new one

pefernan:
This PR adds the migration tool to move old forms to new format. This first PR only copies fields, layout & model settings. Advanced optiosn as formulas, custom layouts, JS callbacks are left on the old form until they are supported on the new forms engine.

This PR also adds a generic Migration Tool that allows to plug other migration tools via SPI and let's the user switch running each plugged migration tool individually or in batch mode.

Since there are a lot of changes. I'd like to split the review:

@Rikkola could you please take a look at the changes for the project migration tool?
@kkufova Hi! @jsoltes told me that you're the one in charge of forms while he's recovering. Could you please take a look? Ping me on IRC and I'll help you.
@romartin there has been some module changes, could you please check the stunner side too? Only package changes, nothing important.

This PR is part of an ensemble, merge with:

https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/720
https://github.com/kiegroup/kie-wb-common/pull/1492
https://github.com/kiegroup/jbpm-designer/pull/754
https://github.com/kiegroup/jbpm-wb/pull/993
https://github.com/kiegroup/kie-wb-distributions/pull/704

PR #1495, opened 2018-03-13
RHBA-514: migration tool window launcher

mareknovotny:
This PR enhances linux launcher to get proper java and removes the setting global CLASSPATH with all jars in lib folder.

Also I added Windows launcher and changed distribution zip assembly to be dist instead of only Linux.

Fixed platform dependent assembly descriptor to use absolute paths for bin and lib.

### DIFF
--- a/kie-wb-common-cli/kie-wb-common-project-migration/README.md
+++ b/kie-wb-common-cli/kie-wb-common-project-migration/README.md
@@ -12,7 +12,7 @@ To build the project, simply run
 
 This will generate a zip file in the target directory:
 
-    kie-wb-common-project-migration-$VERSION-linux.zip
+    kie-wb-common-project-migration-$VERSION-dist.zip
 
 The zip file is a standalone artifact that contains all the necessary dependencies to run the CLI tool.
 
@@ -23,7 +23,7 @@ Installing the Tool
 
 To install the tool, simply unzip it into the desired location, like so:
 
-    unzip -d $INSTALL_DIR kie-wb-common-project-migration-$VERSION-linux.zip
+    unzip -d $INSTALL_DIR kie-wb-common-project-migration-$VERSION-dist.zip
 
 This should create a directory (`kie-wb-common-project-migration-$VERSION`) in the `$INSTALL_DIR`.
 
@@ -42,7 +42,13 @@ Simple Usage
 
 The simplest way to invoke the tool is
 
+Linux -
     $TOOL_DIR/bin/migration-tool.sh -t $NIOGIT
+
+
+Windows -
+    $TOOL_DIR/bin/migration-tool.bat -t $NIOGIT
+    
 
 The tool will migrate projects in-place: when the tool finishes a successful run, the `$NIOGIT` directory will be ready for use with the new workbench.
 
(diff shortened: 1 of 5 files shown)


kie-wb-common-cli/kie-wb-common-project-migration/src/main/sh/migration-tool.sh
```diff
@@ -1,7 +1,35 @@
 #!/bin/sh
+SCRIPTDIR="`pwd`"
 
-SCRIPTDIR="`dirname $0`"
-export CLASSPATH="`find "$SCRIPTDIR/../lib" -name '*.jar' | tr '\n' ':' | sed -E 's/:$//'`"
+CLASSPATH="$SCRIPTDIR/lib/*"
```
mareknovotny: from java 6 it is possible to use wildcard in CLASSPATH and the variable should live only in script to not affect environment outside

kie-wb-common-cli/kie-wb-common-project-migration/src/main/sh/migration-tool.sh
```diff
@@ -1,7 +1,35 @@
 #!/bin/sh
+SCRIPTDIR="`pwd`"
```
jhrcek: I think that `dirname $0` that was there before was better, because it was robust to where the script is executed from.
This will only work when executing this directly from `bin` folder
But if user executes it from the directory above (e.g. ./bin/migration-tools.sh) the classpath will not be assembled correctly on the line below.

mareknovotny: you're right, i overlooked that. I will fix it

kie-wb-common-cli/kie-wb-common-project-migration/README.md
```diff
@@ -42,8 +42,14 @@ Simple Usage
 
 The simplest way to invoke the tool is
 
+Linux -
     $TOOL_DIR/bin/migration-tool.sh -t $NIOGIT
 
+
+Windows -
+    $TOOL_DIR/bin/migration-tool.bat -t $NIOGIT
```
tomasdavidorg: Just a small change. I would replace the `/`slash chars with `\` backslash chars for Windows commands.

kie-wb-common-cli/kie-wb-common-project-migration/README.md
```diff
@@ -55,5 +61,12 @@ The simple invocation will prompt the user to confirm before migration is attemp
...
 
-
+Linux -
+    ```
     $TOOL_DIR/bin/migration-tool.sh -b -t $NIOGIT
+    ```
+
+Windows -
+    ```
+    $TOOL_DIR/bin/migration-tool.bat -b -t $NIOGIT
```
tomasdavidorg: Same here

mbarkley:
[This other PR](https://github.com/kiegroup/kie-wb-common/pull/1492) will have conflicts with this change, but after speaking with @pefernan we think this PR should be merged first and @pefernan will rebase.

mbarkley:
The failing test is unrelated (also failed recently on master) so I'm merging this. Thanks again @mareknovotny.

kie-wb-common-cli/kie-wb-common-project-migration/README.md
mareknovotny: good catch, I can fix it independently

kie-wb-common-cli/kie-wb-common-project-migration/README.md
mareknovotny: @tomasdavidorg here is the fix https://github.com/kiegroup/kie-wb-common/pull/1503

PR #1496, opened 2018-03-13
DROOLS-2372: [DMN Editor] Marshalling/unmarshalling of QNames is broken. Failing Unit Test

manstis:
See https://issues.jboss.org/browse/DROOLS-2372

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/FunctionDefinition.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/FunctionDefinition.java
@@ -28,7 +28,9 @@ import org.kie.workbench.common.dmn.api.property.dmn.QName;
 @Portable
 public class FunctionDefinition extends Expression implements HasExpression {
 
-    public static final QName KIND_QNAME = new QName("{" + DMNModelInstrumentedBase.URI_KIE + "}kind");
+    public static final String DROOLS_PREFIX = "drools";
+
+    public static final QName KIND_QNAME = new QName("{" + DMNModelInstrumentedBase.URI_KIE + "}" + DROOLS_PREFIX + ":kind");
 
     private Expression expression;
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/QNamePropertyConverter.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/QNamePropertyConverter.java
@@ -28,7 +28,7 @@ public class QNamePropertyConverter {
      * @return maybe null
      */
     public static QName wbFromDMN(final javax.xml.namespace.QName qName) {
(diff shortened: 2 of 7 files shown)


manstis:
@tarilabs Here's a failing unit test. It shows the WB ```FunctionDefinition``` does not have the correct ```Kind``` ```QName```. However, please note, that ```roundTripUnmarshalMarshalThenUnmarshalDMN(..)``` for the same DMN file seems to work OK but the result of unmarshalling the ```FunctionDefiniton``` becomes an unwrapped ```Context``` and the outer ```FunctionDefinition``` is lost (even in _your_ Drools DMN model).. Is this correct?

manstis:
@tarilabs Your changed helped and the test now passes. I've also confirmed I can author a Function, set it's type to JAVA, PMML or FEEL, save it and re-open successfully.. 

Your implementation of the QName's string representation is inconsistent with that in ```javax.xml.namespace.QName#toString()``` (See [here](https://docs.oracle.com/javase/7/docs/api/javax/xml/namespace/QName.html#toString()) ```{URI}localPart```). I am therefore a little concerned that we may hit some incompatibilities loading files containing JAVA/PMML Functions authored by other tools... but for now _our_ tooling works OK and we can address any other problems later.

I tried simply changing ```QNamePropertyConverter``` to use the JDK representation but it caused a lot of the ```DMNMarshallerTest``` tests to fail.

tarilabs:
> Your implementation of the QName's string representation is inconsistent with that in javax.xml.namespace.QName#toString() (See here {URI}localPart). I am therefore a little concerned 

That is intentional. For instance a typeRef in DMN v1.1 can be expressed as `feel:string` which is NOT compatible with the JDK `QName.toString()` (as the JDK ignore the `feel:` prefix during toString ).

That is the reason for instance we have the Marshalling utils to deal with QNames.

I believe the only remaining point was to discuss if the QName on the UI need to deal with the namespace, prefix and localname separately instead of today 1 String only; but based on the discussion of today is not really clear to me if it's a big deal or not from a UI side. I believe we can review that the next DMN sync point
/cc @etirelli

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

tarilabs:
Seeing cf06985 I really reiterate we should discuss if we want to have WB QName to split in 3 String. The `drools` is a namespace prefix in the XML, but theoretically it can be anything, as long as it aliases for `http://www.drools.org/kie/dmn/1.1`

manstis:
@tarilabs Happy to do so... just a three String field POJO; URI, prefix and local name? 

Or do you see me needing to emulate more _function_ of ```javax.xml.namespace.QName```?

manstis:
@tarilabs @jomarko But I'd propose this PR is merged and I'll create a new JIRA for the ```QName``` improvements (remember, the existing WB ```QName``` has always been considered - by me at least - as a quick fix when the WB DMN model was put together....)

tarilabs:
I am not vetoing to go ahead and merge this PR as long as between the changes fixes the problem at hand. 

But now that I seen cf06985 I can locate an issue if a xml model by a third party is issued, we have no guarantee the `drools` will be used as the actual prefix, the only thing we can expect is only the namespace `http://www.drools.org/kie/dmn/1.1` for the actual execution semantics on our engine for v1.1.

I believe is also best to discuss the UI implication, when we get a suitable time for a call.

manstis:
Jenkins please retest this.

PR #1501, opened 2018-03-13
AF-1067: [Project Oriented] Import should clone the external reposito…

mbarkley:
…ry, not copy contents


mbarkley:
Some of the code surrounding the `ImportPresenter` may seem a bit over-engineered. I had originally created a second project import screen that would work like importing examples but with history, but after discussion with @porcelli we decided not to do this now.

I left some of the generality in the code because we do plan to add a screen for importing sub-directories of a repository in the future.

mbarkley:
@paulovmr Would you mind reviewing this?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryServiceImpl.java
```diff
@@ -195,6 +205,46 @@ public WorkspaceProject createProject(final OrganizationalUnit activeOrganizatio
                                          mode);
     }
 
+    @Override
+    public WorkspaceProject importProject(final OrganizationalUnit targetOU,
```
paulovmr: Could you please add a test for this?

mbarkley: Added tests in `LibraryServiceImplTest`.

mbarkley:
@paulovmr Added tests for importing projects. Would you mind taking another look?

PR #1508, opened 2018-03-15
 JBPM-6813: Stunner - Interrupting (Start event) RHBA-495 JBPM-6997

alepintus:
Hi @romartin @hasys @pefernan ,
For the start message, start timer , start error and start signal events the isInterrupting checkbox is shown only if the event is inside an event sub-process.
Tests for isInterrupting fields inside an event subprocess added for all start events with this property.

For more information see [this ticket](https://issues.jboss.org/browse/JBPM-6813).

Could you please take a look?
Thanks!


romartin:
ok to test

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/filters/BaseStartEventFilterProvider.java
```diff
@@ -0,0 +1,78 @@
...
+import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
+import org.kie.workbench.common.stunner.core.graph.Edge;
+import org.kie.workbench.common.stunner.core.graph.Element;
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.content.definition.Definition;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+import org.kie.workbench.common.stunner.core.graph.util.GraphUtils;
+import org.kie.workbench.common.stunner.forms.client.formFilters.StunnerFormElementFilterProvider;
+
+public abstract class BaseStartEventFilterProvider implements StunnerFormElementFilterProvider {
```
romartin: Hey @alepintus 
I would remove the `abstract` keyword and make this class "instantiable". In fact if you realize there is also no abstract method declared, which means this can be addresses in a different way. Also, for each new bean it requires to create an implementation of this abstraction. Also if you change the constructor parameters you have to update all the current implementations.... and what about unit tests?? I don't think this makes sense for the goals it provides... hard to maintain and review.

As far as I see  the only difference between types is the result for `getDefinitionType()`, right? In that case you can just declare a `private final Supplier<Class<?>>` in this class, and create a static Factory class that just create new instances of this class and sets the right supplier instance. In case you need to apply also some different logic inside at some point, you can do the same approach by using `Function<T, R>`. I mean the idea is avoid inheritance as much as possible and focus on composition. This way you'll end up with a SINGLE class for handling this filter and a SINGLE class for testing it, plus some factory.

Inheritance is hard to maintain, so IMO the best way to address this is just using a single `StartEventFilterProvider`, or even as much generic as possible, like some `BPMNFilterProvider`, if this applies for different kind of BPMN nodes, and so having another single Factory class which creates or sets the right parameters for each type. The idea is once we need add any new type, avoid coding as much as possible, you know. Same for testing, this approach makes testing difficult and requires more effort.

Does that makes sense?
Thanks!

hasys:
Hi @alepintus,

thank you looks very nice! Just some comments:

- When you add an Event to Subprocess and set **isInterrupting** to _false_ and after that move the Event out of subprocess to Top Level process **isInterrupting** property is still set to _false_, border is still dashed and there are no way to change it back. You should move the Event back to Subprocess to set **isInterrupting** to _true_. I guess it will be good just clear state of **isInterrupting** to default value (_true_) when you move it out of Parent and that's it.
- Error Start Event **isInterrupting** property should be disabled and always should be equals to _true_ according to BPMN 2.0 specification. Is it possible?
- Another minor issue not related to this PR but, probably, will be easy to fix. If not, we can report new issue: Error Start Event _Implementation/Execution_ has different order than other Start Events. See picture below.
image

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/filters/BaseStartEventFilterProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private SessionManager sessionManager;
+    private GraphUtils graphUtils;
```
hasys: This `GraphUtils` instance is used only for one STATIC method invocation. Is it possible to remove this field from the class and just invoke static method like this: `GraphUtils.getParent(node);` ?

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-engine/kie-wb-common-forms-adf-engine-api/src/main/java/org/kie/workbench/common/forms/adf/engine/shared/FormElementFilter.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class FormElementFilter<T> {
+
+    private String elementName;
+
+    private Predicate<T> predicate;
+
+    /**
+     * @param elementName String containing the name of the form element. It does support nested filtering by adding
+     * expresions like "element.childElement" so the filter will apply to the child element.
```
hasys: `expresions` - typo, should be `expressions`.

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-engine/kie-wb-common-forms-adf-engine-api/src/test/java/org/kie/workbench/common/forms/adf/engine/shared/formGeneration/FormGeneratorTest.java
```diff
@@ -19,12 +19,15 @@
...
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.kie.workbench.common.forms.adf.engine.shared.FormElementFilter;
 import org.kie.workbench.common.forms.adf.engine.shared.formGeneration.model.Person;
 import org.kie.workbench.common.forms.adf.engine.shared.test.AbstractFormGenerationTest;
 import org.kie.workbench.common.forms.model.FormDefinition;
 import org.mockito.runners.MockitoJUnitRunner;
 
-;
+;import static org.junit.Assert.assertEquals;
```
hasys: Redundant `;` symbol

pefernan:
@romartin In this case I think having an static factory will make things more complex. The properties panels forms is a general stunner feature (used in bpmn sets, dmn, case management...). The `FormDisplayer` component is using the `StunnerFilterProviderManager` to provide `StunnerFormElementFilterProvider` instances for some of the nodes. This `StunnerFormElementFilterProvider`  are dynamically loaded via CDI ManagedInstance.

If we add a static factory we'll need to implement a mechanism to register the `StunnerFormElementFilterProvider` on that factory for BPMN sets, and maybe for DMN or other sets that might require it.

My 2 cents here.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/filters/BaseStartEventFilterProvider.java
pefernan: @romartin In this case I think having an static factory will make things more complex. The properties panels forms is a general stunner feature (used in bpmn sets, dmn, case management...). The `FormDisplayer` component is using the `StunnerFilterProviderManager` to provide `StunnerFormElementFilterProvider` instances for some of the nodes. This `StunnerFormElementFilterProvider`  are dynamically loaded via CDI ManagedInstance.

If we add a static factory we'll need to implement a mechanism to register the `StunnerFormElementFilterProvider` on that factory for BPMN sets, and maybe for DMN or other sets that might require it.

My 2 cents here.

alepintus:
Hi everybody,
@romartin code changed as suggested .
@hasys , Error Start Event Implementation/Execution order fixed, Error Start Event isInterrupting property is now readonly and about setting isInterrupting to false after that moved out of event subprocess to top Level process could you please create a new Jira Issue?

Thanks!

hasys:
Hi @alepintus,

thank you! Ok, will report it as separate issue.

P.S. There are some failing tests in this PR:
```
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.twinTaskFormIsCorrectlyGenerated
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.testNestedFormsGeneration
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.emptyTaskFormIsCorrectlyGenerated
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.ProcessFormIsCorrectlyGenerated
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.differentIOTaskFormIsCorrectlyGenerated
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.inputTaskFormIsCorrectlyGenerated
org.kie.workbench.common.forms.integration.tests.formgeneration.FormGenerationIntegrationTest.sameIOTaskFormIsCorrectlyGenerated
```
Is it for @pefernan or you can fix it?

hasys:

hasys:

pefernan:
@hasys I also had that integration tests failures on other PR's but they "magically" got fixed... I tried to reproduce them locally but they were only appearing on jenkins. Let's see what the current downstream build says.

hasys:

romartin:

PR #1527, opened 2018-03-22
JBPM-5649: [Stunner] Fixing test compilation.

romartin:
Hey @pefernan @wmedvede 
Sorry something broken once merging this https://github.com/kiegroup/kie-wb-common/pull/1509 (Promised the Jenkins was green)
This is the fix.
Thanks!


romartin:

romartin:
Anyway @hasys @evacchi , I don't understand why the `BPMNDiagramMarshallerBase` has been copy/pasted... now we have this class as a copy of  `BPMNDiagramMarshallerTest`, why? Should we remove one of those, isn't it? Thx!

evacchi:
I think `BPMNDiagramMarshallerBase` it's a partial copy of  `BPMNDiagramMarshallerTest` and both ` BPMNDiagramMarshallerTest` and  `BPMNDirectDiagramMarshallerTest` will be eventually dropped; can @hasys confirm?

hasys:
Hi @evacchi, @romartin,

Yes, @evacchi you are correct. It is partial copy and `BPMNDiagramMarshallerTest` with `BPMNDirectDiagramMarshallerTest` will be dropped as soon as we will finish migration to new tests structure. We will try to do it ASAP, but it is a huge task and we have to do a lot of staff in parallel, so it will take some time. But any way it should be finished before GA (as part of features verification).

PR #1530, opened 2018-03-22
DROOLS-2410: [DMN Designer] Active entries not highlighted correctly when scrolling

manstis:
See https://issues.jboss.org/browse/DROOLS-2410

The problem was that grids "off screen" are no longer registered and hence when they moved "off screen" a "selected" grid could not be found and the ghost not rendered. 

I also fixed a problem I found when if you select a cell in a grid scroll it "off screen" and select a different cell in a different nested grid; then scroll back the original selection remained. This too was caused by "off screen" grids no longer being registered. I now walk the data models and deselect all grids whether "on screen" or "off screen". This is quite a (relatively) slow operation (and was removed elsewhere for performance reasons) but is OK to use when selecting a grid as it's driven by the User's interactions and performance is not critical.
 
Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/256
- https://github.com/kiegroup/kie-wb-common/pull/1530

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/literal/LiteralExpressionUIModelMapper.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/literal/LiteralExpressionUIModelMapper.java
@@ -22,6 +22,7 @@ import java.util.function.Supplier;
 import org.kie.workbench.common.dmn.api.definition.v1_1.LiteralExpression;
 import org.kie.workbench.common.dmn.client.widgets.grid.controls.list.ListSelectorView;
 import org.kie.workbench.common.dmn.client.widgets.grid.model.BaseUIModelMapper;
+import org.kie.workbench.common.dmn.client.widgets.grid.model.GridCellTuple;
 import org.uberfire.ext.wires.core.grids.client.model.GridCellValue;
 import org.uberfire.ext.wires.core.grids.client.model.GridData;
 import org.uberfire.ext.wires.core.grids.client.model.impl.BaseGridCellValue;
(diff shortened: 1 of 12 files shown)


manstis:
Jenkins please retest this.

manstis:
@jomarko I have fixed the reported issue; and another that was more serious and existed on ```master``` (if you created a ```Context``` - or anything that has nested expressions - and add a bunch of rows you could not multi-select the "Undefined Expression" cells). 

This PR now also depends on https://github.com/kiegroup/appformer/pull/256

I also need to add more Unit Tests for the second commit on this PR.. to follow.

manstis:
@jomarko Unit Tests added. This is ready for re-review.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/layer/DMNGridLayer.java
```diff
@@ -144,9 +144,19 @@ public void updatePinnedContext(final GridWidget gridWidget) throws IllegalState
     @Override
     public void select(final GridWidget selectedGridWidget) {
         this.selectedGridWidget = Optional.of(selectedGridWidget);
-        getAllGridWidgets().stream().forEach(GridWidget::deselect);
-        selectedGridWidget.select();
-        batch();
+        boolean selectionChanged = false;
+        for (GridWidget gridWidget : getAllGridWidgets()) {
+            if (!gridWidget.equals(selectedGridWidget)) {
```
jomarko: Could we use `Objects.equals` to avoid eventual NPE?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/layer/DMNGridLayer.java
```diff
@@ -144,9 +144,19 @@ public void updatePinnedContext(final GridWidget gridWidget) throws IllegalState
...
+                selectionChanged = true;
+                gridWidget.deselect();
+            } else if (gridWidget.equals(selectedGridWidget)) {
```
jomarko: Could we use `Objects.equals` to avoid eventual NPE?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/literal/LiteralExpressionUIModelMapper.java
```diff
@@ -22,29 +22,49 @@
...
+                                  columnIndex).setSelectionStrategy((final GridData model,
+                                                                     final int uiRowIndex,
+                                                                     final int uiColumnIndex,
+                                                                     final boolean isShiftKeyDown,
+                                                                     final boolean isControlKeyDown) -> {
+                final GridData parentUiModel = parent.getGridWidget().getModel();
+                return parentUiModel.getCell(parent.getRowIndex(),
+                                             parent.getColumnIndex()).getSelectionStrategy().handleSelection(parentUiModel,
+                                                                                                             parent.getRowIndex(),
+                                                                                                             parent.getColumnIndex(),
```
jomarko: Is it intended to use parentUiModel, parentRowIndex and parentColumnIndex?

manstis: Absolutely; that is really the essence to the fix for the issue where you could not select multiple ```LiteralExpression``` or "Undefined Expression" cells contained in, for example, different rows on a {{Context}}. Whenever you select either a ```LiteralExpression``` or "Undefined Expression" cell the actual selection is now handled by the parent.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/layer/DMNGridLayer.java
manstis: Done.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/layer/DMNGridLayer.java
manstis: Done.

PR #1532, opened 2018-03-23
DROOLS-2301: Increase coverage of policy selector view

jomarko:
See https://issues.jboss.org/browse/DROOLS-2301

@manstis I found some unused private field, so decided also increase coverage during removing it.


jomarko:
@manstis good point, PR updated.

jomarko:
jenkins please retest this

PR #1548, opened 2018-03-27
DROOLS-2424: [DMN Designer] Expression type can not be changed after cleared multiple times

manstis:
See https://issues.jboss.org/browse/DROOLS-2424

I didn't see how I could write a test for this fix really.. but open to ideas. I moved the instantiation of ```parent``` from "inline" to "in the constructor". In Java I don't see what difference it'd make but in GWT it seems to make a world of difference.


manstis:
Jenkins please retest this.

manstis:
@jomarko Thanks for the review.. I will close this PR is due course as I think (hope!) my refactoring of the cache for https://issues.jboss.org/browse/DROOLS-2423 will fix these problems. When my changes are complete I'll test the scenario you illustrate here.

PR #1550, opened 2018-03-27
[AF-1087]: Excluded file names from Library Asset Type Definition. Those files should not be indexed.

adrielparedes:
@mbarkley @manstis Yesterday we talked about about Indexers, and to inject some indexers into LibraryIndexer, but actually, that LibraryIndexer is an indexer, and injecting Instance<Indexer> it was going to add many many indexers that are not neccesary, so I found __LibraryAssetTypeDefinition__ that contains the definition of a Library Asset. So here I create a list of assets that should not be indexed.

I uploaded this so we can continue the conversation here.

Related to:
https://github.com/kiegroup/appformer/pull/269
https://github.com/kiegroup/kie-wb-common/pull/1550
https://github.com/kiegroup/drools-wb/pull/838
https://github.com/kiegroup/optaplanner-wb/pull/269
https://github.com/kiegroup/jbpm-designer/pull/760


kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryAssetTypeDefinition.java
```diff
@@ -76,10 +90,17 @@ public String getSimpleWildcardPattern() {
...
 
     boolean isFolder(final Path path) {
         return Files.isDirectory(Paths.convert(path));
     }
+
+    boolean isExcluded(final Path path) {
+        return excludedFileNames.stream()
+                .filter(excluded -> path.getFileName().equals(excluded))
+                .findAny()
+                .isPresent();
```
mbarkley: FYI, a shortcut for this is `anyMatch`.

adrielparedes: you are right, sorry, I will change it

mbarkley:
So personally, although I liked @manstis idea, I am also fine with this solution.

If you wanted to pursue something more similar to @manstis original suggestion, you could try injecting `Instance<ResourceTypeDefinition>` and figure out a way of filtering which resource types are visible assets.

A few possibilities:

1. An explicit, programmatic white-list of `ResourceTypeDefinition` implementations that should be shown.
2. Use a qualifier so you can inject only visible type definitions (ex. `@Inject @VisibleAsset Instance<ResourceTypeDefinition> defintions`).
3. Add a new `isVisibleAsset()` method on the `ResourceTypeDefinition` that defaults to false; override it where appropriate and filter using that.

Another option, is you could rebase off my indexing PRs and inject `Instance<IndexersFactory> factory`. The only catch there, is that my PR introduces a `CoreIndexer` that indexes everything, and you would want to explicitly ignore that. So you would end up with something like this:

    public class LibraryIndexer implements Indexer {
    // ...
      @Inject Instance<IndexersFactory> factory;
      public boolean supportsPath(Path path) {
        return factory.get()
                  .getIndexers()
                  .stream()
                  .filter(indexer -> !(indexer instanceof CoreIndexer))
                  .anyMatch(indexer -> indexer.supportsPath(path));
      }
    }

Just some alternatives, in case you're interested.

manstis:
@adrielparedes Please advise when the ```anyMatch()``` change is complete and I'll merge.

adrielparedes:
@manstis @mbarkley I'm considering alternative 2 adding VisibleAsset to ResourceTypeDefinitions, so I think I will open a few more PRs to  add this part. I hope by the end of my day

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryAssetTypeDefinition.java
```diff
@@ -29,13 +31,25 @@
 public class LibraryAssetTypeDefinition
         implements ResourceTypeDefinition {
 
+    private static final String PERSISTENCE_XML = "persistence.xml";
+    private static final String KIE_DEPLOYMENT_DESCRIPTOR = "kie-deployment-descriptor.xml";
+    private static final String PROJECT_REPOSITORIES = "project.repositories";
+    private static final String POM_XML = "pom.xml";
+    private static final String KMODULE_XML = "kmodule.xml";
     private Category category;
+    private List<String> excludedFileNames;
```
jhrcek: I think there's no need to create instance field - static one should be sufficient. Also you won't have to initialize in in no-args constructor and can initialize it in static initializer directly

```java
    private static final Set<String> EXCLUDED_FILENAMES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
            "persistence.xml",
            "kie-deployment-descriptor.xml",
            "project.repositories",
            "pom.xml",
            "kmodule.xml"
    )));
```

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryAssetTypeDefinition.java
jhrcek: Even better I'd replace this stream base implementation with a simple hash set and `set.contains(path.getFileName())` as shown [here](https://github.com/jhrcek/kie-wb-common/commit/561268ee17b5ce189c9e2e9d1875f1c9ac76a2ba). I tried quick and dirty [benchmark](https://gist.github.com/jhrcek/b3340cf3f81b74284f8ccf04491c300f) and my version is consistently 10x faster than this stream-based one (even with `anyMatch`.

mbarkley: On my machine the slower of the two is still only ~100ms for 1 million comparisons so I don't think it matters much.

But also this is all moot, since @adrielparedes is going to be updating this PR with a different approach discussed in the comments.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
```diff
@@ -70,7 +80,11 @@ void setModuleService(final KieModuleService moduleService) {
 
     @Override
     public boolean supportsPath(final Path path) {
-        return filter.accept(Paths.convert(path));
+        boolean anyResourceTypeMatch = this.getResourceTypeDefinitionList()
```
mbarkley: If you refactored this one-liner into a method call, you could call it directly from the return statement and take advantage of short-circuiting. Currently you are querying all the type definitions, even if the path is a folder or is hidden.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
```diff
@@ -70,7 +80,11 @@ void setModuleService(final KieModuleService moduleService) {
...
+                .stream()
+                .anyMatch(resourceTypeDefinition ->
+                                  resourceTypeDefinition.accept(convertPath(path)));
```
mbarkley: Is converting the path cheap? Maybe do it once in the containing scope and reuse the value?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/test/java/org/kie/workbench/common/screens/impl/LibraryAssetTypeDefinitionTest.java
```diff
@@ -36,16 +40,23 @@
 
     private boolean isAccepted;
 
-    private LibraryAssetTypeDefinition definition;
+    private LibraryIndexer definition;
 
     @Before
     public void setup() {
-        this.definition = new LibraryAssetTypeDefinition() {
+        this.definition = spy(new LibraryIndexer() {
```
mbarkley: I don't like mixing spy with method overrides in this test. It would be better if you could stick to one approach.

Ideally, you shouldn't have to override methods in your class under test at all. You should encapsulate any behaviour you need to mock in an interface that you can inject. (For example, you can already mock Instance, and you could make a wrapper interface for Files with a default impl that dispatches to the static methods.)

But if you really don't want to do that, I would settle for sticking with either method overrides, or mocking of protected methods, but please not both.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
```diff
@@ -49,15 +55,19 @@
...
     public LibraryIndexer() {
+        resourceTypeDefinitions = new HashSet<>();
     }
 
     @Inject
-    public LibraryIndexer(final LibraryAssetTypeDefinition filter) {
-        this.filter = filter;
+    public LibraryIndexer(@VisibleAsset final Instance<ResourceTypeDefinition> resourceTypeDefinitions) {
+        this.resourceTypeDefinitions = StreamSupport
+                .stream(resourceTypeDefinitions.spliterator(),
```
ederign: @adrielparedes the instance.get occurs here?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
```diff
@@ -70,7 +80,15 @@ void setModuleService(final KieModuleService moduleService) {
 
     @Override
     public boolean supportsPath(final Path path) {
-        return filter.accept(Paths.convert(path));
+        return !(this.isFolder(path) || this.isHidden(path)) && this.isSupportedByAnyResourceType(path);
+    }
+
+    private boolean isSupportedByAnyResourceType(Path path) {
+        org.uberfire.backend.vfs.Path convertedPath = convertPath(path);
+        return this.getResourceTypeDefinitionList()
```
ederign: If it's here, @adrielparedes one question about this the assets are mandatory to be @ApplicationScoped? 

I think so, but I think we maybe should ensure that to avoid a memory leak for instance.get().

wdyt?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
adrielparedes: @ederign it happens in the constructor. The StreamSupport get all instances and transform them to the concrete object I need.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
```diff
@@ -161,12 +179,28 @@ public KObjectKey toKObjectKey(final Path path) {
...
+
+    protected boolean isFolder(final Path path) {
+        return Files.isDirectory(path);
+    }
+
+    protected boolean isHidden(Path path) {
+        return path.getFileName().startsWith(".");
+    }
+
+    protected Set<ResourceTypeDefinition> getResourceTypeDefinitionList() {
```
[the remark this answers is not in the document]
adrielparedes: @jhrcek you are right, let me chage it.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-backend/src/main/java/org/kie/workbench/common/screens/impl/LibraryIndexer.java
adrielparedes: @ederign I think yes, that it is mandatory. But the instance.get happens in the constructor, so I think we are not going to have the memory leak, at least in here. When StreamSupport.collect(toSet()); occur, I have all instances assigned to resourceTypeDefinitions attribute.

mbarkley:
@ederign Are there changes you are still waiting for?

PR #1547, opened 2018-03-27
RHBA-629: Can't compile embedded subprocess when new marshaller is enabled

evacchi:
WIP. This currently fixes the issue but breaks one test `testMarshallMagnetsInlane`
Other tests pass.

depends on: #1522

PR #1554, opened 2018-03-28
RHBA-629: Can't compile embedded subprocess when new marshaller is enabled #1547 

evacchi:
Fixes issue. Multiple instances of the same element were added to the document, when a subprocess was present. 

**Warning** there are still some rendering issues that I haven't sorted out: i.e. edges are messed up (tests pass though -- which means we need more tests...)

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java
@@ -56,7 +56,8 @@ public class ProcessPropertyWriter extends BasePropertyWriter implements Element
     private static final String defaultRelationshipType = "BPSimData";
     private final Process process;
     private final BPMNDiagram bpmnDiagram;
-    private Map<String, BasePropertyWriter> childElements = new HashMap<>();
+    private Map<String, PropertyWriter> childElements = new HashMap<>();
+    private Map<String, LanePropertyWriter> lanes = new HashMap<>();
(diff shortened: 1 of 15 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java
```diff
@@ -55,38 +57,49 @@
...
 
-    public void addChildEdge(BPMNEdge edge) {
-        if (edge != null) {
-            bpmnDiagram.getPlane().getPlaneElement().add(edge);
-        }
-    }
+        BPMNPlane bpmnPlane = di.createBPMNPlane();
+        bpmnDiagram.setPlane(bpmnPlane);
+
+        List<DiagramElement> planeElement = bpmnPlane.getPlaneElement();
```
romartin: mmm just for brainstorming a little bit... I mean no need to change nothing here.
Doing in that way, `planeElement` is obviously mutable,  and the `foreach` used to add the elements in the list is not an intermediate stream operation, so we cannot achieve here laziness, this way concurrency could be an issue if that code is run in parallel at some point, does that makes sense? 
On the other hand, as we're hopefully working with Streams here, we can probably address laziness in the code by ensuring the intermediate functions we'll use will behave as pure functions - Streams ensure the purity of the functions if we can ensure that the arguments (`childElements`) are not mutable, which at this point happens. right? Anyway we want to just map/collect the items in a new list, I mean the anonymous functions to achieve this do not require any variable from the enclosing parent. so laziness can be guaranteed as well. 
This way what about this? -> Removing the `planeElement` instance, and creating a new static function that just given the `p` argument it returns a `Stream<DiagramElement>`, which will contain both shape and edge. This way you can perform a `flatMap` for the  `childElements.values()` to an Stream<DiagramElement>, and just use the `collect` as for the terminate function.

evacchi:
patchset reworked from scratch, it now passes all tests, and should render as expected

evacchi:
@romartin no I don't think so

evacchi:
rebased. THINE TESTS SHALL PASS

PR #1563, opened 2018-03-29
DROOLS-2436: Fix wrong diagram's name match condition.

romartin:
Hey @manstis  @jomarko 

Fixed the issue on the diagram's name matching condition. 

See [DROOLS-2436](https://issues.jboss.org/browse/DROOLS-2436)

Sorry it was my fault, added a small test at least for this. 

@manstis Thanks for pointing out the root issue, this way things are easier... :+1:  

Thanks!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-backend/src/main/java/org/kie/workbench/common/stunner/backend/service/DiagramLookupServiceImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-backend/src/main/java/org/kie/workbench/common/stunner/backend/service/DiagramLookupServiceImpl.java
@@ -83,7 +83,7 @@ public class DiagramLookupServiceImpl
                               final Diagram<Graph, Metadata> item) {
         final Map<String, String> criteriaMap = AbstractCriteriaLookupManager.parseCriteria(criteria);
         final String name = criteriaMap.get(DiagramLookupRequest.CRITERIA_NAME);
-        if (null != name && name.trim().length() > 9) {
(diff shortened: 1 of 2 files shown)


romartin:

romartin:
@manstis  hahah no no... the `9` was due to my finger at the wrong location... :)

kie-wb-common-stunner/kie-wb-common-stunner-backend/src/main/java/org/kie/workbench/common/stunner/backend/service/DiagramLookupServiceImpl.java
```diff
@@ -83,7 +83,7 @@ protected boolean matches(final String criteria,
                               final Diagram<Graph, Metadata> item) {
         final Map<String, String> criteriaMap = AbstractCriteriaLookupManager.parseCriteria(criteria);
         final String name = criteriaMap.get(DiagramLookupRequest.CRITERIA_NAME);
-        if (null != name && name.trim().length() > 9) {
+        if (null != name && name.trim().length() > 0) {
             return name.equals(item.getName());
```
jomarko: Just question due to if condition, shouldn't we return `name.trim().equals(item.getName())` ?

romartin: well don't think so, or at least, in that case you'll have to do trim() as well on the item. So imagine you have a diagram called "my diagram", if you do the trim that will make the equals return false..

PR #1568, opened 2018-04-03
AF-1151: Project's pom.xml file is not validated before saving

tiagobento:
This PR brings back the `PomEditor` class that was removed along with the implementation of the new Settings screen. More than that, this editor now contains a save validation and displays a friendly message when errors occur.

ezgif-2-bfbbb02672

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/PomEditor.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/PomEditor.java
@@ -16,28 +16,36 @@
 
 package org.kie.workbench.common.screens.library.client.util;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import javax.inject.Inject;
 
 import com.google.gwt.user.client.ui.IsWidget;
-import elemental2.dom.DomGlobal;
+import org.guvnor.common.services.project.client.repositories.ConflictingRepositoriesPopup;
 import org.guvnor.common.services.project.client.type.POMResourceType;
+import org.guvnor.common.services.project.service.DeploymentMode;
+import org.guvnor.common.services.project.service.GAVAlreadyExistsException;
 import org.jboss.errai.common.client.api.Caller;
 import org.kie.workbench.common.screens.defaulteditor.client.editor.KieTextEditorPresenter;
 import org.kie.workbench.common.screens.defaulteditor.client.editor.KieTextEditorView;
-import org.kie.workbench.common.services.shared.validation.ValidationService;
+import org.kie.workbench.common.screens.projecteditor.service.PomEditorService;
+import org.kie.workbench.common.widgets.client.callbacks.CommandWithThrowableDrivenErrorCallback;
+import org.kie.workbench.common.widgets.client.callbacks.CommandWithThrowableDrivenErrorCallback.CommandWithThrowable;
+import org.kie.workbench.common.widgets.client.resources.i18n.CommonConstants;
 import org.uberfire.backend.vfs.ObservablePath;
-import org.uberfire.backend.vfs.Path;
 import org.uberfire.client.annotations.WorkbenchEditor;
 import org.uberfire.client.annotations.WorkbenchMenu;
 import org.uberfire.client.annotations.WorkbenchPartTitle;
 import org.uberfire.client.annotations.WorkbenchPartTitleDecoration;
 import org.uberfire.client.annotations.WorkbenchPartView;
-import org.uberfire.client.promise.Promises;
 import org.uberfire.ext.widgets.common.client.ace.AceEditorMode;
 import org.uberfire.lifecycle.OnStartup;
-import org.uberfire.mvp.Command;
 import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.workbench.model.menu.Menus;
+
+import static org.guvnor.common.services.project.service.DeploymentMode.FORCED;
+import static org.guvnor.common.services.project.service.DeploymentMode.VALIDATED;
 
 @WorkbenchEditor(
         identifier = "PomEditor",
@@ -46,52 +54,38 @@ import org.uberfire.workbench.model.menu.Menus;
 
 public class PomEditor extends KieTextEditorPresenter {
 
-    private final Promises promises;
-
-    private final Caller<ValidationService> validationService;
+    private Caller<PomEditorService> pomEditorService;
+    private ConflictingRepositoriesPopup conflictingRepositoriesPopup;
 
     @Inject
     public PomEditor(final KieTextEditorView baseView,
-                     final Promises promises,
-                     final Caller<ValidationService> validationService) {
-
+                     final Caller<PomEditorService> pomEditorService,
+                     final ConflictingRepositoriesPopup conflictingRepositoriesPopup) {
         super(baseView);
-        this.promises = promises;
-        this.validationService = validationService;
+        this.pomEditorService = pomEditorService;
+        this.conflictingRepositoriesPopup = conflictingRepositoriesPopup;
     }
 
     @OnStartup
     public void onStartup(final ObservablePath path,
                           final PlaceRequest place) {
-        super.onStartup(path, place);
+        super.onStartup(path,
+                        place);
     }
 
-    @Override
-    protected void save(final String commitMessage) {
-        promises.promisify(validationService, s -> {
-            Path path = getPathSupplier().get();Ω
-            String content = getContentSupplier().get();
-            return s.validateForSave(path, content);
-        }).then(errors -> {
-
-            if (errors.isEmpty()) {
-                super.save(commitMessage);
-            } else {
-                DomGlobal.console.info("Error!");
-            }
-
-            return promises.resolve();
-        });
-    }
-
-    @WorkbenchPartTitle
-    public String getTitleText() {
-        return super.getTitleText();
+    @WorkbenchMenu
+    public Menus getMenus() {
+        return super.getMenus();
     }
 
     @WorkbenchPartTitleDecoration
     public IsWidget getTitle() {
         return super.getTitle();
+    }
+
+    @WorkbenchPartTitle
+    public String getTitleText() {
+        return super.getTitleText();
     }
 
     @WorkbenchPartView
@@ -100,17 +94,42 @@ public class PomEditor extends KieTextEditorPresenter {
     }
 
     @Override
-    protected Command onValidate() {
-        return super.onValidate();
-    }
-
-    @WorkbenchMenu
-    public Menus getMenus() {
-        return menus;
+    public AceEditorMode getAceEditorMode() {
+        return AceEditorMode.XML;
     }
 
     @Override
-    public AceEditorMode getAceEditorMode() {
-        return AceEditorMode.XML;
+    protected void save(final String commitMessage) {
+        doSave(commitMessage, VALIDATED);
+    }
+
+    private void doSave(final String commitMessage,
+                        final DeploymentMode mode) {
+
+        //Instantiate a new instance on each "save" operation to pass in commit message
+        view.showBusyIndicator(CommonConstants.INSTANCE.Saving());
+
+        pomEditorService.call(getSaveSuccessCallback(view.getContent().hashCode()),
+                              new CommandWithThrowableDrivenErrorCallback(busyIndicatorView,
+                                                                          saveErrorCallbackConfig(commitMessage)))
+                .save(versionRecordManager.getCurrentPath(),
+                      view.getContent(),
+                      metadata,
+                      commitMessage,
+                      mode);
+    }
+
+    private Map<Class<? extends Throwable>, CommandWithThrowable> saveErrorCallbackConfig(final String commitMessage) {
+        return new HashMap<Class<? extends Throwable>, CommandWithThrowable>() {{
+            put(GAVAlreadyExistsException.class, t -> {
+                view.hideBusyIndicator();
+                final GAVAlreadyExistsException e = (GAVAlreadyExistsException) t;
+                conflictingRepositoriesPopup.setContent(e.getGAV(), e.getRepositories(), () -> {
+                    conflictingRepositoriesPopup.hide();
+                    doSave(commitMessage, FORCED);
+                });
+                conflictingRepositoriesPopup.show();
+            });
+        }};
     }
 }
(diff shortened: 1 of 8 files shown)


porcelli:
yes agree with @manstis the pom editor should have the standard validate button and the save should be like any other asset

tiagobento:
@manstis @porcelli From what I understood, the `Save` buttons allow saving regardless of validation errors and the `Validate` buttons are meant for checking if there are validation errors in the case users wants to.

With the old editor, it was not possible to save the `pom.xml` file with errors because [an exception occurred](https://issues.jboss.org/secure/attachment/12432713/Screenshot%20from%202018-03-15%2015-24-26.png) in the process, so the `pom.xml` was always valid.

During the development of this PR I noticed that when the `pom.xml` file is invalid, not even the Library screen opens, which is why I chose to validate and save as a single action on this editor.

I understand it is not ideal to have it not in the standards but I don't think we can allow the `pom.xml` file to ever be invalid. I'm open to suggestions on how to do that. Or we can deal with the invalid `pom.xml` in every place it's used.


- Behavior of the default text editor (allowed `pom.xml` to be invalid)
rhba-554-2

- Behavior of the old editor (exception occurred on errors):
rhba-554-3

manstis:
@tiagobento I see your point. 

Do we know where the error is thrown? 

I suspect it is in KIE's ```MavenParser``` and since it's not handled by ```Builder``` it simply gets propagated to the UI. I wonder if catching in ```Builder``` and adding a ```BuildMessage``` to the ```BuildResults``` would be enough to be able to follow the "standard" (of having a "Validate" button and allowing save of invalid files; with errors shown in the Message Console). We'd need to catch in both "full builds" (the ```Builder.build()``` method) and "incremental builds" (the overloaded ```Builder.addResource(..)``` methods).

Only a thought; I really appreciate the work you've done to add back the ```pom.xml``` editor and solve the issues you found with it's use... but consistency is important and **if** there's an alternative that supports consistency I personally think it should be pursued.

Also, IDK how much of this is "throw away" with "Places" (@porcelli is that the correct term for the fork/merge changes for the future?)

tiagobento:
@manstis That's good insights you've provided. I'll try to analyze how spread out this invalid pom issue is and I'll post my findings here.

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/PomEditor.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @WorkbenchPartTitle
+    public String getTitleText() {
+        return super.getTitleText();
+    }
+
+    @WorkbenchPartView
+    public IsWidget asWidget() {
+        return super.getWidget();
```
jomarko: Do we really need to override these methods which calls just `super.something()`?

tiagobento: As far as I can tell the editors need to have this annotation in some method so that the Activities are properly generated.

jomarko:
@tiagobento could you please update commit message and PR name, [RHPAM-544](https://issues.jboss.org/browse/RHPAM-544) looks be different issue as solved in this PR.

jomarko:
@tiagobento no problem, I have no comments to code, however have not tried manually, leaving the final approval on @barboras7

jomarko:
@tiagobento I  found ot, Barbora is on training, so I will do manual check also, shouldn't be hard to verify

jomarko:

porcelli:
@jomarko I think we can't do much here, this is an error that parser returns. I think we should accept it as is.

porcelli:
@manstis I think @tiagobento has a point where the pom file is quite a different element, as if it has errors... everything is impacted.

tiagobento:
@porcelli Thanks for the comments. Do you think we should further investigate the errors that happen due to an invalid pom.xml or should we just stick with this approach although it's not in the standards of workbench editors?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/PomEditor.java
manstis: @jomarko The annotation is used by AppFormer to generate some internal classes. The super class, ```KieEditor``` is ```abstract``` and lacks these annotations so, as @tiagobento has done, is the standard approach.

manstis:
@tiagobento Did you ever "...analyze how spread out this invalid pom issue is and I'll post my findings here"? I understand your point of view; but the ```pom.xml``` editor is for advanced Users and (used to be - I guess still is?) is difficult to find (you need to switch the Project Explorer view mode and navigate to ```pom.xml```). Therefore if they break something in the ```pom.xml``` they should expect the workbench to stop providing many services and IMO we should be catering for when the ```pom.xml``` cannot be built and fallback where this was essential for operation. That said, there is a flip side, if the ```pom.xml``` editor is so hard to find then it should matter less that it behaves differently to the other editors. 

I'm a pragmatic guy and since @porcelli has the ultimate say (AFAIK he or @ederign are the leads for generic workbench things) I'll go with what they decree.

tiagobento:
@manstis Didn't have the time yet. Was wrapping up some WIP that finally became a PR today and I still need to review a huge one before having the time to look at this PR again.

I agree with you completely in that the Workbench should be capable of handling invalid `pom.xml` files and continue working with limited capabilities after an advanced (or even curious?) user messes up with its data. It should be possible to fix whatever errors someone has input on the `pom.xml` after breaking it, though. Which we currently do not have based on the short analysis I made during the development of this issue.

As you said, @porcelli and @ederign are who decide in these situations so if they think we should handle it the way it's already implemented we could move forward with this PR, but if not we can certainly find a way to make the Workbench more robust in that sense.

manstis:
@tiagobento @ederign @porcelli See https://github.com/kiegroup/kie-wb-common/pull/1615

It looks like the ability to keep using the workbench when the ```pom.xml``` is invalid has already been addressed. Consequentially is there any reason now why the ```pom.xml``` editor cannot follow the same behaviour as other editors?

Rikkola:
The #1615 Still does not let you access the broken pom.xml file in the UI. The reason for this is that if your pom.xml file is broken, you already did a git clone and push or you imported a repository that is not compatible. 

But this editor changes that. Maybe in a broken project where there is a pom.xml file in the root I should give access to this pom file editor for admin users.

tiagobento:
@ederign Created and changed the PRs's titles.

manstis:
Jenkins please retest this.

manstis:

barboras7:
Hi @tiagobento,
I tried the fix and with some invalid information in the pom.xml file, it works. However, when you delete the content of the pom.xml file and hit "Save", you get an error dialog (see attached screenshot). I think that this issue might be also addressed in this PR, what do you think? 

screenshot from 2018-04-13 12-47-23

tiagobento:

tiagobento:
Okay, so there's a lot going on with the `pom.xml` editor. I'll try to organize everything we discussed so we can make a better decision on that.

Here's a timeline:

1. This editor already existed before the new Settings Screen was implemented. This new screen provides a way of editing the `pom.xml`'s content without the need/possibility to interact with the raw XML file. I removed this editor because I thought this new screen would make it unnecessary.

2. `RHBA-554` was created because the `pom.xml` file could still be edited using the default text editor, so I figured I should bring the editor back and add an extra saving step, the validation. What I didn't know was that we allowed other assets to be saved without being validated and that would cause a build failure on the project containing the faulty asset.

3. @manstis Warned that this editor was not in the standards of the Workbench and I replied saying the `pom.xml` was a special case because there were unexpected errors that occurred when we had an invalid `pom.xml` on the project, so he agreed that we could continue with this approach given that this editor is fairly hidden in the Workbench and because of the extra effort it would take to make this editor be in the standards.

4. #1615 came in to play. From its description we are already capable of detecting when an invalid `pom.xml` exists and warn the user on the UI without breaking it. We could take an extra step and redirect the user to the `pom.xml` editor if he wants. Otherwise he could continue operating with limited features like he apparently already can. This would make it possible to have this editor be in the standards, given that it'll be possible to fix the `pom.xml` in the UI if someone breaks it.

5. @barboras7 Found two bugs that will require some extra effort to be fixed.

Having that all said, my suggestion is we standardize this editor and make it possible to fix the `pom.xml` in the UI by implementing what I described on item 4. Of course, I'll go with what's decided.

@manstis If we decide to standardize this editor, do you think the duplicated GAV check should be done whilst saving or validating?

@manstis @porcelli @ederign Could you please give your opinions?

ederign:
@tiagobento I also would allow users to save an invalid pom and when user reopen the projects redirect to #1615 and then place a link in order to fix the invalid pom.

BUT I would like to know what happens if:

Two users are logged on the same project and then one of them breaks the pom:
- Should the other user don't see anything? (but he will start to see his build broken)
or
- if they are in the project view, maybe makes sense the toni's redirect
- what would happen if they have assets open and on a dirty state? They need to have a chance to save their work in invalid projects and then gets the redirect.

(@paulovmr has methods

What do you guys think?

The @barboras7 issues are introduced by this PR or they are a new thing?

tiagobento:
@ederign The issues @barboras7's found are reintroduced in this PR since I'm "resurrecting" the `pom.xml` editor that once existed.

tiagobento:
https://issues.jboss.org/browse/AF-1138 This issue is also related to the discussion we're having here. The new Settings Screen provides an UI to change the contents of the `kie-deployments-descriptor.xml` file, which is also accessible from the Project Explorer. We should also think about what to do in the case this file is saved with invalid data.

Currently a warning appears on the screen and the `Deployments` section disappears from the list of the Settings Screen because it cannot be assembled correctly due to parsing being impossible. The only difference is that after #1617 this file can already be accessed and fixed from the UI.

@ederign @manstis @barboras7 @porcelli Idk if that's a desired behavior or not, so please share your thoughts on that as well.

manstis:
@tiagobento @ederign @barboras7 @porcelli @Rikkola 

This is turning into a can of worms ;-)

IMO, we should standardise how certain things are handled:

1) ```pom.xml``` becomes invalid; when opening the Project Editor it falls back to the XML editor.
2) ```kie-deployments-descriptor.xml``` becomes invalid; when opening the ```Deployments``` section it falls back to the XML editor.
3) (As yet, unmentioned) ```kmodule.xml``` becomes invalid; when opening the ```KIE Bases``` section it falls back to the XML editor.
4) Both 1 and 2 above would be consistent with how an invalid ```Java``` file is handled in the "Data Object Editor". If the User changes the source (either using the "Source" tab or externally and pushing a commit into the workbench) the editor falls back to just the ```Java``` source editor and the "guided" elements cannot be used (until the source is fixed).

Perhaps for 1, 2 and 3 the applicable sections on the LHS of the Project Editor screen are replaced with "pom.xml", "kmodule.xml", "kie-deployments-descriptor.xml" and the corresponding RHS side just shows a XML editor with the content?

project editor

IDK whether the scope of this PR is stretching and really, having the changes merged "as is" would be better for the sake of progress and the issue of how to better handle invalid files (```pom.xml```, ```kmodule.xml```, ```kie-deployments-descriptor.xml``` and possibly others) is raised as a new JIRA with adjustments to the Project Editor screen made following consultation with UX? 

@ederign In response to your specific questions:-

> Two users are logged on the same project and then one of them breaks the pom.

Is this scenario any different to if User A breaks a rule and User B was writing a Test Scenario using the rule? I wonder if there is much we can do. Users A and B would only be affected when opening a new editor (that needs to parse the ```pom.xml``` for classes available for authoring meta-data; or other). We could broadcast a message (to appear in the Messages Console, and possibly a ```NotificationEvent``` that all Users receive)?

> Should the other user don't see anything? (but he will start to see his build broken)

Same comment.

> What would happen if they have assets open and on a dirty state?

It would have little effect (other than if we choose to automatically navigate to the ```pom.xml``` editor - which to be honest I'm not keen on - but I'm probably overstepping my responsibilities as I have little to do with the general navigation etc these days). However if we broadcast a message to the Message Console for all Users it could contain a link to the ```pom.xml``` editor (whether this is a standalone editor or the Project Editor appropriately configured is another question!)

As the old adage goes "[too many cooks spoil the broth](https://dictionary.cambridge.org/dictionary/english/too-many-cooks-spoil-the-broth)" I am happy to step aside and let @ederign @tiagobento and/or @porcelli etc decide the best course of action.

Rikkola:
First about this PR: 

If we expose the pom.xml in the UI so that the user can break it. Then we just need to also let the user fix it in the UI. The pom.xml is special here, because if it is broken the whole project fails to work. You can't even get a list of assets in Library. We can fix this by giving a shortcut link to the pom.xml editor in the Broken Project screen. If you manage to break any other file. You can go and fix the file with the same method that you used to break it. 

Some additional thoughts about Project Editor:

I think we should invest some time and work on how the Project Editor is implemented. Now jBPM, Drools and Optaplanner are different project types and each might need different editor tabs. For sure I know the multimodule support will change the layout depending on if the project is multimodule or not. 

This means we are in a point where we need to start thinking if we just make the Project Editor layout to fit the UI design as fast as possible, or if we invest time on making it easier and smarter to extend for the future. I believe we want to add "facets" for even our own project types, but there might be additional facets needs once we actually start using Maven and we might want to expose optional UI for different Maven plugins. Pluggable Project Editor is something that Appformer would benefit from. Right now everything is hard coded for KIE.

So with all these ideas and plans it might be best to push all ideas to UX and get a UI design that covers all needs. Then think about the code implementation that covers those needs, but does not necessary have all the features on day one. For example the Project Editor layout has changed few times now, next we add multimodule support, then possibly facets. At the end I bet we look back and see how we spend days just moving and recoding everything when a good plan would have saved us work weeks. 

The planning period should have as many cooks as possible so that we don't miss anything.

tiagobento:
@manstis @Rikkola @ederign I really appreciate your involvement on this issue. Thanks for taking the time to read everything and contribute with meaningful replies.

As @manstis said, I believe we would benefit more from merging this PR "as is" and creating a separate JIRA where this discussion can continue. UI/UX team support will obviously be necessary so there's no point in extending the scope of this PR.

@ederign Despite the issues that @barboras7 pointed out (which are more complex than they seem), I think we should go ahead and merge this PR so that we have at least a user-friendly error message for the most common case of errors when editing the `pom.xml` file. WDYT?

ederign:
@tiagobento can you please move this conversation to a Jira? Also would be good to keep this conversation there. Thanks!

ederign:
Let's merge this task and move one. As soon as @porcelli is back we can discuss this with UX.

tiagobento:
Created this JIRA https://issues.jboss.org/browse/AF-1172 so that this conversation can continue.

PR #1584, opened 2018-04-06
JBPM-7121: Form Editor breaks loading forms for non-existing models

pefernan:
A video showing the change: https://issues.jboss.org/secure/attachment/12433665/error_message.mp4

@jsoltes could you please review?


pefernan:

pefernan:

pefernan:
jenkins retest this

jsoltes:
Hello @pefernan,
Looks good! I have just few notes:
1. I remember we talked about disabling all buttons on form except of the Delete button. Did you decide for some different strategy?
2. As for the reason - "This usually means that Task has been removed from the Process..." - I am unable to reproduce this scenario. When I remove a task, then it's taskform isn't generated. How do you reproduce this?
3. I would place the "Show more" as a separate paragraph under the first line.
4. There is a small mistake in sentence "Something wrong happened loading the Form" -> correctly should be  "Something wrong happened while loading the Form".

pefernan:
@jsoltes yes there are some changes, but after a call with UX on Friday it seems that we are going to change the approach, so forget about this PR for now :S

PR #1593, opened 2018-04-09
AF-1120: Support standalone repositories as targets for the migration…

mbarkley:
… tool


mbarkley:
Related PR: https://github.com/kiegroup/appformer/pull/285

mbarkley:
@paulovmr @pefernan This PR adds support to migrate a repository containing projects as an alternative target to an niogit folder. It still only supports file paths as a target (i.e. no git or http URLs), but if the target path is not an niogit directory, the tool will:

1. Create a temporary niogit directory
2. Import the repository
3. Migrate projects to separate repositories
4. Prompt the user for other migrations w/ the previous wizard (currently only form migration shows here)
5. Export the project repositories to an output folder given by the user

If the target path is an niogit folder, the tool behaves exactly as before.

//cc @porcelli @ederign

mbarkley:
I just updated the README. I've removed the documentation of the batch usage, since it is not respected when migrating a repository (the user is always prompted for an output location). I think we should remove the feature.

mbarkley:
Jenkins, please retest this.

PR #1580, opened 2018-04-06
JBPM-6753: Rename server capabilities

jomarko:
Replace terms:
- Rule with Decision
- Planning with Planner

@karreiro would you mind have a look please?

PR #1604, opened 2018-04-10
[7.7.x] JBPM-6753: Rename server capabilities

jomarko:
Replace terms:
- Rule with Decision
- Planning with Planner

This is cherry-pick of #1580 

@karreiro please, would you mind to have a look?


jomarko:
@manstis this is ready for merge

PR #1611, opened 2018-04-10
(6.4.z) [RHBRMS-2597] Unlocalized string "Nothing selected" in data object new field dialog.

elguardian:
Added logic to change the locale of the select component to the one being used by the user.


jomarko:
ok to test

manstis:
@wmedvede Can you confirm this has a JIRA and ACKs for merge into 6.5.x?

PR #1622, opened 2018-04-11
[AF-1103]: Fixed issue that redirects to the list of assets when switching to project view

adrielparedes:
Added **requiresRefresh** attribute to know if changes in context needs to refresh to use the new data. When TechnicalView or BusinessView is clicked it doesn't need to go to Library if any value is changed, like pom.xml. That's why WorkspaceProjectContextChangeEvent has the requiredRefresh attribute set to false. In that case onChange method in LibraryPlaces understand that values changed but it doens't go to library after those changes.

@barboras7 @jhrcek @Rikkola Would you mind to review?


kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
```diff
@@ -739,7 +740,9 @@ public void init(final LibraryPerspective libraryPerspective) {
     public void onChange(WorkspaceProjectContextChangeEvent previous,
                          WorkspaceProjectContextChangeEvent current) {
         if (current.getWorkspaceProject() != null) {
-            if (!current.getWorkspaceProject().equals(previous.getWorkspaceProject())) {
+            if (Utils.hasProjectChanged(previous.getWorkspaceProject(),
+                                        current.getWorkspaceProject()) &&
+                    current.requiresRefresh()) {
```
Rikkola: Personally I am not a fan of a boolean flag for this. In time we might end up needing more and more of these and hard to tell what the `true` or `false` in the constructors stand for.

And thinking about this `onChange`. When do we need to go back to the `ProjectScreen`? When the project changes to another project.

I tried to think how to know if the project changed and the best way would be just to compare if the `repositories` in both `WorkspaceProjects` are equal. This way you can change the name, pom whatever configuration there is and this does not take you to `ProjectScreen`.

adrielparedes: Sounds great, let me modify it, smoke it a bit and I will update the PR. I think you are right with what you said :)

adrielparedes:
jenkins retest please

adrielparedes:

kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/utils/Utils.java
```diff
@@ -116,6 +117,48 @@ public static boolean hasPackageChanged(final Package pkg,
         return !pkg.equals(activePackage);
     }
 
+    /**
+     * A convenience method to compare two Repositories avoiding cluttering code with null checks.
+     * @param repository
+     * @param activeRepository
+     * @return
+     */
+    public static boolean hasRepositoryChanged(final Repository repository,
```
jhrcek: @adrielparedes Nice solution. But both of these static methods beg for some unit tests :-)

adrielparedes:
@Rikkola @jhrcek ready! thanks guys!

adrielparedes:

PR #1625, opened 2018-04-12
JBPM-7032: Stunner - Missing Overview panel

manstis:
See https://issues.jboss.org/browse/JBPM-7032

Tested with showcase-project for both BPMN and CM. Overview is included.

Tested with showcase-standalone for both BPMN and CM. Overview is not included.

Tested with DMN. Overview is not included - as DMN only has a showcase-standalone.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-project-client/src/main/java/org/kie/workbench/common/dmn/project/client/editor/DMNDiagramEditor.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-project-client/src/main/java/org/kie/workbench/common/dmn/project/client/editor/DMNDiagramEditor.java
@@ -22,7 +22,6 @@ import javax.enterprise.event.Event;
 import javax.inject.Inject;
 
 import com.google.gwt.user.client.ui.IsWidget;
-import com.google.gwt.user.client.ui.Widget;
 import org.kie.workbench.common.dmn.api.factory.DMNGraphFactory;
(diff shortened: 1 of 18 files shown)


manstis:
Jenkins please retest this.

manstis:
Jenkins please execute full downstream build (so @romartin has WARs to use).

romartin:
Oh! got confused with another PR and was sure the downstream build was already running.. :O sorry :+1:

manstis:
Hi @jomarko 

Let me check the error message (although it looks like you've selected a "delete" commit; so there is no file and the error message is correct). If all you did was create, save, modify, save, modify, save etc then Stunner may be doing something unusual. Let me check.

As for the "rules" message. The "Overview" widget is shared across all asset types and the text is common across the old Designer and all other asset types. It's probably best addressed as a new JIRA to correct the generic wording.

manstis:
@jomarko @romartin There is a new commit on this PR fixing an issue @jomarko (sort of) found.

During his testing he reported a popup when selecting a "delete" commit. During my investigation I did not created, delete, create, delete (to get the history @jomarko used) but did create, update, create, update (giving a different history). When selecting an older version on the Overview tab an error was thrown (and now _worked around_ in the new commit). The work around is currently used by all other editors - unknowingly - as it masks an issue with the "Project = Repository" changes. I'll raise a new JIRA for @Rikkola (or someone else) to investigate.

manstis:
Jenkins please retest this.

PR #1615, opened 2018-04-11
AF-1143 : Not possible to work with repository with project having invalid/missing pom.xml

Rikkola:
https://issues.jboss.org/browse/AF-1143

Makes it possible to open broken projects, but you can't build, add assets or edit configurations. You do get a warning about the project being broken.

PR #1629, opened 2018-04-13
AF-1123 : Broken project cannot be deleted

Rikkola:
https://issues.jboss.org/browse/AF-1123

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/resources/i18n/LibraryConstants.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/resources/i18n/LibraryConstants.java
@@ -288,9 +288,6 @@ public class LibraryConstants {
     public static final String DeletingProject = "DeletingProject";
 
     @TranslationKey(defaultValue = "")
-    public static final String DeleteProjectSuccess = "DeleteProjectSuccess";
-
-    @TranslationKey(defaultValue = "")
     public static final String WrongProjectConfirmedName = "WrongProjectConfirmedName";
 
     @TranslationKey(defaultValue = "")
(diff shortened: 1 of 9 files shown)


Rikkola:
@barboras7 @paulovmr Follow up for #1615

Rikkola:
I'll group this and the fix for AF-1123 in on PR for 7.5.x and 7.7.x once we have this in.

barboras7:

barboras7:
Hi @Rikkola,
tried this locally by following the steps to reproduce from the AF-1123. What happened is that I get the "The project is broken" screen but there is "Loading assets..." popup that does not disappear and the business central seems to be broken (you can not do anything at all) and you need to refresh the page. 
screenshot from 2018-04-16 16-05-30

Rikkola:
@barboras7 I need to double check this PR has the latest. I had the same problem, but fixed it.

Rikkola:
@barboras7 You are right. Import leaves the popup open.

Rikkola:
@barboras7 Ok. Fixed the import.

barboras7:

barboras7:

barboras7:

barboras7:
jenkins please retest this

barboras7:
@Rikkola, I tested the fix locally and the above mentioned problem is fixed. What happens now is that if you import the broken project as described in the steps to reproduce of the issue, you are not able to access the project settings. You can open the Contributors and Metrics tab, but the Settings tab does not work. Is this intentional behaviour? This way, user is not able to access the pom.xml file.

Rikkola:
Yes. Since you can not fix the pom.xml with the Project Editor and you can't break it either. At this point you can only do this with cloning the repository, so that is how it needs to be fixed.

barboras7:
@Rikkola, yes, that makes sense. Maybe disabling the Settings tab in a broken project might be a good approach so that it is obvious that it should not work for now. Anyway, I'm not sure how much effort would be necessary for this and whether it's worth the time.

Rikkola:

Rikkola:
Jenkins, I'm not going to even ask you test this today since even I can't build anything locally. See you tomorrow.

Rikkola:
All checks passed! This can go in :)

PR #1671, opened 2018-04-23
[AF 1140] Poms Migration tool

desmax74:
@porcelli @jakubschwan feedbacks are welcome


kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,414 @@
...
+        }
+    }
+
+
+    /***************************************** Start PluginRepositories TAG *****************************************/
+
+    private boolean write(Model model, String absolutePath){
+
+        try {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
jhrcek: You're not closing this stream. Please use `try-with-resources` to guarantee stream is closed automatically
```
try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
    ...
}
```

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,414 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(PomEditor.class);
+    private MavenXpp3Reader reader;
+    private MavenXpp3Writer writer;
+    private static final String PROPERTIES_FILE = "PomMigration.properties";
+    public final String TRUE = "true";
+    protected String FILE_URI = "file://";
+    private static final String KIE_VERSION_KEY = "KIE_VERSION";
+    private String kieVersion;
+    private String SCOPE_PROVIDED = "provided";
```
jhrcek: How about making all of these final (or don't use Uppercase names fi they're not constants) to avoid confusion?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,414 @@
...
+    /***************************************** END Build TAG *****************************************/
+    /***************************************** Start Dependencies TAG *****************************************/
+
+    private void updateDependenciesTag(Model model) {
+        model.setDependencies(applyMandatoryDeps(getChangedCurrentDependencies(model.getDependencies())));
+    }
+
+    private List<Dependency> applyMandatoryDeps(List<Dependency> dependencies) {
+        Set<Dependency> uniques = new HashSet<>();
+        uniques.addAll(dependencies);
```
jhrcek: You don't need to call addAll - you can pass the original collection directly to constructor

```
Set<Dependency> uniques = new HashSet<>(dependencies);
```

desmax74: @jhrcek a explanation to avoid a waste of energy, there are a lot of changes because the origin of the branchis the maven compiler branch, the code and files to review are in the module kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration, you could skip the files in others module. Thanks

jhrcek: Ok, I'll focus on that module only.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -49,82 +50,82 @@
 
 public class PomEditor {
 
+    private static final String PROPERTIES_FILE = "PomMigration.properties";
+    private static final String KIE_VERSION_KEY = "KIE_VERSION";
+    public final String TRUE = "true";
     private final Logger logger = LoggerFactory.getLogger(PomEditor.class);
+    //protected String FILE_URI = "file://";
+    protected String FILE_URI = "file:" + File.separator + File.separator;
```
jhrcek: This one is probably wrong - file URL should always have "//" , regardless of what OS we are on.
`protected String FILE_URI = "file://";` was correct.

kie-wb-common-ala/kie-wb-common-ala-wildfly/kie-wb-common-ala-wildfly-provider/pom.xml
```diff
@@ -156,7 +156,7 @@
           <executions>
             <execution>
               <id>ban-duplicated-classes</id>
-              <phase>validate</phase>
+              <phase>accept</phase>
```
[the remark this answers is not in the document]
jhrcek: @desmax74 Even though this is outside of the scope of the review, this should be probably reverted. It looks like you did some renaming refactoring in your IDE (renaming validate to accept), but this renamed all textual occurences of validate -just search for `accept` in this PR and you'll see many invalid replacements in comments etc.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PluginPresence.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.project.migration.cli.maven;
+
+public class PluginPresence {
+
+    private boolean isPresent;
+    private int position = -1;
```
jhrcek: You can just remove this `= -1` because this field is always set by the constructor

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,400 @@
...
+    public Model updatePom(Path pom) {
+        try {
+            Model model = getModel(pom);
+            Build build = getBuild(model);
+            updateBuildTag(build);
+            updateDependenciesTag(model);
+            updateRepositories(model);
+            updatePluginRepositories(model);
+            boolean writed = write(model, pom.toAbsolutePath().toString());
+            if (writed) {
```
jhrcek: typo writed -> written

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomProcessorTest.java
```diff
@@ -0,0 +1,59 @@
...
+    private PomProcessor processor;
+    private String GENERIC_PRJ = "/target/test-classes/generic/";
+
+    @Before
+    public void setUp() {
+        processor = new PomProcessor();
+    }
+
+    @Test
+    @Ignore
```
jhrcek: Please add info about why this is ignored (something based on which someone can decide about when to unignore it, otherwise it will be ignored forever) or remove the annotation.

`@Ignore("Failing because ABC. Can be unignored after XYZ is fixed")`

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomProcessorTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public void processProjectTest() throws Exception {
+        String currentDir = new File(".").getAbsolutePath();
+        Path path = Paths.get("file:" + currentDir + GENERIC_PRJ);
+        Path pathCopy = Paths.get("file:" + currentDir + "copy" + GENERIC_PRJ);
+        FileUtils.copyDirectory(path.toFile(), pathCopy.toFile());
+        try {
+            int pomProcessed = processor.processProject(path.toAbsolutePath().toString());
+            assertTrue(pomProcessed == 1);
```
jhrcek: `asserEquals(1, pomProcessed)` is better

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomProcessorTest.java
```diff
@@ -0,0 +1,59 @@
...
+        } catch (Exception e) {
+            fail(e.getMessage());
+        } finally {
```
jhrcek: Please use `@After` for cleanup instead of finally.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,400 @@
...
+
+    private Properties loadProperties(String propName) {
+        Properties prop = new Properties();
+        InputStream in = getClass().getClassLoader().getResourceAsStream(propName);
+        if (in == null) {
+            logger.info("{} not available with the classloader, skip to the next ConfigurationStrategy. \n", propName);
+        } else {
+            try {
+                prop.load(in);
+                in.close();
```
jhrcek: Please close this stream in finally :)

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,400 @@
...
+            }
+        }
+    }
+
+    /***************************************** Start PluginRepositories TAG *****************************************/
+
+    private boolean write(Model model, String absolutePath) {
+
+        try {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
jhrcek: You're not closing this stream. Please use try-with-resources to guarantee this stream is closed
```
try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
   ...
}
```

kie-wb-common-ala/kie-wb-common-ala-wildfly/kie-wb-common-ala-wildfly-provider/pom.xml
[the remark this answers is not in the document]
desmax74: I'll align with the master to fix all this "spurious" changes

desmax74:
@jhrcek now is more clean and simple to review, applied the previous suggestions :)

desmax74:
@porcelli  Alex checks if this meet all new requirements about the pom changes

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+    </dependency>
+    <!--
+
+    <dependency>
```
jhrcek: Let's not commit commented-out configuration, please.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PromptPomMigrationService.java
```diff
@@ -0,0 +1,84 @@
...
+        String response;
+        do {
+            response = console.readLine("Do you wish to continue? [yes/no]: ").toLowerCase();
+        } while (!validResponses.contains(response));
+        return "yes".equals(response);
+    }
+
+    public String promptForExternalConfiguration() {
+        SystemAccess.Console console = system.console();
+        console.format("WARNING: Please ensure that you have made backups of the directory [%s] before proceeding.\n", niogitDir);
```
jhrcek: Do we need 2nd prompt for backup here? It's already in `promptForBackup()`..

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,301 @@
...
+            updateRepositories(model);
+            updatePluginRepositories(model);
+            boolean written = write(model, pom.toAbsolutePath().toString());
+            if (written) {
+                return model;
+            } else {
+                return new Model();
+            }
+        } catch (Exception e) {
+            System.out.println("Error occurred during POMs migration:" + e.getMessage());
```
jhrcek: We should probably use `SystemAccess` or logging here instead of `System.out.println`. The same on line 104

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
```diff
@@ -0,0 +1,177 @@
...
+            logger.error(e.getMessage());
+            throw new RuntimeException(e.getMessage());
+        }
+    }
+
+    public PomJsonReader(InputStream in) {
+        try {
+            JsonReader reader = Json.createReader(in);
+            pomObject = reader.readObject();
+            reader.close();
```
jhrcek: Please close in finally, or even better use try-with-resources which does that automatically for anything implementing `AutoCloseable`.

```
try (JsonReader reader = Json.createReader(in)) {
    pomObject = reader.readObject();
} catch (Exception e) {
    logger.error(e.getMessage());
    throw new RuntimeException(e.getMessage());
 }
```

desmax74:
@jhrcek @jakubschwan is it ok ?

porcelli:
@jhrcek this is for GA, so we need get it merged sooner than later.
@krisv what are your thoughts about this timeline?

krisv:
@porcelli @jhrcek yes, ideally we would like to see this merged this week but I realize that might be very difficult.  Our code freeze is next Wed so we would really need to be merged by then in 7.7.x or it would miss the CR1 cut.

jhrcek:
Ok, I'll test this locally today and provide feedback asap.

desmax74:
@jhrcek about the blocker and major issues

1) If a git root repo is used a barerepo is copied and no pom.xml are founded in the tmpdir @porcelli in which branch of the https://github.com/desmax74/kie-wb-common/blob/AF-1140/kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-api/src/main/java/org/kie/workbench/common/migration/cli/MigrationApp.java#L40 the PomMigration is mandatory ? 

2) Fixed

3) I could fix adding a logback.xml in the kie-wb-common-cli-api but in the ER5 
RHPAM-7.0.0.ER5 isn't present and the debug msgs aren't present

4) Fixed

5) Fixed

jhrcek:
@desmax74 regarding issue 3: When you run migration in ER5, you'll see this in the output:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Which means that ER5 doesn't have logging implementation (confirmed that by looking into migration-tool/lib/).

Your PR adds `logback-classic `and `logback-core` which ends up in the zip, which leads to those debug messages being printed. I think that adding reasonable logback.xml is a good idea!

jhrcek:
@desmax74 So I tested this locally and the migrated pom.xml are written to the migrated repositories :+1: 

The bad thing though is that the tool still produces crazy amounts of logging output. There's no more  DEBUG output, but there are still tons of info and warning messages - see example output 
[migrationOutput2.txt](https://github.com/kiegroup/kie-wb-common/files/1979052/migrationOutput2.txt)

The main problem with this is that at the end of migration there's a ton of exception messages (probably result of stopping weld container), in which the tool prompts are completely lost (so I can't see things like "Press 4 to exit").

I think we should disable console logging (so that only things logged via `SystemAccess` would be visible) and only redirect logging output to some log file (so customers can potentially send this file for debugging) - wdyt?

desmax74:
@jhrcek The log messages could be reduced with different log level. The exceptions are the output of the automatic builds failed due to the dependencies without version because aren't our dependencies (like internal rh or qe deps), one solution is stop the build on updates on  files or disable the output and write on a log file like you are proposing, if we prevent the automatic build the users could see this problems on the pom only when they run a build, if we not prevent the automatic build they reads this problems on a file later (with a correct appender log configuration) wdyt ? And about the path of the log file, which path we have to use ?

desmax74:
@jhrcek changed in the following way, the console shows only the list of the projects at the end the message "End POMs migration, detailed log available on <user.home>/pom_migration.log", the errors are contained in the file log

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-api/src/main/java/org/kie/workbench/common/migration/cli/MigrationSetup.java
```diff
@@ -28,5 +28,6 @@ public static void configureProperties(final SystemAccess system,
         system.setProperty(JGitFileSystemProviderConfiguration.GIT_NIO_DIR_NAME, niogitDir.getFileName().toString());
         system.setProperty(JGitFileSystemProviderConfiguration.GIT_DAEMON_ENABLED, "false");
         system.setProperty(JGitFileSystemProviderConfiguration.GIT_SSH_ENABLED, "false");
+        //system.setProperty("org.uberfire.watcher.autostart", "false");//uncommnet to disabled the automatic build on update files
```
jhrcek: +1 to uncomment this line - in my opinion it makes sense to disable automatic build after migration.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,161 @@
...
+
+                    WorkspaceProjectService service = weldContainer.instance().select(WorkspaceProjectService.class).get();
+
+                    Collection<WorkspaceProject> projects = service.getAllWorkspaceProjects();
+                    for(WorkspaceProject prj : projects){
+                        processWorkspaceProject(prj, jsonPath);
+                    }
+                }
+                String userHome = System.getProperty("user.home");
+                system.out().println("End POMs migration, detailed log available on "+userHome + "/migration_tool.log");
```
jhrcek: "End POMs migration" -> Finished POMs migration

jhrcek: log available on -> log available in

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,161 @@
...
+               if (systemMigrationWasExecuted()) {
+
+                    WorkspaceProjectService service = weldContainer.instance().select(WorkspaceProjectService.class).get();
+
+                    Collection<WorkspaceProject> projects = service.getAllWorkspaceProjects();
+                    for(WorkspaceProject prj : projects){
+                        processWorkspaceProject(prj, jsonPath);
+                    }
+                }
+                String userHome = System.getProperty("user.home");
```
jhrcek: In my opinion using "user.dir" (the dir from which user started the tool) makes more sense.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-api/src/main/java/org/kie/workbench/common/migration/cli/MigrationSetup.java
desmax74: @jhrcek if this line is uncommented the watching service is stopped and when a pom is changed the automatic build won't start, the result is the migration_tool.log will be without errors, the user will see the errors due to lack of version in the user dependencies only when a build will starts. Is this the desidered behaviour ?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
desmax74: Ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
desmax74: Ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-api/src/main/java/org/kie/workbench/common/migration/cli/MigrationSetup.java
jhrcek: I'm not sure. Let's see what others think about this setting.. @mbarkley ?

pefernan: +1 to uncomment this, not sure what side effects could appears, but I think it would be ok.

desmax74:
jenkins retest this

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/ServiceCDIWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.inject.Named;
+
+import org.guvnor.common.services.backend.util.CommentedOptionFactory;
+import org.guvnor.structure.repositories.Repository;
+import org.uberfire.backend.server.util.Paths;
+import org.uberfire.backend.vfs.Path;
+import org.uberfire.io.IOService;
+
+@ApplicationScoped
+public class ServiceCDIWrapper {
```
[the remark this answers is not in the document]
desmax74: Ok, where do you suggest to place ?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
```diff
@@ -0,0 +1,190 @@
...
+
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Repository;
+import org.apache.maven.model.RepositoryPolicy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * It reads the pom-migrations.json with dependencies, repositories and plugin repositories with the following format
+ * <p>
```
mbarkley: Having the example is good, but it isn't readable as is. I think if you use the `<pre></pre>` tags, the line-wrapping will be preserved in the javadocs.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
```diff
@@ -0,0 +1,190 @@
...
+ * "snapshotEnabled":true,
+ * "snapshotUpdatePolicy":"never"
+ * }
+ * ]
+ * }
+ */
+public class PomJsonReader {
+
+    private final Logger logger = LoggerFactory.getLogger(PomJsonReader.class);
+    private String JSON_POM_FILE;//= "pom-migration.json";
```
mbarkley: Since this isn't a constant, could you change it to camel case?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
```diff
@@ -0,0 +1,190 @@
...
+        JsonReader reader = null;
+        try {
+            fis = new FileInputStream(path);
+            reader = Json.createReader(fis);
+            pomObject = reader.readObject();
+            reader.close();
+        } catch (Exception e) {
+            logger.error(e.getMessage());
+            throw new RuntimeException(e.getMessage());
+        }finally {
```
mbarkley: Some of the spacing is off in this block. Maybe run the formatter?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
```diff
@@ -0,0 +1,190 @@
...
+            }
+        }
+    }
+
+    public PomJsonReader(InputStream in) {
+
+        try (JsonReader reader = Json.createReader(in)) {
+            pomObject = reader.readObject();
+        } catch (Exception e) {
+            logger.error(e.getMessage());
```
mbarkley: You probably want to add the exception as the second argument to keep the stacktrace. Also, in the line below it would be good to add the caught exception as a cause to the thrown exception.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
```diff
@@ -0,0 +1,188 @@
...
+            assertEquals(modelUpdated.getPackaging(),"kjar");
+            assertNotNull(modelUpdated);
+            assertTrue(modelUpdated.getBuild().getPlugins().size() == 1);
+            List<Dependency> deps = modelUpdated.getDependencies();
+            for (Dependency dep : deps) {
+                assertTrue(dep.getVersion() != null);
+            }
+        } catch (Exception e) {
+            logger.error(e.getMessage());
+            fail(e.getMessage());
```
mbarkley: Definitely we'll want to see the stacktrace if this test ever fails. The pattern I'd recommend for this is to manually throw an  `AssertionError` instead of calling `fail`. Then you can add `e` as a cause.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+            Model modelUpdated = editor.updatePom(path, jsonPath.toAbsolutePath().toString());
+            assertNotNull(modelUpdated);
+            assertEquals(modelUpdated.getPackaging(),"kjar");
+            assertTrue(modelUpdated.getBuild().getPlugins().size() == 1);
+            assertTrue(modelUpdated.getDependencies().size() == 6);
+            assertTrue(modelUpdated.getRepositories().size() == 2);
+            assertTrue(modelUpdated.getPluginRepositories().size() == 2);
+        } catch (Exception e) {
+            logger.error(e.getMessage());
```
mbarkley: Same note on showing stracktrace. See comment above.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
```diff
@@ -0,0 +1,188 @@
...
+            assertTrue(original.getRepositories().size() == 0);
+            assertTrue(original.getPluginRepositories().size() == 0);
+
+            Model modelUpdated;
+            modelUpdated = editor.updatePom(path, null);
+            assertTrue(modelUpdated.getGroupId() == null);
+            modelUpdated = editor.updatePom(null, null);
+            assertTrue(modelUpdated.getGroupId() == null);
+        } catch (Exception e) {
+            logger.error(e.getMessage());
```
mbarkley: Same comment as above.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,161 @@
...
+    private String POM_DOT_XML = "pom.xml";
+
+    @Override
+    public String getTitle() {
+        return "POMs migration";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Migrates pom.xml files to format required in 7.7.x and later (community)";
```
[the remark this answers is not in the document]
desmax74: I agree

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,161 @@
...
+                            }
+                        } catch (Exception e) {
+                            system.err().println("Error reading form: " + fileName + ":\n");
+                            e.printStackTrace(system.err());
+                        }
+                }
+                return FileVisitResult.CONTINUE;
+            }
+        });
+        system.out().println("Migrated " + counter[0] + " POMs for prj:"+workspaceProject.getName());
```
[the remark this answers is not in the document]
desmax74: Ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,161 @@
...
+                            if (jsonPath.isEmpty()) {
+                                model = editor.updatePom(visitedPath);
+                            } else {
+                                model =  editor.updatePom(visitedPath, jsonPath);
+                            }
+                            if (!model.getBuild().getPlugins().isEmpty()) {
+                                counter[0]++;
+                            }
+                        } catch (Exception e) {
+                            system.err().println("Error reading form: " + fileName + ":\n");
```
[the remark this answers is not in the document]
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependencyKey.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Objects;
+
+import org.apache.maven.model.Dependency;
+
+public class DependencyKey {
+
+    private Dependency dependency;
+
+    public DependencyKey(Dependency dep) {
+        this.dependency = dep;
```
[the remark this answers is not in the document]
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,343 @@
...
+                || groupID.startsWith(DROOLS_PKG)
+                || groupID.startsWith(JBPM_PKG));
+    }
+
+    /***************************************** End Dependencies TAG *****************************************/
+    /***************************************** Start Repositories TAG *****************************************/
+
+    private void updateRepositories(Model model) {
+        List<Repository> repos = model.getRepositories();
+        if (repos == null || repos.isEmpty()) {
```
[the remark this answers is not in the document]
desmax74: Correct isn't needed

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,343 @@
...
+            }
+        }
+    }
+
+    /***************************************** End Repositories TAG *****************************************/
+    /***************************************** Start PluginRepositories TAG *****************************************/
+
+    private void updatePluginRepositories(Model model) {
+        List<Repository> repos = model.getPluginRepositories();
+        if (repos == null || repos.isEmpty()) {
```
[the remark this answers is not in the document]
desmax74: Correct isn't needed

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,343 @@
...
+                            new String(baos.toByteArray(),
+                                       StandardCharsets.UTF_8));
+            }
+
+            cdiWrapper.write(org.uberfire.backend.server.util.Paths.convert(path),
+                             new String(baos.toByteArray(), StandardCharsets.UTF_8),
+                             "Pom's Migration" + path.toString());
+            return true;
+        } catch (Exception e) {
+            logger.error(e.getMessage());
```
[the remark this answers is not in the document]
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,343 @@
...
+        } finally {
+            try {
+                baos.close();
+            } catch (IOException e) {
+                //suppressed
+            }
+        }
+    }
+
+    private boolean writeToFile(Model model, String absolutePath) {
```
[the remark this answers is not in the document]
desmax74: +1

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
desmax74: Correct

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
desmax74: Ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomJsonReader.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
desmax74: sounds like a GOTO but is ok in a test

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
desmax74: changed in the same way

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
desmax74: changed in the same way

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PromptPomMigrationService.java
```diff
@@ -0,0 +1,68 @@
...
+                }
+            } while (response != null && !valid);
+            return response + File.separator + JSON_POM_FILE;
+        } else {
+            return "";
+        }
+    }
+
+    private boolean jsonExists(String path) {
+        File test = new File(path + File.separator + JSON_POM_FILE);
```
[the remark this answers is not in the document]
desmax74: Just replaced, thanks for the tip :)

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,161 @@
...
+        this.system = system;
+
+        PromptPomMigrationService promptService = new PromptPomMigrationService(system, config);
+        String jsonPath = promptService.promptForExternalConfiguration();
+        Path niogitDir = config.getTarget();
+        system.out().println("Starting POMs migration");
+        if (projectMigrationWasExecuted()) {
+            try {
+                MigrationSetup.configureProperties(system, niogitDir);
+                weldContainer = new Weld().initialize();
```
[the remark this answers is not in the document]
desmax74: Ok

desmax74:
@mbarkley changes applied

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-api/src/main/java/org/kie/workbench/common/migration/cli/MigrationSetup.java
```diff
@@ -28,5 +28,6 @@ public static void configureProperties(final SystemAccess system,
         system.setProperty(JGitFileSystemProviderConfiguration.GIT_NIO_DIR_NAME, niogitDir.getFileName().toString());
         system.setProperty(JGitFileSystemProviderConfiguration.GIT_DAEMON_ENABLED, "false");
         system.setProperty(JGitFileSystemProviderConfiguration.GIT_SSH_ENABLED, "false");
+        system.setProperty("org.uberfire.watcher.autostart", "false");//commen2t to enable the automatic build on update files
```
porcelli: the comment `//commen2t to enable the automatic build on update files` could be removed

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-forms-migration/src/main/java/org/kie/workbench/common/forms/migration/tool/cdi/MigrationServicesCDIWrapper.java
```diff
@@ -40,6 +40,8 @@
 
     private IOService systemIoService;
 
+    //@TODO move in shared place to reuse in all migration tools
```
porcelli: todo? why not move?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
```diff
@@ -0,0 +1,154 @@
...
+import org.uberfire.java.nio.file.FileVisitResult;
+import org.uberfire.java.nio.file.Files;
+import org.uberfire.java.nio.file.SimpleFileVisitor;
+import org.uberfire.java.nio.file.attribute.BasicFileAttributes;
+
+public class PomMigrationTool implements MigrationTool {
+
+    private SystemAccess system;
+    private ToolConfig config;
+    private String POM_DOT_XML = "pom.xml";
```
porcelli: static final?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PromptPomMigrationService.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final String JSON_POM_FILE = "pom-migration.json";
+    private SystemAccess system;
+
+    public PromptPomMigrationService(final SystemAccess system) {
+        this.system = system;
+    }
+
+    public String promptForExternalConfiguration() {
+        SystemAccess.Console console = system.console();
+        Collection<String> validResponses = Arrays.asList("yes", "no");
```
porcelli: instead of "yes" "no" strings, please use Enum

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,311 @@
...
+import org.uberfire.java.nio.file.Path;
+
+public class PomEditor {
+
+    private static final String PROPERTIES_FILE = "PomMigration.properties";
+    private static final String JSON_POM_MIGRATION = "pom-migration.json";
+    private static final String JSON_POM_MANDATORY_DEPS = "pom-mandatory.json";
+    private static final String KIE_VERSION_KEY = "KIE_VERSION";
+    private final Logger logger = LoggerFactory.getLogger(PomEditor.class);
+    private String FILE_URI = "file://";
```
porcelli: static final

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
```diff
@@ -0,0 +1,311 @@
...
+    private MavenXpp3Writer writer;
+    private String kieVersion;
+    private String KIE_PKG = "org.kie";
+    private String JBPM_PKG = "org.jbpm";
+    private String DROOLS_PKG = "org.drools";
+    private String OPTAPLANNER_PKG = "org.optaplanner";
+    private String KJAR_PKG = "kjar";
+    private String POM_PKG = "pom";
+    private String KIE_MAVEN_PLUGIN_ARTIFACT_ID = "kie-maven-plugin";
+    private JSONDTO jsonConf;
```
porcelli: all upper cased should be static final

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
```diff
@@ -0,0 +1,194 @@
...
+    private final Logger logger = LoggerFactory.getLogger(PomEditorTest.class);
+    private PomEditor editor;
+    private String currentDir;
+    private String CURRICULUM_COURSE_PRJ = "/target/test-classes/curriculumcourse/";
+    private String DINNER_PARTY_PRJ = "/target/test-classes/dinnerparty/";
+    private String EMPLOYEE_ROSTERING_PRJ = "/target/test-classes/employee-rostering/";
+    private String EVALUATION_PRJ = "/target/test-classes/employee-rostering/";
+    private String ITORDERS_PRJ = "/target/test-classes/itorders/";
+    private String MORTGAGES_PRJ = "/target/test-classes/mortgages/";
+    private String OPTACLOUD_PRJ = "/target/test-classes/optacloud/";
```
porcelli: uppercase == static final

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PromptPomMigrationService.java
```diff
@@ -0,0 +1,63 @@
...
+        String response;
+        do {
+            response = console.readLine("Do you want provide a path to an external Pom migration json  with dependencies and repositories ? [yes/no]: ").toLowerCase();
```
porcelli: extra whitespace

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/ServiceCDIWrapper.java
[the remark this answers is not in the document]
porcelli: if shared is needed, maybe create a module cli-shared

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependencyKey.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public Dependency getDependency() {
+        return dependency;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
```
porcelli: null check?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependenciesCollection.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.kie.workbench.common.project.migration.cli.maven;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+
+import org.apache.maven.model.Dependency;
+
+public class DependenciesCollection {
```
porcelli: Pojo missing toString, equals and hashCode

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/JSONDTO.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package org.kie.workbench.common.project.migration.cli.maven;
+
+import java.util.List;
+
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Repository;
+
+public class JSONDTO {
```
porcelli: missing toString, equals and hashCode

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PluginPresence.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.project.migration.cli.maven;
+
+public class PluginPresence {
```
porcelli: toString, equals and hashCode

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-forms-migration/src/main/java/org/kie/workbench/common/forms/migration/tool/cdi/MigrationServicesCDIWrapper.java
desmax74: Because it requires works on the Pere's module

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-api/src/main/java/org/kie/workbench/common/migration/cli/MigrationSetup.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PomMigrationTool.java
desmax74: only final, is here to avoid a "pom.xml" on a if check

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PromptPomMigrationService.java
desmax74: Is the pattern used in the other modules

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
desmax74: removed wasn't used

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependencyKey.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-forms-migration/src/main/java/org/kie/workbench/common/forms/migration/tool/cdi/MigrationServicesCDIWrapper.java
porcelli: check with pere, and move it to a common module.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependenciesCollection.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/JSONDTO.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PluginPresence.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PomEditor.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
desmax74: ok

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/PromptPomMigrationService.java
desmax74: fixed

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependenciesCollection.java
```diff
@@ -47,4 +49,30 @@ public void addDependenciesKeys(List<DependencyKey> deps) {
...
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("DependenciesCollection{");
+        sb.append("keys=").append(keys);
+        sb.append('}');
+        return sb.toString();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        PortablePreconditions.checkNotNull("DependenciesCollection", o);
```
mbarkley: Hmmm, this will throw an exception if compared to `null`, which I think is not ideal. I hate to contradict @porcelli, but I think it was okay the way it was before because `instanceof` is always false for null objects, so a null argument would cause the method to exit on line 68.

@porcelli WDYT?

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependencyKey.java
mbarkley: See comment in equals method of dependency.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/JSONDTO.java
```diff
@@ -43,4 +45,35 @@ public JSONDTO(List<Dependency> dependencies, List<Repository> respositories, Li
...
+        sb.append("dependencies=").append(dependencies);
+        sb.append(", repositories=").append(repositories);
+        sb.append(", pluginRepositories=").append(pluginRepositories);
+        sb.append('}');
+        return sb.toString();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        PortablePreconditions.checkNotNull("JSONDTO", o);
```
mbarkley: Same comment as above.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/ServiceCDIWrapper.java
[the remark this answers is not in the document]
desmax74: Is fine the cli-api name, here the common parts of the MigrationServicesCDIWrapper will be moved to the cli-api, then the FormDefinitionSerializer will be moved from MigrationServiceCDIWrapper into FormsMigrationServiceCDIWrapper in the cli-forms-migration because it needs classes from the forms migration module

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/PluginPresence.java
```diff
@@ -32,4 +36,33 @@ public boolean isPresent() {
...
+        final StringBuilder sb = new StringBuilder("PluginPresence{");
+        sb.append("isPresent=").append(isPresent);
+        sb.append(", position=").append(position);
+        sb.append('}');
+        return sb.toString();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        PortablePreconditions.checkNotNull("PluginPresence", o);
```
mbarkley: Same comment as above.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/test/java/org/kie/workbench/common/project/migration/cli/maven/PomEditorTest.java
```diff
@@ -85,9 +87,13 @@ private void testDefault(String prj) {
             for (Dependency dep : deps) {
                 assertTrue(dep.getVersion() != null);
             }
-        } catch (Exception e) {
-            logger.error(e.getMessage());
-            fail(e.getMessage());
+        } catch (IOException ioex) {
+            logger.error(ioex.getMessage(), ioex);
+            throw new AssertionError(ioex.getMessage());
```
mbarkley: What I had in mind here, was this:

     throw new AssertionError("Test failed from error.", ioex);

That way when the test fails you will get the error stacktrace as part of the failure. Same applies to other AssertionErrors below.

kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/java/org/kie/workbench/common/project/migration/cli/maven/DependenciesCollection.java
porcelli: @mbarkley is correct that the use of checkNotNull is not ideal. When I suggested null check was more like what my IDE usually generates:

```java
        if (o == null) {
            return false;
        }
        if (this == o) {
            return true;
        }
        if (!(o instanceof ???)) {
            return false;
        }
        ??? that = (???) o;
        ....
```

desmax74:
jenkins retest this

porcelli:
@desmax74 found a small bug. existing dependencies are loosing the version, example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>guvnor.feature</groupId>
  <artifactId>dtables</artifactId>
  <version>1.0</version>
  <packaging>kjar</packaging>
  <name>dtables</name>
  <dependencies>
    <dependency>
      <groupId>org.group</groupId>
      <artifactId>workitems</artifactId>
      <version>1.0.0-FINAL</version>
    </dependency>
  </dependencies>
</project>
```

results in this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>guvnor.feature</groupId>
  <artifactId>dtables</artifactId>
  <version>1.0</version>
  <packaging>kjar</packaging>
  <name>dtables</name>
  <dependencies>
    <dependency>
      <groupId>com.thoughtworks.xstream</groupId>
      <artifactId>xstream</artifactId>
      <version>1.4.10</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.kie</groupId>
      <artifactId>kie-internal</artifactId>
      <version>7.8.0-SNAPSHOT</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>org.optaplanner</groupId>
      <artifactId>optaplanner-core</artifactId>
      <version>7.8.0-SNAPSHOT</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.12</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.group</groupId>
      <artifactId>workitems</artifactId>
    </dependency>
    <dependency>
      <groupId>org.optaplanner</groupId>
      <artifactId>optaplanner-persistence-jaxb</artifactId>
      <version>7.8.0-SNAPSHOT</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>org.kie</groupId>
      <artifactId>kie-api</artifactId>
      <version>7.8.0-SNAPSHOT</version>
      <scope>provided</scope>
    </dependency>
  </dependencies>
  <build>
    <plugins>
      <plugin>
        <groupId>org.kie</groupId>
        <artifactId>kie-maven-plugin</artifactId>
        <version>7.8.0-SNAPSHOT</version>
        <extensions>true</extensions>
      </plugin>
    </plugins>
  </build>
</project>
```

kie-wb-common-services/kie-wb-common-services-backend/src/main/resources/pom-mandatory.json
```diff
@@ -0,0 +1,10 @@
+{
+  "dependencies":[
+    {"groupId":"org.kie", "artifactId":"kie-api", "version":"${version.org.kie}", "scope":"provided"},
+    {"groupId":"org.kie", "artifactId":"kie-internal", "version":"${version.org.kie}", "scope":"provided"},
+    {"groupId":"org.optaplanner", "artifactId":"optaplanner-core", "version":"${version.org.kie}", "scope":"provided"},
+    {"groupId":"org.optaplanner", "artifactId":"optaplanner-persistence-jaxb", "version":"${version.org.kie}", "scope":"provided"},
+    {"groupId":"junit", "artifactId":"junit", "version":"4.12", "scope":"test"},
+    {"groupId":"com.thoughtworks.xstream", "artifactId":"xstream", "version":"1.4.10", "scope":"test"}
+  ]
+}
```
porcelli: this file needs to be filtered, this is what I get when I run it:

```java
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.myteam</groupId>
  <artifactId>MyTest</artifactId>
  <version>1.0.0</version>
  <packaging>kjar</packaging>
  <name>MyTest</name>
  <description></description>
  <dependencies>
    <dependency>
      <groupId>com.thoughtworks.xstream</groupId>
      <artifactId>xstream</artifactId>
      <version>1.4.10</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.kie</groupId>
      <artifactId>kie-internal</artifactId>
      <version>${version.org.kie}</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>org.optaplanner</groupId>
      <artifactId>optaplanner-core</artifactId>
      <version>${version.org.kie}</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.12</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.optaplanner</groupId>
      <artifactId>optaplanner-persistence-jaxb</artifactId>
      <version>${version.org.kie}</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>org.kie</groupId>
      <artifactId>kie-api</artifactId>
      <version>${version.org.kie}</version>
      <scope>provided</scope>
    </dependency>
  </dependencies>
  <repositories>
    <repository>
      <id>guvnor-m2-repo</id>
      <name>Guvnor M2 Repo</name>
      <url>file:///Users/porcelli/Desktop/wildfly-11.0.0.Final/repositories/kie/global</url>
    </repository>
  </repositories>
  <build>
    <plugins>
      <plugin>
        <groupId>org.kie</groupId>
        <artifactId>kie-maven-plugin</artifactId>
        <version>${version.org.kie}</version>
        <extensions>true</extensions>
      </plugin>
    </plugins>
  </build>
</project>
```

desmax74: @porcelli the version.org.kie was retrieved with https://github.com/desmax74/kie-wb-common/blob/AF-1140/kie-wb-common-cli/kie-wb-common-cli-tools/kie-wb-common-cli-pom-migration/src/main/resources/PomMigration.properties and was filtered in the cli-pom-migration. Added the filtering true in the kie-wb-common-services-backend to here  filter https://github.com/desmax74/kie-wb-common/tree/AF-1140/kie-wb-common-services/kie-wb-common-services-backend/src/main/resources

desmax74: Test added

PR #1679, opened 2018-04-24
JBPM 6491 - [Stunner] Improve usability and user experience when using connectors

tiagodolphine:
When creating sequence flows using gateways, now the default magnet assigned to the nodes consider the position to choose left or right magnet index. This is a quick win, maybe it can have more detailed rules to assign default connectors considering other restrictions, IMO this should be specified and handled on a new JIRA, after the release. WDYT ?
@romartin 
@LuboTerifaj 

Thanks.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/components/toolbox/actions/CreateNodeAction.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/components/toolbox/actions/CreateNodeAction.java
@@ -147,10 +147,12 @@ public class CreateNodeAction extends AbstractToolboxAction {
                                                                targetNode))
                         .deferCommand(() -> addEdge(canvasHandler,
                                                     sourceNode,
+                                                    targetNode,
                                                     connector))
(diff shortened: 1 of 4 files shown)


tiagodolphine:

tiagodolphine:
Jenkins retest this.

tiagodolphine:
Hey @romartin @LuboTerifaj  when you have a chance to take a look on this PR, this a simple change.
Thanks.

tiagodolphine:
Hi @LuboTerifaj if you have a chance to review this one, it is a simple change. 
Thanks.

LuboTerifaj:
Hi @tiagodolphine 
I am not sure what is the prupose of this PR (and its Jira). Since the sequence flows are not ortogonal anymore, it doesn't make sense to me. Can you please clarify it?
Thank you!

romartin:
hey @LuboTerifaj , well yes you're right, but it seems that it results more intuitive by using concrete magnets rather than using the current approach... so that's why finally decided to do this, just for improving the end user experience...

tiagodolphine:
Hey @LuboTerifaj even with the polyline connectors, if you create sequence flow from one gateway to 3 or more tasks for instance, you get the connectors overlapping the task shapes. As I understand for the JIRA description, the main idea is when you have this scenario the source magnet of the gateway should be the one on right and the magnet of the tasks should be the magnet at the left, for all connections. This could  be better for UX.
@romartin I don't know if you have more details about that but this was my understanding when we talk about this task.

Thanks.

tiagodolphine:
oh we answered at the same time :)

LuboTerifaj:

tiagodolphine:
@romartin this is ready to be merged.

PR #1689, opened 2018-04-25
JBPM-7178: [Stunner] Copy and paste lane with content

karreiro:
See https://issues.jboss.org/browse/JBPM-7178

Before:
before_

After:
after_

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/CloneNodeCommand.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/CloneNodeCommand.java
@@ -34,15 +34,18 @@ import org.kie.workbench.common.stunner.core.command.Command;
 import org.kie.workbench.common.stunner.core.command.CommandResult;
 import org.kie.workbench.common.stunner.core.command.impl.AbstractCompositeCommand;
 import org.kie.workbench.common.stunner.core.command.util.CommandUtils;
+import org.kie.workbench.common.stunner.core.definition.clone.CloneManager;
 import org.kie.workbench.common.stunner.core.definition.clone.ClonePolicy;
 import org.kie.workbench.common.stunner.core.graph.Edge;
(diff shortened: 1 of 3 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/CloneNodeCommand.java
```diff
@@ -118,6 +118,18 @@ protected boolean delegateRulesContextToChildren() {
...
 
+    void cloneNodeContentWithProperties(final GraphCommandExecutionContext context) {
+
+        final View cloneContent = getClone().getContent();
+        final View candidateContent = (View) getCandidate().getContent();
+        final Bounds contentBounds = candidateContent.getBounds();
+        final CloneManager cloneManager = context.getDefinitionManager().cloneManager();
+        final Object clonedDefinition = cloneManager.clone(candidateContent.getDefinition(), ClonePolicy.ALL);
+
+        cloneContent.setBounds(contentBounds);
```
tiagodolphine: I think it is better to actually clone the `contentBounds`, I mean creating a new instance with the same properties. Because in the way it is, the same reference to the contentBounds is used, so if it is modified on one point it will be reflected on the both, cloned and the original node. The idea of the `CloneManager` is to handle this kind of thing, so one suggestion would be to insert the logic on the clone manager to clone the bounds as well, WDYT ?

karreiro: @tiagodolphine I wasn't that concerned about this because the bounds are override [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/UpdateElementPositionCommand.java#L124). However, considering the name of the method `cloneNodeContentWithProperties`, I think that it's a good idea to make a real clone.

As we discussed on IRC, let's use the `BoundsImpl.build` instead of the `cloneManager` to keep this PR simpler.

Thank you for the review :-)

karreiro:
Jenkins please retest this.

PR #1652, opened 2018-04-18
[7.7.x] AF-965: Unify KIE Base and Session naming

jomarko:
See [AF-965](https://issues.jboss.org/browse/AF-965)

@ederign would you mind have a look?

This is cherry-pick of #1651

PR #1696, opened 2018-04-26
AF-965: I18n session count label

jomarko:
@jhrcek same comment as in #1652


jomarko:
@cristianonicolai @ederign this is ready for merge if you are fine with code.

PR #1698, opened 2018-04-26
JBPM-7134: Stunner - Do not notify User about no commands to redo

manstis:
See https://issues.jboss.org/browse/JBPM-7134

I removed the use of ```IllegalStateException``` in ```RedoCommandHandler``` however also made all the ```AbstractSelectionAwareSessionCommand``` implementations only invoke the command in response to a keys match **if** the command itself is **enabled**. This IMO is a better fix than simply not throwing the exception.


romartin:

manstis:

cristianonicolai:

PR #1714, opened 2018-04-27
JBPM-7162: Stunner - Make canvas size configurable (via workbench preferences)

wmedvede:


wmedvede:

wmedvede:

wmedvede:

wmedvede:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/resources/org/kie/workbench/common/stunner/core/client/preferences/resources/i18n/StunnerPreferencesConstants.properties
```diff
@@ -20,5 +20,11 @@
 StunnerPreferences.Label=Stunner Preferences
 StunnerPreferences.StunnerDiagramEditorPreferences=Diagram Editor Preferences
 StunnerDiagramEditorPreferences.Label=Diagram Editor Preferences Fields
-StunnerDiagramEditorPreferences.AutoHidePalettePanel.Label=Auto Hide Category Panel
+StunnerDiagramEditorPreferences.AutoHidePalettePanel.Label=Auto hide category panel
 StunnerDiagramEditorPreferences.AutoHidePalettePanel.Help=Enables the automatically hiding of the toolbar panel associated with a Category. E.g. when a category item is selected, or when the mouse leaves the panel.
+StunnerDiagramEditorPreferences.CanvasWidth.Label=Drawing area width
+StunnerDiagramEditorPreferences.CanvasWidth.Help=Sets the width of the editor drawing area. It must be an integer number between 2800 and 5600.
```
jomarko: Would be possoble to replace constants `2800 ...` with something like `{0} ...` and then during construction of the real string use the constants from `CanvasWidthValidator` and `CanvasHeightValidator` ?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/preferences/CanvasSizeValidator.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public ValidationResult validate(String value) {
+        int intValue;
+        try {
+            intValue = Integer.parseInt(value);
+        } catch (NumberFormatException e) {
+            return new ValidationResult(false,
+                                        messagesBundleKeys());
```
jomarko: Just minor thing, but I think NumberFormatException doesn't mean out of the range.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/preferences/StunnerPreferences.java
```diff
@@ -31,6 +31,8 @@
     @Override
     public StunnerPreferences defaultValue(final StunnerPreferences defaultValue) {
         defaultValue.diagramEditorPreferences.setAutoHidePalettePanel(false);
+        defaultValue.diagramEditorPreferences.setCanvasWidth(2800);
```
jomarko: I think we could use `CanvasWidthValidator.MIN_CANVAS_WIDTH`

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/preferences/StunnerPreferences.java
```diff
@@ -31,6 +31,8 @@
...
+        defaultValue.diagramEditorPreferences.setCanvasHeight(1400);
```
jomarko: I think we could use `CanvasHeightValidator.MIN_CANVAS_HEIGHT`

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/resources/org/kie/workbench/common/stunner/core/client/preferences/resources/i18n/StunnerPreferencesConstants.properties
wmedvede: @jomarko The preferences API doesn't have the ability of accepting/processing {xxx} parameters in the validators, so it's not possible at this point.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/preferences/StunnerPreferences.java
wmedvede: @jomarko I don't think so. The MIN_VALUE is the lower bound. i.e., we don't want users to try to set the value lower than this minimum. But instead the by default value is what we "believe" is "the best value" to start working with. Then the user has the freedom of changing.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/preferences/StunnerPreferences.java
wmedvede: @jomarko same as above.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/preferences/CanvasSizeValidator.java
wmedvede: @jomarko I see the point, but see that validations are chained and the preferences api first checks if the value is a valid number, and then "transfers the execution" to the next validator. So if a wrong number is typed, this validator is not executed, and a proper message is automatically set. Since I still need to parse the value, I need to do something in case the exception is throw, but this case will probably never happen. I believe we can live with that. We'll also move to finite canvases as asap, so this canvas size preferences will be removed asap too.

LuboTerifaj:
Hi @wmedvede @jomarko 
I have checked it also manually and it seems that it works properly.
Thanks!

PR #1719, opened 2018-04-30
JBPM-7180: Stunner - Name change widget appear in wrong position for containers

manstis:
See https://issues.jboss.org/browse/JBPM-7180

@tiagodolphine Would you mind taking a look please?

The issue was that the in-place popup was being positioned at the mouse co-ordinate less half the width/height of the containers bounds - which is a really strange thing to do. Ideally the popup should be positioned at the mouse co-ordinate less half the popups height/width; however unfortunately (I tried) the popup height/width is not available until after it is shown and re-positioning it then would lead to flickering.... so I did what was most simple and just positioned at 50% of the CSS min-height/min-width settings. Works much better.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/actions/AbstractCanvasInPlaceTextEditorControl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/actions/AbstractCanvasInPlaceTextEditorControl.java
@@ -46,8 +46,6 @@ import org.kie.workbench.common.stunner.core.client.shape.view.event.TextExitEve
 import org.kie.workbench.common.stunner.core.client.shape.view.event.TextExitHandler;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.ViewEventType;
 import org.kie.workbench.common.stunner.core.graph.Element;
-import org.kie.workbench.common.stunner.core.graph.content.view.View;
-import org.kie.workbench.common.stunner.core.graph.util.GraphUtils;
 import org.uberfire.mvp.Command;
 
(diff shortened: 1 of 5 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/actions/AbstractCanvasInPlaceTextEditorControl.java
```diff
@@ -165,16 +163,13 @@ public void handle(TextExitEvent event) {
...
-        }
-        final double rx = null != size ? size[0] / 2 : 0d;
+        final double offsetX = getTextEditorBox().getDisplayOffsetX();
+        final double offsetY = getTextEditorBox().getDisplayOffsetY();
         getFloatingView()
-                .setX(x - rx)
+                .setX(x)
                 .setY(y)
+                .setOffsetX(-offsetX)
+                .setOffsetY(-offsetY)
```
jomarko: Well, using `-offsetX` and `-offsetY` now ensures invoked inline editor is centered to point of click, however the editor can be resized and then if inline editor is invoked repeatedly then it is not centered. So just idea for discussion, if we use `setOffsetX(offsetX)` and `setOffsetY(offsetY)` then we would get always inline editor left upper corner aligned with point of double click, what do you think? But I think also the current state is fine, so just asking if you considered this variant.

manstis: Hi @jomarko thanks for the review.

I originally removed the offset all together but found, when the popup's top-left appeared at my mouse-pointer, it was a pain to use as I always had to move the mouse pointer after the popup appeared. I guess the "offset" I have is not intended to necessarily centre the popup over the mouse-pointer but just remove the need to move the mouse-pointer after the popup has appeared.

manstis:
@tiagodolphine Awaiting your review please.

PR #1722, opened 2018-04-30
RHPAM-630: Stunner - Can't compile process with reusable subprocess when new marshaller is enabled

evacchi:
Fixes the JIRA issue in the subject


evacchi:
@hasys @romartin please take a look at this when you have time

hasys:
Hi @evacchi, is it possible to write some tests for this change?

evacchi:
I have updated some existing tests because they were -- in fact -- inconsistent, according to engine error:

     `Process 'process' [test.process]: Node 'Subprocess' [1] SubProcess you can only set independent to 'false' only when 'Wait for completion' is set to true.`

these tests actually were with independent="false" AND waitForCompletion="false"

PR #1733, opened 2018-05-02
Stunner - Fix shifted events after migration - fix for bpmn2 exporter…

tsurdilo:
… setting


kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/legacy/Bpmn2JsonUnmarshaller.java
```diff
@@ -198,6 +198,9 @@
     public static final String defaultFontColor = "#000000";
     public static final String defaultSequenceflowColor = "#000000";
 
+    public static final String exporterName = "jBPM Process Modeler";
+    public static final String exporterVersion = "1.0";
```
romartin: hey @tsurdilo , what is this version used for? You think it makes sense to use the current module version (`7.x`) rather than `1.0`? Should we be updating this version on each new release?

tsurdilo: @romartin its not used at all. I think this is OK as othrwise we would need to add source filteering in pom or something which is not needed. Designer will be able to determine that bpmn2 was not generated by itself just from the exporter name.

PR #1735, opened 2018-05-02
[AF-1103-7.7.x]: Fixed issue that redirects to the list of assets when switching to project view

adrielparedes:
Backport to 7.7.x of https://github.com/kiegroup/kie-wb-common/pull/1622

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/utils/Utils.java
+++ b/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/utils/Utils.java
@@ -19,6 +19,7 @@ import org.guvnor.common.services.project.model.Module;
 import org.guvnor.common.services.project.model.Package;
 import org.guvnor.common.services.project.model.WorkspaceProject;
 import org.guvnor.structure.organizationalunit.OrganizationalUnit;
+import org.guvnor.structure.repositories.Repository;
 import org.kie.workbench.common.screens.explorer.model.FolderItem;
 import org.kie.workbench.common.screens.explorer.model.FolderItemType;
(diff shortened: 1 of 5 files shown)


kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/utils/Utils.java
```diff
@@ -116,6 +117,48 @@ public static boolean hasPackageChanged(final Package pkg,
...
     }
 
+    /**
+     * A convenience method to compare two Repositories avoiding cluttering code with null checks.
+     * @param repository
+     * @param activeRepository
+     * @return
+     */
+    public static boolean hasRepositoryChanged(final Repository repository,
+                                               final Repository activeRepository) {
```
jomarko: I think this method can be simplified just to one line:
```java
return !Objects.equals(repository, activeRepository);
```

adrielparedes: @jomarko You are completely right, I didn't see when I wrote it. I followed the other methods so they should be changed too, but 2 things, just to confirm the change. Is Objects available in GWT? Because that could be a reason of why the others methods are like this. And the seconds is that this PR is a backport and the original is already merged. Should I create other PR for the original?

manstis:
@adrielparedes ```java.util.Objects``` is emulated by GWT and can be used client-side.

It was added to Java in 1.7 and added to GWT in 2.6. A lot of the existing Workbench code would have targeted Java 1.6 and GWT 2.5 so would not be able to use it.

adrielparedes:
@manstis @jomarko so, do you think this can be merged? Thanks for the comments

manstis:
@adrielparedes OK with me (I just commented on ```Objects```) up to @jomarko to approve now :-)

PR #1606, opened 2018-04-10
JBPM-7070: Service Tasks - Integrate with new BPMN marshallers

evacchi:
WIP: do not merge yet. Waiting for JBPM-7089

PR #1741, opened 2018-05-03
Stunner - New Marshallers: add target namespace and schema location

evacchi:


hasys:

evacchi:
@hasys I'm pretty sure this will fix all of those warning seen in #1606 
Bonus: it seems to me that old marshallers did suffer from a few of those. Reason: `xsi:schemalocation` did not contain xsd location for DI, DC in the generated XML

evacchi:

PR #1747, opened 2018-05-04
JBPM-7145 Stunner: saving diagram SVG on backend

tiagodolphine:
When a diagram is saved, the SVG referred to it should be generated and persisted on the backend, in the save way as the_ .bpmn2 _files. The SVG file should be tested integrating with the jbpm on the workbench.

It was created a command to handle the saving process `SaveDiagramSessionCommand`, for now it is handling the SVG saving after the Diagram itself is saved on the backend, the idea is to improve and handle all the saving logic within this command on a centralized way, we are on the half way on that and the rest of saving could be a separated JIRA. A refactor was necessary to support this modification, because the saving logic was spread when it is a the Project (integrated with WB) and when it is Standalone.

Other detail is that the SVG for BPMN2 need a specific key for the ids inserted on the SVG groups, so other logic were need to handle that, adding this key on the DefinitionSet in this way, other definition sets may have its own keys to identify the nodes when generating the SVG if necessary. Considering the SVG file name I implemented to have the same pattern as the old desginer the logic is on the `AbstractVFSDiagramService.getDiagramSvgFileName` that is **projectname. processname-svg.svg** .

**Steps to test:**

1. create a valid diagram without any bpmn violation and save it
2. build and deploy (on the authoring menu)
3. check the SVG generated on the WB, menu Manage -> Process Definitions -> select the created process.

This PR depends on appformer https://github.com/kiegroup/appformer/pull/337 and lienzo-core https://github.com/kiegroup/lienzo-core/pull/28.

@romartin 
@LuboTerifaj 

Guys this review I think is dedicated and you may have more information than me how to test the whole flow on the WB to make sure the SVG is generated on the proper way and is compatible and integrated with jpbm, the steps I inserted are what I used on my tests locally, besides checking the  **.niogit**  to view the bpmn and SVG created files.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/annotation/SvgNodeId.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/annotation/SvgNodeId.java
@@ -13,14 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package org.kie.workbench.common.stunner.core.definition.annotation;
 
-package org.kie.workbench.common.stunner.client.widgets.toolbar.impl;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
-import org.kie.workbench.common.stunner.client.widgets.toolbar.Toolbar;
-import org.kie.workbench.common.stunner.client.widgets.toolbar.ToolbarCommand;
-import org.kie.workbench.common.stunner.core.client.session.impl.EditorSession;
-
-public interface EditorToolbar
-        extends Toolbar<EditorSession> {
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface SvgNodeId {
 
 }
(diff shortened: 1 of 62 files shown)


tiagodolphine:
Jenkins, execute full downstream build.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/annotation/SvgNodeId.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface SvgNodeId {
```
romartin: why we need this? Probably we can do some quick talk on Mon and you can give me a few hints about the approach

romartin:
hey @tiagodolphine , sorry after merging the memory leaks fixes this needs rebase... lemme know if I can help

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/annotation/SvgNodeId.java
tiagodolphine: @romartin sure, we can talk about this and I can explain the idea of this approach, this is related to adapters of DefininitionSet, and you have more knowledge on this part, let's to a call today!

tiagodolphine:
@romartin I'm rebasing and applying the necessary merges now. In case I got any problem I will ping you. Thanks.

LuboTerifaj:
Hi @tiagodolphine @romartin 
This PR is too big and I have to go now, so unfortunately I cannot make it today. I will have a look at it on Wednesday. I am sorry for inconvenience.

romartin:
Tested and the SVG generation is fine, but it's missing to highlight the current task shapes.

PR #1760, opened 2018-05-08
[7.7.x] RHPAM-433: Stunner - [IE11] Quick menu icons are not displayed

wmedvede:

PR #1766, opened 2018-05-09
JBPM-6858: Stunner - [IE11] Quick menu icons are not displayed (#1760)

wmedvede:
(cherry picked from commit ec292df23a53dd102aa6cfca812965d365ba2387)


wmedvede:
Hi @romartin @hasys this is the peer PR for the already merged into 7.7.x https://github.com/kiegroup/kie-wb-common/pull/1760.
Would you mind review?
Thanks

jomarko:
Some other QE should check Process designer.

hasys:

wmedvede:
@jomarko I don't know what the final decision will be to be honest. In my opinion, we should directly discontinue support of IE11 for the new Stunner, and just move to Ms. Edge. I believe to continue supporting IE11 for stunner will be a huge penalization and an infinite source of problems. But this decision is not mine. Anyway, so going back to your query by now we moved only the BPMN icons to png, but if we continue IE11 we'll probably have to move DMN icons too. But let's move this PR forward by now.l

hasys:
@wmedvede My guess is some big corporations still using IE11 due to ActiveX support, since MS Edge doesn't support it. So yea, no one have desire to support IE11, but this requirement is going from PMs.

PR #1768, opened 2018-05-09
DROOLS-2533: [DMN Designer] Form Properties panel does not work

manstis:
See https://issues.jboss.org/browse/DROOLS-2533


manstis:
@jomarko There are no tests as this code is entirely "throw away" when we integrate into the workbench.

romartin:
sorry guys that seems to be my fault at some point... thanks for the fix!!

jomarko:
Filed here https://issues.jboss.org/browse/DROOLS-2538

PR #1776, opened 2018-05-10
JBPM-7253: Stunner - Case Management editor fails to open due to missing preferences

romartin:
Hey @hasys @wmedvede 

See https://issues.jboss.org/browse/JBPM-7253, reported by Jeremy.

@wmedvede I changed a bit the approach:
* Instead of a "global" preferences registry, it now considers that can exist different types. By default it provides the `DefaultPreferencesRegistry`, but other types can provide a custom registry in order to setup some custom stuff, as we only have (at least for now) a single `StunnerPreferences` bean type. 
* This way for example CM now provides its own `PreferencesRegistry` implementation
* So from any other components, like the palette, there is no need to keep the reference to the instance as part of the component's state, as it just injects the right registry and obtain the instance when necessary
* IMO this way the code probably becomes "less" coupled - no need to provide public API for the preferences in any component. So whaetever implementation can just create an injection point to the preferences registry (or use the `StunnerPreferencesRegistries`) and obtain the right instance, which also is being shared between other components.

Anyway that are just my thoughts... feel free to share yours guys!! :)

Thanks!


romartin:

romartin:
@wmedvede not really sure if has to be cherry-picked as in 7.7.x CM is not part of the application, and from the BPMN side I tested and the settings were working fine, so IMO there is no need for it.

PR #1791, opened 2018-05-15
RHPAM-1006 & RHPAM-995: Fixed errors once removing some nodes

romartin:
Hey @hasys @krisv @Josephblt 

(Hopefully) I've fixed an issue which it fixes 3 tickets at once (2 are blockers even)!! So it fixes:
* https://issues.jboss.org/browse/RHPAM-1006
* https://issues.jboss.org/browse/RHPAM-995
* https://issues.jboss.org/browse/JBPM-7236

Please guys, can you give **priority to this**? It's an small change that fixes blocker issues. Thanks in advance!!

PS: Also fixed the server startup for showcases when running from the IDE, it was failing. This was already in `master` but not in `7.7.x`, it does NOT hurt anyway as are just a few changes in the `beans.xml` for the showcases, no downstream stuff affected.


romartin:

romartin:

Josephblt:
Hey @romartin. I can reproduce an error similar to @hasys. Except I can do it without using a container.
All you have to do is select only the nodes and none of the connectors and you have the error when you undo. I made a video of it:
https://drive.google.com/file/d/1dNJQEN2l_5hCBMd2750Jb8qUR9zaFH7X/view

romartin:
OK thanks guys ,lemme look into that issues, thanks for testing!

romartin:
Hey guys,

@hasys I cannot reproduce same issue as you mention. So I change the target connection from the end node to some location, but not to any concrete magnet, then remove the end node, which removes the connector as well, and then undoing - it results on the connector being placed on top left as the screenshot you added. On the other hand, I can move the task, and also remove that "wrong" connector and create a new one again, I mean for me, it's not blocking. Anyway this issue is really "old", looking now.. good catch! It's missign to handle the connector locations, when no magnet is assigned, so I can reproduce this on master and even older versions too. Rather than it's not introduced by this fix, I worth fixing it, sooo I'm working on this, will let you know!! 

@Josephblt well tbh the bug you mention is exactly the one I should have fixed! hehe I tested this several times, also tested again now and it works for me... can you please double check? anyway I'll introduce some new commit, so no worries!

Thanks!!

romartin:
Hey @hasys @Josephblt 

I've been testing and most of the issues on connectors appears when at some point you "disconnected" some of the endings.. you can continue working but then undoing and redoing stuff was doing weird things. For example the issue @hasys mention when removing the end event and undoing. In fact, connections are not taking into account parents anyway (that's an already known missing feature to add), so parents should not affect on this either.

These kind of issues were due to some missing capability on the client engine, as when a connection has no ending, it was not really taking care of the location of the connection, and even not calling the right rendering components, as it was considering "nothing" was changed...  So finally implemented support for connections with no magnet attached, by adding a new connection type `Point2DConnection`, and done the right changes to make it work. 

I've tested and now I can "detach" connections, remove the nodes, undo, redo, etc, and everything is fine. Also tested again and it fixes the blockers and tickets mentioned in this PR's description. 

Please can you give it a look again?? Thanks!!

romartin:

romartin:
Yeah tested again the workbench and work good! I :+1: I think that's a really important fix to include for next CR, also not risky, as it does not affects backend side (as BPMN do not allow "empty" connections either) and the amount of changes and tests are a few ones. and safe, right?

So please lemme know asap and let's see if we can include this, if find some but and you think it's not a blocker, also consider approving this one, that fixes important blockers, and reporting. 

Thanks!!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/Point2DConnection.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public String toString() {
+        return "[Point2DConnection at {" + location + "}" + "]";
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof Point2DConnection) {
+            Point2DConnection other = (Point2DConnection) o;
+            return ((null != location && null != other.location) ?
```
hasys: I guess this can be simplified by:
`Objects.equals(location, other.location)`

org.kie.workbench.common.stunner.core.graph.content.view.Point2D contains better comparison mechanism than `==` usage for Double datatype.

romartin:
Hey @hasys @Josephblt 

Thanks for the review, good catches! I guess I've fixed both issues as well, I guess now the delete operations, even with multiple elements, are safe :+1: 

So added a new commit and a new fix (and test) on the lienzo side, see:
* https://github.com/kiegroup/lienzo-core/pull/32
* https://github.com/kiegroup/lienzo-tests/pull/23

Please give it a try ASAP and lemme know!!
Thanks!

romartin:

romartin:
any feedback @hasys @Josephblt about this?

Josephblt:
Hey @romartin. Yeah it's all good for me now.

PR #1800, opened 2018-05-16
[7.7.x] RHPAM-540: Stunner - disable doing operation in view mode

wmedvede:


wmedvede:
Hello  @romartin  @hasys , would you mind review this PR?
Thanks.

romartin:
Hey @hasys 
Let you know this ticket also fixes https://issues.jboss.org/browse/RHPAM-998, so please consider giving a try to rename the process as well. 
Thanks!!

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/ManagedClientSessionCommands.java
```diff
@@ -95,9 +96,28 @@ public void bind(final ClientSession session) {
...
-        commands.clear();
+        clearCommands();
         types.clear();
         sessionCommands.destroyAll();
     }
-}
+
+    private void clearCommands() {
+        commands.forEach(ClientSessionCommand::destroy);
+        commands.clear();
```
jomarko: Asking just because of other changes in this PR, don't we need check `commands != null` ?

wmedvede: By construction this commands variable can never by null

jomarko: ok, will try manually

wmedvede:
Hello @jomarko , yes, this is something we are aware of. The problem is that there are some "custom" widgets, like the "process variables" widget (and some others), etc, that don't have implemented the readonly mode.
So when the form is painted in readonly mode this widgets still appears in editable mode. 
BUT, the model is not affected if the user tries to makes changes in this scenarios. i.e., even when the user changes values on this widgets the model remains readonly. And this changes are cleaned up as soon the user changes to some other version, etc. So it's just UI effect by now. Of course a non desirable one, but at the point we are right now we can't implement the readonly mode in the missing widgets.
So this is something we'll have to live with for this release.  @romartin is aware of this too, and we agreed that with the time being this was the best that could be done. So I'd appreciate that if this are the only drawbacks you approve it.

jomarko:
@wmedvede @hasys thank you

PR #1824, opened 2018-05-23
[AF-1209]: Removed onResourceCopiedEvent because copied asset was shown for all users

adrielparedes:


manstis:
Is this just a quick fix? 

Did we want to stop the User who copied a file from having it open automatically too?

Won't removal potentially re-open the JIRA that lead to the now deleted code being added in the first place (https://issues.jboss.org/browse/RHPAM-565)?

What about ```RepositoryRemovedEvent```, ```RemoveOrganizationalUnitEvent``` and ```RenameModuleEvent```. These are ```@Portable```, raised from the server (so will be received by all clients) and presumably we want to keep the handling of these (for the User who removed a repository, organisational unit or renamed a module?)

adrielparedes:
@manstis yes, this is a quick fix that @krisv needs for 7.7.x, and you are right is a revert from 565. Is is possible to make the revert from 565?

manstis:
@adrielparedes You should be able to perform a ```git revert 000ffb7f07cebb8cc5a41fed54b6edf85a250135``` on ```7.7.x``` if that's what you'd prefer.

adrielparedes:
@manstis thanks, for the tip, I will do revert in both, master and 7.7.x. Can I ask you to review those PRs?

manstis:
@adrielparedes Sure; close this one and submit two new ones containing the reverts. I think that'll make the history cleaner (by all means update the commit comment to reference AF-1209 too).

PR #1835, opened 2018-05-28
JBPM-7267 Some perfomance Improvements on Stunner

tiagodolphine:
- Get shape on Canvas now uses a Map instead of a list
- Removed CDI event on canvas draw
- Selection only applyState on selected elements instead of all elements on canvas

@romartin 
@hasys 
@LuboTerifaj


tiagodolphine:
Jenkins retest this

PR #1848, opened 2018-06-01
added new URL in settings.xml

mbiarnes:
added URL that points to Nexus where the artifacts of dailyBuilds are stored


cristianonicolai:
@mbiarnes would we use https://github.com/kiegroup/kie-wb-common/pull/1853 instead?

mbiarnes:
settings.xml in kie-wb-common/kie-wb-common-services/kie-wb-common-services-backend/src/test/
is not existing any more - therefore this PR can be closed

PR #1867, opened 2018-06-07
Fix compilation error in tests

jhrcek:
..caused by changes from https://github.com/kiegroup/appformer/pull/388


mareknovotny:
@jhrcek Thanks for this quick fix

PR #1890, opened 2018-06-12
DROOLS-2629 : Complex enumerations are not selected in BRL multi value constraint

Rikkola:
https://issues.jboss.org/browse/DROOLS-2629

@manstis @jomarko If you could please review.

https://github.com/kiegroup/kie-soup/pull/47
https://github.com/kiegroup/kie-wb-common/pull/1890

### DIFF
--- a/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilities.java
+++ b/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilities.java
@@ -27,6 +27,7 @@ import org.jboss.errai.bus.client.api.BusErrorCallback;
 import org.jboss.errai.bus.client.api.base.MessageBuilder;
 import org.jboss.errai.bus.client.api.messaging.Message;
 import org.jboss.errai.common.client.api.RemoteCallback;
+import org.kie.soup.commons.util.ListSplitter;
 import org.kie.soup.project.datamodel.oracle.DropDownData;
 import org.kie.workbench.common.services.shared.enums.EnumDropdownService;
 import org.kie.workbench.common.widgets.client.resources.i18n.CommonConstants;
@@ -41,7 +42,6 @@ public class EnumDropDownUtilities {
 
     /**
      * Populate ListBox with values from DropDownData
-     *
      * @param value
      * @param dropData
      * @param isMultipleSelect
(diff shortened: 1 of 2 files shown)


kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilities.java
```diff
@@ -121,42 +121,28 @@ private void fillDropDown(final String value,
...
+                                                           isMultipleSelect);
 
+        int selectedIndexOffset = addItems(listBox);
         boolean selected = false;
-        Set<String> currentValues = new HashSet<String>();
-        String currentValue = value;
-        String trimmedCurrentValue = currentValue;
-        if (isMultipleSelect && trimmedCurrentValue != null) {
-            trimmedCurrentValue = currentValue.replace("\"",
-                                                       "");
```
Rikkola: Removing this bit is what worries me. Is there some situation I am missing here, where this is needed.

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilitiesTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+@RunWith(GwtMockitoTestRunner.class)
+public class EnumDropDownUtilitiesTest {
+
+    @Test
+    public void testCommaseparatedValues() throws Exception {
```
jomarko: By java convention we should use the name: `testCommaSeparatedValues`

kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilities.java
```diff
@@ -121,42 +121,28 @@ private void fillDropDown(final String value,
...
             String val;
-            if (v.indexOf('=') > 0) {
+            if (enumeratedValue.indexOf('=') > 0) {
                 //using a mapping
-                String[] splut = ConstraintValueHelper.splitValue(v);
-                String realValue = splut[0];
-                String display = splut[1];
+                final String[] splut = ConstraintValueHelper.splitValue(enumeratedValue);
+                final String realValue = splut[0];
+                final String display = splut[1];
```
jomarko: I think as google translator don't think `splut` is english word, we shouldn't use it :)

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilitiesTest.java
```diff
@@ -0,0 +1,52 @@
...
+                                                    downData,
+                                                    true,
+                                                    mock(Path.class),
+                                                    listBox);
+
+        verify(listBox).clear();
+
+        verify(listBox).addItem("a");
+        verify(listBox).addItem("\"b, c\"");
+        verify(listBox).addItem("d");
```
jomarko: I thin k we have to improve the test to get real reproducer test, the issue was that values were not selected, so I think we have to add assertion like bellow:
```java
verify(listBox).setItemSelected(0, true);
verify(listBox).setItemSelected(1, true);
verify(listBox).setItemSelected(2, true);

Rikkola:
@jomarko Ok updated and added your tests in.

kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/widget/EnumDropDownUtilities.java
manstis: @Rikkola I can't think of any reason (other than this code is in essence the cause of the bug relating to using enum values containing ```"```.. so removing it should be perfectly valid). IIRC "in" and "not in" operators do something quirky with the values so I'll have a quick play with ```drools-wb```.

manstis: @Rikkola I get test failures for ```EnumDropDownUtilitiesTest``` with this PR locally. I won't test ```drools-wb``` until we're certain the changes in this PR are final; if OK.

manstis: Oh... need ```kie-soup``` too... let me try again (let me go back on PTO!!)

manstis:
@Rikkola The issue reported in the JIRA seem to be resolved; however I noticed (given an enumeration defined as ```Person.name: ['a', '"b, c"', 'd']```) that the generated DRL is ```Person( name == "b, c" )``` whereas should it be ```Person( name == "\"b, c\"" )```? Closing and re-opening the decision table keeps the correct values selected in the table. I also tried defining the BRL fragment to use a literal value for the enumeration and this gave the same results (DRL _possibly_ incorrect, but re-opening and editing the column was OK).

I also checked Guided Rules and the DRL is equally wrong however re-opening the rule also led to the wrong enumeration value being selected in the editor (it selected the first option by default) and the DRL showing the same (i.e. I set it to ```"b, c"``` and re-opening the file selected ```a```).

IDK whether this should be addressed as a new JIRA?

manstis:
@Rikkola ```org.drools.workbench.models.datamodel.rule.builder.DRLConstraintValueBuilder``` is responsible for writing the DRL applicable for the enumeration values and does not add quotes if the value already starts/ends with them.. That code is quite old; so I'd propose it is not fixed for this JIRA. It might need some digging to understand _why_ it works the way it does :-(

manstis:
Seems to have been added for https://bugzilla.redhat.com/show_bug.cgi?id=996964

https://github.com/kiegroup/drools/commit/c37653ae479e87802b288e3d57babcea6d84ac36

Rikkola:
@manstis I might as well take a quick look if I get the new issue you found fixed.

Rikkola:
Ok I took a look at the issue and looks like the problem might not be in the classes that I touched here. I was worried that this fix would conflict with the issue @manstis found, but the list splitter works fine, just some where after setting the cell value we clean up the quotes.

I made a separate ticket so the issue can be tracked.
https://issues.jboss.org/browse/DROOLS-2638

jomarko:
@Rikkola well, looking on RHPAM-555, I think we need cherry-pick into 7.7.x.

PR #1892, opened 2018-06-12
[AF-1242] Error ocurred after project deletion

ederign:
This commit introduced two issues:

- Deletion of a project triggers a null pointer
- When I created two projects I got a ConcurrentModificationException.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectScreen.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectScreen.java
@@ -26,7 +26,6 @@ import elemental2.dom.HTMLElement;
 import elemental2.promise.Promise;
 import org.guvnor.common.services.project.client.security.ProjectController;
 import org.guvnor.common.services.project.context.WorkspaceProjectContextChangeEvent;
-import org.guvnor.common.services.project.context.WorkspaceProjectContextChangeHandler;
 import org.guvnor.common.services.project.model.WorkspaceProject;
 import org.guvnor.messageconsole.client.console.widget.button.ViewHideAlertsButtonPresenter;
 import org.guvnor.structure.client.security.OrganizationalUnitController;
@@ -66,7 +65,7 @@ import org.uberfire.workbench.events.NotificationEvent;
 
 @WorkbenchScreen(identifier = LibraryPlaces.PROJECT_SCREEN,
         owningPerspective = LibraryPerspective.class)
-public class ProjectScreen implements WorkspaceProjectContextChangeHandler {
+public class ProjectScreen {
 
     private Elemental2DomUtil elemental2DomUtil;
     protected WorkspaceProject workspaceProject;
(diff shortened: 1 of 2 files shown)


kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/test/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectScreenTest.java
```diff
@@ -420,9 +418,21 @@ public void titleIsUpdatedWhenContextModuleIsUpdated() throws Exception {
...
+    public void shouldNotChangeProjectAndTitleWhenContextChange() throws Exception {
+
+        presenter.changeProjectAndTitleWhenContextChange(new WorkspaceProjectContextChangeEvent() {
+            @Override
+            public WorkspaceProject getWorkspaceProject() {
+                return null;
+            }
+        });
+
+        verify(view, never()).setTitle("module name");
```
jhrcek: Just a niptick, but I'd use `never()).setTitle(any())` here to make the test more restrictive about what changeProjectAndTitleWhenContextChange implementation can do.

ederign:
Build timed out (after 60 minutes). Marking the build as aborted.

ederign:
please jenkins retest this

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectScreen.java
```diff
@@ -233,6 +224,13 @@ public void onContributorsUpdated(@Observes AfterEditOrganizationalUnitEvent eve
         resolveContributorsCount();
     }
 
+    public void changeProjectAndTitleWhenContextChange(@Observes final WorkspaceProjectContextChangeEvent current) {
+        if (current.getWorkspaceProject() != null) {
+            this.workspaceProject = current.getWorkspaceProject();
+            this.view.setTitle(workspaceProject.getName());
+        }
+    }
+
```
Rikkola: @jhrcek @ederign @adrielparedes 
There is some danger here.

The `WorkspaceProjectContext` was created to keep the client side context in sync. Meaning at any time you can check what the active project is and at the moment when we call `onChange()` the context is already up to date. Anything that cared about the active project would use `WorkspaceProjectContextChangeHandler` to get notified about the changes and the actual event `WorkspaceProjectContextChangeEvent` would just be used to do the actual changes in `WorkspaceProjectContext`.

Here `WorkspaceProjectContextChangeEvent` is observed and this might be before `WorkspaceProjectContext` observes the information and this works as long as nobody adds anything that depends on `WorkspaceProjectContext` into the `ProjectScreen` and `ProjectScreen` and it's children always only uses the `workspaceProject` field declared in it. 

The history behind this is that we previously had bugs that were caused by client side not knowing what the current active project (ou/folder/branch/package) is. Also `WorkspaceProjectContextChangeEvent` was used to both follow updates and update `WorkspaceProjectContext`.

I would say this is OK for now. For the future I think we should rename `WorkspaceProjectContextChangeEvent` to `ApplyChangeOnWorkspaceProjectContextEvent`, remove the `workspaceProject` field from `ProjectScreen` and just listen to changes in context, using the `onChange()` method. Giving the UI one central for context and context changes.

PR #1900, opened 2018-06-14
[AF-1224] Execution UI should point to swagger, instead of endpoint

elguardian:
set the swagger endpoint if the server has the swagger capability


elguardian:
depends on https://github.com/kiegroup/droolsjbpm-integration/pull/1497

PR #1908, opened 2018-06-20
DROOLS-2425: [DMN Designer] Undo of relation column deletion is not proper

manstis:
See https://issues.jboss.org/browse/DROOLS-2425

@romartin I added the cache's life-cycle to ```ManagedSession``` (remember our discussions?!)

The issue was that after saving, closing and re-opening the (Context etc) grid was retrieved from the cache however associated to a new "parent" that lead to sizing/re-sizing issues that updated the size of the "old" (disconnected) parent and not the "current" parent. 

The cache's life-cycle is now tied to that of the ```Session``` and hence destroyed when the editor is closed. I also added a ```DMNViewerSession``` (although it is not accessible at the moment, as it is not possible to open read-only versions until I integrate the editor into a "project showcase" - as opposed to "standalone showcase"). This is because I have to cast the ```Session``` to get access to the new Canvas Control for the cache.

PR #1912, opened 2018-06-22
DROOLS-2644: [DMN Designer] Deletion of the column causes last column width increase

manstis:
See https://issues.jboss.org/browse/DROOLS-2644

The issue was that when a column was resized it considered the maximum of the "proposed" width (could be narrower than it was, when deleting a column) and the current column width. I've changed this to consider the "proposed" width and the minimum required for any editors in the column.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
@@ -90,6 +90,18 @@ public class GridCellTupleTest {
     }
 
     @Test
+    public void testProposeContainingColumnWidthWhenLargerThanExistingEditor() {
+        gridData.setCell(0, 0, () -> new DMNGridCell<>(new ExpressionCellValue(Optional.of(existingEditor))));
+        when(existingEditor.getPadding()).thenReturn(BaseExpressionGrid.DEFAULT_PADDING);
+        when(existingEditor.getMinimumWidth()).thenReturn(200.0);
+        when(gridColumn.getWidth()).thenReturn(100.0);
+
+        tuple.proposeContainingColumnWidth(300.0);
+
+        verify(gridColumn).setWidth(300.0);
+    }
+
+    @Test
     public void testOnResizeSetsColumnWidth() {
         when(gridColumn.getWidth()).thenReturn(100.0);
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
@@ -16,9 +16,13 @@
 
 package org.kie.workbench.common.dmn.client.widgets.grid.model;
 
+import java.util.Optional;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.kie.workbench.common.dmn.client.editors.expressions.types.context.ExpressionCellValue;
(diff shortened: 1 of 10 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
```diff
@@ -66,7 +74,19 @@ public void testProposeContainingColumnWidthWhenSmallerThanExisting() {
...
+    public void testProposeContainingColumnWidthWhenSmallerThanExistingEditor() {
+        gridData.setCell(0, 0, () -> new DMNGridCell<>(new ExpressionCellValue(Optional.of(existingEditor))));
+        when(existingEditor.getPadding()).thenReturn(BaseExpressionGrid.DEFAULT_PADDING);
+        when(existingEditor.getMinimumWidth()).thenReturn(200.0);
+        when(gridColumn.getWidth()).thenReturn(100.0);
+
+        tuple.proposeContainingColumnWidth(50.0);
+
+        verify(gridColumn).setWidth(200.0);
     }
```
jomarko: Would this scenario make sense?
```java
@Test
    public void testProposeContainingColumnWidthWhenLargerThanExistingEditor() {
        gridData.setCell(0, 0, () -> new DMNGridCell<>(new ExpressionCellValue(Optional.of(existingEditor))));
        when(existingEditor.getPadding()).thenReturn(BaseExpressionGrid.DEFAULT_PADDING);
        when(existingEditor.getMinimumWidth()).thenReturn(200.0);
        when(gridColumn.getWidth()).thenReturn(100.0);

        tuple.proposeContainingColumnWidth(300.0);

        verify(gridColumn).setWidth(300.0);
    }
```

manstis:
Hi @jomarko regarding your comment about an issue after save, close and re-open; any issues relating to saving, closing and re-opening should be fixed by https://github.com/kiegroup/kie-wb-common/pull/1908. Did you want to focus your efforts on that PR and once merged we can reconsider these new ones?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
manstis: @jomarko Yes, it would make sense. I'll add.

jomarko:
@manstis thanks, looking on #1908 right now

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/widgets/grid/model/GridCellTupleTest.java
manstis: Test added in new commit (to follow..)

jomarko:
@manstis thanks for updates, I found similar problem if the decision table and relation are used in parallel. However not sure if it is very rare corner case or very common case. Let me know what do you think.
- Open expression editor
- Define top level as Context
- Add two rows to the context
- Specify first row as Decision Table
- Specify second row as Relation
- Insert two columns to the relation, before the exiting column
- Remove the last column of the relation
- Undo the last action
- Notice the width of both Dec. Table Description column and Last Relation column became much bigger that would be needed

jomarko:
screenshot from 2018-06-26 09-59-45

manstis:
@jomarko I've introduce a fix for the last issue you reported.. I now store and restore existing column widths when deleting (and undoing said deletion) columns. This ensures column widths are preserved exactly. I tried with your use-case and it works much better.

The reason the delete/undo was making the grid wider was that when deleting the Relationship column it's last column is made wider automatically to fit the width of the decision table. When the deleted column was restored the proposed width of the grid is the sum of the grids' column widths (however the last had been resized automatically) and hence the new proposed width was greater than it had been originally.

manstis:
@karreiro Could you please review?

PR #1916, opened 2018-06-25
Stunner - Fix project showcase startup on SDM

romartin:
Hey @hasys @tiagodolphine 

This is a fix for the project showcase, as @ederign pointed out there was an error on the pom, it was not starting up on SDM.

Thx


danielzhe:
I just tested here and now is working.

romartin:
Hey @hasys , by SDM we mean get GWT "Super Dev Mode" :) so it means when running the app from command line for example...

PR #1918, opened 2018-06-26
[7.7.x] RHPAM-1107: Selected container gets changed

nmirasch:
ContainerPresenter reacting only on ContainerSpecSelected event when it is the one that is being displayed
Use container.getContainerSpecId() instead container.getContainerName() to map containers
clear status notifications before add new one, added by update actions

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/container/ContainerPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/container/ContainerPresenter.java
@@ -108,6 +108,10 @@ public class ContainerPresenter {
         return view;
     }
 
+    protected void setContainerSpec(ContainerSpec containerSpec){
+        this.containerSpec = containerSpec;
+    }
+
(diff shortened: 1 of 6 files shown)


nmirasch:
@cristianonicolai could you take a look? Thanks!

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/src/main/java/org/kie/workbench/common/screens/server/management/client/container/ContainerPresenter.java
```diff
@@ -72,6 +72,7 @@
     private final Event<ServerTemplateSelected> serverTemplateSelectedEvent;
     private final Event<NotificationEvent> notification;
     private ContainerSpec containerSpec;
+    private ContainerSpecKey selectedContainerSpecKey;
```
cristianonicolai: @nmirasch would the _containerSpecKey_ be different from the _containerSpec_ we already hold as variable?

nmirasch:
@cristianonicolai agree, the have to hold the same data, the only thing that differs is when they are set. I removed selectedContainerSpecKey  and added the setting of the  variable containerSpec at load(final ContainerSpecKey containerSpecKey) to be able to use that from loadContainers(@Observes final ContainerSpecData content) .

nmirasch:
jenkins please retest this

cristianonicolai:
jenkins please retest this

PR #1919, opened 2018-06-26
DROOLS-2628: [DMN Designer] DataInput toolbox actions

manstis:
See https://issues.jboss.org/browse/DROOLS-2628

@romartin The issue was that the ```anyMatch(..)``` in ```isSourceConnectionAllowed(..)``` results in ALL permitted connections matching and hence the target role for all source node types are collected. I've modified the code to get permitted connections whose ```startRole``` matches the source node's role(s) and then collect the ```endRole```. This fixes DMN and fixes all Unit Tests that broke as a result of adding the extra permitted connection in ```DomainLookupFunctionsTest```.


manstis:
@romartin I kept ```isSourceConnectionAllowed(..)``` as it is ```public static``` and might be used elsewhere  and it's purpose is still valid - it checks if a Rule supports a source node role as connection start point.

PR #1926, opened 2018-06-28
JBPM-7360 - Extend Kie Server Controller REST endpoints for RuntimeMa…

cristianonicolai:
…nagementService
Assembly:
- https://github.com/kiegroup/kie-wb-common/pull/1926
- https://github.com/kiegroup/droolsjbpm-integration/pull/1516
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/782

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/pom.xml
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/pom.xml
@@ -291,6 +291,22 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-integ-tests-common</artifactId>
+      <scope>test</scope>
+      <exclusions>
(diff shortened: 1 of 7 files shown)


cristianonicolai:
@sutaakar should be fixed now, thanks.

sutaakar:
@cristianonicolai Unfortunately it still doesn't work as expected. The problem lies in kieServerInstanceManager. RuntimeManagementServiceImpl is using kieServerInstanceManager to retrieve Kie server informations, however for the Workbench there is injected AsyncKieServerInstanceManager.
AsyncKieServerInstanceManager always returns empty response, relying on async callbacks - https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/runtime/AsyncKieServerInstanceManager.java#L219. But since the REST is synchronous then there is returned an empty response to the client.

cristianonicolai:
@sutaakar hopefully I got it right this time :)

PR #1938, opened 2018-07-02
[JPBMP 7384] - Stunner - Add setting for enable/disable HiDPI

danielzhe:
https://issues.jboss.org/browse/JBPM-7384

Depends on kiegroup/lienzo-core#51
Tests: kiegroup/lienzo-tests#33

@romartin Here it goes, Roger. Let me now if there is anything that I could do better. Thanks!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/wires/WiresCanvas.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/wires/WiresCanvas.java
@@ -22,6 +22,7 @@ import java.util.logging.Logger;
 
 import javax.enterprise.event.Event;
 
+import com.ait.lienzo.client.core.config.LienzoCore;
 import com.ait.lienzo.client.core.shape.wires.IConnectionAcceptor;
 import com.ait.lienzo.client.core.shape.wires.IContainmentAcceptor;
 import com.ait.lienzo.client.core.shape.wires.IDockingAcceptor;
@@ -32,6 +33,7 @@ import org.kie.workbench.common.stunner.client.lienzo.canvas.LienzoLayer;
 import org.kie.workbench.common.stunner.client.lienzo.util.LienzoLayerUtils;
 import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvas;
 import org.kie.workbench.common.stunner.core.client.canvas.Canvas;
+import org.kie.workbench.common.stunner.core.client.canvas.CanvasSettings;
 import org.kie.workbench.common.stunner.core.client.canvas.Layer;
 import org.kie.workbench.common.stunner.core.client.canvas.event.CanvasClearEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.CanvasDrawnEvent;
@@ -81,8 +83,8 @@ public abstract class WiresCanvas extends AbstractCanvas<WiresCanvas.View> {
     }
 
     @Override
-    public Canvas initialize(final int width,
-                             final int height) {
+    public Canvas initialize(CanvasSettings settings) {
+        LienzoCore.get().setHidpiEnabled(settings.isHiDPIEnabled());
         return this;
     }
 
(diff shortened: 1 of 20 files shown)


ederign:
@paulovmr can you please talk with Daniel and Roger about preferences scope?

danielzhe:
Thank you for the reviews, @romartin ! I love this kind of code review. :)
I'll do the proposed changes and fix the name of the branches. 

About the scopes, I have to do some tests to see and figure it out. I'll do more tests and see what I discover and how far I go.
Thank you @romartin , @ederign and @paulovmr when you get in touch :D

romartin:
thanks @danielzhe !!

cc @porcelli

porcelli:
@paulovmr will be able to provide all details you'd need, but you can take a look here http://www.uberfireframework.org/docs/uberfireExtensions/preferencesLowLevelStorageAPI.html#per-user-manner

paulovmr:
Hi @romartin and @danielzhe , we do support user scoped preferences. I think the best way to show this is to point at the code. 

We have a `LibraryProjectPreferences` class in kie-wb-common that is a good example. It is basically a POJO with some annotations to allow us to pre-process it in build time, and also generate a UI for the user to modify its values. The `LibraryPreferences` class has an attribute of type `LibraryProjectPreferences` to include it as a child in the preference hierarchy, and also defines a default value for its attributes in the method `LibraryPreferences#defaultValue`.

I think the best way for you to add your preference is to create a class similar to `LibraryProjectPreferences`, let's say `StunnerPreferences`, with your preferences as attributes inside it (hidpiEnabled, width, height, etc). Then create a attribute in `LibraryPreferences` of type `StunnerPreferences`. Also define the default values for your attributes in the method `LibraryPreferences#defaultValue`. By doing this, the preference will be ready to use.

To fetch the preference values in your code, take a look [here](http://www.uberfireframework.org/docs/uberfireExtensions/preferencesBeansAPI.html). Basically you just have to inject your POJO and call the method load (with callbacks when using it in client code).

After all this, to edit your preference in the screen, you just have to go to the admin screen (little cog in the right top corner of the page) and select "Library". There should be an item in the tree with your preference name.

Let me know if you have any questions!

romartin:
Hey @paulovmr , thanks for your explanation. Anyway yeah, if you look at this PR we already have the `StunnerPreferences` bean, it also appears on the admin area as a custom section "Process designer".  But the thing is that we are adding just a new setting (`boolean` type), but we want this setting to be set for each user (user scoped), rather than the already other existing settings that are using the default scope (I guess it's global or per project?)
So I was expecting kinda way to declare the scope for each setting, in some preferences bean, via annotation or whatever... but not sure if that possible. 
Are we on same page? :)
Thanks!

romartin:
ok to test

paulovmr:
@romartin Ok, I should have looked at the code before all the explanation. :) 

So, the admin perspective applies the change globally (global scope) indeed, and you can't customize the scope there (only admin users should have access to it). We don't have yet a screen/perspective for each user to customize its own preferences, although the API supports that already. If you need that, for now the best option I see is to add, for example, one checkbox somewhere (maybe on the editor itself?) where the user can change this setting, and your code calls the preference to change it only for that user:

```
@Inject
StunnerPreferences stunnerPreferences;

// Client:
stunnerPreferences.getDiagramEditorPreferences().setAutoHidePalettePanel(userSetting);
stunnerPreferences.save(stunnerPreferences, successCallback, errorCallback);

// Server:
stunnerPreferences.getDiagramEditorPreferences().setAutoHidePalettePanel(userSetting);
stunnerPreferences.save();
```

romartin:
Hey @paulovmr 
Yes we are no same page now :) Okkk I see, no worries, we'll do it at stunner level in that case.
Thanks!

romartin:
Jenkins please retest this

romartin:

romartin:
hey @tiagodolphine , assigning the review for you as well. QE is not PTO this week and we would like to cherry-pick this to 7.7.x and include it in the next patch release. So please, can you give it a look asap? Thx

tiagodolphine:
@romartin sure I'll do that.

romartin:
Please  @danielzhe can you update this PR's description by specifying the links to the other PRs it depends on? So whoever will merge this will take into account. Thanks!!!

danielzhe:
Depends on [https://github.com/kiegroup/lienzo-core/pull/51](https://github.com/kiegroup/lienzo-core/pull/51)
Tests: [https://github.com/kiegroup/lienzo-tests/pull/33](https://github.com/kiegroup/lienzo-tests/pull/33)

PR #1940, opened 2018-07-03
Updating Stunner Showcase README's with more detailed info

jeremyary:
@mdproctor @ederign @romartin updated as requested to better detail steps needed to get the standalone showcase running and stable.

Can look at the new README over here if it's easier than looking through the diff:
https://github.com/jeremyary/kie-wb-common/tree/showcase_docs/kie-wb-common-stunner


manstis:
ok to test.

manstis:
ok to test.

kie-wb-common-stunner/README.md
```diff
@@ -46,33 +51,167 @@ Here are the instructions for setting up the Stunner environment on **IntelliJ I
...
+5. Select this option, then click *Apply* and/or *OK*.
+
+***Configure the File Header Template***
+1. Open **Preferences** from the **File** (linux/Win) menu or **IntelliJ IDEA** app menu (mac).
+2. In the search field, type '*file and code templates*'.
+3. Select the *Includes* tab.
+4. Select *File Header* in the left-hand list.
+5. Replace the content on the right-hand side with the following, then click *Apply* and/or *OK*:
+
+         /*
```
romartin: minor thing - I would probably prefer referencing the "official" license file to use... it does not usually change, but well given the case, they'll just update that file, this way we don't have to maintain this..

jeremyary: Will do. It looks like by default IntelliJ 15 parses File Header after the package name and doesn't offer a Copyright template, so I need to revisit anyways.

PR #1893, opened 2018-06-12
JBPM-7217 Stunner - Boundary event morphing cause issues

tiagodolphine:
Now it is possible to morph a docked Event on an activity inside a sub-process and change its locations, save and load, and undo/redo without any issues.
The problem was on the `MorphCanvasNodeCommand.java` that has not handling the dock/undock in case morphing a docked node.

@romartin 
@LuboTerifaj 
@hasys

PR #1943, opened 2018-07-04
JBPM-6829 : Stunner - Morphing issues for BPMN activities

tiagodolphine:
Changes related to this PR:
- Morph from tasks to reusable sub-processes
- Morph from reusable sub-processes to tasks
- Service tasks **not**  allowed to be morphed
- Morph between sub-processes (containers) except reusable (that is not a container)
- Fix morph sub-process with children or docked nodes

PS:  I need to update the `MorphCanvasNodeCommandTest.java`  but depending on this https://github.com/kiegroup/kie-wb-common/pull/1893 to be merged before that includes the test class.

@romartin 
@LuboTerifaj

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/adapter/MorphAdapter.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/definition/adapter/MorphAdapter.java
@@ -30,6 +30,11 @@ public interface MorphAdapter<S> extends Adapter {
     <T> Iterable<MorphDefinition> getMorphDefinitions(final T definition);
 
     /**
+     * Returns the morphing definitions for the given DefinitionId and BaseId
+     */
+    Iterable<MorphDefinition> getMorphDefinitions(final String id, final String baseId);
+
+    /**
      * Returns the morphing properties for the given Definition instance, if any.
      */
     <T> Iterable<MorphProperty> getMorphProperties(final T definition);
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/command/MorphCanvasNodeCommand.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/command/MorphCanvasNodeCommand.java
@@ -16,12 +16,14 @@
 package org.kie.workbench.common.stunner.core.client.canvas.command;
 
 import java.util.List;
+import java.util.Optional;
 
 import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
 import org.kie.workbench.common.stunner.core.client.command.CanvasViolation;
 import org.kie.workbench.common.stunner.core.client.shape.EdgeShape;
 import org.kie.workbench.common.stunner.core.client.shape.MutationContext;
 import org.kie.workbench.common.stunner.core.client.shape.Shape;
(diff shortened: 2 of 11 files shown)


tiagodolphine:

tiagodolphine:
Jenkins retest this.

tiagodolphine:
Hey @romartin I think @LuboTerifaj is on PTO, so if this is important to the release we should review and merge ASAP, anyone else to help on this? After merge it is necessary the cherry-pick to 7.7.x.

romartin:
Hey @tiagodolphine 

Thanks, again good job! Did the review and found a couple of things:

* Service tasks cannot be morphed at all. I suspect is due to the classhierarchy and the logic related to this that is not easy to do it, but I guess that is expected to morph service tasks into regular tasks or even re-usable subprocess, although is not possible from those to morph into service task again, only one way. So it depends on how the fix costs we can merge this and implement the service task morphing later or wait for this PR. 

* On the palette now exists a small gap before the reusable sub-process, I suspect that as you changed the class hierarchy, the morphing definitions have changed, and the palette groups by morphing.. soo not even sure yet about how to quickly fix this, ping me if so and well take a look.

screenshot from 2018-07-06 00-06-54

Anyway it's absolutely much better now! The morphing options are making sense and also the menu looks better with no so many icons, so let's try to include this on 7.7.x because it's important... 

Thanks!

tiagodolphine:
@romartin 

- About the Service tasks, what is expected? because in my understanding it was supposed to not be morphed to any other task.  So,  whats is the scenario? I mean it is difficult for now morphing between different service tasks because we load them on runtime... 

- About the palette yeah.. I think this is because the reusable has a different "base class" now... well need to investigate if we can overcome this.

Thanks.

romartin:
Hey @tiagodolphine 

**About service tasks**
Tbh my comment on the ticket was not really clear sorry. IMO should be possible to morph a service task into some regular task (or even reusable subprocess), but not possible to morph it back to any service task again... so only one way. Anyway please @krisv can you confirm? Also notice @tiagodolphine that this "one way" probably makes the impl a bit more difficult .. dunno yet, but we could also create a new ticket for service tasks morphing and go with this code, it's fine for me!

**About the palette "gap"**
The palette relies on a model, so the resulting hierarchy and grouping is given by the model you supply.  By default, the palette model is built by using the category and morph definition for each element, as you changed this a bit, it results on adding the reusable sub-process in a different "group", so it results on a gap on the view side.
I propose to fix it by changing the palette model for BPMN and making the reusable sub-process be part of the same group as the other ones. Here are some comments:
* The key point is on the [ExpandedPaletteDefinitionBuilder](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/components/palette/ExpandedPaletteDefinitionBuilder.java#L137) -> for each element it calls the `morphDefinitionProvider`, which returns the `MorphDefinition`.  As the grouping is based on the id of the MorphDefinition instance returned, here we can return a different/custom instance and make the grouping happen in the way we expect
* So this solution relies on setting a custom `morphDefinitionProvider` function in the `BPMNPaletteDefinitionBuilder`. see [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/components/palette/BPMNPaletteDefinitionBuilder.java#L176).

This way, and only for BPMN, the palette grouping can be customized.

Hope this helps!

tiagodolphine:
@romartin updated the PR the palette issue is fixed, about the service tasks it is better to keep without morphing and open a new JIRA to handle that, this will demand some effort.

tiagodolphine:

tiagodolphine:

tiagodolphine:
Jenkins retest this.

tiagodolphine:
@romartin I think this can be merged, since the respective cherry pick to 7.7.x is already merged.

romartin:
Hey @LuboTerifaj 
Merging this as we already did the review on `7.7.x` during last week. Feel free to open new tickets if you find something related. thanks!

romartin:
@tiagodolphine well the full downstream failed... trying again

romartin:

tiagodolphine:
@LuboTerifaj I haven't created it, so if possible you can do it, plz.
Thx

PR #1946, opened 2018-07-05
DROOLS-2705 : Files having '+' sign in name are getting ignored in Build

Rikkola:
https://issues.jboss.org/browse/DROOLS-2705

@baldimir I see you were marked as the tester.
@manstis You might be the best to take a look.

Thank you.


kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/core/Builder.java
```diff
@@ -398,7 +399,7 @@ private void removeResource(final Path resource) {
     }
 
     private String destinationPath(final Path resource) {
-        final String destinationPath = resource.toUri().toString().substring(projectPrefix.length());
+        final String destinationPath = EncodingUtil.decode(resource.toUri().toString().substring(projectPrefix.length()));
```
Rikkola: Fix is here. The destination path is now decoded.

mareknovotny:
Jenkins retest this

PR #1952, opened 2018-07-05
JBPM-7027 Stunner - docking event to the specific edge of activity

tiagodolphine:
Cherry pick commit f25b5995cfe272c4f594de1ec57d2a291b7defa9 from @romartin fork.

Related https://github.com/kiegroup/lienzo-core/pull/52.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
@@ -22,7 +22,6 @@ import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
@@ -45,6 +45,7 @@ import org.kie.workbench.common.stunner.core.client.canvas.controls.AbstractCanv
 import org.kie.workbench.common.stunner.core.client.canvas.controls.CanvasControl;
 import org.kie.workbench.common.stunner.core.client.canvas.controls.drag.LocationControl;
 import org.kie.workbench.common.stunner.core.client.canvas.controls.keyboard.KeysMatcher;
+import org.kie.workbench.common.stunner.core.client.canvas.controls.select.SelectionControl;
 import org.kie.workbench.common.stunner.core.client.canvas.event.ShapeLocationsChangedEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.selection.CanvasClearSelectionEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.selection.CanvasSelectionEvent;
@@ -58,6 +59,8 @@ import org.kie.workbench.common.stunner.core.client.shape.Shape;
 import org.kie.workbench.common.stunner.core.client.shape.view.HasDragBounds;
 import org.kie.workbench.common.stunner.core.client.shape.view.HasEventHandlers;
 import org.kie.workbench.common.stunner.core.client.shape.view.ShapeView;
+import org.kie.workbench.common.stunner.core.client.shape.view.event.DragEvent;
+import org.kie.workbench.common.stunner.core.client.shape.view.event.DragHandler;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.MouseEnterEvent;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.MouseEnterHandler;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.MouseExitEvent;
@@ -97,6 +100,7 @@ public class LocationControlImpl
     private CommandManagerProvider<AbstractCanvasHandler> commandManagerProvider;
     private double[] boundsConstraint;
     private final Collection<String> selectedIDs = new LinkedList<>();
+    private SelectionControl selectionControl;
 
     protected LocationControlImpl() {
         this(null,
@@ -114,6 +118,7 @@ public class LocationControlImpl
     public void bind(final EditorSession session) {
         // Keyboard event handling.
         session.getKeyboardControl().addKeyShortcutCallback(this::onKeyDownEvent);
+        this.selectionControl = session.getSelectionControl();
     }
 
     private void onKeyDownEvent(final KeyboardEvent.Key... keys) {
@@ -226,6 +231,34 @@ public class LocationControlImpl
                                                 outHandler);
                     registerHandler(shape.getUUID(),
                                     outHandler);
+
+                    //Adding DragHandler on the shape to check whether the moving shape is not selected, th
+                    final DragHandler dragHandler = new DragHandler() {
+                        @Override
+                        public void start(DragEvent event) {
+                            if (selectionControl != null) {
+                                final Element item = canvasHandler.getGraphIndex().get(shape.getUUID());
+                                if (!selectionControl.isSelected(item)) {
+                                    selectionControl.getSelectedItems()
+                                            .forEach(uuid -> selectionControl.deselect(String.valueOf(uuid)));
+                                    selectionControl.select(item);
+                                }
+                            }
+                        }
+
+                        @Override
+                        public void end(DragEvent event) {
+
+                        }
+
+                        @Override
+                        public void handle(DragEvent event) {
+
+                        }
+                    };
+
+                    hasEventHandlers.addHandler(ViewEventType.DRAG, dragHandler);
+                    registerHandler(shape.getUUID(), dragHandler);
                 }
             }
         }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
@@ -21,6 +21,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -45,7 +47,6 @@ import org.kie.workbench.common.stunner.core.client.canvas.controls.AbstractCanv
 import org.kie.workbench.common.stunner.core.client.canvas.controls.CanvasControl;
 import org.kie.workbench.common.stunner.core.client.canvas.controls.drag.LocationControl;
 import org.kie.workbench.common.stunner.core.client.canvas.controls.keyboard.KeysMatcher;
-import org.kie.workbench.common.stunner.core.client.canvas.controls.select.SelectionControl;
 import org.kie.workbench.common.stunner.core.client.canvas.event.ShapeLocationsChangedEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.selection.CanvasClearSelectionEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.selection.CanvasSelectionEvent;
@@ -100,25 +101,27 @@ public class LocationControlImpl
     private CommandManagerProvider<AbstractCanvasHandler> commandManagerProvider;
     private double[] boundsConstraint;
     private final Collection<String> selectedIDs = new LinkedList<>();
-    private SelectionControl selectionControl;
+    private final Event<CanvasSelectionEvent> selectionEvent;
 
     protected LocationControlImpl() {
         this(null,
+             null,
              null);
     }
 
     @Inject
     public LocationControlImpl(final CanvasCommandFactory<AbstractCanvasHandler> canvasCommandFactory,
-                               final Event<ShapeLocationsChangedEvent> shapeLocationsChangedEvent) {
+                               final Event<ShapeLocationsChangedEvent> shapeLocationsChangedEvent,
+                               final Event<CanvasSelectionEvent> selectionEvent) {
         this.canvasCommandFactory = canvasCommandFactory;
         this.shapeLocationsChangedEvent = shapeLocationsChangedEvent;
+        this.selectionEvent = selectionEvent;
     }
 
     @Override
     public void bind(final EditorSession session) {
         // Keyboard event handling.
         session.getKeyboardControl().addKeyShortcutCallback(this::onKeyDownEvent);
-        this.selectionControl = session.getSelectionControl();
     }
 
     private void onKeyDownEvent(final KeyboardEvent.Key... keys) {
@@ -236,13 +239,9 @@ public class LocationControlImpl
                     final DragHandler dragHandler = new DragHandler() {
                         @Override
                         public void start(DragEvent event) {
-                            if (selectionControl != null) {
-                                final Element item = canvasHandler.getGraphIndex().get(shape.getUUID());
-                                if (!selectionControl.isSelected(item)) {
-                                    selectionControl.getSelectedItems()
-                                            .forEach(uuid -> selectionControl.deselect(String.valueOf(uuid)));
-                                    selectionControl.select(item);
-                                }
+                            if (Objects.nonNull(selectionEvent)) {
+                                //select the moving shape, if not
+                                selectionEvent.fire(new CanvasSelectionEvent(canvasHandler, shape.getUUID()));
                             }
                         }
 
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
@@ -47,6 +47,7 @@ import org.kie.workbench.common.stunner.core.client.canvas.Layer;
 import org.kie.workbench.common.stunner.core.client.canvas.command.DefaultCanvasCommandFactory;
 import org.kie.workbench.common.stunner.core.client.canvas.command.UpdateElementPositionCommand;
 import org.kie.workbench.common.stunner.core.client.canvas.event.ShapeLocationsChangedEvent;
+import org.kie.workbench.common.stunner.core.client.canvas.event.selection.CanvasSelectionEvent;
 import org.kie.workbench.common.stunner.core.client.command.CanvasCommand;
 import org.kie.workbench.common.stunner.core.client.command.CanvasCommandFactory;
 import org.kie.workbench.common.stunner.core.client.command.CanvasCommandManager;
@@ -169,6 +170,9 @@ public class LocationControlImplTest {
     @Mock
     private EventSourceMock<ShapeLocationsChangedEvent> shapeLocationsChangedEvent;
 
+    @Mock
+    private EventSourceMock<CanvasSelectionEvent> canvasSelectionEvent;
+
     private CanvasCommandFactory<AbstractCanvasHandler> canvasCommandFactory;
     private ShapeViewExtStub shapeView;
     private LocationControlImpl tested;
@@ -207,7 +211,7 @@ public class LocationControlImplTest {
         when(wiresManager.getSelectionManager()).thenReturn(selectionManager);
         when(selectionManager.getControl()).thenReturn(wiresCompositeControl);
 
-        this.tested = new LocationControlImpl(canvasCommandFactory, shapeLocationsChangedEvent);
+        this.tested = new LocationControlImpl(canvasCommandFactory, shapeLocationsChangedEvent, canvasSelectionEvent);
         tested.setCommandManagerProvider(() -> commandManager);
     }
 
@@ -343,8 +347,7 @@ public class LocationControlImplTest {
         tested.init(canvasHandler);
         tested.register(element);
         tested.deregister(element);
-        verify(shapeEventHandler,
-               times(3)).removeHandler(any(ViewHandler.class));
+        verify(shapeEventHandler, atLeastOnce()).removeHandler(any(ViewHandler.class));
         assertFalse(tested.isRegistered(element));
     }
 
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
@@ -344,7 +344,7 @@ public class LocationControlImplTest {
         tested.register(element);
         tested.deregister(element);
         verify(shapeEventHandler,
-               times(1)).removeHandler(any(ViewHandler.class));
+               times(3)).removeHandler(any(ViewHandler.class));
         assertFalse(tested.isRegistered(element));
     }
 
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
@@ -56,6 +56,8 @@ import org.kie.workbench.common.stunner.core.client.shape.ShapeViewExtStub;
 import org.kie.workbench.common.stunner.core.client.shape.view.HasControlPoints;
 import org.kie.workbench.common.stunner.core.client.shape.view.HasEventHandlers;
 import org.kie.workbench.common.stunner.core.client.shape.view.ShapeView;
+import org.kie.workbench.common.stunner.core.client.shape.view.event.DragEvent;
+import org.kie.workbench.common.stunner.core.client.shape.view.event.DragHandler;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.MouseEnterHandler;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.ViewEventType;
 import org.kie.workbench.common.stunner.core.client.shape.view.event.ViewHandler;
@@ -82,6 +84,7 @@ import org.uberfire.mocks.EventSourceMock;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentCaptor.forClass;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.atLeastOnce;
@@ -230,7 +233,7 @@ public class LocationControlImplTest {
         tested.register(element);
         assertTrue(tested.isRegistered(element));
         verify(shapeView, times(1)).setDragEnabled(eq(true));
-        ArgumentCaptor<BoundingBox> bbCaptor = ArgumentCaptor.forClass(BoundingBox.class);
+        ArgumentCaptor<BoundingBox> bbCaptor = forClass(BoundingBox.class);
         verify(wiresCompositeControl, times(1)).setBoundsConstraint(bbCaptor.capture());
         BoundingBox bb = bbCaptor.getValue();
         assertEquals(1.0d, bb.getMinX(), 0d);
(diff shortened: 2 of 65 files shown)


romartin:

romartin:
Hey @wmedvede @Josephblt 
I'm doing the review for this, but need for another candidate for it.. please feel free to give this a try, as far as I know it affects docking, the align&distribution and some other calculations that affect the bounding box etc. So you could test from the UI: boundaries (docking), trying to add / remove children (containment), resizing, check the distribution guided lines... 
Thanks in advance!

tiagodolphine:

tiagodolphine:
Jenkins retest this.

romartin:
the build is still failing due to compilation errors, not sure yet tbh.. I'll try to run it locally in a while and will give you feedback

romartin:
@tiagodolphine yeah I guess was the first run, where was missing the PR for tests... Iit's building for me. Thanks

romartin:
@tiagodolphine  @mdproctor Well now compared with master, i can reproduce the issues about undo/redo related to docking as well, but the other reported ones are candidates to be introduced by this commits. Will keep doing tests and checks.. And thanks both for the collaboration!!

tiagodolphine:
@romartin about 1, the docking directly from the palette yeah... this could be implemented because the previous impl works betting the closest magnet from the given position.
About 2,3 I think it is bug on Stunner side to be fixed.
The 4,5 probably is bug on lienzo side.

Well, WDYT? we need to have a plan to attack these issues, but I'm afraid it will not be finished and tested until the release.

romartin:
Hey @hasys 
This PR from @tiagodolphine is the one that contains most of the lienzo fixes from @mdproctor PR's. We need to introduce this ASAP in order to be able to be on sync with the blessed, also because it introduces fixings and improvments.
I did already some tests and found a few bugs, would appreciate if you can also do some testing from the UI and check we're not breaking other stuff.
Also wer're missing more tests on the lienzo side, we'll add them, no worries, but makes sense start testing this. 
BTW @tiagodolphine lemme know if I can help, at least on the testing side, I know here are missing quite a few tests... if you identify them, feel free to split and assign some for me, just lemme know and I'll start coding :) Thx

tiagodolphine:
@romartin let's track the issues you found and others that eventually comes, on the JIRA, we can use this JBPM-7027 for that or create a new one.. this PR is aggregating more things than the specified JBPM-7027 I used... so I think it makes sense to update the JIRA with all the stuff and new issues found,  this is easier than create a new JIRA... WDYT ?
Anyway I'll start working on this.

romartin:
Hey @tiagodolphine 
Well I guess some of those issues are due to the changes from this PR, so once you analyze it we can create new tickets as well (the only issue that was already existing on master was the docking undo/redo)

hasys:
Hi @LuboTerifaj, can I ask you to take a look at this PR?

LuboTerifaj:
@hasys 
yes of course, just have to wait for downstream build.

LuboTerifaj:

tiagodolphine:
@romartin @hasys @LuboTerifaj  I have updated the PR with the fixes:

**1. Docking directly from the palette**: now it is working docking a boundary event to an activity, it is attached according to the nearest intersection on the edge. 
**2. Delete docked nodes**: now it is working properly deleting docked nodes and undo/redo
**3. Delete container nodes/ dock parent:** now it is working deleting container nodes with children and nodes with other docked nodes in it, including the undo/redo.
**4. Node Border (docking area) issue:** now the issue with the border when selecting and dragging is fixed.
**5. Zoom/Pan with docked nodes**: this seems to be solved at least after the other fixes I was not able to reproduce
**6. Drag parent with a Docked/Child node selected**: there was a problem when selecting a docked node and dragging the parent, the toolbox was not moving, not it is keeping the dragged shape (parent) selected when dragging.
**7. Fix handler registering/deregistering on AbstractCanvasHandlerRegistrationControl**: basically the handlers were using a Map with the uuid as key, but when registering many handlers to the same shape, the value was being override in this way the deregister was not being executed for all handlers, just for the last one registered, now it is using a set o handlers for each shape.

PS: the tests are going to be updated/implemented but I just updated the PR to than it could be reviewed since this is a important patch.

tiagodolphine:

tiagodolphine:
Jenkins retest this.

romartin:
Comment for @LuboTerifaj  and for me - remember to test this using mediators (zoom, pan) and after scrolling (client vs rel vs abs locations)

LuboTerifaj:
Hi @tiagodolphine , @romartin 
I will have a look at this one on Monday. I have to go now. 
Sorry for inconvenience.
Thanks.

tiagodolphine:
@romartin @LuboTerifaj  updated the PR because I found and fixed these issues:
- Dock a boundary event on a task inside a subprocess than delete the event, undo and try to dock the event to the subprocess, this was duplicating the event on canvas.
-  Toolbox quick animation was appearing when deleting a docked node
-  Call the existing `WiresDockingControl.getCandidateLocation()` to update the candidate location after docking, this is used when docking en event directly from the palette( @romartin this is about the comment you made on the PR https://github.com/kiegroup/lienzo-core/pull/52)

tiagodolphine:
Jenkins retest this.

tiagodolphine:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/SelectionControl.java
```diff
@@ -30,11 +30,11 @@
 
     SelectionControl<C, E> select(final E item);
 
-    SelectionControl<C, E> deselect(final E item);
+    SelectionControl<C, E> deselect(final String item);
```
romartin: hmm it's fine changing the signature but looking at the other methods `select` and  `isSelected`, they have an `E` as argument... looks like a bit confusing moving to `String` here, isn't it? well just my opinion... :) any reason for this?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/AbstractCanvasHandlerRegistrationControl.java
```diff
@@ -107,19 +113,23 @@ public boolean isRegistered(final Element element) {
     }
 
     protected Set<String> getRegisteredElements() {
-        return handlers.keySet();
+        return handlers.entrySet()
+                .stream()
+                .filter(entry ->  Objects.nonNull(entry.getValue()) && !entry.getValue().isEmpty())
```
romartin: minoooor comment and just personal opinion as well - doesn't look easier to read splitting the logical operation (`&&`) into another `filter` intermediate operation, so two filters concatenated?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/AbstractCanvasHandlerRegistrationControl.java
```diff
@@ -107,19 +113,23 @@ public boolean isRegistered(final Element element) {
...
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toSet());
     }
 
     protected boolean isRegistered(final String uuid) {
-        return handlers.containsKey(uuid);
+        return handlers.containsKey(uuid) && !handlers.get(uuid).isEmpty();
```
romartin: can this case be given? I mean having an entry for the given `uuid` but with no registrations?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/DeleteSelectionSessionCommand.java
```diff
@@ -92,29 +102,31 @@ public boolean accepts(final ClientSession session) {
...
-                // Check the results.
-                if (!CommandUtils.isError(result)) {
-                    callback.onSuccess();
-                } else {
-                    callback.onError((V) new ClientRuntimeError("Error deleing elements [message=" +
-                                                                        result.toString() + "]"));
-                }
+                selectionControl.clearSelection();
+
+                timerUtils.executeWithDelay(()-> {
```
romartin: why delaying the operation???

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
```diff
@@ -232,6 +237,34 @@ public void handle(MouseExitEvent event) {
                                                 outHandler);
                     registerHandler(shape.getUUID(),
                                     outHandler);
+
+                    //Adding DragHandler on the shape to check whether the moving shape is not selected, th
+                    final DragHandler dragHandler = new DragHandler() {
```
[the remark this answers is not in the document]
romartin: So the idea is that a CanvasControl (like the selection, location ones) is responsible for providing 
 some capability on top of the canvas, this way controls can add any kind of event handler on shapes that apply, np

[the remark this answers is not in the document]
tiagodolphine: @romartin yes, I can try to add on the SelectionControl, but just to clarify the thing is, when you are trying to move a shape that is not selected, in this case I check if the shape being moved is selected or not, and if not than I select it. So this logic can be on the location in case of the movement on shape or on selection if we think that the moving shape needs to be selected. It depends on it is simpler to understand. 
Do you think it is better to insert this on the SelectionControl?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
```diff
@@ -232,6 +237,34 @@ public void handle(MouseExitEvent event) {
...
+                        @Override
+                        public void start(DragEvent event) {
+                            if (selectionControl != null) {
+                                final Element item = canvasHandler.getGraphIndex().get(shape.getUUID());
+                                if (!selectionControl.isSelected(item)) {
+                                    selectionControl.getSelectedItems()
```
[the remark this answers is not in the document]
tiagodolphine: Agree, I'll do that.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/SelectionControl.java
tiagodolphine: I just changed the signature because to deselect the Element in fact you just need the `key` and if you look at the impl we are just using the UUID for that. In this way this allow us to deselect without having to get the Element on the Graph, in case we don't have a reference to the Element itself were calling is being made. Anyway it was just trying to avoid unnecessary calls, well I can keep the argument as `E` to keep it more consistent, no problem.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/AbstractCanvasHandlerRegistrationControl.java
tiagodolphine: ok, I can do this to make it more clear!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/AbstractCanvasHandlerRegistrationControl.java
tiagodolphine: humm in theory no, it was just to avoid any error/concurrency when deregistering a handler, but since this code has no concurrency, so, I think we can remove the second condition.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/DeleteSelectionSessionCommand.java
tiagodolphine: @romartin hey this is the trick part, the same on the Morphing, remember, the Toolbox appearing on the wrong location, in my understanding this is caused by the animation on the Toolbox to show/hide. Do you have a better solution? In summary it is necessary to perform an action after the animation on the toolbox has finished, in this case Delete the selection, but could be Morphing a node as well.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImplTest.java
```diff
@@ -346,7 +346,7 @@ public void testDeregister() {
         tested.register(element);
         tested.deregister(element);
         verify(shapeEventHandler,
-               times(1)).removeHandler(any(ViewHandler.class));
+               times(3)).removeHandler(any(ViewHandler.class));
```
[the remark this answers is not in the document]
tiagodolphine: Ok

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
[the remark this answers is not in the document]
romartin: ahhh mmmm yes I see... ok makes sense for me :)
On the other hand, I would like to remove sooner than later the method bind on the CanvasControls, in order to avoid references between them, I would prefer or compositing them on top or just, as I propose here too, relying on CDI. So what about, instead of adding the selectionControl reference in this class, just throwing a CDI event for selecting the node with is being drag. So that's fine to keep logic here but trying to decouple both controls in some way... does it makes sense?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/SelectionControl.java
romartin: yes I'm also agree with you on this! I mean for deselecting sounds easier to just use the identifier, as you said, to avoid unnecessary calls from the other components for obtaining the node by uuid... 
Soo I'm just brainstorming, what you think is better - keeping all methods same signature or not? 
@evacchi as you're a lover of the language internals and semantics... wdyt? :)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/DeleteSelectionSessionCommand.java
romartin: mmmmm yes I remember... tbh it was fine when we're busy for the release but at this point I would prefer to do this in some other way, just because this timer at least depends on the animation duration, so "hard-coding" here the milliseconds is a bit tricky yeah.. :)
Just thinking....

romartin: ok I have an idea that would make this easy to handle... so considering:
* Once clicking or selecting some shape, it gets into the SELECTED state, and once you deselect it, or once clearing the selection (which does the call to deselect) it gets into the NONE state
* Actually the default behavior is to use animations for the shape state changes (See ShapeStateAttributeAnimationHandler )
So what about using animations ONLY when getting into the SELECTED state, and do not using them when deselecting (and clearing selection). This way we can ensure that after the call to SelectionControl.clearSelection(), the toolbox has been removed.
Does it makes sense? Should this work for removing those timers? 
Thanks!

romartin: Just for the records - in fact for me the best option would be:
* Using animations when selecting/deselecting/clearing via CDI events
* Do not use animations when calling the SelectionControl API
But I think this is not easy at this point, so proposed the solution above.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LocationControlImpl.java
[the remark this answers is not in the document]
tiagodolphine: @romartin sure, I do prefer events as well :+1: , I'll change the impl.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/SelectionControl.java
evacchi: don't have enough context here. I'd prefer symmetry over asymmetry, but then, maybe you don't need/have the whole object at hand just to "deselect"; compare collection APIs such as

- `add(T)`
 - `remove(id)`

where usually it's enough the key/id to remove the element

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/DeleteSelectionSessionCommand.java
tiagodolphine: @romartin totally agree with you on this, the delay is far from a good resolution, but at least it proves that the issue was about the animation. 
So I'll investigate to solve it on the best way like you said animation when using events and not use animation when using the API, but in fact we can use a parameter on events and API to specify if we want or not animation on that call, and maybe the default is to have animations, WDYT ? anyway I'll investigate if it is feasible to implement on this PR.
I'm just afraid that this PR will take more time than were expecting, but I prefer to make things better if we have time for that.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/SelectionControl.java
tiagodolphine: @romartin @evacchi going to the code, the select in fact just uses the ID as well, so I think we can keep the symmetry by having:
```
SelectionControl<C, E> select(final String uuid);
SelectionControl<C, E> deselect(final String uuid);
```
In general I prefer to keep the interfaces as flexible and simple as they could be.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/DeleteSelectionSessionCommand.java
romartin: @tiagodolphine  ok sounds good for me. I'm not sure if will be easy to fire the de/selection with some options (liek animations), as actually the thing happens on Shape.applyState(ShapeState), so that would imply an API refactor... 
well not sure either, but agree on spending a bit more time on this and trying to fix it ASAP, otherwise some bug for this can appear sooner than later.. :)
Thx!

tiagodolphine:

tiagodolphine:
Jenkins retest this.

tiagodolphine:

tiagodolphine:
Jenkins retest this.

tiagodolphine:
Hey @romartin @hasys , the full downstream is ready it is ok to review. 
I fixed the issues @LuboTerifaj found + roger suggestions + fixes bugs when Deleting nodes with children, docked nodes, childdren with docked nodes, change parent of docked nodes, undo/redo on these cases, so if you give attention to more complex cases like these.. 

Thanks guys!!

tiagodolphine:
@hasys it is missing the tests, I'll work on that as soon as all the issues and refactors are ok, on manual testing.

tiagodolphine:
Hi @LuboTerifaj  thanks for the review.
Follows my comments:

- Resize activity to the smallest size: I think we can open a new JIRA, since this is not related to this PR and this PR is getting to long.

- Boundary event change its position after resizing: I will check this, could be relatead to the changes on PR.

- Mediators: for my tests this is not related to this PR, it seems that the mediators has issues with other stuff, for instance the Align/Distribute get lost on positions as well, so I think it is better to open a new JIRA for these mediators issues.

- Toolbox quick animation :  I will check this condition, because the toolbox hide animation should be disabled, maybe I missed some point.

@LuboTerifaj  @romatin WDYT about these points?

Besides these issues did you get any other related to Docking or any regression that this PR may caused?

Thanks again for the detailed review.

tiagodolphine:
Hi @LuboTerifaj the issue about _Boundary event change its position after resizing_ is fixed, great catch, I need to make a check on lienzo-core, so no modification here.  
_Toolbox quick animation_ I tested and I was not able to reproduce anymore, can you try again, and if it occurs I need the steps.
The full downstream is ok, I tested the war now, so if you have a chance to review again these points.

Thanks.

tiagodolphine:
Jenkins retest this

tiagodolphine:

tiagodolphine:
Jenkins retest this

tiagodolphine:
Hi @LuboTerifaj I resolved the conflicts with master and pushed, the full downstream is ok to review.
Thanks.

tiagodolphine:
Hey @LuboTerifaj thanks for the review, about the quick animation, now I could reproduce it still happening just when delete the event using the toolbox button, when using the Delete key it is not happening I'll investigate it.

Can you please create the JIRA for the issues you found that are not related to this PR.

I'll let you know as soon as I fix the issue with the animation.

Thanks.

tiagodolphine:
@LuboTerifaj issue with quick animation when deleting events from the toolbox is fixed.
Thanks.

LuboTerifaj:
Reported issues:
[JBPM-7558](https://issues.jboss.org/browse/JBPM-7558) - missing the red resize "button" after resizing activity to the smallest size
[JBPM-7559](https://issues.jboss.org/browse/JBPM-7559) - boundary event is displayed twice after zooming canvas

LuboTerifaj:

LuboTerifaj:
Jenkins retest please.

wmedvede:
Hi @tiagodolphine I've tried to build this PR "as is" locally and got build errors, and when I've tried to rebase with master Iv'e got merge errors. Is this PR properly updated with master? I can see git is not complaining, but probably it's still not rebaseable with master.

tiagodolphine:
hey @wmedvede this PR is quite complicated because I need to merge some branches from @romartin in this one and solve conflicts, so, now if you rebase you will get lot of conflicts, so the best way is to just fetch/merge the master into this branch if you can build locally, I did this on the last commit. Anyway to test this PR you can download the WAR file on the jenkins build and run on a local wildfly. 
Thanks.

tiagodolphine:
@LuboTerifaj did you have a chance to check the last update regarding the quick animation?
Thanks.

tiagodolphine:
@LuboTerifaj thanks !
@wmedvede waiting for your feedback to merge this one, just to remember that his one is related to kiegroup/lienzo-core#52 
Thanks.

LuboTerifaj:
Hello @tiagodolphine 
After all, I have found one issue yet, but I think, that it is not directly related to this PR so I have reported it as new one. The issue is present also on master, but with a little bit different behaviour. For more information see: [JBPM-7593](https://issues.jboss.org/browse/JBPM-7593)

tiagodolphine:
Hey @LuboTerifaj great, thanks for reporting, yeah let's work on this new JIRA.

wmedvede:
@tiagodolphine yes, my plan is to check this running. But the question is, the full-down-stream build seems to have failed. It still have artifacts to download, but are this ones safe?

wmedvede:
Hi @tiagodolphine this are my finding after using the built version taken from this PR.

1) When I dock a catching intermediate event, and then try to move/dragg it to the "Embedded Subprocess" I get an exception. The exception seems to start looping and makes the editor unusable.
In the videos below I show two use cases that creates the docked event in two different ways with same results.
https://youtu.be/Y-P2GBh0IZw
https://youtu.be/OtU8jPh1SIM

2) This second video shows a weird effect when we dock an event, and then creates a second event. The second event should ideally be placed near to the source event.

https://youtu.be/TuA0xtB926A

tiagodolphine:

tiagodolphine:
Jenkins retest please.

tiagodolphine:
@wmedvede thanks for the review, good catch, I fixed the issues you found. The last one, related to the position when creating a sequence flow from a docked event was not related to the PR, but I fixed it too since it was a simple change on the 'CanvasLayoutUtils`.

wmedvede:
@tiagodolphine One last thing I could see, but not sure exactly what could be the best solution.
In general intermediate catching events can be used not only as boundary events. They can also be used in the "normal flow" of the process/subprocess. e.g. you can have TaskA -> TimerB -> TaskC. In TimerB the process may want to wait for some time, and continue to TaskC when the required time was elapsed. (just an example, but same use case might occur for some other intermediate catching events). This is fine.

Ok, let's now imagine the user trying to model the use case above. When the use case is modeled in the top level process everything is fine.
But when same use case is modeled in an Embedded Subprocess, the usecase requires "extra steps" to be done.
1) Create the embedded subprocess
2) drag the "intermediate catching event" from the palette into the embedded subprocess.
When we leave the mouse, the node is automatically placed as a docked node. Because it's "automatically" assumed that all intermediate catching events dragged into the embedded subprocess are boundary events.
3) So we need do an extra step for moving the mode from the subprocess boundary, into the desired position inside the process.

The alternative path is:
1) drag the "intermediate catching event" from the palette into the embedded subprocess
2) now drag the node into the embedded subprocess. 
Here an extra step is required but is less confusing.

I believe this is something we can live with, but wondering if this might confuse users. Leave this comment here for the record or in case there's a quick trick that can be done so solve it. If not, I believe we should go ahead.

wmedvede:
@tiagodolphine sorry, but I've found this last error. (still with the build form yesterday)
See video below.
https://youtu.be/tophPn-XK4Y

I think it depends a bit on the position where the node is left. Since once I saw the error it took some time to reproduce, and also there are cases when you model the same process as the video and the error is not produced.
This is the generated process for this case.

[TestProcess.bpmn.txt](https://github.com/kiegroup/kie-wb-common/files/2265661/TestProcess.bpmn.txt)

tiagodolphine:

tiagodolphine:
Hi @wmedvede thanks again for the review !
About your first comment, yeah, now if we want to use the `Intermediate catching event` on a normal flow, not on boundary, than we need an additional step, but this is known and we decided to give preference to dock the event as a "default" action, this could be solved in case the drag/drop from palette is changed and we get the real shape on the canvas, but for now as I said before while dragging we don't have the shape and we don't have a way to show the "blue border" when docking. IMO the most used use case is to dock the intermediate catching event, so it is better than default action to add as child, anyway I think this could be discussed on a JIRA for improvement, this PR is not changing this behavior it just fixes to match the modifications on lienzo.

About the second comment, it is not related but I found a solution, the issue was caused because we were docking a node on canvas before adding the parent, so I changed to check it the parent was added before, take a look on the `DrawCanvasCommand`.
@LuboTerifaj I think it is better you take a look as well after these modifications, please.

Thanks guys!!

tiagodolphine:

LuboTerifaj:
Hi @tiagodolphine , @wmedvede

Oh yes, issue about docking directly from palette was discussed in the past like Tiago already mentioned. Maybe even some Jira exist for that so I would live it as it is.

I will check the rest changes as soon as the full downstrema build will be ready.
I see that it is still pending over 23 hours so I will let to build new one.

Thanks.

LuboTerifaj:

tiagodolphine:
Hi @LuboTerifaj thanks for the review, I think it is better to open a new JIRA related to the node creation via toolbox, this issue is more related to the `CanvasLayoutUtis`, that will be more complicated to solve, I tried a quick solution, maybe it is better to check on the master, the issue should be present there as well. I this way we can proceed with this PR that is quite long now covering many unrelated issues and it is blocking the lienzo sync with blessed one, WDYT ?

evacchi:
my only 2c, please make sure all of this work as intended with the new marshalling backend (it should be already default in master)

LuboTerifaj:
Reported issue: https://issues.jboss.org/browse/JBPM-7621

tiagodolphine:
Hi @LuboTerifaj  as I said on the meeting, I'm getting an issue with new marshallers enabled.
Create Task with a docked event and move it inside a subprocess, save and open the diagram. When opening the task will be not inside the subprocess.

hasys:
Hi @manstis, this PR is ready to merge. Thank you!

@tiagodolphine new marshaller issues are not related to this PR, please report it as separate Jira. Thank you!

hasys:

tiagodolphine:
@manstis @hasys full downstream is ok now.

PR #1955, opened 2018-07-06
DROOLS-2655: [DMN Designer]: Decision Table: Automatically create input columns for each InputData element

manstis:
See https://issues.jboss.org/browse/DROOLS-2655

PR #1963, opened 2018-07-11
DROOLS-2721: [DMN Designer] InputData node can not be connected to BKM node

manstis:
See https://issues.jboss.org/browse/DROOLS-2721

@jomarko I also tested this fix with https://github.com/kiegroup/kie-wb-common/pull/1955 and it worked OK.


jomarko:
Approved without manual check, will manul check as part of https://github.com/kiegroup/kie-wb-common/pull/1955

jomarko:
@manstis user is now not able to store diagram where `InputData` is connected to `BKM` because `BusinessKnowledgeModelConverter` throws and error on this [line](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/BusinessKnowledgeModelConverter.java#L98)

Not sure if the `BusinessKnowledgeModelConverter` has to be updated or this PR reverted.

manstis:
@jomarko I am tempted to revert. The OMG DMNv1.1 Specification states:

> 6.2.2.1 Information Requirement notation
> Information Requirements may be drawn from Input Data elements to Decisions, and from Decisions to other Decisions. They represent the dependency of a Decision on information from input data or the results of other Decisions...

No mention of BKMs at all. Let me chat with @tarilabs to understand.

PR #1725, opened 2018-04-30
JBPM-6730  Stunner - BPMN marshallers re-implementation: switch the default implementation to new

evacchi:
Features:

   - Enables new marshallers by default (you can still opt out with `-Dbpmn.marshaller.experimental=false`)
   - Makes my life miserable

Please merge after 

- [x]  #1724 
- [x] #1722 
- [x] #1650
- [x] #1633 
- [x] #1616  
- [x] #1779
- [x] #1741
- [x] #1732 
- [x] #1793
- [x] #1799
- [x] #1606  (blocked by above)
- [x] #1851
- [x] #1871
- [x] #1922
- [x] #1956
- [x] #1964
- [x] #1968
- [x] #1969
- [x] #1978

PR #1969, opened 2018-07-12
JBPM-7432: Stunner - New Marshallers: Adding a boundary event to a subprocess throws ClassCastException

evacchi:


kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/processes/AbstractProcessConverter.java
```diff
@@ -39,7 +39,7 @@
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toSet;
 
-public class AbstractProcessConverter {
+public abstract class AbstractProcessConverter {
```
romartin: hey @evacchi , IMO this abstraction does not really makes much sense... isn't it? I mean looking at what this class does, it looks like it's a set of helper methods that rely on the `DefinitionsBuildingContext`, not really on the instance state... why not using composition/aggregation instead of inheritance? Just brainstorming! :+1:

evacchi:
Agree. In fact... At the beginning this *was* a delegate. At some point I turned it into a base class because I thought its purpose would have been "clearer". Thanks for the advice will turn this back into a delegate (maybe another PR though)

evacchi:

evacchi:
style broken, fixed. Changed abstract classes to delegates.

evacchi:
see #1725

PR #1971, opened 2018-07-12
Initial effort to clean up code base from Double Brace initialization

hasys:
Hi @romartin,

I started clean up Stunner code base from double brace initialization because this way of initialization lead to generation tons of useless inner classes and can lead to memory leaks.

Please provide feedback is you OK with it. And if yes I will continue my work on it.

CC @evacchi @Josephblt @tiagodolphine @wmedvede

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
@@ -25,9 +25,6 @@ import javax.enterprise.inject.Any;
 import javax.inject.Inject;
 
 import com.google.gwt.user.client.ui.IsWidget;
-import org.gwtbootstrap3.client.ui.AnchorListItem;
-import org.gwtbootstrap3.client.ui.Button;
-import org.gwtbootstrap3.client.ui.ButtonGroup;
 import org.gwtbootstrap3.client.ui.DropDownMenu;
 import org.gwtbootstrap3.client.ui.constants.ButtonSize;
 import org.gwtbootstrap3.client.ui.constants.IconType;
@@ -61,12 +58,11 @@ public class MenuDevCommandsBuilder {
     }
 
     public MenuItem build() {
-        final DropDownMenus.Builder menuBuilder = new DropDownMenus.Builder().addStyleName("pull-right");
+        final DropDownMenu menu = new DropDownMenus.Builder().addStyleName("pull-right").build();
         for (final MenuDevCommand command : devCommands) {
-            menuBuilder.add(new AnchorListItems.Builder(command.getText())
+            menu.add(new AnchorListItems.Builder(command.getText())
                                     .setIcon(command.getIcon())
-                                    .addClickHandler(event -> command.execute())
-                                    .build());
+                             .addClickHandler(event -> command.execute()).build());
         }
 
         final IsWidget group = new ButtonGroups.Builder()
@@ -75,10 +71,8 @@ public class MenuDevCommandsBuilder {
                              .setDataToggle(Toggle.DROPDOWN)
                              .setIcon(IconType.COG)
                              .setSize(ButtonSize.SMALL)
-                             .setTitle("Development")
-                             .build())
-                .add(menuBuilder.build())
-                .build();
+                             .setTitle("Development").build())
+                .add(menu).build();
 
         return MenuUtils.buildItem(group);
     }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
@@ -34,6 +34,10 @@ import org.gwtbootstrap3.client.ui.constants.IconType;
 import org.gwtbootstrap3.client.ui.constants.Toggle;
 import org.jboss.errai.ioc.client.api.ManagedInstance;
 import org.kie.workbench.common.stunner.client.widgets.menu.MenuUtils;
+import org.kie.workbench.common.stunner.core.util.AnchorListItems;
+import org.kie.workbench.common.stunner.core.util.ButtonGroups;
+import org.kie.workbench.common.stunner.core.util.Buttons;
+import org.kie.workbench.common.stunner.core.util.DropDownMenus;
 import org.uberfire.workbench.model.menu.MenuItem;
 
 /**
@@ -57,25 +61,25 @@ public class MenuDevCommandsBuilder {
     }
 
     public MenuItem build() {
-        final DropDownMenu menu = new DropDownMenu() {{
-            addStyleName("pull-right");
-        }};
+        final DropDownMenus.Builder menuBuilder = new DropDownMenus.Builder().addStyleName("pull-right");
         for (final MenuDevCommand command : devCommands) {
-            menu.add(new AnchorListItem(command.getText()) {{
-                setIcon(command.getIcon());
-                addClickHandler(event -> command.execute());
-            }});
+            menuBuilder.add(new AnchorListItems.Builder(command.getText())
+                                    .setIcon(command.getIcon())
+                                    .addClickHandler(event -> command.execute())
+                                    .build());
         }
-        final IsWidget group = new ButtonGroup() {{
-            add(new Button() {{
-                setToggleCaret(false);
-                setDataToggle(Toggle.DROPDOWN);
-                setIcon(IconType.COG);
-                setSize(ButtonSize.SMALL);
-                setTitle("Development");
-            }});
-            add(menu);
-        }};
+
+        final IsWidget group = new ButtonGroups.Builder()
+                .add(new Buttons.Builder()
+                             .setToggleCaret(false)
+                             .setDataToggle(Toggle.DROPDOWN)
+                             .setIcon(IconType.COG)
+                             .setSize(ButtonSize.SMALL)
+                             .setTitle("Development")
+                             .build())
+                .add(menuBuilder.build())
+                .build();
+
         return MenuUtils.buildItem(group);
     }
 
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
@@ -61,8 +61,9 @@ public class MenuDevCommandsBuilder {
         final DropDownMenu menu = new DropDownMenus.Builder().addStyleName("pull-right").build();
         for (final MenuDevCommand command : devCommands) {
             menu.add(new AnchorListItems.Builder(command.getText())
-                                    .setIcon(command.getIcon())
-                             .addClickHandler(event -> command.execute()).build());
+                             .setIcon(command.getIcon())
+                             .addClickHandler(event -> command.execute()).build()
+            );
         }
 
         final IsWidget group = new ButtonGroups.Builder()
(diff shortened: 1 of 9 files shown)


romartin:
Hey @hasys 
Yes in general I'm agree, although it also means there is no need to refactor ALL instances with double brace implementation, right? 

I mean for example creating some new collection and adding primitive types as:
```
new HashMap<String, Object>() {{
                                                    put("init",
                                                        Boolean.TRUE);
                                                    put("internal",
                                                         Boolean.TRUE);
                                                }}
```
This does not tend to leak nothing...

evacchi:
@romartin a **new** anonymous inner class it's created each time this idiom is used. If you use this in 3 different region of the same class, you'd have generate 3 distinct, invisible, anonymous classes! This contributes to grow our codebase for very little gain: the number of lines of code is exactly the same!

What we *can* do instead, is refactor these to constructors, or small static methods.
In java 9 we would use [Map.of(...)](https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#ofEntries-java.util.Map.Entry...-). We cannot depend on JDK 9, but I think we can depend from Guava. In that case we could use [ImmutableMap.of(...)](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableMap.html#of--). And, if not even this is viable, it's really a matter of seconds to roll our own 

```
     import static org... ...Maps.entry;

     Map<Integer,String> map = Maps.ofEntries(
         entry(1, "a"),
         entry(2, "b"),
         entry(3, "c"),
         ...
         entry(26, "z"));

```

manstis:
@romartin @evacchi So it's down to memory usage for (anonymous) ```Class``` definitions not memory usage of instances of classes.. If I understand correctly.

evacchi:
@manstis 


As for leaks, it depends if you, by chance, refer to the outer scope from within the `{{ ... }}`; e.g.: a field in the enclosing class; the compiler *might* inject a reference to the whole outer class **inside** the anonymous inner class you are creating!

e.g. (TBF not 100% of this example, but you get the idea)

```java
class Foo {
  String blah;
  void foo() {
                                           new HashMap<String, Object>() {{
                                                    put("blah",
                                                        this.blah);

                                                }}
  }

}
```

romartin:
Yeah agree with you guys, by using `{{ ... }}` as for the instance creation results on more complexity and can lead to more memory usage and potential leaks. 

But it doesn't mean we're actually leaking, it depends on the way the inner class is implemented, as @evacchi already commented. in fact also depends on the compiler and the JVM optimizations being done under the hood (at bytecode level, JVM memory region, etc), but AFAIK most common is that if you guarantee immutability and concurrency for the memory accessed within the inner class, it should not cause any issues, so you can access constants or thread-safe immutable variables in a safe way, isn't it?

Anyway good too see this refactor @hasys , that's a cool idea :+1:  

Also IMO agree with @evacchi about using static factory methods, although notice that Guava does not provide client side support.... so I would vote for adding the factory methods for Map, etc etc, as you mention, as new stuff in the Stunner core, making it also available for both compilers (shared).  @hasys wdyt?

Thanks guys for the good brainstorming! :+1:

romartin:
@hasys I'm agree with @evacchi on creating static factory methods.... you? Well added @evacchi  and @tiagodolphine as reviewers too, let us know when this is ready for review!! 
Thanks!

hasys:
@romartin ok, so should I create factory method also for Lists and Maps? Where is a suitable place for such classes?

romartin:
Hey @hasys yes, please! As I understand we're just going to create a few factory methods for Maps and Lists, and the implementations that will return will be the default ones, which are also supported by GWT, I would say to create this class/es in the `kie-wb-common-stunner-core-common module`,  this way it will be visible from both server and client side. For example, you can create it on the [util](https://github.com/kiegroup/kie-wb-common/tree/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util) package, up to you! And thanks!!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/Maps.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Maps {
```
wmedvede: @hasys @manstis Wondering if it could make sense to add this class in kie-soup-commons, in the package org.kie.soup.commons.util

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/Sets.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.util;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class Sets {
```
wmedvede: @hasys same for this one.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/util/MapsTest.java
```diff
@@ -0,0 +1,66 @@
...
+        final String VALUE_1 = "some value";
+        final String VALUE_2 = "different value";
+
+        Map<Double, String> map = new Maps.Builder<Double, String>()
+                .put(KEY_1, VALUE_1)
+                .put(KEY_2, VALUE_2)
+                .build();
+
+        assertEquals(2, map.size());
+        assertEquals(map.get(KEY_1), VALUE_1);
```
wmedvede: @hasys silly comment, but usually the expected value goes as the first parameter of the assertEquals.
I believe this could be assertEquals(VALUE_1, map.get(KEY_1));  instead.

hasys: Good point, thank you! Fixed and rebased to master.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/Maps.java
hasys: @wmedvede, yes good point. Fine with me, if other projects leaders find it useful will move it to more general Project/package. Thank you!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/Sets.java
hasys: @wmedvede same here, if others OK with it, I will move, no problems. Thank you!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/Maps.java
manstis: @wmedvede IMO it would be good to push these type of utility classes to as high up as possible.

hasys: Hi @manstis, thank you, does it mean it should be moved to `org.kie.soup.commons.util` (is it part of `kie-wb-common` or it is a part of another repository?) as suggested by @wmedvede or there are even better place for it? Unfortunately it is not clear to me :(

manstis: Hi @hasys @wmedvede is meaning move the classes to [here](https://github.com/kiegroup/kie-soup/tree/master/kie-soup-commons/src/main/java/org/kie/soup/commons/util).

hasys: @manstis ok, great. Thank you, will do it.

hasys: Hi @wmedvede I sent a PR to kie-soap repo: https://github.com/kiegroup/kie-soup/pull/52 when it will be merged I will update this one.

hasys:
Hi @wmedvede, @manstis,

I moved Maps, Sets and Lists to kie-soup repository, see https://github.com/kiegroup/kie-soup/pull/52

Also, I moved all Sets, Lists and Maps related issues to separate PR:
https://github.com/kiegroup/kie-wb-common/pull/2090

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/MenuDevCommandsBuilder.java
```diff
@@ -57,25 +61,25 @@ public void init() {
     }
 
     public MenuItem build() {
-        final DropDownMenu menu = new DropDownMenu() {{
-            addStyleName("pull-right");
-        }};
+        final DropDownMenus.Builder menuBuilder = new DropDownMenus.Builder().addStyleName("pull-right");
```
wmedvede: I would probably have finished the initialization of the DropDownMenu here
final DropDownMenu menu = new DropDownMenus.Builder().addStyleName("pull-right").build();

and instead of doing 
.add(menuBuilder.build()) in the lines below I would have done .add(menu) .
I believe it's more clear in this case. But just a personal preference.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/menu/dev/ShapeSetsMenuItemsBuilder.java
```diff
@@ -55,29 +58,29 @@ public ShapeSetsMenuItemsBuilder(final ShapeManager shapeManager) {
     public MenuItem build(final String title,
                           final String prefix,
                           final Callback callback) {
-        final DropDownMenu menu = new DropDownMenu() {{
-            addStyleName("pull-right");
-        }};
+        final DropDownMenus.Builder menuBuilder = new DropDownMenus.Builder();
```
wmedvede: same here

hasys:
Jenkins retest please

hasys:
Jenkins retest please

hasys:

hasys:
Jenkins retest please

hasys:
Hi @romartin, @manstis,

enforcer plugin fails build due to this errors: 

```[WARNING] Rule 0: de.is24.maven.enforcer.rules.IllegalTransitiveDependencyCheck failed with message:
Found 7 illegal transitive type dependencies in artifact 'org.kie.workbench.stunner:kie-wb-common-stunner-core-common:jar:7.12.0-SNAPSHOT':
1.) org.gwtbootstrap3.client.ui.AnchorListItem
2.) org.gwtbootstrap3.client.ui.Button
3.) org.gwtbootstrap3.client.ui.ButtonGroup
4.) org.gwtbootstrap3.client.ui.DropDownMenu
5.) org.gwtbootstrap3.client.ui.constants.ButtonSize
6.) org.gwtbootstrap3.client.ui.constants.IconType
7.) org.gwtbootstrap3.client.ui.constants.Toggle
```
to be honest, not sure what does it mean and how can I fix it. Do you have an idea?

manstis:
@hasys You need to find the JAR(s) containing each of those classes and ensure its an explicit dependency in whatever applicable modules `pom.xml` (there's an enforcer rule to check we don't use _transient_ dependencies but have them all explicitly declared).

hasys:
Thank you @manstis, fix pushed!

romartin:

hasys:
Hi @romartin,

this pr is ready to merge.

romartin:
Hey @hasys , sorry after merging other PRs from you this needs a fix, there are code conflicts. Thanks!

hasys:
Hi @romartin,

rebased to master and conflicts merged. Let's build make green again! :)

hasys:
Jenkins retest please

hasys:
Jenkins retest please

hasys:

hasys:
Jenkins retest please

hasys:

PR #1976, opened 2018-07-16
AF-1405: Can No Longer Download Text Files From Business Central

karreiro:
See:
- https://issues.jboss.org/browse/RHPAM-1248
- https://issues.jboss.org/browse/AF-1405

demo

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/419
- https://github.com/kiegroup/kie-wb-common/pull/1976
- https://github.com/kiegroup/drools-wb/pull/889


karreiro:
Thank you, everyone, for the comments.

jomarko:
Hi @romartin @manstis @karreiro @hasys.

During review of #1984 I realized, stunner based editors now have two download buttons for the same. See the `Download` and `Download ad DMN` in the attached screenshot. I think we have these options, could you please provide your thoughts?

- Keep both of them
- Keep separately
  - Download as JPG, PNG, SVG, PDF
  - Remove Download as DMN, just current Download will remain
- Remove current Download, keep all Download as items

hasys:
Hi @jomarko, @manstis, @karreiro, @romartin,

I guess @wrnkld should decide how it should look like.

manstis:
I think Stunner based editors (with multiple download options) should override the _default_ download toolbar button. Obviously the new _default_ download button was added after Stunner's button so there was little Stunner could do about it... but IMO we should simply override the default with a specific where applicable. A new JIRA for sure.

jomarko:
Well, filed issue here https://issues.jboss.org/browse/JBPM-7461

PR #1977, opened 2018-07-16
Stunner - Clean dependencies for lienzo-extensions module

romartin:
Hey @manstis @hasys 

As @mdproctor noticed, the `stunner-lienzo-extensions` module, which was supposed to depend only on lienzo, was also including also a few dependencies from uf/errai, and should be not present.

It was only because the use of the class `org.uberfire.mvp.Command`, so I refactored this for `Runnable`. This way the dependency tree for this module is the excepted one. 

On the other hand I remember some talks with you @manstis about this - you was proposing not to use `Runnable` for client side stuff, is it? If so, which other functional interface can we use?

Thanks!!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-lienzo-extensions/pom.xml
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-lienzo-extensions/pom.xml
@@ -45,38 +45,6 @@
       <scope>provided</scope>
     </dependency>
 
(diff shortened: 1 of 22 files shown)


romartin:

manstis:
@danielezonca IIRC you've moved to JDK's ```Consumer``` interface to replace use of AF's ```Command```?

manstis:
Thanks @danielezonca I forgot the specifics.

romartin:
ok great thanks @manstis @danielezonca :+1:

PR #1988, opened 2018-07-19
DROOLS-2566: [DMN Editor] Integrate Editor into Workbench

manstis:
See https://issues.jboss.org/browse/DROOLS-2566

This also incorporates https://issues.jboss.org/browse/DROOLS-2748

@romartin I had to include a change to ```WorkItemDefinitionClientRegistry``` to only clear the registry if one was found (otherwise the DMN Editor was throwing a ```NPE``` when the session was destroyed).

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/pom.xml
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/pom.xml
@@ -164,6 +164,12 @@
       <artifactId>kie-soup-commons</artifactId>
     </dependency>
 
+    <!-- Drools DMN model -->
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-model</artifactId>
+    </dependency>
+
   </dependencies>
 
 </project>
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/pom.xml
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/pom.xml
@@ -164,12 +164,6 @@
       <artifactId>kie-soup-commons</artifactId>
     </dependency>
 
-    <!-- Drools DMN model -->
-    <dependency>
-      <groupId>org.kie</groupId>
-      <artifactId>kie-dmn-model</artifactId>
-    </dependency>
-
   </dependencies>
 
 </project>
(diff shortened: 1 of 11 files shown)


manstis:
Part of an ensemble:

- https://github.com/kiegroup/kie-wb-common/pull/1988
- https://github.com/kiegroup/kie-wb-distributions/pull/781

manstis:
@jomarko @karreiro Could you please endeavour to review today?

@etirelli has requested this integration be in 7.9.0.Final.

manstis:
Removed @romartin as he's approved the ```WorkItemDefinitionClientRegistry``` [here](https://github.com/kiegroup/kie-wb-common/pull/1989).

kie-wb-common-dmn/kie-wb-common-dmn-api/pom.xml
```diff
@@ -164,12 +164,6 @@
       <artifactId>kie-soup-commons</artifactId>
     </dependency>
 
-    <!-- Drools DMN model -->
-    <dependency>
-      <groupId>org.kie</groupId>
-      <artifactId>kie-dmn-model</artifactId>
```
jomarko: I see due to this change, we have to reintroduce the  [same constants](https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-model/src/main/java/org/kie/dmn/model/v1_1/DMNModelInstrumentedBase.java), what I think is not ideal. Could you please explain slightly why we can't depend on this module more?

manstis: @jomarko Yes, I wasn't keen on it either. I did it because GWT was failing to compile ```kie-wb-distributions``` (which was weird as both the DMN showcase webapp and ```drools-wb-webapp``` worked OK). Anyway, I've reverted this specific change in a new commit and found a fix on https://github.com/kiegroup/kie-wb-distributions/pull/781 to keep GWT happy (third commit).

kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/authz/PermissionTreeSetup.java
```diff
@@ -22,10 +22,12 @@
 
 import org.guvnor.structure.client.security.OrganizationalUnitTreeProvider;
 import org.guvnor.structure.client.security.RepositoryTreeProvider;
+import org.kie.workbench.common.workbench.client.EditorIds;
```
karreiro: I think that we could remove this line.

manstis: Removed. Thanks.

romartin:
@manstis btw I guess that the bug you mention on the description here, the one you fixed, is already being fixed by @tiagodolphine , just FYI :+1:

PR #1991, opened 2018-07-19
[AF-1406] Asset cannot be created in default package

ederign:
@Rikkola @tomasdavidorg 	

I'm not 100% sure if this is the right solution or I just 'shoot the messenger'.

But what happens is:

If the default package is select at first, all good because the value will be set to "default".

But if we change the package to another one, and then come back to default the select value will be "" and we cannot lookup the selected package.

@Rikkola this issue is for Friday EOD, could you please take a look if this is the best option or should I do something else? 

screenshot 2018-07-19 17 45 29

screenshot 2018-07-19 17 45 44

### DIFF
--- a/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/handlers/PackageListBoxViewImpl.java
+++ b/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/handlers/PackageListBoxViewImpl.java
@@ -68,7 +68,10 @@ public class PackageListBoxViewImpl
         return options;
     }
 
-    KieSelectElement.Option newOption(final String name, final String value) {
+    KieSelectElement.Option newOption(final String name, String value) {
+        if (value.isEmpty()) {
+            value = name;
+        }
         return new KieSelectElement.Option(name, value);
     }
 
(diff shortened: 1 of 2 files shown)


tomasdavidorg:
jenkins retest this

kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/handlers/PackageListBoxViewImpl.java
```diff
@@ -68,7 +68,10 @@ public void setUp(final String activePackage,
         return options;
     }
 
-    KieSelectElement.Option newOption(final String name, final String value) {
+    KieSelectElement.Option newOption(final String name, String value) {
+        if (value.isEmpty()) {
+            value = name;
+        }
         return new KieSelectElement.Option(name, value);
```
jomarko: Do we really have to store something to the method parameter? Can't we do this?
```java
return new KieSelectElement.Option(name, value.isEmpty() ? name : value);
```

ederign: In this case yes. I want to make it clear in the code.

ederign:
jenkins retest this

ederign:
thanks @tomasdavidorg : https://github.com/kiegroup/kie-wb-common/pull/1992 7.7.x

ederign:
and thanks @Rikkola  :)

ederign:
jenkins retest this.

PR #2005, opened 2018-07-26
JBPM-7505 - Only update view when event is related to same container …

cristianonicolai:
…spec


bsremac:
@cristianonicolai I have built kie-wb off of the PR, but the same error is still being thrown.
This is the entire reproducer screencast:
[Screencast from 30-07-18 00:50:54.zip](https://github.com/kiegroup/kie-wb-common/files/2239259/Screencast.from.30-07-18.00.50.54.zip)

The environment used was: EAP 7.1.3 and the latest Kie Server 7.10.0 snapshot.

cristianonicolai:
@bsremac I think it has to do with the fact that my changes were still based on _7.9.0-SNAOSHOT_, I've rebased it to be included into _7.10.0-SNAPSHOT_ now, let me know if it works for you now.

bsremac:
@cristianonicolai It seems that was the problem indeed. After rebuilding and testing it again, the error is not showing anymore.

PR #2008, opened 2018-07-27
DROOLS-2564: [DMN Designer] Data-types: Grid: General

manstis:
See https://issues.jboss.org/browse/DROOLS-2564


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditor.java
```diff
@@ -94,10 +101,26 @@ public void exit() {
...
+            final Node node = (Node) element;
+            if (node.getContent() instanceof Definition) {
+                final Definition definition = (Definition) node.getContent();
+                hasExpression.ifPresent(e -> {
+                    if (e.equals(definition.getDefinition())) {
+                        view.setReturnToDRGText(Optional.ofNullable((HasName) definition.getDefinition()));
+                    }
+                });
+            }
+        }
```
jomarko: The three inner blocks make the code hard readable, but I was not able to come up with more readable code. So probably we can keep as it is with one expcetion. We do not need the line `final Node node = (Node) element;`. Also we should minimize the dangerous of NPE by `Objects.equals(e, definition.getDefinition())`  The code will then looks like:
```java
final Element<?> element = event.getElement();
        if ((element instanceof Node)) {
            if (element.getContent() instanceof Definition) {
                final Definition definition = (Definition) element.getContent();
                hasExpression.ifPresent(e -> {
                    if (Objects.equals(e, definition.getDefinition())) {
                        view.setReturnToDRGText(Optional.ofNullable((HasName) definition.getDefinition()));
                    }
                });
            }
        }
```

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorTest.java
```diff
@@ -158,4 +173,23 @@ public void testOnCanvasFocusedSelectionEvent() {
 
         verify(testedEditor).exit();
     }
+
+    @Test
+    public void testOnCanvasElementUpdated() {
```
jomarko: We should test different node update!

manstis:
@jomarko This is OK for review now thank-you.

jomarko:
e.g. Node renamed but decision table has still output named with old name. 
screenshot from 2018-08-01 12-46-51

manstis:
@jomarko Thanks for the review. Regarding the issue with the grid header not updating until clicked on I have a fix for that in DROOLS-2792 (no PR yet, as it's work in progress). I will move the fix from there to this PR and let you know when you can re-review.

manstis:
@jomarko New commit added fixing the reported issue.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/layer/DMNGridLayerProducer.java
```diff
@@ -17,18 +17,33 @@
...
     @DMNEditor
     public DMNGridLayer getGridLayer() {
         return layer;
     }
+
+    DMNGridLayer makeGridLayer() {
+        return new DMNGridLayer();
+    }
+
+    @SuppressWarnings("unused")
```
karreiro: I think that this line could be removed.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditor.java
```diff
@@ -94,10 +102,25 @@ public void exit() {
         });
     }
 
-    public void onCanvasFocusedSelectionEvent(@Observes CanvasSelectionEvent event) {
+    @SuppressWarnings("unused")
```
karreiro: I think that this line could be removed.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/layer/DMNGridLayerProducer.java
manstis: It prevents IDEA from reporting the ```event``` parameter is not used.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditor.java
manstis: It prevents IDEA from reporting the ```event``` parameter is not used.

manstis:
@karreiro Thank-you for your comments about the use of ```Optional```. I agree that the DMN editor code is littered with many _optional_ parameters however in my defence I _think_ I could either keep the optional parameter or, everywhere the methods are invoked, add a check for whether a parameter exists for a expression grid type and call one of the overloaded methods. The base expression grid code is generic for different expression types and some have all the parameter types and some do not.

However please do create a JIRA (under EPIC https://issues.jboss.org/browse/BAPL-970) and we can revisit the use of optional at some future date.

PR #2018, opened 2018-08-06
JBPM-7596: Stunner - New Marshallers: default bpmn.marshaller.experimental must be true

evacchi:
please @hasys @LuboTerifaj take a look at this ASAP, otherwise the default to the marshallers will be still FALSE !!

You can verify it is actually true. When you open stunner for the first time you should read something like this on the console:

```
16:14:57,450 INFO  [org.kie.workbench.common.stunner.bpmn.backend.BPMNBackendService] (default task-57) bpmn.marshaller.experimental = true
```


evacchi:
apparently an issue on my machine

PR #2038, opened 2018-08-15
DROOLS-2777: Guided Decision Table is changing date field value based on the timezone

karreiro:
See:
https://issues.jboss.org/browse/DROOLS-2777

---

Demo:
- First, we need to set the `user.timezone` property:

- ..and then, we can change the timezone to **"Europe/Vilnius"** in the server (`systemsetup -settimezone Europe/Vilnius`):

- ..or even change to **"Europe/Prague"** (`systemsetup -settimezone Europe/Prague`):

..the client side is always using the timezone from the server (from `user.timezone`), so the displayed value will always be consistent.

cc @etirelli 

---

Part of an ensemble:
- https://github.com/kiegroup/kie-wb-common/pull/2038
- https://github.com/kiegroup/drools-wb/pull/911

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/preferences/KieTimeZonePreferencesLoader.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/preferences/KieTimeZonePreferencesLoader.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.services.backend.preferences;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TimeZone;
+
+import org.guvnor.common.services.backend.preferences.ApplicationPreferencesLoader;
+import org.kie.workbench.common.services.shared.preferences.ApplicationPreferences;
+
+public class KieTimeZonePreferencesLoader implements ApplicationPreferencesLoader {
+
+    @Override
+    public Map<String, String> load() {
+
+        final Map<String, String> preferences = new HashMap<>();
+        final int offset = getOffset();
+        final String offsetValue = Integer.toString(offset);
+
+        preferences.put(ApplicationPreferences.KIE_TIMEZONE_OFFSET, offsetValue);
+
+        return preferences;
+    }
+
+    private int getOffset() {
+        final TimeZone timeZone = getTimeZone();
+        return timeZone.getOffset(0);
+    }
+
+    private TimeZone getTimeZone() {
+        final String timezone = getSystemPropertyTimeZone();
+
+        if (timezone == null || timezone.isEmpty()) {
+            return TimeZone.getDefault();
+        } else {
+            return TimeZone.getTimeZone(timezone);
+        }
+    }
+
+    String getSystemPropertyTimeZone() {
+        return System.getProperty("user.timezone");
+    }
+}
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/preferences/KieTimeZonePreferencesLoaderTest.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/preferences/KieTimeZonePreferencesLoaderTest.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.services.backend.preferences;
+
+import java.util.Map;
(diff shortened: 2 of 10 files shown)


etirelli:
@karreiro I still don't think this is the correct approach, as we should not be doing any timezone conversion at all for date values (no time) at authoring time. 

But ok, how does this solution work if 2 users in different timezones are using the same server in a different timezone?

manstis:
@karreiro I wonder whether using a similar approach to what you have implemented is more suitable. Instead of getting a time-zone offset and adjusting the time-zones etc on the browser before formatting the {{Date}} with the adjusted time-zone; can we not just capture the server-size time-zone as a property and use this in the browser when formatting the {{Date}} object? Just a thought.. given @etirelli comment above.

karreiro:
@etirelli Considering the following scenario:
    - server in the "Europe/Vilnius" timezone;
    - user A in the "Europe/Prague" timezone;
    - user B in the "America/Sao_Paulo".

User A and user B will see on the screen the date using "Europe/Vilnius" timezone.

---

@manstis Capturing the timezone from the server, in order to use it on the client side was my first idea. However, the API that we have available on client side (`com.google.gwt.i18n.client.DateTimeFormat`, `com.google.gwt.i18n.client.TimeZone`) is not flexible enough to covert a date based on a simple `String` property (like "UTC" or "America/Sao_Paulo"). Thus, I'm getting the offset (relative to UTC) from the server to be able to convert dates in the client side (using the `com.google.gwt.i18n.client.TimeZone`). Technically is the "same" thing, just a little bit more prolix, right?

jomarko:
I would like to review, are we able to give me some time?

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/preferences/KieTimeZonePreferencesLoader.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private int getOffset() {
+        final TimeZone timeZone = getTimeZone();
+        return timeZone.getOffset(0);
+    }
+
+    private TimeZone getTimeZone() {
+        final String timezone = getSystemPropertyTimeZone();
+
+        if (timezone.isEmpty()) {
```
jomarko: Documentation of `System.getProperty()` says that `null` can be returned if wanted system property is not set. I think we should make the if more defensive:

```java
if (timezone == null || timezone.isEmpty()) {
...
```

karreiro:
@manstis @jomarko Changes applied. I fixed issue too (now, when the user selects a date in the date picker, the timezone is persisted correctly - no matter the timezone in the client side).

karreiro:
Jenkins please retest this

kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/preferences/KieTimeZonePreferencesLoaderTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private KieTimeZonePreferencesLoader loader;
+
+    @Before
+    public void setup() {
+        loader = new KieTimeZonePreferencesLoader();
+    }
+
+    @Test
+    public void testGetTimeZoneWhenTimeZoneIsHST() {
+        System.setProperty("user.timezone", "HST");
```
jomarko: I think we need to add `@After` or `@AfterClass` method that will restore `"user.timezone"` system property to original value.

kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/test/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCellTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private PopupDateEditCell cell;
+
+    private boolean autoHidePartnerAdded = false;
+
+    private boolean autoHidePartnerRemoved = false;
+
+    @Before
+    public void setup() {
+        System.setProperty("user.timezone", "Europe/Vilnius");
```
jomarko: Similar comment about restoring the original value.

kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/test/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCellTest.java
```diff
@@ -0,0 +1,198 @@
...
+    private DatePicker datePicker;
+
+    @GwtMock
+    private ValueUpdater<Date> valueUpdater;
+
+    @GwtMock
+    private SafeHtmlRenderer<String> renderer;
+
+    @Captor
+    private ArgumentCaptor<ShowHandler> showHandlerCaptor;
```
jomarko: Please remove, it is never used.

kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/test/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCellTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Captor
+    private ArgumentCaptor<HideHandler> hideHandlerCaptor;
```
jomarko: Please remove, it is never used.

kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/test/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCellTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Captor
+    private ArgumentCaptor<Date> dateCaptor;
+
+    private PopupDateEditCell cell;
+
+    private boolean autoHidePartnerAdded = false;
```
jomarko: Please remove, it is never used.

kie-wb-common-widgets/kie-wb-decorated-grid-widget/src/test/java/org/kie/workbench/common/widgets/decoratedgrid/client/widget/cells/PopupDateEditCellTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private boolean autoHidePartnerRemoved = false;
```
jomarko: Please remove, it is never used.

jomarko:
@karreiro please see my comments and I can approve, manual check passed.

karreiro:
@jomarko Changes applied. (I've proffered to remove the `"user.timezone"` from the tests to avoid collateral effects.. since we're in the unit test level here, I think that it's fair enough approach).

Thank you.

PR #2045, opened 2018-08-20
JBPM-7645 Stunner: New Marshallers - task w/ docked node in subprocess

evacchi:
This fixes the issue. 

Please notice that the underlying issue is that processing nodes in the sequence coming from the XML may lead to unexpected results, when the ordering is not the "right" one. In this case, we had 2 issues

**Marshalling**. Docked nodes are rendered (added) to the XML representation *before* the nodes they were docked to (e.g., the event before its parent task); the generated XML looked like this:

```xml
      <bpmn2:boundaryEvent id="_D37CB404-5190-4CF4-92A1-E5BDBFDD3276"
                           drools:boundaryca="true"
                           drools:dockerinfo="126.0^61.0|"
                           name="Event"
                           attachedToRef="_6A26F0A2-3368-4769-B9E9-A6290530ED8F">

        <bpmn2:timerEventDefinition id="_7K3JlqRjEeimWcuZ5UDpfg"/>
      </bpmn2:boundaryEvent>

      <bpmn2:businessRuleTask id="_6A26F0A2-3368-4769-B9E9-A6290530ED8F" name="Task">
      </bpmn2:businessRuleTask>
```

**Unmarshalling**.  As a rule of thumb, nodes are usually processed in the order they are read from XML; thus commands to the CommandManager, are issued in that specific order. Therefore, we were

1. adding a boundary event
2. setting its position
3. adding its parent task
4. updating its position

this somehow caused coordinates to be wrong/rendered wrong (I don't know exactly)

Please notice this is a bug (IMO this is a bug, we should not rely to the order of the serialization) both in the new and the old marshallers.

So, what we are doing now in the new marshallers is the following:

**Marshalling**: ensure that docked nodes (events) always come **after** other kind of nodes, both in processes and subprocesses. This should not be necessary, but it ensures backwards compatibility with the old marshallers

**Unmarshalling**: ensure that docked nodes are always processed **after** other nodes in processes and subprocesses. If they come *before* other nodes, we defer execution of their command (i.e. we reorder the converted nodes so that commands are sent in the right order)  

cc: @tiagodolphine @hasys @romartin

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java
@@ -112,7 +112,13 @@ public class ProcessPropertyWriter extends BasePropertyWriter implements Element
 
     public void addChildElement(PropertyWriter p) {
         this.childElements.put(p.getElement().getId(), p);
-        process.getFlowElements().add(p.getFlowElement());
+        // compatibility fix: boundary events should always occur at the bottom
+        // otherwise they will be drawn at an incorrect position on load
+        if (p instanceof BoundaryEventPropertyWriter) {
+            process.getFlowElements().add(p.getFlowElement());
+        } else {
+            process.getFlowElements().add(0, p.getFlowElement());
+        }
 
         ElementParameters simulationParameters = p.getSimulationParameters();
         if (simulationParameters != null) {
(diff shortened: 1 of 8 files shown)


evacchi:
Jenkins please retest this

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java
```diff
@@ -112,7 +112,13 @@ public BPMNDiagram getBpmnDiagram() {
 
     public void addChildElement(PropertyWriter p) {
         this.childElements.put(p.getElement().getId(), p);
-        process.getFlowElements().add(p.getFlowElement());
+        // compatibility fix: boundary events should always occur at the bottom
+        // otherwise they will be drawn at an incorrect position on load
+        if (p instanceof BoundaryEventPropertyWriter) {
```
romartin: this code cam be shared across process/subprocess property writers, isn't it? In some helper method or abstraction works for me...

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/SubProcessPropertyWriter.java
```diff
@@ -54,7 +54,13 @@ public SubProcessPropertyWriter(SubProcess process, VariableScope variableScope)
 
     public void addChildElement(PropertyWriter p) {
         this.childElements.put(p.getElement().getId(), p);
-        process.getFlowElements().add(p.getFlowElement());
+        // compatibility fix: boundary events should always occur at the bottom
+        // otherwise they will be drawn at an incorrect position on load
+        if (p instanceof BoundaryEventPropertyWriter) {
```
romartin: code duplication

hasys:

evacchi:
Jenkins please retest this

evacchi:

evacchi:
failures unrelated @hasys can QE take a look?

evacchi:
@romartin [merge it good](https://www.youtube.com/watch?v=j_QLzthSkfM)

tiagodolphine:
@evacchi lol for the **merge it good** video.
Thanks for the PR and the explanation.

PR #2048, opened 2018-08-21
PLANNER-1159: Improve categories

yurloc:
- Add text label for the Optimization category.
- ~Move Package resource type to Model category.~
- Improve Package resource type icon.

Related PRs:
kiegroup/appformer/pull/438
kiegroup/kie-wb-common/pull/2048
kiegroup/drools-wb/pull/914
kiegroup/optaplanner-wb/pull/296

## Add Asset (Model)
screenshot from 2018-08-20 16-53-43

## Add Asset (All)
categories_package

## Old Package icon
categories_package_icon


yurloc:
Poll: Should we move Package to Model category?

manstis:
@yurloc I'm less keen on "Package" being moved to "Model" as it's also a way to categories rules, processes etc and not just Data Objects (which, IIUC, is what "Model" represents along with some peripheral asset types).. but again.. the decision should probably be agreed with the original requester and not an executive engineering decision.

kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/src/main/java/org/kie/workbench/common/screens/projecteditor/client/type/PackageResourceTypeDefinition.java
```diff
@@ -0,0 +1,61 @@
+package org.kie.workbench.common.screens.projecteditor.client.type;
```
Rikkola: License.

Rikkola:
In one of the UX/Design calls it came up that package should be something you add or set when creating a new asset. Since it isn't really a resource, but before that work gets done this is fine by me as long as it is checked with UX.

ederign:
@wrnkld could you please take a look on this package question?

wrnkld:
I agree that package isn't really an asset but I think it is fine for now and shouldn't hold up this PR.

PR #2049, opened 2018-08-21
BAQE-550: Stabilize community tests on jenkins

domhanak:
Removed check that maven repo ends with ".m2/repository"
Added isNull check & isDirectory check.

### DIFF
--- a/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/main/java/org/kie/workbench/common/services/backend/compiler/impl/utils/MavenUtils.java
+++ b/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/main/java/org/kie/workbench/common/services/backend/compiler/impl/utils/MavenUtils.java
@@ -17,6 +17,7 @@
 package org.kie.workbench.common.services.backend.compiler.impl.utils;
 
 import java.io.ByteArrayInputStream;
+import java.io.File;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -48,6 +49,7 @@ import org.uberfire.java.nio.file.Paths;
 public class MavenUtils {
 
     private static final Logger logger = LoggerFactory.getLogger(MavenUtils.class);
+    private static final String JENKINS_SETTINGS_XML_FILE = "JENKINS_SETTINGS_XML_FILE";
 
     public static List<Artifact> resolveDependenciesFromMultimodulePrj(List<String> pomsPaths) {
         MavenXpp3Reader reader = new MavenXpp3Reader();
@@ -126,6 +128,17 @@ public class MavenUtils {
         }
 
         return repository;
+    }
+
+    public static String getSettingsFile() {
+        String jenkinsFile = System.getenv().get(JENKINS_SETTINGS_XML_FILE);
+        if (jenkinsFile != null) {
+            logger.info("Using settings.xml file provided by JENKINS:{}", jenkinsFile);
+            return jenkinsFile;
+        } else {
+            logger.info("Using local settings.xml file.");
+            return new File("src/test/settings.xml").getAbsolutePath();
+        }
     }
 
     private static String getTempRepo() {
--- a/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/test/java/org/kie/workbench/common/services/backend/compiler/ClassLoaderProviderTest.java
+++ b/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/test/java/org/kie/workbench/common/services/backend/compiler/ClassLoaderProviderTest.java
@@ -56,12 +56,15 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 public class ClassLoaderProviderTest {
 
+    private static final Logger logger = LoggerFactory.getLogger(ClassLoaderProviderTest.class);
+    private static final String MAVEN_MAIN_SKIP_PROPERTY = "maven.main.skip";
+    private static final String REPOSITORY_PREFIX = "repo";
+    private static final String REPOSITORY_DIR_NAME = "dummy";
+
     private String mavenRepo;
-    private Path tmpRoot;
-    private Path tmp;
+    private Path tmpRepositoryRoot;
+    private Path tmpRepository;
     private Path uberfireTmp;
-    private final String MAVEN_MAIN_SKIP = "maven.main.skip";
-    private Logger logger = LoggerFactory.getLogger(ClassLoaderProviderTest.class);
 
     @Rule
     public TestName testName = new TestName();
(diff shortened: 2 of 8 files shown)


domhanak:
@winklerm @desmax74 Please take a look.

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/test/java/org/kie/workbench/common/services/backend/compiler/utils/MavenUtilsTest.java
```diff
@@ -62,19 +64,32 @@ public void resolveDependenciesFromSinglePrjTest() {
...
         });
     }
 
     @Test
     public void getMavenLocalRepoDirTest() {
         String testRepo = MavenUtils.getMavenRepoDir(MavenRepos.LOCAL);
-        assertThat(testRepo).endsWith(".m2/repository/");
+        File mavenRepository = FileUtils.getFile(testRepo);
+
+        SoftAssertions.assertSoftly(softly -> {
```
jhrcek: TL;DR: simplify to `assertThat(mavenRepository).isDirectory();`

Just 3 nitpicks about assertj usage:
1. The whole point of `SoftAssertions.assertSoftly` is that you don't have to thing about calling assertAll at the end (it does it [internally for you](https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/SoftAssertions.java#L159))

2. You don't event need soft assertions for doing multiple asserts on one object, you can just do
```java
assertThat(x).assertionMethod1()
             .assertionMethod2()...;
```
3. The call to `isNotNull` is redundant, because most assertj methods that rely on object being non-null [call it automatically](https://github.com/joel-costigliola/assertj-core/blob/749aac4d4e967dfb2b5806dd1794156a55d7325d/src/main/java/org/assertj/core/internal/Files.java#L202).

domhanak: let me rebase man

domhanak:
_Checking the failures_

jhrcek:
@domhanak nice. Your assert seems to have found a bug probably introduced by [latest changes](https://github.com/kiegroup/kie-wb-common/commit/27db878452da58721cb38d87f307a6381d08c9d8)

desmax74:
@domhanak The MavenUtils ask to Aether the url to use, this url couldn't exists on the FS if Maven isn't yet executed, at the first run checks the url of the Maven repo and if is not present is created.
Instead, about the wrong url this is a problem to fix because now we are reading the M2_REPO from the jenkins env and maybe this could conflict with Aether used in the MavenUtils (implemented before the check of the Jenkins env about M2_REPO, implemented two weeks ago), I'll fix next week.

domhanak:
jenkins retest this

domhanak:
jenkins retest this

domhanak:
@jhrcek Looks like the 3rd build got lucky :) Let's leave this open until we discuss this possible bug with Max and get more information.

domhanak:
@desmax74  Do you have any updates here? I updated the tests to check if the return path is a directory and removed the exist(".m2/repository") check. Let me know if it is OK and desired outcome of the tested method.

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/test/java/org/kie/workbench/common/services/backend/compiler/ClassLoaderProviderTest.java
```diff
@@ -205,6 +206,59 @@ public void getResourcesFromADroolsPRJ() throws Exception {
...
+        assertThat(res.getDependenciesAsURI()).hasSize(4);
+
+        KieModuleMetaData kieModuleMetaData = new KieModuleMetaDataImpl((InternalKieModule) kModule,
+                                                                        res.getDependenciesAsURI());
+
+        assertThat(kieModuleMetaData).isNotNull();
+
+        List<String> resources = CompilerClassloaderUtils.getStringFromTargets(tmpRoot);
+        assertThat(resources).hasSize(3);
+        TestUtil.rm(tmpRoot.toFile());
```
jhrcek: This cleanup is already in the `@After` method in this class. Please remove it from the test (Actually while you're at it, please remove it from all the tests from this class, because it's already in the After :)

domhanak:
jenkins retest this

desmax74:
@domhanak fix for the envVars https://github.com/kiegroup/kie-wb-common/pull/2067

domhanak:
Updated with changes that should fix majority of issues from here: https://rhba-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/BxMS/job/RHPAM-7.1-nightly/job/baseline/job/blessed-community-tests-bcentral-baseline/lastCompletedBuild/testReport/

@desmax74 @jhrcek Feedback is welcome. Just fyi I will be probably adding a couple of commits.

domhanak:
@desmax74 If you check 06f9754 commit, I added the -s argument with path to settings into some calls, where I deemed it relevant and valid. Wdyt?

desmax74:
@domhanak is correct the fix, is the same present in the core module

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/main/java/org/kie/workbench/common/services/backend/compiler/impl/utils/MavenUtils.java
```diff
@@ -128,6 +130,17 @@ public static String getMavenRepoDir(MavenRepos repo) {
...
     }
 
+    public static String getSettingsFile() {
+        String jenkinsFile = System.getenv().get(JENKINS_SETTINGS_XML_FILE);
+        if (jenkinsFile != null) {
+            logger.info("Using settings.xml file provided by JENKINS:{}", jenkinsFile);
+            return jenkinsFile;
+        } else {
+            logger.info("Using local settings.xml file.");
+            return new File("src/test/settings.xml").getAbsolutePath();
```
domhanak: @desmax74  I am not sure about this `else` branch pointing to "src/tests/" folder. Is that desired? I was not sure if this is correct behavior for production environment, so I wanted to discuss it in PR. What do you think about it? Maybe we could let user set some property -> "org.kie.compiler.settings" that we could check here and if it is not set, we can include some example-production-settings.xml as a default one.

domhanak:
jenkins retest this

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/main/java/org/kie/workbench/common/services/backend/compiler/impl/utils/MavenUtils.java
desmax74: In the production environment we have the env vars of the if branch, the else is for local development

domhanak: So if there is no property named JENKINS_SETTINGS_XML_FILE in env vars, or this property is not set. What happens?

desmax74: The setting.xml was the default because the first time the test ran on jenkins we found the different repos used and the settings.xml was added, which other default do you want to use on jenkins ?

domhanak: This is `MavenUtils` I expected it to be used differently then the `MavenTestUtils`. Imho `MavenUtils` is for production environment. For testing purpose I am completely fine with returning "src/tests/settings.xml" file, but imagine when user deploys workbench and starts to use it, builds a project, he does not have JENKINS_SETTINGS_XML_FILE set, so the setting file from "src/test/" is going to be used and I am not sure this is what we want. 
I believe there is this property "kie.maven.settings.custom" that can be used for this. Imho we should let user use his custom settings.xml file, in case he is using some special artifacts for his project. Wdyt?

desmax74: Why are you changing this to fix the test ? The settings.xml is needed to fix the test where isn't present, but in the code the user add the settings when is required in the args of CompilationRequest, in a user code isn't present a env variable like in our jenkins. About the env "kie.maven.settings.custom" isn't a good idea because our API provides the ability to change the maven repo used for each build and using an env var force the compiler to use to a single version of the settings, better to pass in the compilationRequest as a param.

domhanak:
@porcelli Alex, could you please review and check the conversation here. We would like to hear your input on this.

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/main/java/org/kie/workbench/common/services/backend/compiler/impl/utils/MavenUtils.java
desmax74: To fix the tests on the service module we have to introduce the settings.xml in the executors (local and remote) @porcelli @domhanak

porcelli: +1 to have settings as executor parameters, in regard the `else` branch, -1 for it. Reference test infra in production code isn't a good practice.

domhanak:
Closing this as changes introduced are incorporated into different PR.

PR #2056, opened 2018-08-23
DROOLS-2578 [DMN Designer] Add use of 'DMNModelInstrumentedBase.setParent(..)'...

tarilabs:
...to marshaller

[DMN Designer] Add use of 'DMNModelInstrumentedBase.setParent(..)' to
marshaller

/cc @etirelli @manstis @jomarko

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DefinitionsConverter.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DefinitionsConverter.java
@@ -43,12 +43,19 @@ public class DefinitionsConverter {
         result.getNsContext().putAll(dmn.getNsContext());
 
         for (org.kie.dmn.model.api.ItemDefinition itemDef : dmn.getItemDefinition()) {
-            ItemDefinition itemDefConvered = ItemDefinitionPropertyConverter.wbFromDMN(itemDef);
-            result.getItemDefinition().add(itemDefConvered);
+            ItemDefinition itemDefConverted = ItemDefinitionPropertyConverter.wbFromDMN(itemDef);
+            if (itemDefConverted != null) {
+                itemDefConverted.setParent(result);
+            }
+            result.getItemDefinition().add(itemDefConverted);
         }
 
         for (org.kie.dmn.model.api.Import i : dmn.getImport()) {
-            result.getImport().add(ImportConverter.nodeFromDMN(i));
+            Import importConverted = ImportConverter.nodeFromDMN(i);
+            if (importConverted != null) {
+                importConverted.setParent(result);
+            }
+            result.getImport().add(importConverted);
         }
 
         return result;
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/Association.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/Association.java
@@ -18,7 +18,6 @@ package org.kie.workbench.common.dmn.api.definition.v1_1;
 import java.util.HashSet;
 import java.util.Set;
 
-import org.jboss.errai.common.client.api.annotations.MapsTo;
 import org.jboss.errai.common.client.api.annotations.Portable;
 import org.jboss.errai.databinding.client.api.Bindable;
 import org.kie.workbench.common.dmn.api.property.dmn.Id;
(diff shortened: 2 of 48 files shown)


tarilabs:
```
Test Result (1 failure / +1)
org.kie.workbench.common.services.backend.compiler.DefaultMavenCompilerTest.buildWithCloneTest
```

Jenkins test again

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DefinitionsConverter.java
```diff
@@ -43,12 +43,19 @@ public static Definitions wbFromDMN(final org.kie.dmn.model.api.Definitions dmn)
         result.getNsContext().putAll(dmn.getNsContext());
 
         for (org.kie.dmn.model.api.ItemDefinition itemDef : dmn.getItemDefinition()) {
-            ItemDefinition itemDefConvered = ItemDefinitionPropertyConverter.wbFromDMN(itemDef);
-            result.getItemDefinition().add(itemDefConvered);
+            ItemDefinition itemDefConverted = ItemDefinitionPropertyConverter.wbFromDMN(itemDef);
```
manstis: Should we cover ```drgElement```, ```artifact```, ```elementCollection``` and ```businessContextElement``` too?

tarilabs: Concerning `drgElement`, `artifact`, it was previously requested to manage them separately, in the Node of the graph, which is done at a DMNMarshaller level.
Concerning `elementCollection` this is basically to depict DRDs but it is not clear to me how should be managed from the stunner pov.
Concerning `businessContextElement` this is another non-execution element, but if it's covered by the editor is just matter to understand how to most appropriately convert from the stunner PoV (so as a graph node, or as a child element here?)

Nevertheless I would suggest all the above is a different scope than the current, which is covering setParent for existing converter as per DROOLS-2578, and I would suggest to manage as a separate JIRA/+PR once the Stunner requirements is identified?

manstis: Thanks for the explanations @tarilabs 

I agree, we can worry about the other items (```elementCollection``` and ```businessContextElement```) when the UI handles them and we understand what our (UI) requirements are better.

tarilabs:
Hi, just a follow-up --can this be merged given it's approved, please?

manstis:
@tarilabs Waiting on @jomarko to approve before merge.

A lot of the "workbench" related repositories need PRs to be both _peer_ and _QE_ reviewed before merge.

tarilabs:
Any estimation on this, please? Just trying to avoid having the need of rebase due to potential conflicts. Thanks!

tarilabs:
@jomarko 

> unguearded occurence in DMNMarshaller.connectRootWithChild is it fine?

I would say Yes because connectRootWithChild is never called with either of the params as null since original definition of the method

tarilabs:
Can I have the error message please?

jomarko:
@tarilabs sure, sorry I forgot to attach. However not saying much.
- Ui error
```
Uncaught exception: (TypeError) : Cannot read property 'Lyb' of null
```
- Server log error, however present also on master
```
08:39:06,931 ERROR [org.kie.dmn.core.compiler.DMNEvaluatorCompiler] (EJB default - 2) No expression defined for name 'Decision-1' on node 'Decision-1'
```

tarilabs:
> `08:39:06,931 ERROR [org.kie.dmn.core.compiler.DMNEvaluatorCompiler] (EJB default - 2) No expression defined for name 'Decision-1' on node 'Decision-1'`

I cannot see how this is related with this change.

Moreover the reported error looks like the workbench is trying to compile the DMN model, and if a Decision is missing an Expression (which allegedly given the reported step was indeed not defined) it is normaly that DMN model compilation will fail.
Is the Workbench now trying to compile the DMN model when is being opened?
How this is impacted by this serialization work, please?
@manstis can you kindly advise?

manstis:
@jomarko Hi, was the error in ```kie-(drools)-wb``` or something else please?

@tarilabs The ```kie-(drools)-wb``` **will** add the DMN file to ```KieBuilder``` and perform an incremental compilation of the project to get the list of errors (in the project) to show to Users (like your IDE shows a "problems" panel so does the Workbench).

I'll take a look at this PR this morning.

jomarko:
@manstis as usually, I used `kie-wb-distributions/kie-drools-wb-parent` module.

manstis:
@tarilabs @jomarko The error is caused by marshalling of the _enriched_ (parent property) model from the server to the client. I'll need to dig a bit more as to what is causing it and why; but it's something relating to a ```null``` value somewhere....

PR #2070, opened 2018-08-29
DROOLS-2590: [DMN Designer] Remove fallback from QNameFieldConverter.toWidgetValue(..)

manstis:
See https://issues.jboss.org/browse/DROOLS-2590


jomarko:
@manstis please see my acceptance test in jira, it seems to me this PR introduces troubles with using external dmn asset.

PR #2072, opened 2018-08-29
[AF-1467] Upgrade to WildFly 13

ederign:
https://github.com/jboss-integration/jboss-integration-platform-bom/pull/421


jomarko:
@ederign Hi I think we should create PR also for kie-docs and mention it there in release notes. What do you think?

ederign:
@jomarko sure, but this will be only merged after 7.1.

ederign:
This will be replaced by a wf14 upgrade done by @adrielparedes

PR #2082, opened 2018-09-03
JBPM-7696 - Add Process Administration to settings page

cristianonicolai:


bsremac:
@cristianonicolai When trying to access _Admin >> Process Administration_, I'm getting an error:
`Uncaught exception: java.lang.NullPointerException Caused by: null`
Screenshot:
screenshot from 2018-09-03 14-14-10

cristianonicolai:
@bsremac yep, got the same issue here, somehow it works fine using the showcase but not the final dist. Still checking.

cristianonicolai:
@bsremac should work now, please give it another go.

bsremac:
@cristianonicolai There's one more issue when it comes to entering an arbitrary numeric value.
If after entering a non-allowed value (e.g. 55) and clicking _Save_ button immediately after it, although the help message shows up, saving navigates from _Process Administration_ page back to _Settings_ view and the confirmation message pops up as if the change was accepted, although it was rejected.

I suppose the expected behavior would be to stay on the page until a valid number is entered.

cristianonicolai:
Assembly:
- https://github.com/kiegroup/appformer/pull/480
- https://github.com/kiegroup/kie-wb-common/pull/2082
- https://github.com/kiegroup/jbpm-wb/pull/1192

bsremac:
@cristianonicolai The issue described above is still present. The user can enter whatever value they want and click _Save_ button, which drives them away from the page with the notification message saying that the change was successfully saved (although the error message pops up below the input field in case of an invalid input for just a moment before the page is closed and the value is left unchanged).
I would suggest moving the input validation on save and prevent leaving the page until a valid value is entered. That way that information popup about the allowed values can be useful during the same visit of the page and not only after realizing the change you entered hasn't been propagated to pagination options :)

cristianonicolai:
@bsremac yep, sorry I haven't shared the details on that. If you take a look in other settings sections you can see that this error is consistent across all. This needs to be fixed in the generic preferences editor. I would prefer to isolate that into its own JIRA, wdyt?

bsremac:
@cristianonicolai As long as this issue is recognized as such I have nothing against fixing this as a separate effort. 
Would it be ok for you to open a new JIRA on this since you have more knowledge about the background? I wouldn't be able to provide any meaningful description which would give more details and guidance to somebody in the foundation team (as this would be their area, right?)

cristianonicolai:
@bsremac feel free to open it and yes, perhaps @paulovmr could help with that.

bsremac:
@cristianonicolai @paulovmr JIRA opened to track this issue https://issues.jboss.org/browse/AF-1605

cristianonicolai:
@bsremac thanks.

PR #2081, opened 2018-08-31
DROOLS-2946: [DMN Designer] Data-types: Data-type selector broken

manstis:
See https://issues.jboss.org/browse/DROOLS-2946

PR #2083, opened 2018-09-03
Tests cleanup following intensive custom data type dialog development

jomarko:
Dependent on #2081 

@manstis @karreiro please have a look.


jomarko:
@manstis this is ready for merge.

PR #2085, opened 2018-09-04
DROOLS-2919: [DMN Designer] Data-types: Clean-up

manstis:
See https://issues.jboss.org/browse/DROOLS-2919

I've left as three commits to make review of the different clean-ups easier.

I will squash when merged.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
@@ -173,60 +173,6 @@ public abstract class BaseExpressionGrid<E extends Expression, D extends GridDat
 
     protected abstract boolean isHeaderHidden();
 
-    protected Function<GridCellTuple, Command> newCellHasNoValueCommand() {
-        return (gc) -> new DeleteCellValueCommand(gc,
-                                                  () -> uiModelMapper,
-                                                  gridLayer::batch);
-    }
-
-    protected Function<GridCellValueTuple, Command> newCellHasValueCommand() {
-        return (gcv) -> new SetCellValueCommand(gcv,
-                                                () -> uiModelMapper,
-                                                gridLayer::batch);
-    }
-
-    protected Function<GridCellTuple, Command> newHeaderHasNoValueCommand() {
-        return (gc) -> new DeleteHeaderValueCommand(extractEditableHeaderMetaData(gc),
-                                                    gridLayer::batch);
-    }
-
-    protected Function<GridCellValueTuple, Command> newHeaderHasValueCommand() {
-        return (gcv) -> {
-            final String title = gcv.getValue().getValue().toString();
-            return new SetHeaderValueCommand(title,
-                                             extractEditableHeaderMetaData(gcv),
-                                             gridLayer::batch);
-        };
-    }
-
-    protected Function<GridCellTuple, Command> newHeaderHasNameHasNoValueCommand() {
-        return (gcv) -> {
-            final CompositeCommand.Builder<AbstractCanvasHandler, CanvasViolation> commandBuilder = new CompositeCommand.Builder<>();
-            commandBuilder.addCommand(new DeleteHeaderValueCommand(extractEditableHeaderMetaData(gcv),
-                                                                   () -> {
-                                                                       gridLayer.batch();
-                                                                       getNodeUUID().ifPresent(uuid -> refreshFormPropertiesEvent.fire(new RefreshFormProperties(sessionManager.getCurrentSession(), uuid)));
-                                                                   }));
-            getUpdateStunnerTitleCommand("").ifPresent(commandBuilder::addCommand);
-            return commandBuilder.build();
-        };
-    }
-
-    protected Function<GridCellValueTuple, Command> newHeaderHasNameHasValueCommand() {
-        return (gcv) -> {
-            final String title = gcv.getValue().getValue().toString();
-            final CompositeCommand.Builder<AbstractCanvasHandler, CanvasViolation> commandBuilder = new CompositeCommand.Builder<>();
-            commandBuilder.addCommand(new SetHeaderValueCommand(title,
-                                                                extractEditableHeaderMetaData(gcv),
-                                                                () -> {
-                                                                    gridLayer.batch();
-                                                                    getNodeUUID().ifPresent(uuid -> refreshFormPropertiesEvent.fire(new RefreshFormProperties(sessionManager.getCurrentSession(), uuid)));
-                                                                }));
-            getUpdateStunnerTitleCommand(title).ifPresent(commandBuilder::addCommand);
-            return commandBuilder.build();
-        };
-    }
-
     @SuppressWarnings("unchecked")
     public Consumer<HasName> clearDisplayNameConsumer(final boolean updateStunnerTitle) {
         return (hn) -> {
@@ -322,14 +268,16 @@ public abstract class BaseExpressionGrid<E extends Expression, D extends GridDat
                                                      newCellHasValueCommand());
     }
 
-    public TextBoxSingletonDOMElementFactory getHeaderHasNameTextBoxFactory() {
-        return new TextBoxSingletonDOMElementFactory(gridPanel,
-                                                     gridLayer,
-                                                     this,
-                                                     sessionManager,
-                                                     sessionCommandManager,
-                                                     newHeaderHasNameHasNoValueCommand(),
-                                                     newHeaderHasNameHasValueCommand());
+    protected Function<GridCellTuple, Command> newCellHasNoValueCommand() {
+        return (gc) -> new DeleteCellValueCommand(gc,
+                                                  () -> uiModelMapper,
+                                                  gridLayer::batch);
+    }
+
+    protected Function<GridCellValueTuple, Command> newCellHasValueCommand() {
+        return (gcv) -> new SetCellValueCommand(gcv,
+                                                () -> uiModelMapper,
+                                                gridLayer::batch);
     }
 
     public TextAreaSingletonDOMElementFactory getHeaderTextAreaFactory() {
@@ -340,6 +288,20 @@ public abstract class BaseExpressionGrid<E extends Expression, D extends GridDat
                                                       sessionCommandManager,
                                                       newHeaderHasNoValueCommand(),
                                                       newHeaderHasValueCommand());
+    }
+
+    protected Function<GridCellTuple, Command> newHeaderHasNoValueCommand() {
+        return (gc) -> new DeleteHeaderValueCommand(extractEditableHeaderMetaData(gc),
+                                                    gridLayer::batch);
+    }
+
+    protected Function<GridCellValueTuple, Command> newHeaderHasValueCommand() {
+        return (gcv) -> {
+            final String title = gcv.getValue().getValue().toString();
+            return new SetHeaderValueCommand(title,
+                                             extractEditableHeaderMetaData(gcv),
+                                             gridLayer::batch);
+        };
     }
 
     protected EditableHeaderMetaData extractEditableHeaderMetaData(final GridCellTuple gc) {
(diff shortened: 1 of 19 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/NameAndDataTypeHeaderMetaData.java
```diff
@@ -21,23 +21,51 @@
...
+
+        HasTypeRef hasTypeRef = expression.get();
+        final DMNModelInstrumentedBase base = hasExpression.asDMNModelInstrumentedBase();
+        if (base instanceof HasVariable) {
+            final HasVariable hasVariable = (HasVariable) base;
+            hasTypeRef = hasVariable.getVariable();
+        }
+
+        this.hasTypeRef = hasTypeRef;
+    }
```
jomarko: I think the new constructor should use the existing one and we should test this new constructor.
```java
public NameAndDataTypeHeaderMetaData(final HasExpression hasExpression,
                                         final Optional<E> expression,
                                         final Optional<HasName> hasName,
                                         final Consumer<HasName> clearDisplayNameConsumer,
                                         final BiConsumer<HasName, Name> setDisplayNameConsumer,
                                         final BiConsumer<HasTypeRef, QName> setTypeRefConsumer,
                                         final CellEditorControlsView.Presenter cellEditorControls,
                                         final NameAndDataTypeEditorView.Presenter headerEditor) {
        this(hasName,
             getTypeRefOfExpression(expression, hasExpression),
             clearDisplayNameConsumer,
             setDisplayNameConsumer,
             setTypeRefConsumer,
             cellEditorControls,
             headerEditor);
    }

    private static <E extends Expression> HasTypeRef getTypeRefOfExpression(final Optional<E> expression,
                                                                            final HasExpression hasExpression) {
        HasTypeRef hasTypeRef = expression.get();
        final DMNModelInstrumentedBase base = hasExpression.asDMNModelInstrumentedBase();
        if (base instanceof HasVariable) {
            final HasVariable hasVariable = (HasVariable) base;
            hasTypeRef = hasVariable.getVariable();
        }

        return hasTypeRef;
    }
```

jomarko: Also thinking if we need to pass both `expression` and `hasExpression` as the `hasExpression` has method `getExpression`. Are the `expression` and `hasExpression.getExpression` different things?

jomarko:
@manstis manual check passed, please have a look on my comments to code and I can approve.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
```diff
@@ -310,14 +268,16 @@ public TextBoxSingletonDOMElementFactory getBodyTextBoxFactory() {
...
-    public TextBoxSingletonDOMElementFactory getHeaderHasNameTextBoxFactory() {
-        return new TextBoxSingletonDOMElementFactory(gridPanel,
-                                                     gridLayer,
-                                                     this,
-                                                     sessionManager,
-                                                     sessionCommandManager,
-                                                     newHeaderHasNameHasNoValueCommand(),
-                                                     newHeaderHasNameHasValueCommand());
+    protected Function<GridCellTuple, Command> newCellHasNoValueCommand() {
+        return (gc) -> new DeleteCellValueCommand(gc,
```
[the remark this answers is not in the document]
manstis: Done.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
```diff
@@ -310,14 +268,16 @@ public TextBoxSingletonDOMElementFactory getBodyTextBoxFactory() {
...
+                                                  () -> uiModelMapper,
+                                                  gridLayer::batch);
+    }
+
+    protected Function<GridCellValueTuple, Command> newCellHasValueCommand() {
+        return (gcv) -> new SetCellValueCommand(gcv,
```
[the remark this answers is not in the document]
manstis: Done.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/NameAndDataTypeHeaderMetaData.java
manstis: @jomarko This is a very good question! 

I had a quick look around and the only place I see we need to separate the two is in ```UndefinedExpressionGrid.onExpressionTypeChanged(..)``` where we instantiate a ```GridCellValueTuple``` containing a new instance of an editor for the selected Expression type. This is then passed to a command to set the ```Expression``` on the ```HasExpression``` instance. However we cannot instantiate the editor without having (at this point) the dis-associated ```HasExpression``` and ```Expression``` (as it has not actually been associated until the command executes).

Sadly that means we cannot use ```HasExpression.getExpression()``` in the various Expression grid editors' ```BaseExpressionGrid.initialiseUiColumns()``` methods (that create the ```NameAndDataTypeHeaderMetaData``` objects) as the association (between ```HasExpression``` and ```Expression```) has not been established.

I could possibly improve ```NameAndDataTypeHeaderMetaData``` (and no doubt other places) to use a ```Supplier<? extends Expression>``` _internally_ (i.e. not a constructor parameter) that calls ```HasExpression.getExpression()``` on demand (which, as far as my quick review suggests,    will always be after the association has been made). However I'd rather create a new JIRA (under the "Code improvements" banner) to address outside of this PR. The analysis and impact of this proposal is quite far reaching. WDYT?

manstis:
@jomarko New commit containing changes added.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/NameAndDataTypeHeaderMetaData.java
manstis: Constructor changes and new tests added.

jomarko: Thanks for deep analysis. As you and the length of the explanation suggest, lets address in separate jira.

jomarko: https://issues.jboss.org/browse/DROOLS-2977

manstis:
Jenkins please retest this.

@karreiro Please review when you're bored one afternoon :-)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
```diff
@@ -173,70 +173,28 @@ protected void doInitialisation() {
...
-
     @SuppressWarnings("unchecked")
-    public Consumer<HasName> clearDisplayNameConsumer() {
-        return (hn) -> sessionCommandManager.execute((AbstractCanvasHandler) sessionManager.getCurrentSession().getCanvasHandler(),
-                                                     newHasNameHasNoValueCommand(hn).build());
+    public Consumer<HasName> clearDisplayNameConsumer(final boolean updateStunnerTitle) {
+        return (hn) -> {
+            final CompositeCommand.Builder commandBuilder = newHasNameHasNoValueCommand(hn);
+            if (updateStunnerTitle) {
+                getUpdateStunnerTitleCommand("").ifPresent(commandBuilder::addCommand);
```
karreiro: I think that this line is not covered by tests 👆

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
```diff
@@ -173,70 +173,28 @@ protected void doInitialisation() {
...
 
     @SuppressWarnings("unchecked")
-    public BiConsumer<HasName, Name> setDisplayNameConsumer() {
-        return (hn, name) -> sessionCommandManager.execute((AbstractCanvasHandler) sessionManager.getCurrentSession().getCanvasHandler(),
-                                                           newHasNameHasValueCommand(hn, name).build());
+    public BiConsumer<HasName, Name> setDisplayNameConsumer(final boolean updateStunnerTitle) {
+        return (hn, name) -> {
+            final CompositeCommand.Builder commandBuilder = newHasNameHasValueCommand(hn, name);
+            if (updateStunnerTitle) {
+                getUpdateStunnerTitleCommand(name.getValue()).ifPresent(commandBuilder::addCommand);
```
karreiro: This one is not covered by tests too... in fact `clearDisplayNameConsumer` and `setDisplayNameConsumer` are pretty similar, maybe we could extract both to a single method 🤔

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/NameAndDataTypeHeaderMetaData.java
```diff
@@ -54,6 +74,18 @@ public NameAndDataTypeHeaderMetaData(final Optional<HasName> hasName,
         this.setTypeRefConsumer = setTypeRefConsumer;
     }
 
+    private static <E extends Expression> HasTypeRef getTypeRefOfExpression(final Optional<E> expression,
+                                                                            final HasExpression hasExpression) {
+        HasTypeRef hasTypeRef = expression.get();
```
karreiro: Here we have an _"unsafe"_ `.get()`, instead of calling `.get()` what do you think about calling `orElseThrow(..) ` explaining why `expression` is always present here? :-)

Something like this (with a better error message):
```java
HasTypeRef hasTypeRef = expression.orElseThrow(() -> new UnsupportedOperationException("'expression' is never null, since this is called after (..) (?)"));
```

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
manstis: It is now!

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
manstis: It is now. I looked at refactoring, as you suggested, however they execute different commands and have different return types so I think to leave "as is". If you feel really strongly about it I am always open to proposals.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/NameAndDataTypeHeaderMetaData.java
manstis: Done as you propose. It *should* never happen as a grid that has a ```NameAndDataTypeHeaderMetaData``` should *always* have an ```Expression``` (the only exception is ```UndefinedExpressionGrid``` but that does not have any headers). I've not written a test for this reason.... but again.. if you insist...

manstis:
Jenkins please retest this.

PR #2093, opened 2018-09-07
JBPM-7689: Focus canvas after node selection in DMN navigator dock

jomarko:
@manstis @hasys please have a look, one more case to check.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/decision/factories/DecisionNavigatorBaseItemFactoryTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/decision/factories/DecisionNavigatorBaseItemFactoryTest.java
@@ -27,6 +27,7 @@ import org.junit.runner.RunWith;
 import org.kie.workbench.common.dmn.client.decision.DecisionNavigatorItem;
 import org.kie.workbench.common.dmn.client.decision.DecisionNavigatorPresenter;
 import org.kie.workbench.common.stunner.core.api.DefinitionManager;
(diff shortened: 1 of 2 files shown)


hasys:
Nice catch! Looks good to me, thank you @jomarko

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/decision/factories/DecisionNavigatorBaseItemFactoryTest.java
```diff
@@ -55,7 +56,8 @@
 import static org.mockito.Mockito.when;
 
 @RunWith(GwtMockitoTestRunner.class)
-public class DecisionNavigatorBaseItemFactoryTest {
+public class
+DecisionNavigatorBaseItemFactoryTest {
```
manstis: We don't usually split this on to two lines.. but it's cosmetic (I'll be surprised if the checkstyle plugin passes!)

hasys: @manstis It should pass, checkstyle doesn't check stuff like this. It checks not used imports, tab symbols, if/for/while without curly braces and things like this. All stylish moments delegated to IDE formatter.

jomarko: I am surprised how does this get in? :D will update definitely. Thanks for spotting.

jomarko:
@manstis please merge when you have a moment.

PR #2097, opened 2018-09-07
DROOLS-2941: [DMN Designer] - Save external DMN model

manstis:
See https://issues.jboss.org/browse/DROOLS-2941

@tarilabs Can you approve the use of ```DescriptionPropertyConverter.dmnFromWB(..)``` for **ALL** ```Description``` instances (I found a few that simply copied the value from DMN to WB).

@jomarko @tarilabs This PR removes all default _empty_ instances of properties on ```DMNElement``` objects that were being created by the WB model. There were a few related changes in the client code but nothing too serious.

I tested with the @tarilabs  minimal reproducer DMN file that highlights the issue. Tests now succeed and doing a file comparison between the original and "round-tripped" files are now very much more comparable.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/BusinessKnowledgeModel.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/BusinessKnowledgeModel.java
@@ -85,7 +85,7 @@ public class BusinessKnowledgeModel extends DRGElement implements HasVariable,
              new org.kie.workbench.common.dmn.api.property.dmn.Description(),
              new Name(),
              new InformationItem(),
-             new FunctionDefinition(),
+             null,
              new BackgroundSet(),
              new FontSet(),
              new RectangleDimensionsSet());
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/DMNDiagram.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/DMNDiagram.java
@@ -65,15 +65,15 @@ public class DMNDiagram extends DMNModelInstrumentedBase implements DMNDefinitio
         add("dmn_diagram");
     }};
 
-    @PropertySet
-    @FormField
-    @Valid
-    protected Definitions definitions;
-
     @Property
     @FormField(readonly = true)
     @Valid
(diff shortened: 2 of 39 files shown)


jomarko:
@manstis lace model came from https://redhat.trisotech.com/ -> Trisotexh Examples -> DMN modeler

Postponing my review until kie-wb-distributions works again.

manstis:
@jomarko @tarilabs Unit test changed to use @tarilabs minimal reproducer.

jomarko:
@manstis could you also please have a look on this https://github.com/manstis/kie-wb-common/pull/34 and check why the added test fails?

manstis:
@jomarko I imported ```DROOLS-2941.dmn``` into Trisotech and executed the model.

It only returns "a" when executed with an empty _context_. Trisotech's editor reports there are no _inputs_ and hence you cannot set a value in the _context_. I won't pretend to understand how this model evaluates perhaps @tarilabs would be kind enough to educate us :-) 

If I remove [this](https://github.com/manstis/kie-wb-common/pull/34/files#diff-e60839568850ea8b772f18a29a2dc30aR1175) from the test is passes.

manstis:
@jomarko The "Description" issue is unrelated to this JIRA; but https://issues.jboss.org/browse/JBPM-7166

When you update the Decision's description, click on the canvas (i.e. de-select the Decision) and click on the Decision once more the description is indeed empty however the Description for the child InformationItem has the description you set earlier.

manstis:
@tarilabs @jomarko OK for re-review.

@tarilabs You have an open (resolved) question about the LACE model.

@jomarko I've answered your questions.

tarilabs:
@manstis 
> If I remove this from the test is passes.
If that line was wrote by me originally it was clearly a mistake, I might have inadvertently commited a line making some tests. The DMN hardcodes a Decision as "a" value and indeed no inputs are required.

manstis:
@tarilabs Could you please approve? (Now we use your minimal reproducer DMN file).

tarilabs:
> however for me it doesn't make sense, why if I have some DMN model, that hardcode the result to "a", why I can not set some variable to DMN context. Could you please clarify?

Really sorry I missed this @jomarko . The point here is that you are not setting "some variable", you are setting _the_ variable for which there is also the Decision in the model. Hence by the spec, is like you are supplying a pre-computed partial result. As the engine evaluates all the Decision, it actually find the one for the case was already present (as supplied in input) hence it re-uses that result. Then the input was inconsistent with the test (test expect it to be "a" but you forced it to be "b"), hence the failure was the expected behavior.

In other words, we have a model which for Decision `A Vowel` always answer `a`. You are supplying a precomputed result for the same, valued as `b`. The test checks the `A Vowel` results in `a` but because of the previous step is actually `b`. Inconsistency and test fail.

PR #2115, opened 2018-09-12
[AF-1392] Unable to import filesystem bare git repository in Windows

porcelli:
Path (String) does not work well in windows. Wrapping the string in an URI.


tomasdavidorg:
I tested this on windows machine but it does not work for me. I get similar exception: `Unable to complete your request. The following exception occurred: Illegal character in opaque part at index 2: C:\white-list-source`
screenshot from 2018-09-14 12-40-19

elguardian:
theres is a small diff between URI.create and new URI.... that messed the PR.

elguardian:
it should be c:/white-list-source or file:///c:/white-list-source (that would do)

PR #2129, opened 2018-09-17
DROOLS-2956: IE11 Prevent error due to unsupported element.remove

jomarko:
This PR is corollary of issues:
- https://issues.jboss.org/browse/DROOLS-2816
- https://issues.jboss.org/browse/DROOLS-2818
- https://issues.jboss.org/browse/DROOLS-2820

@manstis @danielzhe @karreiro would you mind to have a look?

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListView.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListView.java
@@ -110,8 +110,10 @@ public class DataTypeListView implements DataTypeList.View {
 
         for (int i = 0; i < subDataTypeRows.length; i++) {
             final Element item = subDataTypeRows.getAt(i);
-            cleanSubTypes(item.getAttribute(UUID_ATTR));
-            item.remove();
+            if (item != null && item.parentNode != null) {
+                cleanSubTypes(item.getAttribute(UUID_ATTR));
+                item.parentNode.removeChild(item);
+            }
         }
     }
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListViewTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListViewTest.java
@@ -41,6 +41,7 @@ import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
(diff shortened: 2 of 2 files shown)


jomarko:
jenkins please retest this

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListView.java
```diff
@@ -110,8 +110,10 @@ void cleanSubTypes(final String uuid) {
 
         for (int i = 0; i < subDataTypeRows.length; i++) {
             final Element item = subDataTypeRows.getAt(i);
-            cleanSubTypes(item.getAttribute(UUID_ATTR));
-            item.remove();
+            if (item != null && item.parentNode != null) {
```
karreiro: @jomarko Could you share the steps to reproduce the `NullPointerException` that could be raised here?

I think that `item` and `item.parentNode` are never null here  in the _real life_. We can mock the scenario in a unit test, but considering the DOM tree and the `subDataTypeRows`, I think that these check would never be really used.

Am I missing something? :-)

danielzhe: @karreiro  I guess you're right. I did the same check in other PR, but now that you mentioned maybe it's not really necessary.

danielzhe: @jomarko FYI: looks like this PR fixes [DROOLS-2956](https://issues.jboss.org/browse/DROOLS-2956).

jomarko: @karreiro you don't miss anything. It is just my defensive approach. I was looking for javadoc [here](https://static.javadoc.io/com.google.elemental2/elemental2-dom/1.0.0-RC1/elemental2/dom/NodeList.html) and [here](https://static.javadoc.io/com.google.elemental2/elemental2-dom/1.0.0-RC1/elemental2/dom/Element.html) but I didn't find if `item.parentNode` and `getAt()` return `non null` values.

jomarko: @danielzhe thanks, will recheck that issue and if this PR fix it, I will update commit message and PR title.

jomarko: @karreiro thanks for discussion. I totally agree that defensive approach is not always the most appropriate. Sometimes we really want the error to be shown to user because `NPE` means fatal error somewhere.

However I believe cleaning widget is not case when user should be informed that there was some problems with `HTML` manipulation.

PR #2130, opened 2018-09-17
RHPAM-1541: Stunner - it is possible to corrupt diagram and lose all made changes

wmedvede:


wmedvede:
Hello @tiagobento , @LuboTerifaj 
This is the quick-win fix we talked about in Stunner meeting. Would you mind review?
Thanks!

p.s. as it was discussed this fix is not the "optimal" one, but's still a quick-win for 7.1 if we can review and merge on time. The fix is basically preventing a node to point to itself. Since the rules we have still don't model this scenarios, a helper class was introduced for doing the "rule check". A think the change is more or less isolated and doesn't break other existing stuff, so if possible we should try to add it to the release. But let's see if there are time.

wmedvede:
@tiagobento, @LuboTerifaj forgot to mention, If you can prioritize this guys it'l be great. 
Thank you!

wmedvede:
Hello @jomarko , any chance you can take a quick look at this since  @LuboTerifaj won't be able to review today?
Thanks!

jomarko:
@wmedvede ok, building locally.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelper.java
```diff
@@ -0,0 +1,43 @@
...
+public class ConnectionCommandHelper {
+
+    public static CommandResult<RuleViolation> checkSelfReferenceViolations(final Node<? extends View<?>, Edge> sourceNode,
+                                                                            final Node<? extends View<?>, Edge> targetNode) {
+        final String sourceId = sourceNode != null ? sourceNode.getUUID() : null;
+        final String targetId = targetNode != null ? targetNode.getUUID() : null;
+        final GraphCommandResultBuilder resultBuilder = new GraphCommandResultBuilder();
+        if ((sourceId != null || targetId != null) && Objects.equals(sourceId,
+                                                                     targetId)) {
+            resultBuilder.addViolation(new RuleViolationImpl("A node can not be connected to itself "));
```
jomarko: Isn't here one additional ending white space? Shouldn't we use `"A node can not be connected to itself"` instead of `"A node can not be connected to itself "`?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelperTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public void testWhenTargetIsNull() {
+        assertFalse(CommandUtils.isError(ConnectionCommandHelper.checkSelfReferenceViolations(sourceNode,
+                                                                                              null)));
+    }
+
+    @Test
+    public void testSourceIsNull() {
+        assertFalse(CommandUtils.isError(ConnectionCommandHelper.checkSelfReferenceViolations(null,
+                                                                                              targetNode)));
+    }
```
jomarko: Do we want to check as source so target are null case?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelperTest.java
```diff
@@ -0,0 +1,76 @@
...
+        assertFalse(CommandUtils.isError(ConnectionCommandHelper.checkSelfReferenceViolations(sourceNode,
+                                                                                              targetNode)));
+    }
+
+    @Test
+    public void testWhenNodesAreEqual() {
+        when(targetNode.getUUID()).thenReturn(SOURCE_UUID);
+        assertTrue(CommandUtils.isError(ConnectionCommandHelper.checkSelfReferenceViolations(sourceNode,
+                                                                                             targetNode)));
+    }
```
jomarko: Do we want to check also error message in this case?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelper.java
wmedvede: It's added by intention for making sure the resulting message is properly spaced form the automatically appended strings.
Since this is a temporal fix I'd vote for letting it live this way.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelperTest.java
wmedvede: Not sure if I understand. This case tests the use case when the target source is null. And we have another use case that tests when the target is null. ¿?

jomarko: Ok, I think we should add test case:
```java
public void testBothNodesNull() {
        assertFalse(CommandUtils.isError(ConnectionCommandHelper.checkSelfReferenceViolations(null,null)));
    }
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelper.java
jomarko: Ok, one more question, is it ok that error message won't be I18n in this way?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/test/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelperTest.java
wmedvede: It can be done, but I think we can live without this check. Unless there are any other mayor suggested changes to do, I'd prefer to go ahead without such minor tweaks. Mainly because I'm afraid that if we have to wait for the jenkns ci system to "re-build" this PR finish we won't be able to include this fix for the 7.1 :(.
Let's wait for @tiagobento and see if there are mayor changes.

jomarko: @wmedvede sure, you are right. All I am asking/commenting/discussing are minor things. I won't block PR if you decide not to implement those points as manual check looks well.

However I don't agree if the only reason is time pressure. Then is better to say we won't fix this, we have too much work and this needs to be done in next iteration properly.

wmedvede:
Sorry, I believe I've added the wrong Tiago to the thread :(.  @tiagodolphine , would you mind take a look at this PR?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelper.java
tiagodolphine: @wmedvede minor thing here, there is no need to check the  `sourceId/targetId != null` because the `Objects.equals` aldready considers it.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/SetConnectionSourceNodeCommand.java
```diff
@@ -116,8 +117,13 @@ public SetConnectionSourceNodeCommand(final Node<? extends View<?>, Edge> source
         final Node<View<?>, Edge> sourceNode = (Node<View<?>, Edge>) getSourceNode(context);
         final Edge<View<?>, Node> edge = (Edge<View<?>, Node>) getEdge(context);
         final Node<? extends View<?>, Edge> lastSourceNode = edge.getSourceNode();
-        // Only check for rules in case the connector's source node is a different one.
-        if ((null == lastSourceNode && null != sourceNode) ||
+        //7.1 fix, by now we just ban a node from referring to itself by checking that source node and target nodes
```
tiagodolphine: I think it is worth to open a JIRA to solve the issue with the permanent solution, to keep track of this, and in the future remove this.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/command/impl/ConnectionCommandHelper.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.kie.workbench.common.stunner.core.command.CommandResult;
+import org.kie.workbench.common.stunner.core.graph.Edge;
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.command.GraphCommandResultBuilder;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+import org.kie.workbench.common.stunner.core.rule.RuleViolation;
+import org.kie.workbench.common.stunner.core.rule.violations.RuleViolationImpl;
+
+public class ConnectionCommandHelper {
```
tiagodolphine: A quick javadoc to explain the need for this class will be great to track this in the future. And IMO it is better to move the class to another package since it is not a command itself.

tiagodolphine:

wmedvede:
suggestions were added accordingly

wmedvede:
don't know why the full downstream has failing status now. I saw it green yesterday night.
anyway, all the failing tests are not related with this PR.

wmedvede:
jenkins retest this

PR #2152, opened 2018-09-27
Improve way we aquire free port for maven compiler tests

domhanak:
Add `setResuseAddress(true)` so there is no timeout period after it is closed.
Improve the exception handling.

@porcelli @desmax74 @jhrcek  Here is and alternative to https://github.com/kiegroup/appformer/pull/469 let me now your opinions.

### DIFF
--- a/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-testutil/src/main/java/org/kie/workbench/common/services/backend/compiler/TestUtilGit.java
+++ b/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-testutil/src/main/java/org/kie/workbench/common/services/backend/compiler/TestUtilGit.java
@@ -18,7 +18,6 @@ package org.kie.workbench.common.services.backend.compiler;
 import java.io.IOException;
 import java.net.ServerSocket;
 
-import org.junit.Assert;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -27,17 +26,29 @@ public class TestUtilGit {
     private static Logger logger = LoggerFactory.getLogger(TestUtilGit.class);
 
     public static int findFreePort() {
-        int port = 0;
+        ServerSocket socket = null;
         try {
-            ServerSocket server = new ServerSocket(0);
-            port = server.getLocalPort();
-            server.close();
+            socket = new ServerSocket(0);
+            socket.setReuseAddress(true);
+            int port = socket.getLocalPort();
+            try {
+                socket.close();
+            } catch (IOException e) {
+                // Ignore IOException on close()
+            }
+            logger.debug("Found free port {}", port);
+            return port;
         } catch (IOException e) {
-            Assert.fail("Can't find free port!");
+            // nop = ok
+        } finally {
+            if (socket != null) {
+                try {
+                    socket.close();
+                } catch (IOException e) {
+                    // Ignore IOException on close()
+                }
+            }
         }
-        if(logger.isDebugEnabled()) {
-            logger.debug("Found free port " + port);
-        }
-        return port;
+        throw new IllegalStateException("Could not find a free TCP/IP port to start git-daemon.");
     }
 }


kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-testutil/src/main/java/org/kie/workbench/common/services/backend/compiler/TestUtilGit.java
```diff
@@ -27,17 +26,29 @@
...
+                } catch (IOException e) {
+                    // Ignore IOException on close()
+                }
+            }
         }
-        if(logger.isDebugEnabled()) {
-            logger.debug("Found free port " + port);
-        }
-        return port;
+        throw new IllegalStateException("Could not find a free TCP/IP port to start git-daemon.");
```
domhanak: Let me know if we should call `Assert.fail()` here instead.

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-testutil/src/main/java/org/kie/workbench/common/services/backend/compiler/TestUtilGit.java
```diff
@@ -27,17 +26,29 @@
...
+            }
+            logger.debug("Found free port {}", port);
+            return port;
         } catch (IOException e) {
-            Assert.fail("Can't find free port!");
+            // nop = ok
+        } finally {
+            if (socket != null) {
+                try {
+                    socket.close();
```
jhrcek: `ServerSocket` implements `AutoCloseable` which means you can use try-with-resources to close it

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-testutil/src/main/java/org/kie/workbench/common/services/backend/compiler/TestUtilGit.java
jhrcek: I think IllegalStateException is ok here

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-testutil/src/main/java/org/kie/workbench/common/services/backend/compiler/TestUtilGit.java
```diff
@@ -18,26 +18,28 @@
...
     public static int findFreePort() {
-        int port = 0;
-        try {
-            ServerSocket server = new ServerSocket(0);
-            port = server.getLocalPort();
-            server.close();
+        try (ServerSocket socket = new ServerSocket(0)) {
+            socket.setReuseAddress(true);
+            int port = socket.getLocalPort();
+            try {
```
jhrcek: This try socket.close() can be removed - the socket will be closed in the (compiler generated) finally block - that's the whole point of try-with-resources :)

domhanak: this here is a sanity check that, the port is really free :)

domhanak:
@jhrcek Hmm nice a failure on the first try :D Let's rerun the build and see if it is still a random occurence.

Roses are red,
violets are blue,
ignorance is bliss,
Jenkins test this

jhrcek:
@domhanak Check the console log:
```
2018-09-27 05:35:48,145 [main] INFO o.k.w.c.s.b.c.BaseCompilerTest - Git port used:34654
--
2018-09-27 05:42:34,833 [33268f2d-8a46-42e2-8aea-b318122983cd] INFO o.a.m.c.event.ExecutionEventLogger - ------------------------------------------------------------------------
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 12.322 sec <<< FAILURE! - in org.kie.workbench.common.services.backend.compiler.DefaultMavenCompilerTest
buildWithAllDecoratorsTest(org.kie.workbench.common.services.backend.compiler.DefaultMavenCompilerTest)  Time elapsed: 0.041 sec  <<< ERROR!
org.uberfire.java.nio.IOException: java.io.IOException: Failed to open server socket for /127.0.0.1:34654
```
The port is allocated 5:35 and then 7 minutes later another test is trying the same port, which is already allocated. I'm not sure if this PR will improve this problem in any way.

domhanak:
Imho we should use ExternalResource from JUnit in some way, but I need to take a deeper look on how it works if we want to fix the tests only :)

jhrcek:
I don't think `ExternalResource` is going to help in this effort. What it gives you you is the ability to call some logic before / after test class on an object annotated with `@Rule`. If we knew what logic to put there, the problem could have been solved a long time ago. But I'm not going to try stop you from trying ;-)

porcelli:
as @jhrcek said, I won't stop you to try, but I don't think you can improve this.

the reason for that is, by design, the port assignment is executed only once and it'll live during the JVM life, and this assignment will happen when the JGitFileSystemProvider is referenced in the JVM, so no matter what you do... if any code reached it before, the port will be already assigned.

PR #2160, opened 2018-09-28
JBPM-7718 improve labels on sequence flows

tiagodolphine:
- Changing the sequence flow text style to the same as the SVG shapes (same values as css when generating the SVGs shapes)

- Centralize on a configuration class, the text style for Non-SVG shapes, now used only by Sequence flows on BPMN.

@romartin @LuboTerifaj  @wmedvede 
can you please help to review.

Related PR:


tiagodolphine:

tiagodolphine:
Hi @romartin 
Now if you take a look on the code the `StunnerTextPrefeerences` is **NOT** an exposed preference to the user, just use the term preference, so in this way this is only set internally (but if we think in the future it is better to expose than it will be just a matter of annotations configurations basically, or setting it using forms), and for now the values set for bpmn are the same as the CSS, but as constants. Reading the CSS and setting didn't work, i tried on many different ways, and throws exception when building the GWT so I just set the same values, probably we can work generating dynamic code in the same way as the SVG shapes generation, but this can bring some more complexity and will couple the non-svg with the svg, to get the style, WDYT ?  any other idea?

Anyway, the idea of centralizing the text styling properties (that is the default ones, used for all shapes, this could be set through forms in the future as I said) is to use them on the shapes injecting while creating the Shape Definitions, instead of hardcoding the same values on all the definitions (now there was only  on sequence flows, but we can apply to other definitions, even the SVG ones). 
But if you think it is better to keep hardcoded the values on the Definitions like it was on the `SequenceFlowConnectorDef`, I can revert the changes, just let me know, np.
Sure we can discuss more about this on hangouts !

The details of this issue are on this JIRA https://issues.jboss.org/browse/JBPM-7718 
but mainly it was about styling and wrapping the labels, the wrapping part is on lienzo.

Thanks

tiagodolphine:
Jenkins retest this

tiagodolphine:

tiagodolphine:
@hasys @romartin the build failed, I'm taking a look on this.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/preferences/StunnerTextPreferences.java
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.client.preferences;
+
+/**
+ * Centralize the default Text properties used by Stunner.
+ * Now this is not exposed to be configured by the user through the UI.
```
romartin: minor comment about the "now this is not exposed....", I guess the "now" word doesn't make sense any more :)

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/preferences/StunnerTextPreferences.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface StunnerTextPreferences {
```
romartin: IMO I don't see the need for an interface here, it's just a bean with some fields....

tiagodolphine:
Jenkins retest this

tiagodolphine:

tiagodolphine:

tiagodolphine:
@romartin @hasys this is green now, ok to review.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/preferences/StunnerTextPreferences.java
tiagodolphine: @romartin the idea of the interface is to allow different sets (bpmn, dmn, ...) to expose different specific text preferences, WDYT? anyway, I can remove it, np.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/preferences/StunnerTextPreferences.java
tiagodolphine: sure, I'll remove it.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/preferences/StunnerTextPreferences.java
romartin: I propose this as the "text preferences" is in fact just data object, it does not imply any logic, as for example the `BasePreference` does (which it does provides `load`, etc methods). So we don't expect different "implementations", we just expect different values for the fields.
Also the `StunnerPreferences` and `StunnerDiagramEditorPreferences` are classes with some state, this way by making it a class we will keep some coherence as well.
IMO this can be a class and the BPMN, DMN etc, can just register a new class instance but using different values for the fields, if any, isn't it?
Anyway no worries, just brainstorming, not a problem for me doing this way :)

hasys:
Hi @LuboTerifaj, @tiagodolphine

new lines doesn't work for all nodes, not sure is it should be resolved in this Jira or in separate one.

tiagodolphine:

tiagodolphine:
Jenkins retest this

tiagodolphine:
@romartin @LuboTerifaj updated according to what was discussed with Matthew. Plus I applied the Line break, this is an easy change we can apply to other shapes labels if necessary.
@hasys ^

LuboTerifaj:
Hi @tiagodolphine 

The first issue I have mentioned is on Marshallers side. Using old marshaller, the issue is not present.

romartin:
ok thanks @LuboTerifaj , good catches! I know @tiagodolphine that those issues are not probably related to this task, but please can you try to give them a quick look? At least try to identify and see if makes sense fixing them here or not. 
Thanks!!

tiagodolphine:

tiagodolphine:
@evacchi hey can you please take a look on this last commit https://github.com/kiegroup/kie-wb-common/pull/2160/commits/e7b9e7390c755fcadfbcf07f25c3184fdc1edd88 I need to change the sequence flow marshalling becase names with multiline were not working.
Thanks.

tiagodolphine:
@romartin @LuboTerifaj I've just pushed some changes to fix the multiline and the alignment when changing the labels text.
Thanks.

tiagodolphine:
@LuboTerifaj build is ok to test.
Thanks.

evacchi:
@tiagodolphine if that works (I don't recall SequenceFlow being a FlowElement, but I'm probably wrong), then the change looks fine !

romartin:
Please @LuboTerifaj  @hasys remember this is a blocker for 7.2, so starting to be urgent.. thanks in advance!!

tiagodolphine:
Hi @LuboTerifaj thanks for the review. For now I think it is better to let in this way, since we need to merge it for the 7.2, if necessary we can improve on a new JIRA.
@romartin I think we can proceed and merge.
Thanks !

PR #2168, opened 2018-10-05
Use <target> instead of deprecated <tasks> in antrun plugin config

jhrcek:
Fixes the warning
"[WARNING] Parameter tasks is deprecated, use target instead"

See maven-antrun-plugin [docs](http://maven.apache.org/plugins/maven-antrun-plugin/run-mojo.html#tasks)


jhrcek:
@manstis could you please check and merge? Cleaning up some build warnings..

PR #2171, opened 2018-10-05
JBPM-7604 Stunner - Showing case management properties for cases only

tiagodolphine:
Hide / Show Case Management Properties whether the diagram is under a case project and the `adhoc` property is set to **true**

To test it is necessary 2 conditions:
- the existence of **.caseproject** file on the root project filesystem on the .niogit, on the WB just create a case project should be enough because it creates this file.
- the property adhoc set to true or false on the forms

Since the adhoc condition can be changed at runtime, you can test setting to true/false and see the Case Management properties be shown/hide.

@romartin 
@LuboTerifaj

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/service/ProjectType.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/service/ProjectType.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.bpmn.service;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+@Portable
+public enum ProjectType {
+
+    BPMN(null),
+    CASE(".caseproject");
+
+    ProjectType(String fileName) {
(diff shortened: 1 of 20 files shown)


tiagodolphine:

romartin:
@pefernan adding you as a reviewer as it updates some code on the Stunner side but related to forms... just please double check it makes sense for you. Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/service/ProjectType.java
```diff
@@ -0,0 +1,42 @@
...
+public enum ProjectType {
+
+    BPMN(null),
+    CASE(".caseproject");
+
+    ProjectType(String fileName) {
+        this.fileName = fileName;
+    }
+
+    private String fileName;
```
manstis: Isn't it `fileExtension`?

tiagodolphine: Hi @manstis in fact the file name is `.caseproject`, this is a trick part, you can see here https://github.com/kiegroup/jbpm-wb/blob/master/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/casemgmt/service/CaseProjectServiceImpl.java when a new Case Project is created on the WB, it creates this file, on the root of the repo, and this is used to identify when creating Assets under a case project. Anyway, this is the way it works, and this PR is to make Stunner bpmn to support case variables under a case project.

tiagodolphine:
@romartin PR updated solving conflicts with master.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/service/ProjectType.java
manstis: Thanks for the explanation @tiagodolphine

tiagodolphine:

tiagodolphine:
jenkins retest this

romartin:
Hey @tiagodolphine , I added just a single review comment, any feedback about it? I think it makes sense fixing it... Also still missing @LuboTerifaj 's review. Let me know whatever. thx!

tiagodolphine:
Hey @romartin I saw your comment I didn't have time to take a deeper look, but I think if necessary I can improve that inserting a callback on the `loadDiagramType` combining with the `workItemDefinitionService.load`.  But in this way, it would be executed sequentially one after the other, the way it is today the calls to backend are being made in parallel and may be faster.

PR #2184, opened 2018-10-12
AF-1456 POM Cleanup mechanism

desmax74:
POM Cleanup mechanism
Bootstrap: https://github.com/kiegroup/droolsjbpm-build-bootstrap
Compiler task to receive the Set<String> with the unused deps founded in the project executing the dependency:analyze in memory

@ederign @jhrcek  @domhanak @porcelli


ederign:
@jhrcek this feature will be an 'experimental feature' for the community. 

The feature that will be included in the product is the 'dynamic adding of the dependencies'. 

If we have this unreliability on the plugin results, my suggestion would be to try to run the plugin, remove the dependencies, build the project and if the build is broken revert it with no changes. wdyt? (/cc @porcelli @desmax74)

jhrcek:
@ederign I think this approach will not work for runtime dependencies for exactly the same reasons that maven `dependency:analyze` doesn't work for them: they're not needed for compilation, only at runtime.
Imagine a project with dependency on jdbc driver. `mvn dependency:analyze` will report it as unused, so you remove it. Maven build will succeed, because it's not using any classes from it in the source code. And then you run the project and it will fail with `ClassNotFound`

desmax74:
@jhrcek the problematic deps using scope  `runtime` or `provided` and in this case we could skip the removal if they are present in the unusedDeps provided by `dependency:analyze` , wdyt ?

ederign:
@jhrcek got it, thanks for the clarification.

We have some options here:

- Rollback and don't provide this feature
- Provide this under as 'experimental feature' hidden on settings and as experimental, users will know the implications and 'hope for the best' (including that we can do what @desmax74 suggested and also mark the 'already know' problematic dependencies in order to avoid the exclusion
- Figure out another way to do this: at first, we thought about analyzing java classes imports in order to figure out dependencies. But I think even with this, we still can fall on runtime problem issue isn't?

@porcelli thoughts?

jhrcek:
> Figure out another way to do this: at first, we thought about analyzing java classes imports in order to figure out dependencies. But I think even with this, we still can fall on runtime problem issue isn't?

Yes, the same problem with that approach (also think e.g. `@Inject` - the implementation of the injected field doesn't need to be on the compile classpath but we'll miss it at runtime).

My general thoughts about this feature:
We can't even solve this problem (of unused / used but undeclared) dependencies in our (kiegroup) codebase. There are hundreds of poms and many of them contain unused dependencies / use undeclared dependencies. If we want to automate something like this for user projects, I'd first suggest you try to do it manually for couple of our projects and see how problematic something like that can be. Instead of adding half-working feature and then paying maintenance burden for it I'd recommend not implementing it in the first place.

PR #2192, opened 2018-10-16
RHPAM-1534 - Stunner - Not possible to use Undo action immediately

hasys:
Hi @romartin, @LuboTerifaj,

I investigated this issued and there are my findings:

**Short version:**
There are two EventHandlers for `MouseUp` event: on **lienzo-core** side and on **Stunner** side. Both handlers are executed when we finishing drag an element, but **lienzo** side handler is **_first_** in the queue of handlers and it execute check of command registry are there any commands to undo or not.

**Stunner** handler executes **_after_** **lienzo-core** handler and this handler adding new command to command registry and that's why first action doesn't unlock undo button, but works for second and following actions (and that's why it is possible to undo first action if button already unlocked).

As _temporary_ solution I added new event after new command added to registry and handle it in undo command. As long term solution I see **refactoring of Undo/Redo** commands.

My idea is fire RegistryChanged event in Client registry (probably we should create child of the existing registry to use GWT events only on client side and "vanilla" existing simple registry on server side) and handle this event in Undo/Redo commands.

So Undo/Redo commands won't be observer `any command executed`, but `registry changed`. (also `registry cleared` and `Redo/Undo command`  executed for `Undo/Redo` commands accordingly.

In my opinion it will simplify overall Undo/Redo functionality and will make it more robust. WDYT about it?

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/event/registration/RegisterChangedEvent.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/event/registration/RegisterChangedEvent.java
@@ -18,6 +18,6 @@ package org.kie.workbench.common.stunner.core.client.canvas.event.registration;
 
 import org.uberfire.workbench.events.UberFireEvent;
 
-public class CommandRegisteredEvent implements UberFireEvent {
+public class RegisterChangedEvent implements UberFireEvent {
 
 }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManager.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManager.java
@@ -22,6 +22,7 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Event;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
@@ -29,6 +30,7 @@ import org.kie.workbench.common.stunner.core.client.api.SessionManager;
 import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
 import org.kie.workbench.common.stunner.core.client.canvas.event.mouse.CanvasMouseDownEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.mouse.CanvasMouseUpEvent;
+import org.kie.workbench.common.stunner.core.client.canvas.event.registration.CommandRegisteredEvent;
 import org.kie.workbench.common.stunner.core.client.session.event.SessionDestroyedEvent;
 import org.kie.workbench.common.stunner.core.client.session.event.SessionOpenedEvent;
 import org.kie.workbench.common.stunner.core.command.Command;
@@ -58,17 +60,19 @@ public class RequestCommandManager extends AbstractSessionCommandManager {
 
     private final SessionManager sessionManager;
     private Stack<Command<AbstractCanvasHandler, CanvasViolation>> commands;
+    private Event<CommandRegisteredEvent> commandRegisteredEvent;
 
     private boolean roolback;
 
     protected RequestCommandManager() {
-        this(null);
+        this(null, null);
     }
 
     @Inject
-    public RequestCommandManager(final SessionManager sessionManager) {
+    public RequestCommandManager(final SessionManager sessionManager, Event<CommandRegisteredEvent> commandRegisteredEvent) {
         this.sessionManager = sessionManager;
         this.roolback = false;
+        this.commandRegisteredEvent = commandRegisteredEvent;
     }
 
     @Override
@@ -218,6 +222,7 @@ public class RequestCommandManager extends AbstractSessionCommandManager {
                 getRegistry().register(new CompositeCommand.Builder<AbstractCanvasHandler, CanvasViolation>()
                                                .addCommands(commands.stream().collect(Collectors.toList()))
                                                .build());
+                commandRegisteredEvent.fire(new CommandRegisteredEvent());
             }
             LOGGER.log(Level.FINEST,
                        "Current client request completed.");
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManager.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManager.java
@@ -16,13 +16,12 @@
 
 package org.kie.workbench.common.stunner.core.client.command;
 
+import java.util.ArrayList;
 import java.util.Stack;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
-import javax.enterprise.event.Event;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
@@ -30,7 +29,6 @@ import org.kie.workbench.common.stunner.core.client.api.SessionManager;
 import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
 import org.kie.workbench.common.stunner.core.client.canvas.event.mouse.CanvasMouseDownEvent;
 import org.kie.workbench.common.stunner.core.client.canvas.event.mouse.CanvasMouseUpEvent;
-import org.kie.workbench.common.stunner.core.client.canvas.event.registration.CommandRegisteredEvent;
 import org.kie.workbench.common.stunner.core.client.session.event.SessionDestroyedEvent;
 import org.kie.workbench.common.stunner.core.client.session.event.SessionOpenedEvent;
 import org.kie.workbench.common.stunner.core.command.Command;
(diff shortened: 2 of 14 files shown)


hasys:
**Technical details:**
`LienzoHandleManager` from **Lienzo-core** catches `MouseUpEvent` first and than fire two more events:

`MouseUpEvent` → `NodeMouseUpEvent` → `NodeDragEndEvent`

`NodeDragEndEvent` handled by 3 handlers `WiresShapeHandlerImpl`, `MagnetManager$Magnets`, `WiresContainer` and those 3 classes execute 4 commands:
- `UpdateChildNodeCommand`
- `UpdateChildNodeCommand`
- `UpdateElementPositionCommand`
- `UpdateElementPositionCommand`

Every command fire `CanvasCommandExecutedEvent`, it triggers `UndoSessionCommand#onCommandExecuted` 4 times. So, **state of the register handled 4 times but all 4 times register is empty**, because only Stunner handler will add there a new command.  

When `LienzoHandleManager` finish its work, `LienzoPanelView` from **kie-wb-common-stunner-widgets** starting handle `MouseUpEvent`. It fires `CanvasMouseUpEvent` and `CanvasMouseUpEvent` finally catched by `RequestCommandManager` which adding new command to registry.

But after it `UndoSessionCommand` doesn't check registry until next `CanvasCommandExecutedEvent` will be fired.

hasys:

hasys:
Hi @LuboTerifaj @romartin, just kind reminder :)

romartin:
Hey @hasys , does this bug you mention still apply? From the demo session we did today, I remember @tiagodolphine showing it fixed. Should we proceed anyway or is it now not necessary?
Thanks anyway for the help!

hasys:
Hi @romartin,

yes it is still there. As I described issue looks small but it isn't and it is better to do bigger refactoring in my opinion, in this PR we have a "quick fix". If we agreed on refactoring I can do it, just let's create a Jira issue and add it to a Sprint.

Steps To Reproduce:
1. Startup master snapshot (or any other without this fix)
2. Create new process
3. Move Start node one time (you will see that Undo button is still **Disabled**)
4. Move Start node one more time (you will see that Undo button is **Enabled**)
5. Press Undo button (Start event will be moved back, Undo button is still **Enabled**)
6. Press Undo button one more time (Start event will be moved to initial position).

Actual Result:
After move 3. Undo button will be **Enabled** and you will be able to Undo.

Some notes:
If you will put a break point to Undo register, after first move it will be checked 4 times but button is still disabled because stack is empty. If you will have more nodes stack will be checked even more times, but with same result (Undo button won't be enabled). Refactoring will fire Undo Stack check just once per operation over command and will always work. So, after refactoring Undo/Redo system should be more stable, robust and faster.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManager.java
```diff
@@ -58,17 +60,19 @@
...
     private boolean roolback;
 
     protected RequestCommandManager() {
-        this(null);
+        this(null, null);
     }
 
     @Inject
-    public RequestCommandManager(final SessionManager sessionManager) {
+    public RequestCommandManager(final SessionManager sessionManager, Event<CommandRegisteredEvent> commandRegisteredEvent) {
```
tiagodolphine: the attribute Event can be final as the SessionManager

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/test/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManagerTest.java
```diff
@@ -80,7 +85,7 @@ public void setup() throws Exception {
         when(editorSession.getCanvasHandler()).thenReturn(canvasHandler);
         when(editorSession.getCommandRegistry()).thenReturn(commandRegistry);
         when(editorSession.getCommandManager()).thenReturn(commandManager);
-        this.tested = new RequestCommandManager(clientSessionManager);
+        this.tested = new RequestCommandManager(clientSessionManager, commandRegisteredEvent);
```
tiagodolphine: it is missing the `verify` of the commandRegisteredEvent fire on the tests.
You can use the `EventSourceMock` to mock the event.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/UndoSessionCommand.java
```diff
@@ -110,6 +111,12 @@ void onCommandUndoExecuted(final @Observes CanvasUndoCommandExecutedEvent comman
         checkState();
     }
 
+    void onCommandAdded(final @Observes CommandRegisteredEvent commandRegisteredEvent) {
```
tiagodolphine: missing some test for this new method.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/command/RequestCommandManager.java
```diff
@@ -218,6 +222,7 @@ private void complete() {
                 getRegistry().register(new CompositeCommand.Builder<AbstractCanvasHandler, CanvasViolation>()
                                                .addCommands(commands.stream().collect(Collectors.toList()))
                                                .build());
+                commandRegisteredEvent.fire(new CommandRegisteredEvent());
```
romartin: IMO makes more sense firing the `CommandRegisteredEvent` from the registry component (`getRegistry().register`), instead of from this `RequestCommandMananger` class. So any other component that is using the registry API will also take benefit from this, as the event will be fired internally.

hasys:
Hi @tiagobento, @romartin,

sorry for the delay, finally got a chance to finish this PR. I rebased to master, applied all Tiago's comments and did refactoring as Roger suggested.

Thank you!

hasys:
Hi @romartin, @tiagodolphine,

I updated DMN Editor to follow new Default Editor API to fix build and rebased to master again.

CC @jomarko

hasys:

romartin:
Hey @hasys 

I have this PR in mind - but I'm working on a few issues on some commands now and also will try to give this a look. I think there should be no need for adding new events but need to test your approach and debug a few more locally to be sure... so will let you know ASAP, and thanks a lot for collaborating, as usually, just I need to refresh and understand this again :+1:  

Thanks

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/event/registration/RegisterChangedEvent.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.client.canvas.event.registration;
+
+import org.uberfire.workbench.events.UberFireEvent;
+
+public class RegisterChangedEvent implements UberFireEvent {
```
romartin: well just my opinion... but doesn't sounds better renaming this to something like `CommandRegistryUpdatedEvent` or `CommandRegistryChangedEvent` ??

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/registry/impl/ClientCommandRegistry.java
```diff
@@ -0,0 +1,62 @@
...
+import org.kie.workbench.common.stunner.core.client.canvas.event.registration.RegisterChangedEvent;
+import org.kie.workbench.common.stunner.core.command.Command;
+
+/**
+ * The client implementation for the CommandRegistry type.
+ * Note: The Stack class behavior when using the iterator is not the expected one, so used
+ * ArrayDeque instead of an Stack to provide right iteration order.
+ */
+@Dependent
+public class ClientCommandRegistry<C extends Command> extends CommandRegistryImpl<C> {
```
romartin: So that's fine.. but I guess we can do it in some easier way rather than creating a custom `CommandRegistry` implementation for client side.... 
You can also move the new `RegisterChangedEvent` type into the `core-common` module, and just throw the new event instance on the actual `CommandRegistryImpl` implementation. 
This way there will be no need to refactor so many classes (at least now) and introduce more complexity by providing custom client side bean type resolutions...  for example, no need for refactoring the `AbstractProjectEditor`'s, no need to create this class, etc. 
The only downside is that the event should be fired on backend side, in case some registry instance is used on backend, which is NOT the case, anyway nobody else is listening to this event on backend.
WDYT?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/event/registration/RegisterChangedEvent.java
romartin: Also another comment on this class - this event has no arguments, as for example the `CanvasCommandExecutedEvent` which has the `canvasHandler` (UUID), so please ensure you test it as well with more than one Stunner editor open, as I think the event observers will be listening although the editor not being focused... just try to execute and undo things in two different processes, also change the current editor and click on the undo/redo commands, and check each one is undoing/redoing the right commands, as for each editor (not being "mixed").
Thanks!

jomarko:
I briefly checked DMN, undo and redo seems still working to me.

hasys:
Hi @romartin,

good point about several processes, I found some issues, but **I am able to reproduce it with master as well**. Probably, we should fix all issues in this PR, but it will take little bit more time.

About merging client and backend registries, if you think it is better to fire events on backend as well, we can do it, but not sure what is the benefit especially when refactoring is already done and doesn't harm (in my opinion).

Found issue with several open editors:
* State of the toolbar doesn't updated when you switch processes
* Redo command can share some state between processes:
  * Open two processes **A** and **B**
  * Model process **B** = (Start->Task->End)
  * Undo End and Task for process **B**
  * Switch to process **A** and click on Redo
    * Actual result: Task will appear in top left corner of process **A**, error will be displayed (see picture below)
    * Expected Result: Redo button will be disabled. If process A has any actions to redo, only related to process A commands will be redo.

image

romartin:
hey @hasys 

Good to see you back!!! 

TBH you're right, let's avoid firing events on backend and also because the classes are already there, I was just trying to reduce the amount of client customizations we have on the whole codebase... but anyway, I'm agree with you, it makes sense, ti's already done and even will behave better at performance terms. So lemme fire a new build and if it's fine, will merge, thanks!

Also no worries, better please report the bug above about multiple editors and let's do it in another PR, probably some other things will be affected and we have to test it well... thanks!!

romartin:
Jenkins please retest his

romartin:
The J build is failing due to unrelated tests, merging.

hasys:
Hi @romartin,

thank you. For the record, I filed an issue:
* [JBPM-8207](https://issues.jboss.org/browse/JBPM-8207) - Redo command is shared between windows

PR #2200, opened 2018-10-19
DROOLS-2932: Grid Keyboard control

jomarko:
Dependent on #2324 

Ensemble with:
- kiegroup/appformer#518
- kiegroup/drools-wb#968

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
@@ -73,12 +73,14 @@ import org.kie.workbench.common.stunner.forms.client.event.RefreshFormProperties
 import org.uberfire.ext.wires.core.grids.client.model.GridCellValue;
 import org.uberfire.ext.wires.core.grids.client.model.GridColumn;
 import org.uberfire.ext.wires.core.grids.client.model.GridData;
+import org.uberfire.ext.wires.core.grids.client.util.ColumnIndexUtilities;
 import org.uberfire.ext.wires.core.grids.client.util.CoordinateUtilities;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.GridWidget;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.NodeMouseEventHandler;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.columns.RowNumberColumn;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.impl.DefaultGridWidgetCellSelectorMouseEventHandler;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.grids.GridRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.selections.SelectionExtension;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.GridSelectionManager;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.impl.GridLayerRedrawManager;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.pinning.GridPinnedModeManager;
@@ -522,6 +524,26 @@ public abstract class BaseExpressionGrid<E extends Expression, D extends GridDat
             doAfterHeaderSelectionChange(uiHeaderRowIndex, uiHeaderColumnIndex);
         }
 
+        return isSelectionChanged;
+    }
+
+    @Override
+    public boolean adjustSelection(final SelectionExtension direction,
+                                   final boolean isShiftKeyDown) {
+        final boolean isSelectionChanged = super.adjustSelection(direction, isShiftKeyDown);
+        if (isSelectionChanged) {
+            if (getModel().getSelectedCells().size() > 0) {
+                final GridData.SelectedCell selectedCell = getModel().getSelectedCells().get(0);
+                doAfterSelectionChange(selectedCell.getRowIndex(),
+                                       ColumnIndexUtilities.findUiColumnIndex(getModel().getColumns(),
+                                                                              selectedCell.getColumnIndex()));
+            } else if (getModel().getSelectedHeaderCells().size() > 0) {
+                final GridData.SelectedCell selectedHeraderCell = getModel().getSelectedHeaderCells().get(0);
+                doAfterHeaderSelectionChange(selectedHeraderCell.getRowIndex(),
+                                             ColumnIndexUtilities.findUiColumnIndex(getModel().getColumns(),
+                                                                                    selectedHeraderCell.getColumnIndex()));
+            }
+        }
         return isSelectionChanged;
     }
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
@@ -75,6 +75,7 @@ import org.uberfire.ext.wires.core.grids.client.model.GridColumn;
 import org.uberfire.ext.wires.core.grids.client.model.GridData;
 import org.uberfire.ext.wires.core.grids.client.util.ColumnIndexUtilities;
 import org.uberfire.ext.wires.core.grids.client.util.CoordinateUtilities;
+import org.uberfire.ext.wires.core.grids.client.widget.dom.HasDOMElementResources;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.GridWidget;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.NodeMouseEventHandler;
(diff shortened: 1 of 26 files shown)


jomarko:
@manstis you are definitely right, thanks for review, I will investigate slightly but have feeling that will close the PRs and reassign to someone with higher knowledge.

manstis:
@jomarko I think you should keep the PRs. The items I raised are undecided and can easily be the topic of additional JIRA (if they are requested or required). If you manage to get the focus to work after the cell editor closes I think this would be good to merge.

jomarko:
Status update of able to do:
- navigate in data cells, edit data cell, navigate header cells, edit some header cells

Status update of unable to do:
- dive into cell with inner editor, invoke popup for cell with undefined expression type, edit some header cells

Note, if user wants to finish edit mode of regular data cell, he has to press SHIFT+TAB, not just TAB as intended originally. 

@manstis please have a look as discussed on meeting today.

jomarko:
@manstis Ok, I adapted `EditablePopupHeaderMetaData.destroyResources` to hide popup with name and data type in blur event.

Now trying to show popup at correct coordinates, try to reduce code duplication and will write tests.

manstis:
Use the startEditing(row, column, ..) methods a d positioning should be taken care of automatically.

jomarko:
@manstis  great pointer!

manstis:
@jomarko Please see for [fixed](https://gist.github.com/manstis/a35556247dda096874114e9e89c3556f) test (failing on this PR).

jomarko:
@manstis not sure about comment above, my PR is not complete, marked as not ready for merge, I am not done with new tests and updates of existing, just wanted feedback about code architecture and manual check feedback, nevermind, will proceed and update also tests before I ask you for review, thanks.

manstis:
@jomarko Overall this seems to work well (and the code looks fine). A couple of observations:-

- I like the "dive up/down" navigation for the nested grids
- It'd be good to be able to exit editing _normal_ cells with `<tab>` (and not just `<shift>-<tab>`)
- You need something like the following on `BaseExpressionGrid` to ensure the "Properties Panel" is updated as you move around with the cursor keys.

```
    @Override
    public boolean adjustSelection(final SelectionExtension direction,
                                   final boolean isShiftKeyDown) {
        final boolean hasSelectionChanged = super.adjustSelection(direction, isShiftKeyDown);
        final GridData.SelectedCell selectedCellOrigin = getModel().getSelectedCellsOrigin();
        doAfterSelectionChange(selectedCellOrigin.getRowIndex(),
                               selectedCellOrigin.getColumnIndex());
        return hasSelectionChanged;
    }
```

manstis:
Hi @jomarko I was asked to re-review https://github.com/kiegroup/appformer/pull/518 by @ederign so thought I'd see how the whole ensemble works. FYI, here's some feedback for the DMN editor with the latest code changes - some I propose are new JIRA (assigned to me), some you might like to consider for these PRs:-

If multiple cells are selected <ENTER> edits first selected - should it not edit any?

Decision Table Hit Policy header cell does not respond to <ENTER>.

Decision Table with multiple Output columns. Arrow keys to header of output columns is OK; however if 2nd row has "focus" then you cannot move to Input column or Description column. Need to move to 1st row of Output columns to move left/right to Input/Description columns.

Nested context (context with a context). The 'inner' context you can navigate into an invisible "header". This is an issue in `ContextGrid` (not your keyboard code) as if the grid is nested **NO** header meta-data should be added. See `InvocationGrid.initialiseUiColumns()` where a header meta-data is only added if `nesting == 0` (this needs to be done in `ContextGrid` to only add meta-data if `nesting == 0`). I checked other grids and they were OK. Please raise a JIRA for me.

Issue with `LiteralExpressionGrid` and `UndefinedExpressionGrid`. If these are selected it's not possible to navigate to other cells. This is an issue for me. If you remember I spoke to you about how I used to select the parent cell if either of these grids were selected however following the change for "drill down into properties" this no longer happens. This means when you select either of the grids the 'selected grid' is only that one and hence keyboard navigation appears to not work (from a Users perspective it looks like the cell parent is selected however in reality it is the nested grid). You need to press <ESCAPE> to navigate to 'parent grid' and then keyboard works OK. To edit either a `LiteralExpression` or `UndefinedExpression` the User needs to press <ENTER> twice: 1st 'enters nested grid', 2nd edits cell. Please raise a JIRA for me (I did try reinstating the old code however it then led to issues with "nested properties" as the parent grid was selected). I need to think about it!!!

Following "single click" to edit a cell; when you click on (for example) a Context "Name" cell the popover appears. Using the keyboard correctly navigates to other cells however the popover remains visible. I guess we should hide the popover when the keyboard navigation runs?

Otherwise looking good :-)

jomarko:
@manstis 4/6 comments should be incorporated now, jiras for you and tests pending.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
```diff
@@ -588,6 +590,27 @@ public boolean selectHeaderCell(final int uiHeaderRowIndex,
...
+                                   final boolean isShiftKeyDown) {
+        final boolean hasSelectionChanged = super.adjustSelection(direction, isShiftKeyDown);
+        if (hasSelectionChanged) {
+            if (getModel().getSelectedCells().size() > 0) {
+                final GridData.SelectedCell selectedCell = getModel().getSelectedCells().get(0);
+                doAfterSelectionChange(selectedCell.getRowIndex(),
+                                       ColumnIndexUtilities.findUiColumnIndex(getModel().getColumns(),
+                                                                              selectedCell.getColumnIndex()));
+            }
+            if (getModel().getSelectedHeaderCells().size() > 0) {
```
manstis: Is an `..} else if(getModel().getSelectedHeaderCells().size() > 0) {...` better?

manstis: (the important bit is **else** in between the body selected cells and header selected cells)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/KeyboardOperationEditGridCell.java
```diff
@@ -0,0 +1,95 @@
...
+            final int uiHeaderRowIndex = selectedCell.getRowIndex();
+            final int uiColumnIndex = ColumnIndexUtilities.findUiColumnIndex(model.getColumns(),
+                                                                             selectedCell.getColumnIndex());
+
+            final GridColumn<?> column = model.getColumns().get(uiColumnIndex);
+
+            if (column instanceof DMNGridColumn) {
+                ((DMNGridColumn) column).startEditingHeaderCell(uiHeaderRowIndex);
+            }
+            if (column instanceof RowNumberColumn) {
```
manstis: I think it'd simply be better to make `DecisionTableRowNumberColumn extend DMNGridColumn<DecisionTableGrid, Integer> implements IsRowDragHandle` rather than `RowNumberColumn` then all this boiler plate can be removed.
```
public class DecisionTableRowNumberColumn extends DMNGridColumn<DecisionTableGrid, Integer> implements IsRowDragHandle {

    private static final double DEFAULT_WIDTH = 50.0;

    public DecisionTableRowNumberColumn(final Supplier<HitPolicy> hitPolicySupplier,
                                        final Supplier<BuiltinAggregator> builtinAggregatorSupplier,
                                        final CellEditorControlsView.Presenter cellEditorControls,
                                        final HitPolicyPopoverView.Presenter editor,
                                        final Optional<String> editorTitle,
                                        final DecisionTableGrid gridWidget) {
        super(new RowNumberColumnHeaderMetaData(hitPolicySupplier,
                                                builtinAggregatorSupplier,
                                                cellEditorControls,
                                                editor,
                                                editorTitle,
                                                gridWidget),
              new IntegerColumnRenderer(),
              gridWidget);
        setMovable(false);
        setResizable(false);
        setFloatable(true);

        setWidthInternal(DEFAULT_WIDTH);
    }
}
```
This would also require a couple of changes elsewhere:
* `CommandUtils` L#41 `.filter(column -> column instanceof IsRowDragHandle)`
* `BaseExpressionGridTheme` L#90 `} else if (column instanceof IsRowDragHandle) {`

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/BaseExpressionGrid.java
jomarko: Agree, thank you.

jomarko:
Failed tests:
```
org.kie.workbench.common.screens.server.management.backend.EmbeddedControllerIT.testEmbeddedEndpointsUsingRest
org.kie.workbench.common.screens.server.management.backend.EmbeddedControllerIT.testEmbeddedEndpointsUsingWebSocket
org.kie.workbench.common.stunner.bpmn.backend.service.indexing.BpmnFileIndexerTest.testBpmnIndexing
```

jomarko:
jenkins please retest this

jomarko:
@kiegroup/gatekeepers please merge

PR #2202, opened 2018-10-22
DROOLS-3137: [DMN Designer] Data Types Type - Constraints - MVP component

karreiro:
See https://issues.jboss.org/browse/DROOLS-3137

demo

(the GIF is a little bit outdated, I've improved the tooltip message after some discussions with Stetson and Liz)


karreiro:
@jomarko @manstis Thank you for the review.

---

Jozef, I don't think that this properties have a strong relation, I've checked the DMN spec and this is the description of the Allowed Answers property in the Decision Node:
screen shot 2018-10-24 at 11 23 58 am

..and this is the description of the Allowed Values for an Item Definition:
screen shot 2018-10-24 at 11 26 52 am

Even their types are different, so I don't think that these properties have a strong relation.

PR #2204, opened 2018-10-23
DROOLS-3143 [DMN Designer] DMN unmarshal from 1.1, 1.2; marshal as 1.2

tarilabs:
I have not yet renamed package which hold the old `v1_1` reference in the hope to simplify the review, we can either leave it unchanged or refactor the package rename later once we agree on the actual code ;)

/cc @jomarko @manstis @etirelli 

please notice this requires https://github.com/kiegroup/drools/pull/2116


tarilabs:
Jenkins test again

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -173,98 +178,98 @@ public Graph unmarshall(final Metadata metadata,
                     if (ir.getRequiredInput() != null) {
                         String reqInputID = getId(ir.getRequiredInput());
                         Node requiredNode = elems.get(reqInputID).getValue();
-                        Edge myEdge = factoryManager.newElement(UUID.uuid(),
+                        Edge myEdge = factoryManager.newElement(ir.getId() != null ? ir.getId() : UUID.uuid(),
```
karreiro: I've noticed that this condition happens many times in this class, maybe we could extract this to something like:
```java
private String getNonNullUUID(final String id) {
    return Optional.ofNullable(id).orElse(UUID.uuid());
}
```
..and re-use this in the lines 181, 191, 202, 212, 224, 234, 247, 257, 267 and 293.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -366,17 +371,41 @@ private void connectEdge(final Edge edge,
...
+                                .findFirst();
+        }
+        if (dmnEdge.isPresent()) {
+            DMNEdge e = dmnEdge.get();
+            Point source = e.getWaypoint().get(0);
+            connectionContent.setSourceConnection(MagnetConnection.Builder.at(source.getX(), source.getY()));
+            Point target = e.getWaypoint().get(e.getWaypoint().size() - 1);
+            connectionContent.setTargetConnection(MagnetConnection.Builder.at(target.getX(), target.getY()));
+            if (e.getWaypoint().size() > 2) {
+                List<Point> sublist = e.getWaypoint().subList(1, e.getWaypoint().size() - 2);
```
karreiro: I think we could cover this branch with tests.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -410,31 +444,57 @@ public String marshall(final Diagram<Graph, Metadata> diagram) throws IOExceptio
...
+                        final ViewConnector connectionContent = (ViewConnector) e.getContent();
+                        if (connectionContent.getSourceConnection().isPresent() && connectionContent.getTargetConnection().isPresent()) {
+                            Connection sourceConn = (Connection) connectionContent.getSourceConnection().get();
+                            Connection targetConn = (Connection) connectionContent.getTargetConnection().get();
+                            DMNEdge dmnEdge = new org.kie.dmn.model.v1_2.dmndi.DMNEdge();
+                            dmnEdge.setId("dmnedge-" + e.getUUID());
+                            dmnEdge.setDmnElementRef(new QName(e.getUUID()));
+                            dmnEdge.getWaypoint().add(PointUtils.point2dToDMNDIPoint(sourceConn.getLocation()));
+                            for (ControlPoint cp : connectionContent.getControlPoints()) {
+                                dmnEdge.getWaypoint().add(PointUtils.point2dToDMNDIPoint(cp.getLocation()));
```
karreiro: I think we could cover this line by tests too.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -537,14 +633,15 @@ private static void applyBounds(final RectangleDimensionsSet dimensionsSet,
 
     private static void applyBackgroundStyles(final BackgroundSet bgset,
                                               final DMNShape result) {
+        if (!(result.getStyle() instanceof DMNStyle)) {
+            return;
```
karreiro: This logic is not covered by tests.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -473,59 +533,95 @@ private void internalAugment(Stream<DMNShape> drgShapeStream, Id id, Bound ul, R
...
-            result.setFontStyle(FontSetPropertyConverter.dmnFromWB(d.getFontSet()));
+            applyFontStyle(d.getFontSet(), result);
         }
-
         return result;
     }
 
+    private static void applyFontStyle(FontSet fontSet, DMNShape result) {
+        if (!(result.getStyle() instanceof DMNStyle)) {
+            return;
```
karreiro: This logic is not covered by tests.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -304,23 +302,28 @@ public void testLoan() throws IOException {
         assertFalse(dmnResult.getMessages().toString(), dmnResult.hasErrors());
     }
 
+    @Test
+    public void temp() {
```
karreiro: I think we can remove this method, right? 🙃

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -342,49 +345,45 @@ public void test_diamond() throws IOException {
...
 
         DMNShape myname = findShapeByDMNI(ddRoot, "_4cd17e52-6253-41d6-820d-5824bf5197f3");
         assertBounds(500, 500, 100, 50, myname.getBounds());
-        assertColor(255, 255, 255, myname.getBgColor());
-        assertColor(0, 0, 0, myname.getBorderColor());
-        assertEquals(0.5, myname.getBorderSize().getValue(), 0);
-        assertDMNStyle("Open Sans", 24, 1, 255, 0, 0, myname.getFontStyle());
+        assertColor(255, 255, 255, ((DMNStyle) myname.getStyle()).getFillColor());
+        assertColor(0, 0, 0,       ((DMNStyle) myname.getStyle()).getStrokeColor());
+        //        assertEquals(0.5, myname.getBorderSize().getValue(), 0);
```
karreiro: I think we could remove this line..

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -342,49 +345,45 @@ public void test_diamond() throws IOException {
...
 
         DMNShape prefix = findShapeByDMNI(ddRoot, "_e920f38a-293c-41b8-adb3-69d0dc184fab");
         assertBounds(300, 400, 100, 50, prefix.getBounds());
-        assertColor(0, 253, 25, prefix.getBgColor());
-        assertColor(253, 0, 0, prefix.getBorderColor());
-        assertEquals(1, prefix.getBorderSize().getValue(), 0);
-        assertDMNStyle("Times New Roman", 8, 2.5, 70, 60, 50, prefix.getFontStyle());
+        assertColor(0, 253, 25, ((DMNStyle) prefix.getStyle()).getFillColor());
+        assertColor(253, 0, 0, ((DMNStyle) prefix.getStyle()).getStrokeColor());
+        //        assertEquals(1, prefix.getBorderSize().getValue(), 0);
```
karreiro: ..and this one

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -342,49 +345,45 @@ public void test_diamond() throws IOException {
...
 
         DMNShape postfix = findShapeByDMNI(ddRoot, "_f49f9c34-29d5-4e72-91d2-f4f92117c8da");
         assertBounds(700, 400, 100, 50, postfix.getBounds());
-        assertColor(247, 255, 0, postfix.getBgColor());
-        assertColor(0, 51, 255, postfix.getBorderColor());
-        assertEquals(2, postfix.getBorderSize().getValue(), 0);
-        assertDMNStyle("Arial", 10, 1.5, 50, 60, 70, postfix.getFontStyle());
+        assertColor(247, 255, 0, ((DMNStyle) postfix.getStyle()).getFillColor());
+        assertColor(0, 51, 255, ((DMNStyle) postfix.getStyle()).getStrokeColor());
+        //        assertEquals(2, postfix.getBorderSize().getValue(), 0);
```
karreiro: ..and this one

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -342,49 +345,45 @@ public void test_diamond() throws IOException {
...
 
         DMNShape mydecision = findShapeByDMNI(ddRoot, "_9b061fc3-8109-42e2-9fe4-fc39c90b654e");
         assertBounds(487.5, 275, 125, 75, mydecision.getBounds());
-        assertColor(255, 255, 255, mydecision.getBgColor());
-        assertColor(0, 0, 0, mydecision.getBorderColor());
-        assertEquals(0.5, mydecision.getBorderSize().getValue(), 0);
-        assertDMNStyle("Monospaced", 32, 3.5, 55, 66, 77, mydecision.getFontStyle());
+        assertColor(255, 255, 255, ((DMNStyle) mydecision.getStyle()).getFillColor());
+        assertColor(0, 0, 0, ((DMNStyle) mydecision.getStyle()).getStrokeColor());
+        //        assertEquals(0.5, mydecision.getBorderSize().getValue(), 0);
```
karreiro: ..and this one

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -342,49 +345,45 @@ public void test_diamond() throws IOException {
...
+        assertDMNStyle("Monospaced", 32, 55, 66, 77, (DMNStyle) mydecision.getStyle());
     }
 
-    private void assertDMNStyle(String fontName, double fontSize, double fontBorderSize, int r, int g, int b, DMNStyle style) {
-        assertEquals(fontName, style.getFontName());
+    private void assertDMNStyle(String fontName, double fontSize, int r, int g, int b, DMNStyle style) {
+        assertEquals(fontName, style.getFontFamily());
         assertEquals(fontSize, style.getFontSize(), 0);
-        assertEquals(fontBorderSize, style.getFontBorderSize(), 0);
+        //        assertEquals(fontBorderSize, style.getFontBorderSize(), 0);
```
karreiro: ..and, finally, this one :-)

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
```diff
@@ -401,8 +400,12 @@ private static void assertColor(int r, int g, int b, Color color) {
...
     }
 
-    private DMNShape findShapeByDMNI(org.kie.workbench.common.dmn.backend.definition.v1_1.dd.DMNDiagram root, String id) {
-        return root.getAny().stream().filter(shape -> shape.getDmnElementRef().equals(id)).findFirst().get();
+    private DMNShape findShapeByDMNI(org.kie.dmn.model.api.dmndi.DMNDiagram root, String id) {
+        return root.getDMNDiagramElement().stream()
+                   .filter(DMNShape.class::isInstance)
+                   .map(DMNShape.class::cast)
+                   .filter(shape -> shape.getDmnElementRef().getLocalPart().equals(id))
+                   .findFirst().get();
```
karreiro: When we have a `.get` with no `isPresent` check, I think that's a good idea to explain why the value will always be present and do something like this:
```java
    private DMNShape findShapeByDMNI(org.kie.dmn.model.api.dmndi.DMNDiagram root, String id) {
        return root.getDMNDiagramElement()
                   .stream()
                   .filter(DMNShape.class::isInstance)
                   .map(DMNShape.class::cast)
                   .filter(shape -> shape.getDmnElementRef().getLocalPart().equals(id))
                   .findFirst()
                   .orElseThrow(() -> new UnsupportedOperationException("the explanation here"));
    }
```

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: OK.
Mentioning for repeated should be just fine ;)

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: It is not indeed, because by DMN XSD design this will not happen just by making use of the standard without extensions.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: ref reply

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
tarilabs: Oops thanks for spotting! :)

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
tarilabs: Oops yes, as mentioned DMNDI does not support for border so I am not persisting this data.

tarilabs: (mentioning repeated should fine ;) )

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerTest.java
tarilabs: This is a test utility method, so I believe it would be straight-forward to understand any error, but anyway addressing as requested.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: ok

tarilabs:
For some reason this github page is not seeing https://github.com/tarilabs/kie-wb-common/commit/e839ba1ce7bfd5d04a78c60562993f813fbb0a11

tarilabs:
I'll try rebase in the hope this github page will notice https://github.com/tarilabs/kie-wb-common/commit/e839ba1ce7bfd5d04a78c60562993f813fbb0a11 has been added.

jomarko:
@tarilabs found one more issue during PR review [DROOLS-3185](https://issues.jboss.org/browse/DROOLS-3185)

tarilabs:
Okay I believe maybe I got where the problem is:
```
[INFO] 	at org.kie.workbench.common.dmn.backend.definition.v1_1.dd.PointUtils.point2dToDMNDIPoint(PointUtils.java:27)
[INFO] 	at org.kie.workbench.common.dmn.backend.DMNMarshaller.marshall(DMNMarshaller.java:472)
```

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -390,9 +392,25 @@ private void setConnectionMagnets(final Edge edge, String dmnEdgeElementRef, org
...
             Point source = e.getWaypoint().get(0);
-            connectionContent.setSourceConnection(MagnetConnection.Builder.at(source.getX(), source.getY()));
+            final Node sourceNode = edge.getSourceNode();
+            if (null != sourceNode) {
+                View<?> sourceView = (View<?>) sourceNode.getContent();
+                double xSource = xOfBound(upperLeftBound(sourceView));
+                double ySource = yOfBound(upperLeftBound(sourceView));
+                connectionContent.setSourceConnection(MagnetConnection.Builder.at(source.getX() - xSource, source.getY() - ySource)); // Stunner connection x,y is relative to shape
+            } else { // fallback:
+                connectionContent.setSourceConnection(MagnetConnection.Builder.at(source.getX(), source.getY()));
```
karreiro: I think this branch is not covered by tests.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -390,9 +392,25 @@ private void setConnectionMagnets(final Edge edge, String dmnEdgeElementRef, org
...
             Point target = e.getWaypoint().get(e.getWaypoint().size() - 1);
-            connectionContent.setTargetConnection(MagnetConnection.Builder.at(target.getX(), target.getY()));
+            final Node targetNode = edge.getTargetNode();
+            if (null != targetNode) {
+                View<?> targetView = (View<?>) targetNode.getContent();
+                double xTarget = xOfBound(upperLeftBound(targetView));
+                double yTarget = yOfBound(upperLeftBound(targetView));
+                connectionContent.setTargetConnection(MagnetConnection.Builder.at(target.getX() - xTarget, source.getY() - yTarget)); // Stunner connection x,y is relative to shape
+            } else { // fallback:
+                connectionContent.setTargetConnection(MagnetConnection.Builder.at(target.getX(), target.getY()));
```
karreiro: ..and this one.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -464,16 +482,48 @@ public String marshall(final Diagram<Graph, Metadata> diagram) throws IOExceptio
                     if (e.getContent() instanceof ViewConnector) {
                         final ViewConnector connectionContent = (ViewConnector) e.getContent();
                         if (connectionContent.getSourceConnection().isPresent() && connectionContent.getTargetConnection().isPresent()) {
-                            Connection sourceConn = (Connection) connectionContent.getSourceConnection().get();
-                            Connection targetConn = (Connection) connectionContent.getTargetConnection().get();
+                            Point2D sourcePoint = ((Connection) connectionContent.getSourceConnection().get()).getLocation();
+                            Point2D targetPoint = ((Connection) connectionContent.getTargetConnection().get()).getLocation();
+                            if (sourcePoint == null) { // If the "connection source/target location is null" assume it's the centre of the shape.
+                                View<?> sourceView = (View<?>) e.getSourceNode().getContent();
```
karreiro: This whole branch is not covered by tests.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -464,16 +482,48 @@ public String marshall(final Diagram<Graph, Metadata> diagram) throws IOExceptio
...
+                                }
+                                sourcePoint = Point2D.create(xSource, ySource);
+                            } else { // If it is non-null it is relative to the source/target shape location.
+                                View<?> sourceView = (View<?>) e.getSourceNode().getContent();
+                                double xSource = xOfBound(upperLeftBound(sourceView));
+                                double ySource = yOfBound(upperLeftBound(sourceView));
+                                sourcePoint = Point2D.create(xSource + sourcePoint.getX(), ySource + sourcePoint.getY());
+                            }
+                            if (targetPoint == null) { // If the "connection source/target location is null" assume it's the centre of the shape.
+                                double xTarget = xOfBound(upperLeftBound(view));
```
karreiro: This branch is not covered too.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: indeed because a fallback, as explained by @manstis this is a workaround to a Stunner issue

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: ref reply

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: Disagree. This is the branch covered by the functional test. I would be happy to spare @jomarko manual functional test if there was a way.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: ref reply.

tarilabs:
I have added an additional commit which resolve an issue as the unmarshalling code needed to be transformed into a 2-pass, as Stunner rely on relative positioning for the connection of the Edges.
Then I've done the following.
I included 1 InputData and 1Decision and an InformationRequirement edge.
I have saved and reopened multiple time.
Between re-open, before saving again, I changed the location of the "magnet" to be either of the center, left, right, etc. and mixing the same on the 2 Shapes/Nodes.
I done at least 1 combination of each different, and once Saved and re-opened, it was preserved where I did expect it.
Finally, I did the magnet like this:
image
and I checked the XML was consistent.
I also checked on the Partner editor:
image

@jomarko please notice that the Partner editor make a change if the edge crosses the shape, to avoid the crossing --or at least this is what I noticed.
For example, if you do on our editor something like:
image
the XML will be correct:
image
but on the Partner editor it will render as:
image

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
jomarko: Not sure what functional test do you mean @tarilabs?

tarilabs:
Functional test as a synonym of manual, non-programmable test, executed by human not the machine

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
karreiro: @tarilabs If this line does not deserve to be tested, could we remove it?

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
karreiro: I just ran all unit tests with a coverage tool and this branch continues uncovered.

tarilabs: > I just ran all unit tests with a coverage tool and this branch continues uncovered.

@karreiro I believe I was not able to explain myself.

I have _no_ way to cover with a programmatic test this branch.

This branch is actually covered when @jomarko performs the tests manually, by hand.

This branch cover the requirements as explained by @manstis by the way Stunner framework works, rather its current limitation as explained in https://issues.jboss.org/browse/DROOLS-3184?focusedCommentId=13652929&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13652929, when creating a Diagram for DMN.

Insisting of running a code-coverage tool won't sort any effect, until there is a way to make a DMN Diagram in a programmatic way based on the Stunner canvas.
I was told when originally providing the marshaller that there is no feasible way to automate the test which today @jomarko is doing manually, by hand.

That is the reason why test in the codebase starts with a DMNDiagram in xml form, and de-serialize this into a Stunner Diagram, to cycle in the Marshaller, and finally perform assertion; please notice all test case start from a DMN xml for this reason.

So unless there is a way now to programmatically instrument Stunner to script it like "Create a Decision shape, click on constellation button other-decision, drag shape to this location x,y, etc." there will be no way to cover this **specific** branch from a purely code perspective. As this branch is meant to cover the Stunner limitation from link above, so it would occur only and only when the Diagram is created directly from Stunner.

Was I able to fully explain this time why code-coverage metric is not relevant for this case?

Naturally when there are no Stunner limitations, I was able to cover anyway with cycling the model via the marshaller.

manstis: We can create Stunners graph programmatically to represent what would have been created manually. No problem. I tend to agree with Guilherme that we _could_ write unit tests to cover the code branches.

tarilabs: Cool ! 
So this is new since the original test case I built upon the current test base.
So I am missing the script/list of API calls which performs the click on the constellation button, along with the basic one to instrument in order to "click on Decision", "Drag here" etc.
Is there a test case I can copy, please?

tarilabs: (by "copy" I don't mean I expect a test case to be there already providing all I need, but something to take at least "inspiration" from ;) )

manstis: Heh, I didn't say there was a builder API or scripting facility... but what we can definitely build rather manually is the Stunner 'Graph' object with Nodes and Edges representing what can be constructed using the UI. Definitely a bunch of boiler plate but definitely doable.

manstis: I can help (on Monday) for sure.. I think @jomarko may have some tests already that go from 'Graph' to XML and back again.. I'll need to logon to my laptop.. on Monday!

tarilabs: :(
It was sounding too good to be true :(

So this confirms there is no realistic way to automate the manual test.

So this means to cover this branch I will have to re-create manually hand-crafted/hard-coded the same behavior of the null limitation as explained in the linked JIRA.

If it's really a necessity I can do, but I would suggest the title should have a very well written documentation (in-code) that once Stunner limitation is overcome, the test ought to be removed or might need again manual adjustment.

Please let me know if this code-coverage metric is a strict requirement, given is to cover an edge-case (of the Stunner limitation of the null Point2D)

manstis: ... I have 2 PRs waiting for this to be merged.. so it's in my interest to get this one in

tarilabs: ok

manstis: Is it an edge case?!?! It's easily produced from the UI so we have to expect users to create the scenario... and so should probably be tested.

tarilabs: > Is it an edge case?
On linked JIRA:
> Undefined i.e. null (IDK why, need to raise an issue with the Stunner people)
My understanding this is an issue with Stunner currently using the constellation button rather than using the palette. I have assumed then this will be solved from Stunner framework eventually to align with the palette behavior.
But I may have simply misunderstood the explaination then?

tarilabs: covered.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
tarilabs: covered.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
manstis: @tarilabs Thank-you for taking the time for the new test.

tarilabs: > for the new test.

No problem, thank you to you @manstis for explaining which strategy to pursue in order to complete code-coverage.

There is still some things which I don't understand about the framework, but I would suggest best to also discuss them at the first chance we have a call.

jomarko:
@tarilabs would you mind to use Logger instead of `System.out.println`? I am not going to do detail re-review. Check just few cases of diagrams where magnet points were used.

PR #2217, opened 2018-10-30
DROOLS-3111: [DMN Designer] Decision table generation fails if custom input data connected

manstis:
See https://issues.jboss.org/browse/DROOLS-3111

@jomarko [This](https://github.com/manstis/kie-wb-common/pull/35) change broke the fix that originally covered DROOLS-3111.. This PR fixes the fix :-)


jomarko:
@manstis sorry I broke things, starting review.

manstis:
@jomarko np.. we both verified your original PR to my repository... good it was spotted. 

No blame apportioned :-)

jomarko:
@manstis ok, so I am done with review, basically it works, however I think we have some inconsistency in extracting itemComponents for definition of columns. See my acceptance test notes in [DROOLS-311](https://issues.jboss.org/browse/DROOLS-3111). Not saying we can improve this, just wanted to discuss before approving.

PR #2233, opened 2018-11-01
DROOLS-3223 : Clear alert panel when a test run succeeds  

Rikkola:
https://issues.jboss.org/browse/DROOLS-3223


Rikkola:
@manstis I bet your the most popular person in the team today. If you could please merge this in.

PR #2246, opened 2018-11-06
DROOLS-3248: [DMN Designer] QName: Default to be 'Any' not 'string'

danielzhe:
DROOLS-3248: [DMN Designer] QName: Default to be 'Any' not 'string'

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/property/dmn/QName.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/property/dmn/QName.java
@@ -36,7 +36,7 @@ public class QName implements DMNProperty {
 
     public QName() {
         this(NULL_NS_URI,
-             BuiltInType.STRING.getName());
+             BuiltInType.ANY.getName());
     }
 
     public QName(final String namespaceURI,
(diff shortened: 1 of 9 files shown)


manstis:
@danielzhe It's probably _good practice_ to run `mvn clean install -Dgwt.compiler.skip=true` before submitting PRs to check all tests pass... and even `mvn clean install` (both on the root of `kie-wb-common-dmn`).

manstis:
@jomarko well that's unexpected!

@tarilabs @etirelli does your compiler need changes? I'd have thought changing the default typeRef used in the Workbench from one built in type to another would work flaelessly.

danielzhe:
@manstis My bad. I did it, but not in the root. I'm fixing the test.

danielzhe:
Jenkins please retest this

manstis:
@tarilabs @etirelli bump.. 

Strange the compiler complains about use of built in type "Any" and  Not "any".

tarilabs:
> Strange the compiler complains about use of built in type "Any" and Not "any".

Can I have please the DMN model file exhibiting this issue, or a line of code reference, please?
So I check exactly the error message, and if any bug.
Thanks!

danielzhe:
@tarilabs I'll send you one example to e-mail. Notice that in this PR we're changing the "any" to "Any", so maybe there is something more we need to do in order for this works.

manstis:
@tarilabs I found a DMN file with either `Any` or `any` gives the error reported by @jomarko 

I'm attaching three files (all the same other than the typeRef; `any`, `Any` and `string`). 

Only `string` compiles.

[DROOLS-3248.zip](https://github.com/kiegroup/kie-wb-common/files/2558987/DROOLS-3248.zip)

manstis:
Jenkins please retest this.

PR #2251, opened 2018-11-08
Reuse version.org.wildfly from kie-parent

jhrcek:
@desmax74 please check. We should not re-declare version that's already present on master. Also, having a (part of) version number in property name is not a good idea.


desmax74:
Good to fix the changes introducted with https://github.com/kiegroup/kie-wb-common/pull/2147

jhrcek:
@manstis could you merge, please?

manstis:
@desmax74 Could you please _approve_ this PR (rather than just +1 it). Thanks.

desmax74:
@jhrcek  @manstis  I'm not a reviewer to approve this issue

PR #2256, opened 2018-11-08
DROOLS-3249: [DMN Designer] Add support for optional typeRefs

danielzhe:
DROOLS-3249: [DMN Designer] Add support for optional typeRefs


kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/property/dmn/types/BuiltInType.java
```diff
@@ -32,7 +32,8 @@
     //Requested by Edson Tirelli
     ANY("any"),
     DATE("date"),
-    CONTEXT("context");
+    CONTEXT("context"),
+    UNDEFINED("<Undefined>");
```
jomarko: I am not sure, shouldn't we use `unknown` instead of `<Undefined>`? Asking because of [1].

[1]
https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/types/BuiltInType.java#L44

manstis: Good point.. let's defer to @etirelli as he asked the UI to show `<Undefined>` but obviously it's simple enough to change.

manstis:
Jenkins please retest this.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/property/dmn/types/BuiltInType.java
manstis: @jomarko @etirelli Thinking about this `unknown` appears to be an _alias_ for `Any` (as is `any`) which is quite different to `<Undefined>`.. this is when the `typeRef` is `null` vs the other proceeding values.

danielzhe: So, as I understand:
`unkown, Any, any` = empty
`<Undefined>` = null

manstis: I think it's more akin to:
- `unknown`, `any`, `Any` = `Object`
- `<Undefined>` = `null`

danielzhe:
@manstis Ok, should I change from `Any` to `<Undefined>` as default in this PR?

manstis:
@danielzhe Yes please.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/DataTypePickerWidget.java
```diff
@@ -62,7 +62,12 @@
 
     static final String CSS_DISPLAY_NONE = "none";
 
-    static final Comparator<BuiltInType> BUILT_IN_TYPE_COMPARATOR = Comparator.comparing(o -> o.getName());
+    static final Comparator<BuiltInType> BUILT_IN_TYPE_COMPARATOR = Comparator.comparing(o -> {
+        if (o == BuiltInType.UNDEFINED) {
```
manstis: Nice way to ensure `<Undefined>` always appears first!

PR #2268, opened 2018-11-12
[Outdated] AF-1453 DataModeller Editor Dynamic Dependency Management

desmax74:
Handled persistable nature with JPA, depends omn https://github.com/kiegroup/appformer/pull/541
@ederign @pefernan @domhanak 

**REPLACED** by: https://github.com/kiegroup/kie-wb-common/pull/2364


ederign:
@desmax74 we need three more things for this PR:

- on maven backend PR (AF-1452) hide the Dynamic dependency mechanism  behind an experimental feature (disable by default)
- test this PR with the UI and with an empty pom, create a new datamodeller object persistable and make sure that the jpa is added to the pom file.
- build the pom to check if JPA is the only required dependency for this.

PR #2274, opened 2018-11-14
JBPM-7956 Stunner - fix standalone showcase

tiagodolphine:
`BPMNGraphFactoryImpl` is causing an issue with dependency injection after the PR merge regarding the Case Management Wizard for BPMN.

Another fix were added regarding diagram creation on a _Non Project_  case, where the Id and Name were set to `null`, testing with the Standalone showcase.

@romartin @xiezhang7 
@LuboTerifaj


romartin:
Hey @tiagodolphine @hasys this is also failing on the 7.14.x branch, can we please cherry-pick this fix as well? Does that makes sense for you guys?

romartin:
Tested - merghing.

PR #2280, opened 2018-11-19
RHPAM-1681: Stunner - any change on the canvas scroll process to left-top corner in IE11

romartin:
Hey @jomarko 

Looking at the VCS history I see you created this stuff, so as I'm fixing a bug here, would like you to double check. 

The issue is about calling the `canvas.focus()` method on each `element` registration, in the `AbstractCanvasShortcutsControlImpl`. When opening an existing process in IE11 is causing some issues, as the canvas has not been yet well initialized. Also it's not correct because if you open a process with, let's say 100 elements, it will call `focus` also 100 times...

Anyway, probably this call is quite old as the canvas is being focused anyway when removing it. Please can you double check?

It depends on https://github.com/kiegroup/lienzo-core/pull/72

CC @hasys ^^

Thanks in advance!
Roger

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/keyboard/AbstractCanvasShortcutsControlImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/keyboard/AbstractCanvasShortcutsControlImpl.java
@@ -45,7 +45,6 @@ public abstract class AbstractCanvasShortcutsControlImpl extends AbstractCanvasH
 
     @Override
     public void register(Element element) {
-        focusCanvasToActivateKeyboardCanvasControl();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


romartin:

hasys:
I tested this PR and it looks and works good, thank you @romartin!

Just one more idea about possible fix: is it possible to combine 1 and 2 options? I mean disable append feature only for IE11 since this feature is not critical for users, and there are not so much users of IE11.

jomarko:
Just want to clarify, the append feature itself works, however changes in this ensemble causes it doesn't work immediately after element is dragged form palette to the canvas. After drag operation user has to re-select the element, regardless the browser.

romartin:
Hey @jomarko no worries at all! I'm just trying to fix it, and good catch! I didn't noticed that dragging from the palette was not focusing the canvas... so lemme try to fix this as well guys. Will let you know in a bit
Thanks!

romartin:
Hey @jomarko @hasys 

Just fixed the issue @jomarko mentions when dragging nodes from the palette. Now I think the code is more robust and clear - the idea is that the canvas is being focused once a node is being selected, which happens after dragging a new one from the palette or after using the quick menu (toolbox). This call also happens just once per request.

Please this is a blocker, can you review asap!? Thanks in advance!

romartin:

hasys:
Full downstream build fail looks unrelated, but will rerun it to be sure that we won't break build.

PR #2272, opened 2018-11-13
AF-1704: Case Management & Experimental integration

pefernan:
- Solved problems on Experimental Features Editor caused by using same names & descriptions
- API changes on Stunner Resource types to solved the issue with Case Management Resource Type, which wasn't well registered to the experimentals engine.

@romartin @hasys could you please take a look?

This is part of an ensemble, please merge with:
https://github.com/kiegroup/appformer/pull/527
https://github.com/kiegroup/kie-wb-common/pull/2272

PR #2283, opened 2018-11-19
RHPAM-1706: Stunner - Diagram of running instance doesn't highlight nodes

wmedvede:


wmedvede:
Hello @romartin @tiagobento @hasys ,
would you mind review this PR?

Note: as we spoke in the Stunner meeting the PR is a quick fix for letting BPMN work well. So I've basically added a simple extension to let BPMN work in the way it worked before, and CM work in with the code that was commented. So CM won't fail (in case it could fail because of the commented code) and BPMN will work as it did.

I've checked it in 7.14.x branch and the fix works well, i.e., BPMN is woking well.
Regarding CM I could see that the case management editor is not working in 7.14.x branch. Independently of the changes on this PR CM is not working in 7.14.x. As soon as CM file is create the following error is rasied "Activity not found" and the case management editor is not opened.

I've chatted with @xiezhang7  and it seems like the following PR https://github.com/kiegroup/kie-wb-common/pull/2272 fixes CM that in master. But I don't see this PR in 7.14.x, and not sure if it should/must be merged a the point we are now.

Anyway, my PR must still be merged to get the "highligted nodes issue fixed"

romartin:
Great thanks @wmedvede . Although the CM is broken for the next release, that's not a priority, so let's go ahead with this. Please @xiezhang7, as commented, try to fix this on master branch when possible.
Thanks!

romartin:

wmedvede:
Hi @hasys , it seems like we have the two greens now.
Would you mind take a look please since this is high  prioritary.

Thanks!

xiezhang7:
Thanks @wmedvede !
@romartin I think this pull request also fixes case modeler, should I cherry-pick the commit and create a pull request to master?

wmedvede:
Hi, @romartin @xiezhang7 , yes I think we should cherry-pick this into master to keep both branches synchronized but also given the comments/pointers we have there that highlights that a more polished solution is pending. I didn't it yet since given the hurries for the 7.14.x.
But it'l be much appreciated @xiezhang7 if you can manage to cherry-pick this into master.
Thanks!

wmedvede:
@xiezhang7 Ideally we should have it in before this Friday EOD, if you find any issue or can't take it please let me know and I can do it.

xiezhang7:
Thanks @wmedvede .

PR #2286, opened 2018-11-20
AF-1717: ProfilePreferences should not use a String constant

jesuino:
It breaks GWT superdevmode. Also, adding license headers.


manstis:
@Rikkola Fancy giving it a try?

Rikkola:
Will do

kie-wb-common-profile/kie-wb-common-profile-api/src/main/java/org/kie/workbench/common/profile/api/preferences/ProfilePreferences.java
```diff
@@ -29,7 +45,7 @@ public ProfilePreferences(Profile profile) {
 
     @Override
     public ProfilePreferences defaultValue(ProfilePreferences defaultValue) {
-        String profileStr = System.getProperty(INITIAL_PROFILE_PROPERTY, Profile.FULL.name());
+        String profileStr = System.getProperty("org.kie.workbench.profile", Profile.FULL.name());
```
mareknovotny: So if you use the string instead of the constant, are you going to fix/digg deeper why it is not working? as now constant is useless if you keep it there, at least adding a comment there is useful to save the memory, wdyt?

PR #2290, opened 2018-11-20
RHDM-797: Use other place request constructor

jomarko:
This PR changes used constructor to avoid using "[null]" as identifier of org.uberfire.mvp.impl.PathPlaceRequest
The identifier is used in meassge telling user what place can't be opened

@manstis 

Ensemble with kiegroup/appformer#539

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
@@ -662,7 +662,7 @@ public class LibraryPlaces implements WorkspaceProjectContextChangeHandler {
     }
 
     PathPlaceRequest createPathPlaceRequest(final Path path) {
-        return new PathPlaceRequest(path, Collections.emptyMap());
+        return new PathPlaceRequest(path, path.getFileName());
     }
 
     void closeLibraryPlaces() {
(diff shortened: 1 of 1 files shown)


manstis:
@jomarko Can you please explain how this fixes anything?

The constructor you use still [calls](https://github.com/kiegroup/appformer/blob/master/uberfire-api/src/main/java/org/uberfire/mvp/impl/PathPlaceRequest.java#L47) into the single parameter constructor giving the `PlaceRequest` and `identifier` of "[[null]](https://github.com/kiegroup/appformer/blob/master/uberfire-api/src/main/java/org/uberfire/mvp/impl/PathPlaceRequest.java#L41)"

jomarko:
@manstis Ok List of identifiers I can use form `LibraryPlaces` doesn't contain identifier of specific asset, in other words, I think we should either remove the label from `ActivityNotFoundView` or show there `path` instead of `editorId`. We are sending correct `edittorId` in `placeRequest` sent during asset creation. However asset creation is disabled if asset is disabled in workbench. Do you see my point?

jomarko:
@manstis ok, lets try filename

manstis:
@jomarko @ederign Thinking about this; it's probably better to remove the `[null]` label from the popup view. My reasoning being (a) the `identifier` is used inside `PlaceManager` (or IIRC it's `ActivityCache`) as a key in a map.. changing it to use a different key for each unhandled file *could* affect things we don't have tests for (do you want to take the risk!?!), (b) the User will know what the file is as they either uploaded it or clicked on it so showing the file name adds no value. WDYT?

jomarko:
@manstis I agree with every single word.

PR #2297, opened 2018-11-21
DROOLS-3334: [DMN Designer] Data Types - 'Structure' Data Types are being set as 'Undefined'

karreiro:
See https://issues.jboss.org/browse/DROOLS-3334

---

48725237-b5e43b00-ec12-11e8-91d1-e57085350435


karreiro:
@manstis I don't think we need to create any JIRA.. for the `7.14.x` branch we just reverted the commit that introduced the problem (https://github.com/kiegroup/kie-wb-common/pull/2282).

manstis:
@karreiro OK, I understand (it's just that https://issues.jboss.org/browse/DROOLS-3334 is marked as a "blocker" - that's why I asked.. perhaps change it to "Major"?)

karreiro:
@manstis Done ;)

PR #2302, opened 2018-11-22
[7.14.x] DROOLS-3067: Verify autolayout is applied when opening a diagram

jomarko:
Cherry-pick of #2301


jomarko:
@manstis would you mind to have  a look?

karreiro:
@jomarko @manstis I don't think that we need a PR for this in the `7.14.x` (the JIRA is not a blocker and the change is not crucial).

jomarko:
@karreiro Hi, sorry I can not agree this time. I see your point that jira is not blocker. However this PR adds just tests , that passed on jenkins, so there is no chance to introduce regression into next product build by this commit. 

Furthermore I need this test to formally mark https://issues.jboss.org/browse/BAPL-1087 as verified. In sign of for this jira I need to mention automated test that check if feature is invoked on asset opening. I need taht test on 7.14.x as the BAPL is targeted for 7.2 product.

Please let me know if my explanation persuaded you we can merge this also into 7.14.x.

PR #2303, opened 2018-11-22
DROOLS-3343 Refactor Message/MessageImpl to avoid rely on impl

tarilabs:
depends on https://github.com/kiegroup/drools/pull/2163


tarilabs:
Jenkins test again

PR #2305, opened 2018-11-22
[AF-1715-migration-import]: Added null checking condition for credentials

adrielparedes:
## Description

With @pefernan found that exist the possibility that if crytp:passw contains a null value when it tries to get credentials it fail. So if any credential value is null, credentials are not set.

## Related to:
* https://github.com/kiegroup/appformer/pull/545
* https://github.com/kiegroup/kie-wb-common/pull/2305

@pefernan @ederign would you mind to review?


pefernan:

pefernan:

PR #2319, opened 2018-11-30
[JBPM-7834] Provide OpenShiftStartupStrategy - WB Integration

rhtevan:
Providing a solution for WorkBench integration with KieServer OpenShiftStartupStrategy by implementing a new ServerTemplateOpenShiftStorage.

Related PR: https://github.com/kiegroup/droolsjbpm-integration/pull/1611


kiereleaseuser2:

kiereleaseuser2:

kiereleaseuser2:

kiereleaseuser2:

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/pom.xml
```diff
@@ -0,0 +1,159 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.workbench.screens</groupId>
+    <artifactId>kie-wb-common-server-ui</artifactId>
+    <version>7.15.0-SNAPSHOT</version>
```
mswiderski: we are already on 7.16.0-SNAPSHOT

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateConverter.java
```diff
@@ -0,0 +1,176 @@
...
+
+        template.setId(id);
+        template.setName(id);
+        template.addServerInstance(new ServerInstanceKey(id, id, id, url));
+
+        for (KieContainerResource conRes : state.getContainers()) {
+            Map<Capability, ContainerConfig> configs = new EnumMap<>(Capability.class);
+
+            KieScannerResource scanner = conRes.getScanner();
+            if (scanner != null)
```
mswiderski: could you please wrap the if content with {}

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateConverter.java
```diff
@@ -0,0 +1,176 @@
...
+                    if (KieServerConstants.PCFG_KIE_SESSION.equals(item.getName()))
+                        pcfg.setKSession(item.getValue());
+                    if (KieServerConstants.PCFG_MERGE_MODE.equals(item.getName()))
+                        pcfg.setMergeMode(item.getValue());
+                    if (KieServerConstants.PCFG_RUNTIME_STRATEGY.equals(item.getName()))
+                        pcfg.setRuntimeStrategy(item.getValue());
+                }
+            }
+            if (pcfg != null)
+                configs.put(Capability.PROCESS, pcfg);
```
mswiderski: same here wrap the if content with {}

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateConverter.java
```diff
@@ -0,0 +1,176 @@
...
+    public static KieServerState toState(@NotNull ServerTemplate template) {
+        String id = template.getId();
+        KieServerState state = new KieServerState();
+        KieServerConfig config = new KieServerConfig();
+        Set<KieContainerResource> containers = new HashSet<>();
+
+        config.addConfigItem(new KieServerConfigItem(KieServerConstants.KIE_SERVER_ID, id, String.class.getName()));
+
+        if (template.getServerInstance(id) != null)
+            config.addConfigItem(new KieServerConfigItem(KieServerConstants.KIE_SERVER_LOCATION,
```
mswiderski: wrap if content in {}

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
```diff
@@ -0,0 +1,103 @@
...
+    public ServerTemplateOpenShiftStorage() {}
+
+    @Inject
+    public ServerTemplateOpenShiftStorage(KieServerStateCloudRepository repo) {
+        this.repo = repo;
+    }
+
+    @Override
+    public ServerTemplate store(ServerTemplate serverTemplate) {
+        if (exists(serverTemplate.getId()))
```
mswiderski: wrap if content with {}

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
```diff
@@ -0,0 +1,103 @@
...
+            throw new IllegalArgumentException("Server template with id " + serverTemplate.getId() + " is already stored");
+        repo.create(ServerTemplateConverter.toState(serverTemplate));
+        return ServerTemplateConverter.fromState(repo.load(serverTemplate.getId()));
```
mswiderski: why not returning same template instance instead of doing toState and fromState?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public List<ServerTemplateKey> loadKeys() {
+        return repo.retrieveAllKieServerIds()
```
mswiderski: since templates are loaded quite often to be shown in UI maybe it would make sense to implement some local cache of the templates to make it more efficient?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public boolean exists(String identifier) {
+        return repo.exists(identifier);
+    }
+
+    @Override
+    public ServerTemplate update(ServerTemplate serverTemplate) {
+        // No implementation needed for OpenShift based storage
```
mswiderski: why we don't need implementation here? Update means adding or removing KIE Containers (kjars)...

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
```diff
@@ -0,0 +1,103 @@
...
+        return serverTemplate;
+    }
+
+    @Override
+    public ServerTemplate delete(String identifier) {
+        // Can delete only detached CM, meaning not having an associated DC
```
mswiderski: can you explain what do you mean by this comment? Will it then stop all pods with that config map?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateConverter.java
```diff
@@ -0,0 +1,176 @@
...
+
+    public static KieServerState toState(@NotNull ServerTemplate template) {
+        String id = template.getId();
+        KieServerState state = new KieServerState();
+        KieServerConfig config = new KieServerConfig();
+        Set<KieContainerResource> containers = new HashSet<>();
+
+        config.addConfigItem(new KieServerConfigItem(KieServerConstants.KIE_SERVER_ID, id, String.class.getName()));
+
+        if (template.getServerInstance(id) != null)
```
mswiderski: template can have many server instances attached to it, do we consider that for OpenShift it will always be just one - all pods hidden behind Openshift service/router?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
```diff
@@ -0,0 +1,103 @@
...
+import org.kie.server.services.openshift.impl.storage.cloud.KieServerStateCloudRepository;
+import org.kie.workbench.common.screens.server.management.backend.utils.EmbeddedController;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ApplicationScoped
+@EmbeddedController
+@Alternative
+@Priority(100)
+public class ServerTemplateOpenShiftStorage implements KieServerTemplateStorage {
```
mswiderski: I think we should move the bulk of the implementation to droolsjbpm-integration/kie-server-parent/kie-server-controller module so it can be used both in workbench and standalone controller. Although in droolsjbpm-integration we should not use CDI and the kie-wb-common specific interfaces. Instead that should be in this repo extending the one coming from Kie Server controller modules. Does it make sense?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
rhtevan: Researched the codebase a bit, found out that the only references is from KieServerController connect and disconnect. Since OpenShift-KieServer does not connect to Controller, implementation to 'update' method appears to be unnecessary. In addition, any updates to attached KieServerConfig (Meaning that it is used by KieServer Pod) should go through KieServerService API. 

https://github.com/rhtevan/droolsjbpm-integration/blob/365ddd2f858637c7c026b16c2279602ceed29044/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java#L57-L81

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
rhtevan: To deal with 'attached' ServerTemplate (meaning that the KieServerConfig has been used/attached with running KieServer Pod), the implementation should be read-only and retrieving KieServeConfig (aka KieServerState ConfigMap) via OpenShift API. The write/update channel should go through KieServeService API. 

The implementation of 'store' and 'delete' are targeted to manage new/detached ServerTemplate.

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateConverter.java
rhtevan: All KieServer instance Pods are virtualized behind OpenShift route/service. Under such circumstance, KieServerState is equivalent to ServerTemplate with only 1 server instance which is represented by the virtualized KieServerService endpoint.

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
rhtevan: IMHO, this implementation is to provide an adapter for Workbench to retrieve KieServerState data  from OpenShift API server, and most importantly to have the virtualized KieServerService endpoint (aka KIE_SERVER_LOCATION) so as to go through existing channel to manage KieServer instances. 

Not quite sure how standalone controller would benefit from it. The OpenShift-ified KieServer instances are suppose to be managed unanimously via the virtualized KieServerService API. What could be the additional capabilities offered by standalone controller in such case? Regardless how many KieServer instances they will be managed as if there were just one KieServer instance running.

cristianonicolai: I agree with @mswiderski, ideally, this should be shared with the standalone controller. It ensures that we offer flexibility in the deployment/architecture.  A standalone controller can be used independently ( no business-central at all ) or shared/scaled in a different model. Which can still be valuable to have the Openshift storage.

mswiderski: @rhtevan the point of standalone controller is similar to what workbench is offering but much lighter and without UI so it can be an easy way to get it included in custom admin apps. So it has the same capabilities and thus could be really good fit for OpenShift deployments.

Another thing is that it can be easier to just rely on controller api (either REST or UI) instead of OpenShift itself as it bring bit more context from business automation domain while OpenShift API is rather low level and might not be a good fit for process admins.

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateConverter.java
mswiderski: makes sense, thanks

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
mswiderski: >  The write/update channel should go through KieServeService API.

I would say we should have the controller to be also responsible for updating/deleting server templates (or its containers). That's the whole point of controller that it gives you unified api to manage your KIE Servers - regardless if they are deployed on OpenShift or wherever...

Frankly speaking I don't see any difference when it comes to updating config map for given server template - is it done by controller or by Kie Server (one of pods). In my opinion when we use OpenShift with config map as the configuration then REST api should be actually disabled to manipulate state and use either controller that will make the changes or do it via openshift tools.

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
mswiderski: isn't that adding or removing kie containers will make the server template to be updated? @cristianonicolai wdyt?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
mswiderski: +1

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
rhtevan: Correct me if I'm wrong. Currently, KieServerController **_delegates_** its KieContainer management capability to KieServer via KieServer REST API. (https://github.com/rhtevan/droolsjbpm-integration/blob/95212c25d32b4a93b127f360a526c98dab501f35/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java#L152-L157)

_If KieServer REST API were disabled, KieServerController might not be able to manage KieContainers._

That's what I meant in terms of 'write channel'. Once a KieServerConfig (aka ConfigMap) being attached to KieServer Instances/Pods, it is better to be manipulated by only one 'channel'. That's the reason why we have limited to read-only implementation to 'Attached KieServerConfig' at ServerTemplateOpenShiftStorage. (STOSS)

The term of 'Detached KieServerConfig' used here is more aligned with current 'ServerTemplate' concept. STOSS is allowed to 'delete' and 'create' it. Do we want to extend such capability to say, 'Delete' - Tearing down running KieServer instances/pods; 'Create' - Creating not only KieServerConfig but also spin up KieServer Pods/Instances?

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
rhtevan: Implemented

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
cristianonicolai: correct, for instance _SpecManagementServiceImpl.updateContainerConfig_ uses the _update_ method. I would agree that it needs to be implemented.

rhtevan: Implemented

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend-openshift/src/main/java/org/kie/workbench/common/screens/server/management/backend/openshift/storage/ServerTemplateOpenShiftStorage.java
rhtevan: @mswiderski which module you'd like to move it into, or another new OpenShift module?

mswiderski: A nie module here https://github.com/kiegroup/droolsjbpm-integration/tree/master/kie-server-parent/kie-server-controller

rhtevan: Created a new module and associated PR (https://github.com/kiegroup/droolsjbpm-integration/pull/1670)
For this repo (kie-wb-common), we have only a slight modification to current test case since we do need to have the new ServerTemplateOpenShiftStorage class passing current test cases whichever it is still relevant. 

@mswiderski What would be the best option to package OpenShiftStorage into Workbench and swap out ServerTemplateVFSStorage?

mswiderski: @rhtevan I was thinking to make use of this http://docs.jboss.org/weld/reference/1.1.16.Final/en-US/html_single/#d0e5774 so we can easily exclude the ServerTemplateVFSStorage and then enhance the producers in kie-wb-common to look up the OpenShift one somehow. Maybe create new interface in kie-controller-api module that we can implement on the openshift controller to allow easy injection 
@cristianonicolai you are more into details of the embedded vs standalone controller setup here, what do you think? We can't do much with CDI in droolsjbpm-integration module itself to not make it too dependent on cdi api...

cristianonicolai: This storage probably needs to be defined at the _KieServerEmbeddedControllerProducer_ in there you could switch depending on system properties, for instance, to decide which implementation to use.

rhtevan:
@mswiderski @cristianonicolai @sutaakar 
Implemented a solution to allow dynamically switching between ServerTemplateVFSStorage and ServerTemplateOCPStorage by system property. 

Observed few dependency and packaging related issues:

1. Fabric8 kubernetes client classes are packaged within kie-wb-common-ala-openshift-client module (https://github.com/rhtevan/kie-wb-common/blob/c86fc213f84e0915b263305b37951e1af2ab63c6/kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-client/pom.xml#L129-L176); it causes either duplicates or missing classes. 
2. New dependencies being introduced into kie-wb-common module (i.e. kie-server-controller-openshift --> kie-server-service-openshift --> kie-server-service-common --> Bootstrap, KieServerImpl) result in subtle unexpected behavior. 

Any suggestions? (For example, is it possible to move Bootstrap class into a new module?)

mswiderski:
> Fabric8 kubernetes client classes are packaged within kie-wb-common-ala-openshift-client module (https://github.com/rhtevan/kie-wb-common/blob/c86fc213f84e0915b263305b37951e1af2ab63c6/kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-client/pom.xml#L129-L176); it causes either duplicates or missing classes.

I'd ask for @porcelli support here. Not sure that ala is in use so maybe an option to upgrade the version of Kubernetes client if it's not

> New dependencies being introduced into kie-wb-common module (i.e. kie-server-controller-openshift --> kie-server-service-openshift --> kie-server-service-common --> Bootstrap, KieServerImpl) result in subtle unexpected behavior.

what is this unexpected behaviour?

rhtevan:
@mswiderski The KieServerStateOpenShiftRepository has some dependencies to kie-server-services-common module, such as KieServerState

https://github.com/rhtevan/droolsjbpm-integration/blob/91eb60a59e554e8a255ce948b0c6dd69505599de/kie-server-parent/kie-server-services/kie-server-services-openshift/src/main/java/org/kie/server/services/openshift/impl/storage/cloud/KieServerStateOpenShiftRepository.java#L36-L37

mswiderski:
@rhtevan ok, we have two options
- move the KieServerState and the utils to kie-server-api
- move Bootstrap to https://github.com/rhtevan/droolsjbpm-integration/tree/91eb60a59e554e8a255ce948b0c6dd69505599de/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common

since bootstrap is actually web layer thingy anyway might be worth to move this to rest-common actually, wdyt?

rhtevan:
@mswiderski Tends to agree with the second option. Make sense to move this, Bootstrap, web tier thingy, away from kie-server-service-common module. 
Is it ok to introduce such change into this PR (https://github.com/kiegroup/droolsjbpm-integration/pull/1670)? Thanks!

mswiderski:
@rhtevan yes, go ahead and update that PR

rhtevan:
@mswiderski Apparently by only moving Bootstrap away is not sufficient. As the singleton nature of KieServerLocator, KieServerImpl will be initialized anyway once the class being loaded. We have to done something like this.  

https://github.com/rhtevan/droolsjbpm-integration/blob/e52ef3f6de52a1ae5b0ed9392c429e617a635d1f/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerLocator.java#L20-L33

porcelli:
f8 dependency.. as mentioned this is for old guvnor ala, and as pointed out by @errantepiphany in a different thread, this is unlikely to be working.

in a point in time would be nice to have the provisioning mechanism available back in BC (the current UI is there, just doesn't work =/)

mswiderski:
@porcelli so if it does not work then there is nothing bad in upgrading as long as it compiles isn't it? Just trying to not hold back features because there is some implementation that is not working but requires some old version ... if I got it right

porcelli:
@mswiderski sure, go ahead @rhtevan !

rhtevan:
@porcelli @mswiderski @sutaakar @errantepiphany 
QE has confirmed that there is no compilation issue after upgrade f8 k8s client. Actually, the client lib version has already been upgraded (https://github.com/rhtevan/droolsjbpm-build-bootstrap/blob/081ccba2b14f84ae4c08dd5e3daf1c6898fe6aa4/pom.xml#L232) by previously merged PR (https://github.com/kiegroup/droolsjbpm-integration/pull/1611). 

The issue is that the shade plugin from ala-openshift-client module (https://github.com/rhtevan/kie-wb-common/blob/f20f9177462a9d6e3e951333d26b6f236b306d70/kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-client/pom.xml#L128-L176) packages duplicate and exclude classes. It results in issues, i.e. 'NoClassDefFound' exception. 

Can that module be excluded temporary from WB WAR for now?

mswiderski:
@rhtevan I’d say go ahead or remove the shade plugin. @porcelli any objections?

porcelli:
if it doesn't break what we have working: no objections :)

rhtevan:
Removed the shade plugin.

rhtevan:
@errantepiphany 

https://github.com/rhtevan/kie-wb-common/blob/56296ab892e5fa62ce4eb8fc31963a9c851cf87e/kie-wb-common-ala/pom.xml#L23-L25

rhtevan:
Fixing the dependency issue for integration test case.

https://github.com/rhtevan/kie-wb-common/blob/56296ab892e5fa62ce4eb8fc31963a9c851cf87e/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/test/java/org/kie/workbench/common/screens/server/management/backend/AbstractControllerIT.java#L109

rhtevan:
@sutaakar rebased with master

rhtevan:
Dependent PRs:
- https://github.com/kiegroup/droolsjbpm-integration/pull/1670

Associated PRs:
- https://github.com/kiegroup/droolsjbpm-integration/pull/1705

kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-provider/src/main/java/org/guvnor/ala/openshift/access/OpenShiftClient.java
```diff
@@ -153,75 +151,89 @@ private void createProject(String prjName) {
...
-            .endRoleBinding()
-            .done();
+        /**
+         * Commented out the implementation due the dependent fabric8 client class, DoneablePolicyBinding,
+         * has been removed from the lib API after upgrade from 4.1.0 to 4.1.1.
+         * Since ala-openshift module does not compatible with OpenShift 3.x, this quick fix is to
+         * make it compile only.
+         * Related JIRA JBPM-7834: Provide OpenShiftStartupStrategy - WB Integration
+         */
+//        Resource<PolicyBinding, DoneablePolicyBinding> bindingResource =
```
porcelli: please just delete, we don't need the commented code

kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-provider/src/main/java/org/guvnor/ala/openshift/access/OpenShiftClient.java
```diff
@@ -153,75 +151,89 @@ private void createProject(String prjName) {
...
-            .endRoleBinding()
-            .endRoleBinding()
-            .done();
+        /**
+         * Commented out the implementation due the dependent fabric8 client class, DoneablePolicyBinding,
+         * has been removed from the lib API after upgrade from 4.1.0 to 4.1.1.
+         * Since ala-openshift module does not compatible with OpenShift 3.x, this quick fix is to
+         * make it compile only.
+         * Related JIRA JBPM-7834: Provide OpenShiftStartupStrategy - WB Integration
+         */
```
porcelli: please just delete, we don't need the commented code

kie-wb-common-ala/pom.xml
```diff
@@ -19,7 +19,7 @@
     <module>kie-wb-common-ala-services-rest</module>
     <module>kie-wb-common-ala-ui</module>
     <module>kie-wb-common-ala-docker-provider</module>
-    <module>kie-wb-common-ala-wildfly</module>
+    <module>kie-wb-common-ala-wildfly</module>
```
porcelli: please remove this change to avoid pollute the pr

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/utils/ControllerExtension.java
```diff
@@ -33,18 +35,40 @@
...
-        if (a != null && ControllerUtils.useEmbeddedController() == false) {
-            LOGGER.info("Removing embedded controller class {} from CDI context", event.getAnnotatedType().getJavaClass());
-            event.veto();
+        if (a != null) {
+            if (ControllerUtils.useEmbeddedController()) {
+                /**
+                 * Select server template storage implementation for embedded controller
+                 * based on system properties loaded by ControllerUtils
+                 */
+                if (ControllerUtils.isOpenShiftSupported()) {
```
porcelli: missing test for this

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/storage/ServerTemplateOCPStorage.java
```diff
@@ -0,0 +1,80 @@
...
+import org.kie.server.controller.api.model.spec.ServerTemplateKey;
+import org.kie.server.controller.api.storage.KieServerTemplateStorage;
+import org.kie.server.controller.openshift.storage.OpenShiftServerTemplateStorage;
+import org.kie.workbench.common.screens.server.management.backend.utils.EmbeddedController;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ApplicationScoped
+@EmbeddedController
+public class ServerTemplateOCPStorage implements KieServerTemplateStorage {
```
porcelli: please unit test for this

rhtevan: The unit test class is located here

https://github.com/rhtevan/droolsjbpm-integration/blob/e64a0e513fa311dcb1965b18aedbdd03396f356e/kie-server-parent/kie-server-controller/kie-server-controller-openshift/src/test/java/org/kie/server/controller/openshift/storage/OpenShiftServerTemplateStorageTest.java#L40

kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-provider/src/main/java/org/guvnor/ala/openshift/access/OpenShiftClient.java
rhtevan: Updated as per suggestion

kie-wb-common-ala/kie-wb-common-ala-openshift/kie-wb-common-ala-openshift-provider/src/main/java/org/guvnor/ala/openshift/access/OpenShiftClient.java
rhtevan: Updated as per suggestion

kie-wb-common-ala/pom.xml
rhtevan: Updated as per suggestion

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/main/java/org/kie/workbench/common/screens/server/management/backend/utils/ControllerExtension.java
rhtevan: Added new test, https://github.com/rhtevan/kie-wb-common/blob/3749ab02550a38214a99b602f3f7ca01880f4140/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/test/java/org/kie/workbench/common/screens/server/management/backend/utils/ControllerExtensionTest.java#L79-L85

kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-backend/src/test/java/org/kie/workbench/common/screens/server/management/backend/utils/ControllerExtensionTest.java
```diff
@@ -61,6 +64,17 @@ public void testEmbeddedAnnotationWithStandaloneController() {
         verify(annotatedType).veto();
     }
 
+    @Test
+    public void testEmbeddedAnnotationWithServerTemplateOCPStorage() {
+        System.setProperty(CFG_KIE_CONTROLLER_OCP_ENABLED, "true");
```
porcelli: this is fragile for jenkins... could you make this code more flexible?

what we use to do is create in the bean a constructor that get a map as parameter, so you manually instantiate with the value you want inside the map. and the original constructor, get the System.getProperties() as parameter to the next controller.

so we avoid jenkins failures.

rhtevan: Revised implementation and test case as per suggestion.

rhtevan: @porcelli  Alex, see if everything is ok from your side?
@sutaakar Karel, QE feedback is required for Alex to go ahead with merge.

mswiderski:
Jenkins retest this

PR #2320, opened 2018-11-30
Remove test of non existing class

jomarko:
Randomly spotted this obsolete class, @manstis would you mind to have a look?


manstis:
Waiting for Jenkins to be green before merging... just in case something else extends it!

jomarko:
@manstis sure think to wait for the green build! Thank you.

jomarko:
@manstis please merge.

PR #2321, opened 2018-11-30
AF-1727: Case Management NewResourceHandler wrongly visible with PLANNER_AND_RULES profile

jesuino:
Moved the filter where it should be. It solves 1727 and also the duplicated NewResourceHandler for Case management legacy.


jesuino:
@ederign tests were added in https://github.com/kiegroup/kie-wb-common/pull/2321/commits/48ed5b638b2a0b0453c59adf30a03d2d00882cbe

ederign:
@jhrcek can you please take a look? (if you are not the right person, could you please assign it to someone else?)

PR #2327, opened 2018-12-04
DROOLS-2932: Make edit selected cell logic general

jomarko:
@manstis @gitgabrio This contains changes from previous ensembles and the last commit make the edit selected cell action general.


jomarko:
@gitgabrio please let us know if you are going to have a look or it is fine to merge without your approval.

gitgabrio:
@jomarko 
I'm reviewing

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/KeyboardOperationEditGridCell.java
```diff
@@ -36,15 +34,10 @@ public KeyboardOperationEditGridCell(final GridLayer gridLayer) {
...
-            // just one data cell can be selected before edit operation is invoked
-            return;
-        }
-
-        super.editCell(gridWidget);
+    public boolean perform(final GridWidget gridWidget, final boolean isShiftKeyDown, final boolean isControlKeyDown) {
+        final boolean needToRedraw = super.perform(gridWidget, isShiftKeyDown, isControlKeyDown);
 
+        final GridData model = gridWidget.getModel();
         if (model.getSelectedCells().size() == 1) {
```
gitgabrio: @jomarko @manstis @ederign 
This looks bad, IMHO. It implicitly means that you can have more then one selected cell while editing one, that is mostly counter-intuitive from UI point of view. 
Moreover, it also means that if more then one cell is selected, the code inside the "if" is never invoked.
*If* it is impossible (for reason outside the current snippet) to have more then one selected cell during editing, or if *it should not* be possible, please explicitly block the flow at the very beginning of this method, otherwise fix the implementation.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/KeyboardOperationEditGridCell.java
```diff
@@ -36,15 +34,10 @@ public KeyboardOperationEditGridCell(final GridLayer gridLayer) {
...
-    protected void editCell(final GridWidget gridWidget) {
-        final GridData model = gridWidget.getModel();
-        if (model.getSelectedCells().size() > 1) {
-            // just one data cell can be selected before edit operation is invoked
-            return;
-        }
-
-        super.editCell(gridWidget);
+    public boolean perform(final GridWidget gridWidget, final boolean isShiftKeyDown, final boolean isControlKeyDown) {
+        final boolean needToRedraw = super.perform(gridWidget, isShiftKeyDown, isControlKeyDown);
```
gitgabrio: @jomarko @manstis @ederign 
The name "needToRedraw" is misleading. Method's and variables names are important, and there are rule to follow: the code we are writing is meant to be read and understood by others.
A method named "perform" and returning a "boolean" is expected to return "true" if the "perform" is successful, "false" otherwise.
"needToRedraw" implicitily define an action to be performed "after", but without reading all the code around (the (super.perform implementation and the methods calling the current one) is impossible to understand why redrawing is needed. 
Wording differently, "redrawing" is a choice/responsibility of the "caller" code, not of the method implementation. 
Since the variable is the result of a "perform" method, it should be named "performed" (I personally prefer "toReturn" for the variables that have to be returned by methods, so it is clear at first sight the meaning/scope of the variable, but it is my personal taste).
Knowing to be pedantic, I think this are the details that makes a code easily  readable and manageable in the long term.

jomarko: Hi @gitgabrio,

I named the variable according to [javadoc](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/KeyboardOperation.java#L49) of the method.

However am fine to improve the name. Renaming to `changesToBerendered`. 

Will push as separate commit. Thank you for your comment.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/KeyboardOperationEditGridCell.java
jomarko: Hi @gitgabrio,

The "selected cell count check" is valid here due to [BaseGridWidgetKeyboardHandler](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidgetKeyboardHandler.java)

This `BaseGridWidgetKeyboardHandler` recognizes pressed key and calls bound operation. This is the single responsibility of this class. This handler can not check number of selected cells , because it can be specific for each operation. Lets say it makes sense allow select multiple cells for erase content operation on the other hand it makes sense to require just one cell for editing. That is why I check count of selected cells here.

I can not move the block to appformer module, because here I am handling dmn specific cell.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/grid/columns/KeyboardOperationEditGridCell.java
gitgabrio: @jomarko 
Ok, thanks! I'll take the chance for a personal observation - it is just my opinion, and I'm not asking nothing from your side and I know it does not depends on you. The issue you faced is that the signature of the method ("perform") seems to tell you that an operation has been (successfully) executed or not, but reading the javadoc (as you did) you discover that there is a sort of "side-effect" to invoke. Again, IMHO, this is an inconsistence between the name of the method (that it is not really clear, IMHO) and the scope of it, and one has to read javadoc (or, sometimes worse, actual implementation) to fully understand it, that is a lot of waste of time (and really error prone).

PR #2330, opened 2018-12-05
DROOLS-3395: [DMN Designer] Two copies of node are created

manstis:
See https://issues.jboss.org/browse/DROOLS-3395

This PR simply moves registration of commands to a method that is overridden by the DMN Editor.

This prevents Errai from invoking the `@PostConstruct` method twice.


manstis:
Jenkins please retest this.

PR #2333, opened 2018-12-05
DROOLS-2883: [DMN Designer] Add a Search bar in the Data Type dialog for search top level custom Data Types / DROOLS-3030: [DMN Designer] Expand / Collapse all

karreiro:
See:
- https://issues.jboss.org/browse/DROOLS-2883
- https://issues.jboss.org/browse/DROOLS-3030

---

**DROOLS-2883:**
search

**DROOLS-3030:**
expand-collapse


manstis:
Jenkins please retest this (now the BPMN2 file indexer test is fixed)

jomarko:
@karreiro postponing my review until you address Michael's comments to avoid spotting the same.

karreiro:
Thank you, @manstis.

Regarding the issue when the user adds a new Data Type. I'm disabling the search in this scenario. When users are filtering data types, they are looking for something, and when they are creating a new top-level DT, they don't need to keep the old search.

The bug related to the "Remove" operation is solved now too. The search bar exposed an existing issue, the [DROOLS-3329](https://issues.jboss.org/browse/DROOLS-3329). It's fixed now.

Finally, the NPE with expand/collapse was related to the DROOLS-3329 too.

2018-12-11 11 42 38 am

Thus, changes applied :-) Thanks!

---

@jomarko Feel free to take a look now ;-)

jomarko:
@karreiro HI, found one minor issue on edge [DROOLS-3426](https://issues.jboss.org/browse/DROOLS-3426). And here are some comments with questions, proposals, let me know if I should file jiras for them.
- Expand/Collapse doesn't execute action if filtering is active, could we disable (grey color) them when filtering is active
- The allowed edit actions when filtering is active. It is tightly connected with [DROOLS-3015](https://issues.jboss.org/browse/DROOLS-3015). I see filtering as the way for finding a data type that I want to update in some way. The problem is when I find a data type and update it the result is visible to me just if it mathes the current filter. I think we need to define in more advance level what should happen when user filter for some data type and starts edit it, let me know what do you think.

Otherwise works fine. Tested manually, still need to check code.

jomarko:
Code looks fine

karreiro:
@manstis Thank you ;-)

---

@jomarko Thank you!

Regarding the "Expand/Collapse" improvement, I think we can handle this in a separated JIRA.

Also, about the filter behavior (we the user updates a DT), personally I think that the current operation looks like something I expect as a user. But, we can create a UX JIRA to have a broader discussion there about it (but, for now, I think that we're fine).

PR #2306, opened 2018-11-23
JBPM-7852 Stunner - Nested sub-process is not created well and validation fails

tiagodolphine:
- Rules Evaluation on connection with nested sub-process
- Fixed position when creating nodes with the quick menu on nested sub-process (thanks to @romartin )
- Fixed drag/drop elements from the palette on nested sub-process including containment and dock.
- Fix showcase that is broken, just removing the ShowcaseBPMNDefinitionSetResourceType.java

@romartin @LuboTerifaj

PR #2339, opened 2018-12-07
JBPM-8039 - It is not possible to connect nodes inside of swimlane

hasys:
Hi @romartin, @tiagodolphine, @LuboTerifaj,

I found a regression caused by https://github.com/kiegroup/kie-wb-common/pull/2306

This is a fix and fixed tests.

In short new Embedded subprocess bounds exceeded tests was extended by https://github.com/kiegroup/kie-wb-common/pull/2306 for all types of sub-processes and BPMNDiagram, but Lane were forgotten.

But it is not possible just add Lane to this rule, because it can be a situation when you want to connect nodes from top level process with nodes inside of Swimlane and it is a valid case. Another valid case is connecting two nodes from different Swimlanes.

So, I removed BMPNDiagram from checked parents and extended rule to ignore nodes without checked parent as it was before https://github.com/kiegroup/kie-wb-common/pull/2306.

Thank you!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/ParentsTypeMatchPredicate.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/ParentsTypeMatchPredicate.java
@@ -66,6 +66,10 @@ class ParentsTypeMatchPredicate implements BiPredicate<Node<? extends View<?>, ?
         final Optional<Element<?>> parentInstanceA = getParentInstance(nodeA, parentType);
         final Optional<Element<?>> parentInstanceB = getParentInstance(nodeB, parentType);
 
+        if (!parentInstanceA.isPresent() && !parentInstanceB.isPresent()) {
+            return true;
+        }
+
         return (parentInstanceA.isPresent() && parentInstanceB.isPresent())
                 && (Objects.equals(parentInstanceA.get(), parentInstanceB.get()))
                 && (hasParent(nodeA, parentInstanceA.get()) && hasParent(nodeB, parentInstanceB.get()));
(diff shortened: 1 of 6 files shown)


LuboTerifaj:

tiagodolphine:

hasys:
Hi @tiagodolphine, @romartin, just rebased my commit to master to fix inconsistency with master branche

Jenkins retest please

romartin:
ok thanks @tiagodolphine , waiting for your feedback

hasys:

LuboTerifaj:

hasys:
Hi @tiagodolphine,

war files are generated successfully and you can test it now, thank you!

tiagodolphine:
Hi @hasys sure, I'll do that.
Thanks.

tiagodolphine:
Hi @hasys I when a new connection is created from a node from the top level process or between swimlanes it is ok. I got one situation that is not working, if you have a sequence flow already created and try to move on of the nodes outside the swimlane it is not allowed, probably because of some rule.
thanks.

hasys:
Hi @tiagodolphine, @LuboTerifaj,

thank you, good catch Tiago! But I, probably, will vote for moving your founds to separate issue as Lubo suggested. Because this issue looks little bit more complicated.

Issue caused by this check: https://github.com/kiegroup/kie-wb-common/blob/d8e87770ca97b5cbc5854837355b236f661f8a56/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/FilteredParentsTypeMatcher.java#L86

When we moving nodes, Stunner check is parent of the moving node is the same as before, and this is not the same mechanism as it is for Sequence flows. In short it always checks parent (so rule with listed checked parents we have for sequence flows is not applied here).

For now this module is Abstracted and work not with BPMN graph but with some abstracted Stunner graph, so we have no clue is it Diagram or Lane, we just know is it the same Parent or not.

So solution can be:
* Add dependency to BPMN API and check parent node is it a Lane or a BPMNDiagramImpl and ignore if needed. But this approach is breaking abstraction of the module and we will work directly with BPMN nodes, not sure but probably it is not desirable.
* Another solution can be a some refactoring and moving this rule from stunner-core to BPMN Designer space.
* Probably, also can be a solution to remove this check at all and rely on Sequence flow restrictions, but it should be tested how possible it is (also it can cause some issues for other editors).

So, I suggest to merge this PR as it is simple solution and fixes core issue, and move your finding to another Jira issue and handle it separately.

P.S. Also, I found another issue: if you have Start->Task->End and select all nodes, you can't move all of them out of the parent container due to same rule. (but, probably, it is even more complicated case because every single node tested separately and we will need to identify is entire graph is moving and not a random group of nodes).

CC @romartin

tiagodolphine:
@hasys yeah I agree we can work on a new JIRA for it. 
And about the last issue, yeah this is complicated because we evaluate the rules separately as you said, I think @romartin already opened I JIRA for that, once I investigated how to solve, but the solution will be complicated, anyway, we need to fix this at some point when it becomes a priority.
Thanks !

hasys:
Hi @romartin,

I moved discussed issue to separate Jira: [JBPM-8069](https://issues.jboss.org/browse/JBPM-8069) - it is not possible to move in/out of swimlane connected node.

Test fails are not related to this issue fix, so I guess this PR can be merged, but I will try rerun build with "Green" hope :)

Thank you!

hasys:

PR #2341, opened 2018-12-11
[7.14.x] RHPAM-1773: Can't attach Boundary Events to Service Tasks in Stunner.

romartin:
Hey @hasys 

This is the fix for the blocker issue - https://issues.jboss.org/browse/RHPAM-1773

It fixes the docking issues with Service Tasks, now it's possible to do it as well as with other tasks.

PS: The change/fix is really small, but it was necessary to refactor the return type for method `DefinitionAdapter#getId`, from `Strung` to `DefintionId`,which is a new type that allows "dynamic" ids (eg: service tasks). So this is while a few files are changed but due to the refactor on the API.

Thanks!


romartin:

romartin:
Please @hasys , this one is a blocker and next week I'll be on PTO, so I would like to leave in peace with it... can you please give priority to this one?? I don't want last time worries... :) 
Thanks in advance!

romartin:
ping @hasys

romartin:
Great, thanks @hasys , merging and applying your comments in the PR for master, which I'm going to create right now. Thanks!! :+1:

PR #2347, opened 2018-12-13
DROOLS-3431: [DMN Designer] Error after reopening

manstis:
See https://issues.jboss.org/browse/DROOLS-3431

@karreiro I changed things around a bit to also set the `Diagram` on the `DecisionNavigatorPresenter`. The `Diagram` is available before the `Session` and before Stunner attempts to render the Graph (and hence trigger the `CanvasElementAddedEvent` events that lead to the undesirable behaviour reported by @jomarko).


manstis:
I've tested with both `kie-wb-common-dmn-webapp` and `drools-wb-webapp`... all OK.

PR #2348, opened 2018-12-13
[AF-1750]: Added Clusted event to sync spaces.

adrielparedes:
# Description

Now observing AfterDeleteOrganizationalUnitEvent instead of RemoveOrganizationalUnitEvent.

# Related PRs
* https://github.com/kiegroup/appformer/pull/575
* https://github.com/kiegroup/kie-wb-common/pull/2348

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
@@ -39,8 +39,8 @@ import org.guvnor.common.services.project.model.WorkspaceProject;
 import org.guvnor.common.services.project.service.WorkspaceProjectService;
 import org.guvnor.common.services.project.social.ModuleEventType;
 import org.guvnor.messageconsole.client.console.MessageConsoleScreen;
+import org.guvnor.structure.events.AfterDeleteOrganizationalUnitEvent;
 import org.guvnor.structure.organizationalunit.OrganizationalUnit;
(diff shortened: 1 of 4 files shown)


kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/util/LibraryPlaces.java
```diff
@@ -353,16 +355,21 @@ private boolean isRepoForActiveProject(RepositoryRemovedEvent repositoryRemovedE
                 .isPresent();
     }
 
-    public void onOrganizationalUnitRemoved(@Observes final RemoveOrganizationalUnitEvent removedOrganizationalUnitEvent) {
-        if (isLibraryPerspectiveOpen()) {
+    public void onOrganizationalUnitRemoved(@Observes final AfterDeleteOrganizationalUnitEvent removedOrganizationalUnitEvent) {
+        if (isLibraryPerspectiveOpen() && isSpaceActive(removedOrganizationalUnitEvent)) {
```
jhrcek: I'm wondering why we even need the method `isSpaceActive` and this additional check `&& isSpaceActive(removedOrganizationalUnitEvent)`  when it's doing similar check as the `.filter(active -> active.equals(removedOrganizationalUnitEvent.getOrganizationalUnit()))` below.
Can it happen that you can have 2 different OrganizationalUnits that are equal but don't have the same identifier?

I think we should remove the newly added private isSpaceActive and remove that && check and it will behave the same.

adrielparedes: My fault! I upload something I shouldn't, the only fix was changing the event, but I made a mistake, I will fix it right now, sorry

PR #2349, opened 2018-12-13
JBPM-7888: case modeler - the containment area should be all the column

xiezhang7:


xiezhang7:
Hi @romartin @LuboTerifaj @hasys , this is the pull request for [JBPM-7888](https://issues.jboss.org/browse/JBPM-7888). Thanks.

LuboTerifaj:

romartin:
Hey @xiezhang7 , the WARs are not being generated, please can you check? Thx!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java
```diff
@@ -0,0 +1,54 @@
...
+                                        boolean resizable) {
+        super(name, svgPrimitive, width, height, resizable);
+    }
+
+    @Override
+    protected Optional<MultiPath> makeDropZone() {
+        return Optional.of(new MultiPath().rect(0.0,
+                                                getHeight(),
+                                                getWidth(),
+                                                1350.0));
```
romartin: is that hard-coded value expected somewhere?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/wires/CaseManagementContainmentControl.java
```diff
@@ -198,12 +208,17 @@ private WiresManager getWiresManager() {
         return getShape().getWiresManager();
     }
 
-    private Integer getShapeIndex() {
-        if (!(getParent() instanceof CaseManagementShapeView) || getShape() == null) {
+    Integer getShapeIndex() {
```
romartin: let's avoid use of `null` here, if possible... you can refactor it by using for example `OptionalInt`

LuboTerifaj:
Regarding not generated wars, there is a compilation failure. It cannot find symbol class AbstractReusableSubprocess, but I don't know what it is caused by. I will try to rebuild it again.

LuboTerifaj:

LuboTerifaj:
Hello @xiezhang7 ,

A rebuild didn't help. Full downstream build is still failing.
Can you have a look at the issue please?

Thank you.

xiezhang7:

xiezhang7:
Hi @LuboTerifaj @romartin , I have fixed the build issue. Thanks.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/wires/CaseManagementContainmentControl.java
xiezhang7: Hi @romartin , I have refactored the code to use OptionalInt.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java
xiezhang7: Hi @romartin , the hard-coded value is supposed to be the height of the diagram (i.e. the height of the drop zone). 
I am not sure how to get the height of the diagram in this context. In addition, if this value can be changed interactively by users, I don't know how to update the drop zone height of all the existing Stages dynamically. So I used a hard-coded value. 
1350 is bigger than the initial height of the diagram. And the height of a Subprocess/UserTask is 118 (103 height + 15 gap), so I think 1350 (can contain 12 Subprocesses/UserTasks) should be big enough in normal cases.

romartin: Hey @xiezhang7 

I see the idea behind hardcoding this yes, but I think it makes sense to think about it and do not use any hardcoded number here.

Here is my proposed approach - The height of the drop zone is not always the diagram's one. I think it makes sense to use the height of the `parent` shape. This way, in case embedded stages, it will display using the parent's stage height, otherwise if no `parent`, it can then use the actual `canvas`/`layer` height. You can also set whatever "gap" etc on top of this values. 

So you could just check the wires parent shape, if any, and apply that logic... WDYT? does it makes sense?

xiezhang7:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java
xiezhang7: Hi @romartin , Ivo once said it should not be possible to add a stage to a stage, which means there should not be any embedded stages. So I changed the code to use the canvas height (retrieved from [ScratchPad](https://github.com/kiegroup/kie-wb-common/blob/6360632e468a4784afcd9106aedd445a08908b5c/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java#L59), I am not sure if there is a better way to get the canvas height in this context). And if the canvas height changes, the drop zone will be re-created when [getDropZone](https://github.com/kiegroup/kie-wb-common/blob/6360632e468a4784afcd9106aedd445a08908b5c/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java#L54) is called.

romartin: Hey @xiezhang7 

Well using the height retrieved from the `ScratchPad` is not strictly correct... but good try anyway! :+1: 

Some comments:
1.  The `ScratchPad` thing is like a "helper" object that lienzo provides for each `viewport`. It's basically a canvas `Context2D` wrapper that developers can use for its own use, but in fact lienzo itself is never handling it., I mean, for example, if you change the height of the canvas, the `ScratchPad` instance related to its `viewport` is not "resized" by the lienzo core itself. So just keep the idea that we can not use it to obtain attributes from the "real" canvas (viewport)
2.  Also notice that when you obtain the `height` value (from the viewport or the scratchpad), it returns an `int` type. On the other hand, when you set the `height` for a shape in lienzo, for example a `Rectangle`, the value is a `double` type. This is because the `int` type is used when talking about "pixels" in a screen, on the other hand the `double` type is used when talking about the "dots" on the canvas. For example when applying a transform mediator to a certain layer, for example scaling, the `height` value of the canvas (pixels) remains the same one (the "widget" remains same size), but the `absolute height` (relative to actual transforms) is bigger, as the shapes require more "dots"

So given the above arguments, it results on:
1. Instead of obtain the height value from the `scratchpad` instance, you should obtain the one from the `viewport` (`getLayer().getHeight()`)
2. Once given the `height` in pixels above, you have to transform it using the actual `viewport's  transform` instance and use the resulting value as for the height of your ghost shape. You can find an example of how to calculate this [here](https://github.com/kiegroup/lienzo-core/blob/master/src/main/java/com/ait/lienzo/client/widget/panel/scrollbars/ScrollablePanel.java#L135). The result from this is that the ghost shape must be same height as the actual canvas (layer), even when applying zoom mediators.

Please give this a try and lemme know if it makes sense... :+1: 
Thanks!

romartin: If it does not exist, which I'm not use, you can probably create some static helper method than given a canvas/layer it returns the calculations above, so we can reuse it....

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java
xiezhang7: Thanks @romartin . 
Unfortunately, I could not get the `height` of the diagram from `viewport`, because  `viewport.getLayer()` returns `null` in this context. 
I noticed that the `height` value can be obtained from the `bounds` property of [LienzoBoundsPanel](https://github.com/kiegroup/kie-wb-common/blob/02c1d435c826a0d26e752308656cb898168a27c2/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/canvas/CaseManagementCanvasView.java#L40), which is a property of [CanvasView](https://github.com/kiegroup/kie-wb-common/blob/02c1d435c826a0d26e752308656cb898168a27c2/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/canvas/CaseManagementCanvasView.java#L50). `LienzoBoundsPanel` calculates its own `bounds`, and the value is a `double` . So I changed the code to get the value from the `LienzoBoundsPanel`.

xiezhang7:

xiezhang7:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/test/java/org/kie/workbench/common/stunner/cm/client/wires/VerticalStackLayoutManagerTest.java
```diff
@@ -56,33 +56,33 @@ public void checkLayout() {
 
     @Test
     public void orderChildrenInsertShape0At0() {
-        // Insert to the end
+        //Mouse Y needs to be before target index Y as insertion happens when Shape is dragged before target
```
LuboTerifaj: I don't understand this sentence. Can you please clarify it to me? Thanks.

xiezhang7: Hi @LuboTerifaj , 

"Mouse Y" is the second argument of the method `addShapeAtIndex`. This sentence means, in order to drag and drop a shape before a target shape, when dropping the shape, the y-coordinate of the mouse cursor should be a little smaller than the y-coordinate of the top of the target shape.

In this particular test case, in order to test drag and drop a shape before the target shape `shapes.get(0)`, the second argument of method `addShapeAtIndex` (i.e. "Mouse Y") is set to `shapes.get(0).getY() - PADDING` which is a little smaller than the y-coordinate of the target shape `shapes.get(0)`.

This sentence was not created by me. But I changed it to "Insert to the end" and corrected the test cases, when the drop zone was not working. 
Since the drop zone is fixed now, I changed the comments as well as the test cases back to what they used to be.

Thanks.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/shape/view/CaseManagementStageShapeView.java
romartin: hehe yeah really well done @xiezhang7 :+1:

romartin:
Jenkins please retest this

romartin:
Hey @xiezhang7 
Sorry but I just forgot how to enable the "experimental" features, in order to test the generated WAR on this PR. It was some system property right?

LuboTerifaj:
Hey @romartin 
Yes, it is. You have to run business-central with "-Dappformer.experimental.features=true" and then enable it in settings.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/test/java/org/kie/workbench/common/stunner/cm/client/wires/VerticalStackLayoutManagerTest.java
LuboTerifaj: Thank you for clarification.

LuboTerifaj:
@xiezhang7 
Have you changed something? I see that there is new full downstream build but I see only one commit. So in case that you force pushed here some changes, I don't know what exactly you have changed.

Thank you.

romartin:
Hey @xiezhang7 @LuboTerifaj 

Sorry for the delay on this PR guys, but there is still no way to test it, now that I enabled the experimental features, when adding the permission for the (New) `Case modeler` editor, I understand that due to a bug, all permissions for the `Editors` tab just disappear, and I'm not able to test the case modeler, no asset appears on the list...

Anyway as @LuboTerifaj already approved I'll proceed to merge. 

Thanks!

PR #2352, opened 2018-12-17
DROOLS-3344: [DMN Designer] Data Types - UX improvements in the Data Type List View

karreiro:
# DROOLS-3344

See https://issues.jboss.org/browse/DROOLS-3344

This PR introduces two king of enhancements: **I) The introduction of keyboard shortcuts** in the Data Types tab, and **II) Visual changes** to clarify some concepts to the user.

### Navigation based on arrows:
Now users can use <kbd>↑</kbd> and <kbd>↓</kbd> to navigate between Data Types, and expand/collapse by using <kbd>→</kbd>/<kbd>←</kbd>, respectively:
navigation-arrows

Also, when users are searching for some DT, they can just press <kbd>Tab</kbd> to navigate in the results (ahh.. but they can press <kbd>↓</kbd> to go to the first result too):
tab-arrows

BTW.. the navigation can be disabled by pressing <kbd>Esc</kbd>.

### Updating Data Types:
When users want to update a Data Type, they just need to press <kbd>Ctrl</kbd> + <kbd>E</kbd> to activate the edit mode, and then, they can save the changes by pressing <kbd>Ctrl</kbd> + <kbd>S</kbd>:
edit-mode-dtrl-e

However, if they don't want to update anything, they can close the edit mode by pressing <kbd>Esc</kbd>:
edit-mode-by-esc

### Creating Data Types:
Now the flow for creating Data Types is improved to avoid the use of the mouse. When users want to create a nested DT, they can just press <kbd>Ctrl</kbd> + <kbd>N</kbd>, or <kbd>Ctrl</kbd> + <kbd>U</kbd> for a DT above, or even <kbd>Ctrl</kbd> + <kbd>D</kbd> for a DT below:
many-things-creation-easy

### List of shortcuts:
<kbd>↑</kbd> - Go to the next Data Type
<kbd>↓</kbd> - Go to the previous Data Type
<kbd>→</kbd> - Expand the Data Type
<kbd>←</kbd> - Collapse the Data Type
<kbd> Ctrl </kbd> + <kbd>E</kbd> - Activate edit mode
<kbd> Ctrl </kbd> + <kbd>S</kbd> - Save the Data Type
<kbd> Ctrl </kbd> + <kbd>B</kbd> - Create a nested Data Type
<kbd> Ctrl </kbd> + <kbd>U</kbd> - Create a Data Type above (up)
<kbd> Ctrl </kbd> + <kbd>D</kbd> - Create a Data Type below (down)
<kbd> Ctrl </kbd> + <kbd>Backspace</kbd> - Delete the Data Type
<kbd>Esc</kbd> - Exit

## Visual changes:
screen shot 2018-12-17 at 1 59 57 pm

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/586
- https://github.com/kiegroup/kie-wb-common/pull/2352


ederign:
Awesome PR description!

manstis:
@karreiro If you can fix the conflict I'll endeavour to review before Christmas :-)

(However we'll need to wait for the New Year for @jomarko to review too).

karreiro:
@manstis I'll keep this PR as rebased as possible to avoid difficult conflicts in the future. But, since it needs @jomarko approval, feel free review it on 2019 🎊.

karreiro:
@manstis Thanks for the feedback :-)

I agree that we need to choose a single approach for handling this scenario. I will be at the meeting ☺️

When I was implementing this, I quickly checked the approach on `uberfire-wires-core-grids`, but it looked coupled with grids/canvas stuff (I didn't check Stunner approach.. but I assume that it's more canvas guided too). Thus, I decided to create the `DataTypeShortcuts.java` which is coupled with Elemental2 classes and doesn't rely on deprecated APIs (like [KeyboardEvent.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)).

Probably the best solution will be the fourth one, even more generic, providing mechanisms to handle the three scenarios 🙃

Regarding the <kbd>Ctrl</kbd> + <kbd>N</kbd>, I didn't notice it because here I press <kbd>Cmd</kbd> + <kbd>N</kbd> to open a new tab. Maybe, we could use <kbd>Ctrl</kbd> + <kbd>I</kbd> (inner).

manstis:
@karreiro I worry changing the use of <kbd>Ctrl</kbd> + <kbd>N</kbd> to another combination simply moves the problem. Did you try attaching the keypress handler to a DOM element instead (e.g. the root panel for the Data Types widget)? I guess a different (as yet, unused) combination is a simple workaround.

manstis:
@karreiro Seems that it's [not possible](https://stackoverflow.com/questions/38838302/any-way-to-override-ctrln-to-open-a-new-window-in-chrome) to use <kbd>Ctrl</kbd> + <kbd>N</kbd> in a _web application_... so an alternative combination seems the best (only!) way to go.

jomarko:
I finished manual check. Seems work pretty fine. Just these points in my mind.
- Agree with @manstis , **Ctrl+N** doesn't work, opens new browser window
- UI doesn't inform user about possibility to use keyboard. Could we please add short sentence into **Custom Data Types** paragraph and maybe in future put this info at proper place after discussion with UX team?
- Very minor thing but kind of confused/surprised me. When I searched for some data type, then navigated to N-th result (was a structure), then tried expand and collapse two actions we done, cursor in search box was moved one character left/right respectively and expand/collapse button of the structure was changed respectively. If I remember correctly, search results are always expanded so maybe would be fine to disable expand/collapse keyboard shortcuts when search results are shown.

Code review is ongoing.

jomarko:
During review of updates for Data Type Flash messages I tried the scenario below. I think when Some Flash Message is shown, just keyboard shortcut for saving or canceling editing should be enabled.

- Create new entry [1] and navigate to this entry by arrows
- Press Ctrl+D to insert entry below, invoke edit mode by Ctrl+E [2]
- Rename to same as first entry
- Press Ctrl+S [3]
- Press Ctrl+D, we will reach invalid state with two same named entries [4]

[1]
screenshot from 2019-01-03 08-56-16

[2]
screenshot from 2019-01-03 08-57-55


[3]
screenshot from 2019-01-03 08-58-14

[4]
screenshot from 2019-01-03 08-59-26

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeListShortcuts.java
```diff
@@ -0,0 +1,145 @@
...
+
+        view.reset();
+    }
+
+    void onBackspace() {
+        getCurrentDataTypeListItem().ifPresent(DataTypeListItem::remove);
+    }
+
+    void onCtrlS() {
+        getFocusedDataTypeListItem().ifPresent(DataTypeListItem::saveAndCloseEditMode);
```
jomarko: I am thinking if we shouldn't replace this `getFocusedDataTypeListItem` with `getVisibleDataTypeListItems` because when I compare **Ctrl+S** with **Escape**
- **Escape** Cancel edit mode for all entries that are in edit mode
- **Ctrl+S** Save just the entry that was focused by keyboard arrows navigation, even when more entries are in edit mode

I think am back to my old concern that edit mode shouldn't be allowed for multiple entries in parallel.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeListShortcutsTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.kie.workbench.common.dmn.client.editors.types.listview.DataTypeListItem;
+import org.mockito.Mock;
+
+import static java.util.Arrays.asList;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(GwtMockitoTestRunner.class)
+public class DataTypeListShortcutsTest {
```
jomarko: I see that in implementation you use `ifPresent` for all implemented shortcuts however I would like to ask you confirm no action is done by adding negative tests as you did for an **Escape**.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeShortcutsTest.java
```diff
@@ -0,0 +1,521 @@
...
+        assertTrue(shortcuts.isInputEvent(event));
+    }
+
+    @Test
+    public void testIsInputEventWhenDropdownIsOpened() {
+
+        doReturn(false).when(shortcuts).isTargetElementAnInput(any());
+        doReturn(true).when(shortcuts).isDropdownOpened();
+
+        assertTrue(shortcuts.isInputEvent(event));
```
jomarko: This scenario brought me to this scenario:
- Navigate to some entry
- Invoke its edit mode
- Expand **Type** dropdown
- Save by **Ctrl+S**
- Drowpdown will stay opened

screenshot from 2019-01-03 09-30-09

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeListShortcutsTest.java
karreiro: @jomarko I see two different scenarios.

**Scenario I:**
```java
final Optional<Foo> foo = getFoo();

if (foo.isPresent()) {
    foo.get().bar();
} else {
    resetApp();
}
```

**Scenario II:**
```java
final Optional<Foo> foo = getFoo();
foo.ifPresent(Foo::bar);
```

In the first scenario, we have some logic happening when the `foo` instance is empty, thus it's important to validate it.

In the second scenario, we don't have any logic. The responsibility for testing the two possible states for the `foo` instance belongs to the `Optional` class.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeShortcutsTest.java
karreiro: Thank for reporting, it's fixed now :-)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeListShortcuts.java
karreiro: I understand our mindset, however I see the <kbd>Alt</kbd> + <kbd>S</kbd> as a more "destructive" thing, compared to the <kbd>Esc</kbd> (exit action).

So, I think think we could keep it as is for now, and have a deeper discussion with Liz and Matthew about it [here](https://issues.jboss.org/browse/DROOLS-3494).

karreiro:
@manstis Thank you for the review and for the raised concerns. I've changed the all shortcut to use <kbd>Alt</kbd>, instead of <kbd>Ctrl</kbd>. I think that this solves our concerns for this PR :-)

---

@jomarko Thank you for the review and for reported issues. I've solved both (the one related to the dropdown and the one related to validations).

Answering the questions:

> _I) Agree with @manstis , Ctrl+N doesn't work, opens new browser window_
Yeah, me too. Now that we're using <kbd>Alt</kbd> + <kbd>N</kbd>, I think we're fine. However, I talked to Liz this morning, and she's having some discussions with Matthew regarding better possible combinations. For now, I would keep it as is, and I would wait for their feedback.

> _II) UI doesn't inform user about possibility to use keyboard. Could we please add short sentence into Custom Data Types paragraph and maybe in future put this info at proper place after discussion with UX team?_
I think we should wait for Liz feedback. I've create [this JIRA](https://issues.jboss.org/browse/DROOLS-3494), and probably we're going to have some mockups soon :-)

> _Very minor thing but kind of confused/surprised me. When I searched for some data type, then navigated to N-th result (was a structure), then tried expand and collapse two actions we done, cursor in search box was moved one character left/right respectively and expand/collapse button of the structure was changed respectively. If I remember correctly, search results are always expanded so maybe would be fine to disable expand/collapse keyboard shortcuts when search results are shown._
I've raised a JIRA for this [here](https://issues.jboss.org/browse/DROOLS-3495). Thanks.

karreiro:
@manstis Thank you for the feedback. As we discussed, I've brought all shortcuts to <kbd>Ctrl</kbd> again. Now, we're using <kbd>Ctrl</kbd> + <kbd>B</kbd> for creating a nested DT (I know that's not a very good combination, but since Liz is thinking about better shortcuts, I think that we could keep it as is for now).

jomarko:
@karreiro thank you

manstis:
@karreiro Works well, thanks.. however a couple of suggestions for improvement. When I <kbd>Ctrl</kbd> + <kbd>S</kbd> or <kbd>Esc</kbd> the _current_ row becomes deselected and I need to use the cursor keys to again highlight a row before the keyboard shortcuts can be used again. It'd be _nice_ if the current row remained highlighted so I could continue to use the shortcuts. Also, it'd be nice if I could click on a row to highlight it and keyboard shortcuts work (at the moment I need to use the cursor keys to highlight a row first, even if I've clicked on a row).

karreiro:
@manstis Very pertinent suggestions, thanks. I'm going to address them to JIRAs to improve it soon.

PR #2355, opened 2018-12-18
DROOLS-3427: [DMN Designer] Connections between nodes throws an error if user moves the mouse too fast

danielzhe:
This needs https://github.com/kiegroup/lienzo-core/pull/80 to tests not to fail.
Edit: not required anymore.


manstis:
@danielzhe Could you please explain the cause of the issue and how this fixes it?

danielzhe:
@manstis  When we are creating a new connection between nodes, we create a `transientShapeView` and the magnets for that node, that we destroy when the connection is finally created. But in this class we were only destroying the `transientShapeView` and not the magnets. So the magnets were still listen to the mouse move event even after the user clicked and finished the creation of the connection, but since were only destroying the `transientShapeView`, the event in magnets fail because there is no more `transientShapeView` (it's destroyed).
I think that's it. @romartin

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/components/drag/ConnectorDragProxyImplTest.java
```diff
@@ -0,0 +1,157 @@
+package org.kie.workbench.common.stunner.client.lienzo.components.drag;
```
manstis: License please.

danielzhe:
Yes, we found the root of the issue. Thank you @romartin for the help!

romartin:

romartin:
@danielzhe great job, thanks! 
Running the full build now, once all green lemme or @manstis know and we'll merge it. Thanks!

danielzhe:
Jenkins please retest this

jomarko:

PR #2364, opened 2018-12-21
AF-1453 DataModeller Editor Dynamic Dependency Management

desmax74:
Aligned with AF-1452 with Configuration inside classes
Appformer AF-1452
https://github.com/kiegroup/appformer/pull/541

@ederign @domhanak

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
@@ -17,11 +17,12 @@
 package org.kie.workbench.common.screens.datamodeller.backend.server;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.StringTokenizer;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
 import javax.enterprise.inject.Any;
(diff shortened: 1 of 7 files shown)


kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java
```diff
@@ -233,6 +243,13 @@ public Path createJavaFile(final Path context,
             dataObjectCreatedEvent.fire(new DataObjectCreatedEvent(currentModule,
                                                                    dataObject));
 
+            Object currentValue = options.get("persistable");
+            boolean isPersistable = Boolean.valueOf(currentValue != null ? currentValue.toString() : null);
+            if (isPersistable) {
+                addPomDependencyEvent.fire(new AddPomDependencyEvent(new HashSet(Arrays.asList(new JPADependencyType())),
```
wmedvede: Hi @desmax74 , I believe the comments I've added to the closed https://github.com/kiegroup/kie-wb-common/pull/2268, still apply. Se below please:

I'm not sure if this asynchronous pom.xml dependency update will be reliable.
See that when a .java file is created, at least until now, an incremental project build is produced. So I'm afraid this asynchronous dependency update might lead into variants of the following error prone scenarios:

1. a new brand project is created. (with no JPA dependencies added)
2. a new persistable data object is created
3. in line 239 the new created file with JPA stuff is saved.
4. at some point an incremental build is launched due to the just created file. 
5. the AddPomDependencyEvent(EnumSet.of(DependencyType.JPA), currentModule.getPomXMLPath())) is raized

But, nobody knows which of 4) or 5) will be executed first. I would say that with high probability 4) is executed first. And build errors will be produced.

In parallel at the same point after 5) is raised the DataModelerEditor is trying to open the just created file, and for this it's also required that the project was built for having all the necessary metadata produced by the DataModelOracle, etc.

I'd strongly suggest that this pom.xml dependencies adding is done in a non asynchronous fashion and before the just created file is saved at line ioService.write(nioPath, "content of the new file")

Something like this.

Object currentValue = options.get( "persistable" );
boolean isPersistable = Boolean.valueOf( currentValue != null ? currentValue.toString() : null );

if (isPersistable) {
then here if the pom.xml don't have JPA added yet, please do a synchronous adding of the required JPA dependencies.
}

ok, now proceed with the ioService.write(path, "content of the new file")
and the rest of the existing code.

Thanks! and please don't forget adding me to future PRs related with data modeller stuff, thank you.

desmax74: @ederign wdyt about a direct call instead of event ?

ederign: +1

cristianonicolai:
jenkins please retest this

domhanak:
@desmax74 Is this still relevant? Maybe we should close this.

desmax74:
No more relevant

PR #2368, opened 2019-01-02
DROOLS-3430: [DMN Designer] [Edge] Update Name via Name And DataType dialog

danielzhe:


danielzhe:
I'm forcing the blur event to be fired when the hide() is called. I did some tests in other browsers to see if the event is fired two times doing this change, but it only called once, so I think there is no side effect.

PR #2375, opened 2019-01-04
DROOLS-3175: Reuse existing code for cell selection

jomarko:
@manstis would you mind to have a look?

I am working on shortcuts for context menus and I am in phase of reusing as much general code from uberfire as possible. 

Asking you as author of the touched class what scenario could be affected negatively by this change.


jomarko:
@manstis thank you, good point, done.

PR #2377, opened 2019-01-04
JBPM-8083: BPMN2 process causes Stunner to crash with EMF Error

wmedvede:


wmedvede:
Hello @romartin @hasys , would you mind review this PR?
Thanks

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/util/BPMNGraphUtils.java
```diff
@@ -0,0 +1,199 @@
...
+import org.kie.workbench.common.stunner.bpmn.definition.property.dataio.AssignmentsInfo;
+import org.kie.workbench.common.stunner.bpmn.workitem.ServiceTask;
+import org.kie.workbench.common.stunner.core.graph.Graph;
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+import org.kie.workbench.common.stunner.core.util.StringUtils;
+
+import static org.kie.workbench.common.stunner.core.util.StringUtils.isEmpty;
+
+public class BPMNGraphUtils {
```
romartin: Just a minor comment, IMO given the goals for the methods in this class I would change it to another classname a bit more related to BPMN variables or so... does that makes sense?  Again, just sharing my opinion :)

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/util/BPMNGraphUtils.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private static List<VariableUsage> findVariableUsages(String variableName, Node node) {
+        List<VariableUsage> result = new ArrayList<>();
+        AssignmentsInfo assignmentsInfo = null;
+        String variableInfo;
+        String displayName = null;
+        if (StringUtils.isEmpty(variableName)) {
+            return result;
+        }
+        if (node.getContent() instanceof View) {
```
romartin: Again, nothing really blocking probably for some... but I think could be better trying to improve the readability and maintenance of this code, also it can be simplified in some way I guess...

Maybe it's easier to manage all these conditions by assuming the node is being cased to `Node<View<BPMNDefinition>, Edge>` on the `graph.getNodes()` stream iteration and the provide in some way static Function instances that return, for  example the `DataIOSet` member for some events and tasks, and other `Function` that return the `AssignemtsInfo` instance for a `DataIOSet` instance... I mean, instead of adding if/else expressions, trying to use `Streams` and functional interfaces (`Functions`) in order to achieve same logic in a more clear way.

Eg:
```

StreamSupport.stream(nodes.spliterator(), false)
    .filter(BPMNGraphUtils::isBPMNDefinition)
    .map(BPMNGraphUtils::findVariableUsages)
    ...

private static boolean isBPMNDefinition(Node node) {
return node.getContent() instanceof View &&
        ((View) node.getContent()).getDefinition() instanceof BPMNDefinition;
}

public static AssignmentsInfo findVariableUsages(Node<View<BPMNDefinition>, Edge> node) {
// TODO: Iterate over some set of static functions on this class which can be expressed via lambdas in a single line
}

```

WDYT??

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/util/BPMNGraphUtils.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+        String encodedVariablesList = encodedParts[4];
+        if (!isEmpty(encodedVariablesList)) {
+            String[] variablesList = encodedVariablesList.split(",");
+            String[] variableDefParts;
+            String variableName;
+            VariableUsage variableUsage;
+            String unPrefixedVariableDef;
+            VariableUsage.USAGE_TYPE usageType;
+            for (String variableDef : variablesList) {
```
romartin: Same here, would be nice refactoring this in a more "functional" way, by using Streams and some functions/lambdas....

hasys: Agreed, probably it will make sense to move most of the logic from this method to static factory inside of `VariableUsage` class. Something like `VariableUsage.parseFromString(String)`

hasys:
Hi @wmedvede, code looks good, but I agree with @romartin, that probably it will be good to refactor it little bit.

Will check runtime now.

krisv:
@hasys your first item is valid but I think is a different issue imho, I would keep that separate.  The other two are imho valuable improvements on the usability.  Could we maybe track them seperately as well so at least this can get merged first (and we can then decide the priority of these other improvements)?

wmedvede:
Hi @hasys regading your queries, this are mi opinions/queries:

1) When the parsing fails, the XML process editor is still opened like with the other parsing error use cases. 
In general the pattern with the Converters is that when "something" fails at parse time, an exception is thrown, and the parsing is stopped. So it's not trivial or direct to recognize different parsing error scenarios and try to do recoveries or a sort of "on parsing" refactoring. In fact, as soon a potential Converter fails, the Graph is not constructed. i.e., so we don't have any model, we have only the low level exception thrown by the given Converter.
I agree that it could be intresting to have a sort of on-parsing refactoring, but I don't see a direct solution right now. Probably we can add a sort of hook for managing this use case. Open to here more comments opinions for the rest.

2) you mean to add the node id to the parsing error message thrown by the converter when the file is being opened?

3) Regarding to remove/update variable usages. I understand you mean here at edition time, right?
i.e. to have a sort of "refactoring" facility that let's the user automatically refactor the variable usages along the process?
If this is the case, we can invest some time to see if we can do it. But I'm not confident that all the use cases can be covered and that the refactoring can be done safely. Specially for cases where the process variable is used e.g. in process scripts. Since we have a bunch of supported script languages, etc. I believe that if can't provide a safe refactoring, it's better to not do it :(.

hasys:
@krisv @wmedvede, in my opinion first and the third points are very similar, in case of we will remove variables usage from XML (Graph). The only difference is WHEN - when we open a "corrupted" file or when we are trying to "corrupt" it.

@wmedvede sounds reasonable what you are mention about Scripts for third point. But, probably, just XML (Graph) fix should be enough for now, because we still doesn't deal with removed variables in scripts anyway, but XML "corruption" is an issue right now.

@wmedvede about 2 point yes not just notify user, that it is not possible to remove variable declaration because it is in use in some other nodes, but also display name/id where this node is used. I guess it is possible since we have a full Graph in our hands here.

wmedvede:
@hasys @krisv 
1 and 3 are "similar", but different class of problems in terms of complexity and amount of work. In 1, as mentioned, the pattern we have right now for the Converters don't support this sort of on parsing refactoring and error recovering, etc. We need to think a bit here, etc.

In 3, the scenario is completely different, here we have an already parsed model and "complete" control about whats going on.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/util/BPMNGraphUtils.java
wmedvede: sure, I'll take a look, but if you look more insight you'll see that all nodes are not processed in the same way, some nodes has AssignmentsInfo and some hasn't. Anyway, i'll take a look. This is way all is resolved in the List<VariableUsage> findVariableUsages(String variableName, Node node) that I already have

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/util/BPMNGraphUtils.java
wmedvede: @romartin this iteration here is also tricky, note that when a "part" is being processed, we might have to update the state of an already processed "part", since a variable can be found first as an input variable, and later in a different position of the "Stream" as an output one... and we need to update the status.
So we'll still need to have a sort of global Map<String, VariableUsage> variableUsages that can be accessed by the stream processing. So more or less (unless setting up something more complex and probably less readable), we'll endup just changing a 

for (String variableDef : variablesList) {
    doing something here
}
with 
variablesList.forEach(variableDef -> { and doing the same here...}

For sure everything can be done in functional style, but to be honest, I'm not in favor of forcing using functional approach when can have something simple and readable. This sort of "low level" parsing are some times good cases for using more "old style" coding in my opinion.

wmedvede:
@hasys @romartin suggestions that makes more sense for me where applied. 
Regarding adding the "on file opening refactoring", or "on edition refactorings", etc, I think it's better to talk on Stunner meeting. We'll go faster.

romartin:
Hey @wmedvede @hasys Please as commented before, let's move this separate discussion into a new JIRA and so let me know when this is ready to merge. Thanks!

hasys:
Just check that everything is fine after new commit were pushed.

wmedvede:
@hasys @romartin same here, downstrm buld failed due to unrelated tests, but artifacts were generated.

romartin:
no worries @wmedvede , but is this ready to merge guys? CC @hasys

wmedvede:
> Just check that everything is fine after new commit were pushed.
> 
> Jenkins execute full downstream build

@hasys is the final check with the generated .wars pending?

hasys:
Hi @wmedvede,

looks good now, thank you!

I reported 2 and 3 issue. About 1-st one I guess we can leave current one (JBPM-8083) opened, or just clone it. WDYT?

Reported issues:
* [JBPM-8083](https://issues.jboss.org/browse/JBPM-8143) - provide a button to remove/update process variable declaration automatically for all nodes
* [JBPM-8144](https://issues.jboss.org/browse/JBPM-8144) - add node name/ID to the message dialog were process variable is used

wmedvede:
@hasys I believe it's better to close JBPM-8083 and open a new one for 1 (whatever solution can be found) and move this PR forward. Otherwise we won't be able to close and merge this PR, well, we can merge, but we'll lose tracking, and we want this for 7.3.
Thanks!

hasys:
@wmedvede, @romartin, FYI I cloned Jira issue and modified it's description:

* [JBPM-8146](https://issues.jboss.org/browse/JBPM-8146) - BPMN2 process causes Stunner to crash with EMF Error

PR #2378, opened 2019-01-04
JBPM-8098 Stunner - Canvas becomes empty after saving the diagram

tiagodolphine:
When saving the diagram, the SVG image is generated and this process uses a Wrapped `Context2D` to generate the SVG. 
It is necessary to draw the original `Context2D` from Canvas after the SVG generation is complete, otherwise, the canvas remains empty.

@romartin I think this was removed during some refactor necessary for the infinite canvas (https://github.com/kiegroup/kie-wb-common/commit/1082e1994c9ed231384984b8bb8b9da2205b56fe#diff-22baa68f555ac85b7f154466ae47e825L76)

@LuboTerifaj


tiagodolphine:

PR #2372, opened 2019-01-04
DROOLS-3406: [DMN Designer] Decision Service splitter should be movable

manstis:
See https://issues.jboss.org/browse/DROOLS-3406

Tests to follow.. this is "for early eye-ball".. so I don't waste time writing tests if a serious issue with the code changes is found on preliminary review. This PR adds a movable splitter to the "Decision Service" node and handles persistence of its value. I have followed guidelines from @romartin (and existing code) about adding the new _decorator_.

@romartin I changed Stunner to support multiple SVG Factories (so I have one for _most_ DMN shapes and a specific one for _Decision Service_). This allowed me, IMO, to better distinguish between construction of the different Def, Views etc (otherwise I needed my Factory to contain an `if` statement to distinguish between the two - using the `String` _name_ of a node type.. and personally I thought being able to handle this with different factories was better!)

Part of an ensemble:
- https://github.com/kiegroup/lienzo-core/pull/81
- https://github.com/kiegroup/kie-wb-common/pull/2372

PR #2382, opened 2019-01-08
DROOLS-3488 DMN DecisionService DMNDI position Encapsulated/Output Decis

tarilabs:
... DMN DecisionService DMNDI wrong positioning for Encapsulated/Output
Decision

/cc @manstis @jomarko

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
@@ -96,6 +96,7 @@ import org.kie.workbench.common.stunner.core.graph.Node;
 import org.kie.workbench.common.stunner.core.graph.content.Bounds.Bound;
 import org.kie.workbench.common.stunner.core.graph.content.relationship.Child;
 import org.kie.workbench.common.stunner.core.graph.content.view.BoundImpl;
+import org.kie.workbench.common.stunner.core.graph.content.view.BoundsImpl;
 import org.kie.workbench.common.stunner.core.graph.content.view.Connection;
 import org.kie.workbench.common.stunner.core.graph.content.view.ControlPoint;
 import org.kie.workbench.common.stunner.core.graph.content.view.MagnetConnection;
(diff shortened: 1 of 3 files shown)


manstis:
@tarilabs I'd prefer to defer this until https://github.com/kiegroup/kie-wb-common/pull/2372 is reviewed and merged as it fixes issues I found with the positioning of encapsulated/output decisions during my work on the movable splitter... did this PR fix anything else that I may have missed?!

tarilabs:
> did this PR fix anything else that I may have missed?!
This I don't know.

It is fine for me to hold this PR if necessary.

This PR fixes the problem where the DMNDI was not respecting the DMN specification: by the spec the positioning is absolute. When a DRGElement is a Child (I believe) the positioning is now relative (to the Decision Service) hence the marshalled DMNDI was not really interchange-able with say Partner editors to cross-check the positioning on the screen/diagram, as per linked JIRA.

manstis:
@tarilabs OK, so it seems like you've fixed the same as me.. I found round-tripping _our_ `DecisionService` node had a few issues when imported into other tools and fixed them in the PR I've referenced. Since my PR includes changes to marshalling the divider line etc I'd personally prefer to keep my PR as the _master_ for these changes. However let's keep this "on the radar" so we can be sure everything is fixed.

jomarko:
@manstis please, do I understand correctly that no review needed here from me ? I assume #2372 fixes the same as this one, right? Please tell me if I should review also this.

manstis:
@jomarko Yes, no review required of this PR; however it'd be good to verify #2372 contains fixes that this PR was raised for - i.e. creating a Decision Service with "output" and "encapsulated" decisions and connectors in between nodes "in" the Decision Service and "outside" the Decision Service are correct when a model created in our editor is imported into another vendors tooling and vice-versa (I checked and it all appears OK IMO).

manstis:
@jomarko If you could find a little time to verify the positioning of "output" and "encapsulated" decisions within a Decision Service are correctly maintained when saving and re-opening a DMN model that includes a Decision Service (with "output" and "encapsulated" decisions and various connectors both within the DS and to external nodes) in both our editor and when exported to another vendors this PR can be closed. #2372 is now merged so you can just test with `master` at some point in time.

tarilabs:
This problem was actually resolved as part of DROOLS-3406

PR #2388, opened 2019-01-11
DROOLS-3373: [DMN Designer] Need Palette glyph/icon and Canvas shape for Decision Service

manstis:
See https://issues.jboss.org/browse/DROOLS-3373

screenshot from 2019-01-14 10-48-32


jomarko:
One more point to discuss, if you open some DMN diagram where decision point is used, then you just move the splitter, diagram is marked as dirty but when you press save button, designer says no changes to save, could you have a look also on this please?

manstis:
@jomarko I could not replicate your first issue (magnets not showing) following the steps you provided (however I do notice you're using an old version of this PR.. following comments from UX the implementation changed and some of the code changes have been removed).

Regarding your second issue I have included the "Y-splitter" in the `equals(..)` and `hashCode()` methods (it had not been added). This would explain your findings and, with the change made, I have been unable to replicate.

Please re-check at your convenience.

PR #2390, opened 2019-01-12
JBPM-8113: Review the Forms Annotation processors

pefernan:
@ederign  could you please take a look?

@mbiarnes you'll need this to update the ipbom!

@romartin this change affects the forms annotations processor, could you please check that it doesn't affect stunner properties panels?

This is part of an ensemble, please merge with:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/911
https://github.com/kiegroup/kie-wb-common/pull/2390


pefernan:

pefernan:

pefernan:

pefernan:
unrelated test failures...

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors-tests/src/test/resources/org/kie/workbench/common/forms/adf/processors/models/Parent.expected
```diff
@@ -0,0 +1,229 @@
...
+import org.kie.workbench.common.forms.adf.service.definitions.elements.FormElement;
+import org.kie.workbench.common.forms.adf.service.definitions.layout.LayoutColumnDefinition;
+import org.kie.workbench.common.forms.adf.service.definitions.layout.LayoutDefinition;
+import org.kie.workbench.common.forms.model.FieldDefinition;
+import org.kie.workbench.common.forms.model.TypeKind;
+import org.kie.workbench.common.forms.model.impl.TypeInfoImpl;
+
+@Generated("org.kie.workbench.common.forms.adf.processors.FormDefinitionsProcessor")
+@ApplicationScoped
+public class ModuleFormGenerationResourcesProvider implements FormGenerationResourcesProvider {
```
romartin: hey @pefernan , not sure if that's new or just refactored... also not really complaining at this point, but lemme give you my feedback on this. 
Not really sure how this is being used, but if I understand correctly you're comparing the source files generated (by the tested annotation processors) against these `.expected` ones, isn't it? If so , I don't think this is the best way... because of on some changes on indentations, license updates ... that will imply also updating the `.expected` files on the test scope, right?
Again, if that's the case, nothing to complain anyway... but you could also consider some different approach. For example, creating the class on the test runtime and trying to load in the current classloader (test) scope, this way you can just then create a new instance of the generated class at runtime, and do the right assertions and calls to each method as well. I mean focusing on the goal of testing the behavior / results of this class, rather than comparing the sources... 
Again - just brainstorming :+1:

pefernan: @romartin yes the `.expected` files are only for testing purposes. We've been using this approach in different places on the workbench and agree that there might be some problems when changing licences or code style. Do you have any example of annotation processor test?

ederign:
@jstastny-cz when you have a chance, can you please take a look on this?

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors-tests/src/test/resources/org/kie/workbench/common/forms/adf/processors/models/Parent.expected
romartin: @pefernan sorry I think I don't have an example of this right now :( anyway it depends on each context.. I mean generating the source file and compiling it should not be an issue, as all the dependencies are on the classpath as well. From that point, the difficult thing is about mocking all the context that this class requires for runtime, like the FormsContext or things like that.
Anyway ping me if you need help, we can do some quick try on this, np

pefernan: @jstastny-cz I'll try to answer by parts:
#1 We can be sure that the `.expected` files are correct, they have been generated using the same engine (before this change) and I have verified them. Just a clarification, the annotation engine is actually working well, this changes are preventing an issue that will appear when updating to the next ip-bom version (8.4) which will happen soon.

#2 Yes they are being compared by string equality, this is the same approach we've been taking doing on different  places in our platform. Must say that the copyright is static on the .ftl template (see https://github.com/kiegroup/kie-wb-common/blob/3acd1938a4a7b8fa6a974ba8743e98756e300745/kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors/src/main/resources/org/kie/workbench/common/forms/adf/processors/templates/FormGenerationResourcesProvider.ftl#L2). 

#3 I don't see the problem to change the `.expected` If we change the engine. I think it's ok to assume that if we change a something (.flt, annotations, annotation processor... ) we must invest some time to change the tests.

@ederign @jstastny-cz @romartin I agree that the dynamic apporach (generating the testing classes dynamically, launch the processors etc...) would be much better, but looks like something is out of this PR scope. Don't make me wrong, I'll be happy to work on that as soon as we have more free time.

jstastny-cz: @pefernan thanks for thorough explanation.

I understand and agree with your points for 1 and 2.

ad 3) My concern is, whether there are actually tests that run the generated classes. (Let's say we forget to merge PR with processor changes, but we change the engine in a way, that the format of acceptable generated classes has changed. Do we fail asap?)

I'd be happy to join a discussion about the dynamic approach in future.

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors-tests/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.workbench.forms</groupId>
+      <artifactId>kie-wb-common-forms-adf-processors</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
```
jstastny-cz: Usually a file ends with a newline. Can we add it here, unless there's a different approach in this project.

pefernan:
@jstastny-cz can you take a look again?

kie-wb-common-forms/kie-wb-common-forms-core/kie-wb-common-forms-adf/kie-wb-common-forms-adf-processors-tests/src/test/resources/org/kie/workbench/common/forms/adf/processors/models/Parent.expected
pefernan: @jstastny-cz the main problem is that the generated code on compile time will be wrong and the compilation will fail.

PR #2391, opened 2019-01-14
DROOLS-3513: [DMN Designer] Literal Expression cells clip multi-line expression

manstis:
See https://issues.jboss.org/browse/DROOLS-3513

This PR makes the `LiteralExpressionGrid` height based on the expression text.


jomarko:
@manstis thanks for merge, question about expression ending with empty new lines, see https://github.com/manstis/kie-wb-common/pull/38

jomarko:
For multilines expressions, there is shown scrollbar when editing, however it can be moved just with mouse, the rolling mouse button doesn't work, is it expected please? I am not use if it is doable as we "render" scrollbars in canvas.

manstis:
@jomarko I've pushed a new commit fixing the line break issues... I'll look into your other items now.

manstis:
@jomarko Regarding, "For multilines expressions, there is shown scrollbar when editing..."

This behaviour is expected - but may not be desirable. I observed when using the mouse wheel over the `TextArea` (of an "in cell edit" operation) the `TextArea` was destroyed. This is done by [this](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/widgets/panel/DMNGridPanel.java#L48) and is to hide the `TextArea` when the grid is scrolled using the mouse wheel (for when it overflows the editor height and the grid editor as a whole as a scrollbar).

Perhaps it could be the subject of a new JIRA - as it has not been introduced by this PR?

manstis:
@jomarko This PR has been updated to cover all applicable grids.

It also includes a little refactoring to differentiate between the two different sub-classes of `BaseGridRow` (one for rows containing `LiteralExpression` and the other, renamed from `DMNGridRow` to `ExpressionEditorGridRow` that better reflects its use).

PR #2392, opened 2019-01-14
[DROOLS-3518] Propagating generic-type info to AsyncPackageDataModel

gitgabrio:
@jomarko @Rikkola @manstis 
Scope oif this PR is to propagate generic-type information to client-side, inside AsyncPackageDataModelOracle/AsyncPackageDataModelOracleImpl

### DIFF
--- a/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/DataModelOracleUtilities.java
+++ b/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/DataModelOracleUtilities.java
@@ -242,6 +242,8 @@ public class DataModelOracleUtilities {
         dataModel.setAllPackageElements(oracle.getAllExtensions());
         dataModel.setGlobalTypes(oracle.getPackageGlobals());
         dataModel.setPackageNames(oracle.getModulePackageNames());
+
+        dataModel.setModuleFieldParametersTypes(oracle.getModuleFieldParametersType());
     }
 
     public static void populateDataModel(final PackageDataModelOracle oracle,
--- a/kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/AsyncPackageDataModelOracleImplTest.java
+++ b/kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/AsyncPackageDataModelOracleImplTest.java
@@ -49,6 +49,8 @@ import org.uberfire.client.callbacks.Callback;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.anyList;
@@ -98,6 +100,8 @@ public class AsyncPackageDataModelOracleImplTest {
         oracle.addSuperTypes(createSuperTypes());
         oracle.addPackageNames(createPackageNames());
         oracle.addCollectionTypes(createCollectionTypes());
+
+        oracle.addModuleFieldParametersTypes(createModuleFieldParametersTypes());
 
         oracle.filter(imports);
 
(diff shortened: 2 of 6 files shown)


kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/DataModelOracleUtilities.java
```diff
@@ -242,6 +242,8 @@ public static void populateDataModel(final PackageDataModelOracle oracle,
         dataModel.setAllPackageElements(oracle.getAllExtensions());
         dataModel.setGlobalTypes(oracle.getPackageGlobals());
         dataModel.setPackageNames(oracle.getModulePackageNames());
+
+        dataModel.setModuleFieldParametersTypes(oracle.getModuleFieldParametersType());
```
jomarko: Shouldn't we add this also to the second `populateDataModel` implementation?

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/AsyncPackageDataModelOracleImplTest.java
```diff
@@ -335,6 +339,15 @@ private PackageDataModelOracleIncrementalPayload createPersonPayload() {
         return collectionTypes;
     }
 
+    private Map<String, String> createModuleFieldParametersTypes() {
+        Map<String, String> moduleFieldParametersTypes = new HashMap<>();
+        moduleFieldParametersTypes.put("org.test.Person#phones",
```
jomarko: According to the javadoc, I would expect key "phones" instead of "org.test.Person#phones", could you please explain.

gitgabrio: @jomarko 
Javadoc is misleading. I asked to @Rikkola (and confirmed with debugger): actual key is "ClassName"#"propertyName" (I was worried because if the javadoc was true, there would be no way to differentiate properties with the same name in different classes).

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/DataModelOracleUtilities.java
gitgabrio: @jomarko 
Sorry I do not follow you. That maps exists (currently) only on the server-side implementation (PackageDataModelOracleImpl inherit it from ModuleDataModelOracleImpl) but it is not propagated client-side.
So, to have it at client-side I had to
1) copy it from PackageDataModelOracleImpl to PackageDataModelOracleBaselinePayload
2) copy it from PackageDataModelOracleBaselinePayload to AsyncPackageDataModelOracleImpl

I do not have time to go further deep inside DMO features, so I avoided the "incremental" part to keep it simple; beside that, I tried to follow @Rikkola suggestions.
Probably him and @manstis could confirm my approach or show me the mistake.

kie-wb-common-widgets/kie-wb-common-ui/src/test/java/org/kie/workbench/common/widgets/client/datamodel/AsyncPackageDataModelOracleImplTest.java
Rikkola: Can we fix the JavaDoc?

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/DataModelOracleUtilities.java
Rikkola: I had doubts about the same thing and it looks to me like the answer is no, there is no need.

jomarko: Yes I asked via direct communication means gabriele to have a look on https://github.com/kiegroup/kie-wb-common/blob/f5ce1596f5fdf562cb9c50ead40b57231de6b3bd/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-api/src/main/java/org/kie/workbench/common/services/datamodel/model/PackageDataModelOracleBaselinePayload.java#L48, he found out it can be reused. I think he will update PR soon.

gitgabrio:
@jomarko @Rikkola @manstis 
As Jozef find out, this PR is not needed: the generic info I need are available with the 
AsyncPackageDataModelOracle.getParametricFieldType(final String factType, final String fieldName);

Rikkola:
@jomarko good work

jomarko:
Thank you for quick collaboration. :mag:

PR #2394, opened 2019-01-14
JBPM-8069 Stunner - it is not possible to move in/out of swimlane connected node

tiagodolphine:
Now If you have connected node you can move it outside of swimlane (or bring it inside of any swimlane).
This should not be possible with subprocesses, that should be validated as well.

I made some modifications now it seems to be working better, however, I think this part deserves a refactoring in the future to make it more simple and clear, now it is quite complicated to follow and debug the rules evaluation for this case. 

@romartin 
@hasys

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/FilteredParentsTypeMatcher.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/util/FilteredParentsTypeMatcher.java
@@ -43,21 +43,20 @@ public class FilteredParentsTypeMatcher
     private final NodeCompositePredicate parentsMatchPredicate;
     private final Optional<Element<? extends Definition<?>>> candidateParent;
     private final Optional<Node<? extends Definition<?>, ? extends Edge>> candidateNode;
+    private final Optional<Class<?>> parentType;
 
     public FilteredParentsTypeMatcher(final DefinitionManager definitionManager,
                                       final Element<? extends Definition<?>> candidateParent,
-                                      final Node<? extends Definition<?>, ? extends Edge> candidateNode) {
+                                      final Node<? extends Definition<?>, ? extends Edge> candidateNode,
+                                      final Optional<Class<?>> parentType) {
         this.candidateParent = Optional.ofNullable(candidateParent);
         this.candidateNode = Optional.ofNullable(candidateNode);
         this.parentsTypeMatchPredicate = new ParentsTypeMatchPredicate(new FilteredParentByDefinitionIdProvider(definitionManager),
-                                                                       new FilteredHasParentPredicate());
+                                                                       new FilteredHasParentPredicate(),
+                                                                       parentType.orElse(null));
         this.parentsMatchPredicate = new NodeCompositePredicate(parentsTypeMatchPredicate,
                                                                 new ParentsMatchForCandidatePredicate());
-    }
-
-    public FilteredParentsTypeMatcher forParentType(final Class<?> parentType) {
-        this.parentsTypeMatchPredicate.forParentType(parentType);
-        return this;
+        this.parentType = parentType;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


tiagodolphine:

romartin:
hey @tiagodolphine , just a clarification from your comment - the only restriction about sequence flows and nodes inside same container does not apply to all kinds of subprocesses, just applies to the embedded one, the others should behave as lanes on the containment context, if I recall correctly...

tiagodolphine:
hey @romartin good point, so it is better to validate embedded and other subprocess and all containers in general.
Thanks !

tiagodolphine:
@romartin I just made some tests and the restriction is now applying to Embedded, Event and Multiple instance subprocesses. So, this is not correct? I mean according to the spec?

hasys:
Hi @romartin, @tiagodolphine,

in my understanding Sub-process is a black box. Viz BPMN 2.0 specification (page 173):
> A **Sub-Process** is an **Activity** whose internal details have been modeled using **Activities**, **Gateways**, **Events**, and **Sequence Flows**. A **Sub-Process** is a graphical object within a **Process**, but it also can be “opened up” to show a lower-level **Process**. **Sub-Processes** define a contextual _scope_ that can be used for attribute visibility, transactional _scope_, for the handling of _exceptions_ (see page 275 for more details), of **Events**, or for _compensation_ (see page 302 for
more details).

There is one exception - you can be able to connect Node from top level directly to Start Event inside of Sub-process, but it is currently not supported by Runtime Engine, so there are no needs to implement it in the Designer.

See also: [JBPM-7056](https://issues.jboss.org/browse/JBPM-7056), [JBPM-7053](https://issues.jboss.org/browse/JBPM-7053).

So, in my opinion described behavior (you can move connected nodes in/out of swimlane but not in/out of any types of sub-processes) is sufficient.

tiagodolphine:
@hasys ok, IMO makes sense as well.

I got an issue when creating sequence flows with the quick menu when testing the generated war, I'll fix it and update the PR.

tiagodolphine:
hi @hasys yes I got these issues when testing the war now, don't know what causes the issues maybe one last minute modification before the last commit, I'll fix this and let you know when this is updated.

hasys:
@tiagodolphine, great, thank you.

Just, please, take my comment about tests into account, in my opinion it is very important to catch such an issue during tests execution. If  it is not possible now, we should find a way how to do it.

Thank you!

tiagodolphine:
@hasys yes, and as I commented before all this part IMO deserves a refactoring to make it simple and more clear, and this can allow the tests easier to cover all the cases.
For now, I'll check the existing tests to see what can be done.

tiagodolphine:

tiagodolphine:
Hi @hasys @romartin I've pushed some changes to fix the problem. Now it is ok for review.
Thanks.

romartin:
hey guys
What @hasys says it's really important, as we noticed with these changes, just by introducing some minor issue it results on most of the containment broken, so I'm agree on improving the tests from that side. Probably it can be reported as a separate task, and try to achieve it ASAP. The idea should be, on the test scope, importing the "real" rules for each domain, or at least BPMN, at execute some commands and check all operations are not failing due to rule evaluations. The issue right now is that we're testing unit pieces, which is fine, but some more high level tests with the real rules and operations should be a nice improvement and we'll be much safer on each future commit.

tiagodolphine:
Hi @romartin @hasys yeah, that is a good point we should have a way to have integration tests, I mean, testing, for instance, the scope for bpmn, simulating real use cases that the user may have, that is our "manual tests" that we do every time to avoid regressions. The unit tests are great but it is difficult to cover errors on a bigger scope, anyway, we can discuss more about it to get ideas on how to improve our tests.

Back to the PR, if you guys have a chance to review it ASAP, next two weeks I'll be on PTO I want to close this before.

Thanks guys.

romartin:
Please @hasys @tiagodolphine remember to report the issues commented above. Thanks!

hasys:
Just for the record:
@tiagodolphine, @romartin, I checked the added unit tests and they are failing when I reverted changes, so it looks like now they will catch an issue in the future. The only issue comes to my mind is - those unit tests are very hard to read and understand, we need to add some docs to it or refactor in a way it will be easier to understand.

Also, I reported mentioned issues:
* [JBPM-8141](https://issues.jboss.org/browse/JBPM-8141) - It is not possible to move to swimlane more than one node at a time
* [JBPM-8140](https://issues.jboss.org/browse/JBPM-8140) - cross-border node appended create element under swimlane
* [JBPM-8142](https://issues.jboss.org/browse/JBPM-8142) - It is not possible to append node to cross-border element

PR #2397, opened 2019-01-15
DROOLS-2697: [DMN Designer] Improve readability of node labels

danielzhe:
This PR improves de readability and UX of the node labels by doing the following:
1. It uses a new default text truncate strategy. If the text of the labels doesn't full fits inside the node, the text is truncated and appended with "...". This affects everything, not only DMN editor;

2. For the DMN editor, the font's border (stroke) are removed. Due to a to a current limitation in Lienzo that forces the stroke to be drawn, we're drawing the stroke with alpha 0 to achieve this.

3. The maximum width of the nodes are now 600px instead of 200px;

4. The text inside nodes are now truncate and resized live as the user resize the node.

**This PR depends on this PR:** https://github.com/kiegroup/lienzo-core/pull/84


danielzhe:

manstis:
@danielzhe It's also probably a good idea to remove `FontSet.FontBorderSize` property all together too (and the necessary "clean up" i.e. delete `FontBorderSize` class and all use thereof). WDYT?

danielzhe:
@manstis Sorry, I've forgot the push the commit about the dynamic resizing. Pushed now.

danielzhe:
> @danielzhe It's also probably a good idea to remove `FontSet.FontBorderSize` property all together too (and the necessary "clean up" i.e. delete `FontBorderSize` class and all use thereof). WDYT?

I don't think we need to do this now and in this PR, but I can if you think is necessary.

I'm checking the Decision Service.

manstis:
> I don't think we need to do this now and in this PR, but I can if you think is necessary.

Well.. it's not used anymore, so unless you do as a new JIRA and PR "straight afterwards" it'll probably get forgotten.. not that leaving it does any harm as we don't expose the property in the Properties Panel... so I guess a new JIRA will be enough to clean up later. I'll leave the final decision to you!

danielzhe:
@manstis I'll clean it now. I like to delete things anyway. :D

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresTextDecoratorTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.runner.RunWith;
+import org.kie.workbench.common.stunner.client.lienzo.shape.view.ViewEventHandlerManager;
+import org.mockito.Mock;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.mockito.Mockito.when;
+
+@RunWith(LienzoMockitoTestRunner.class)
+public class WiresTextDecoratorTest {
```
manstis: Do we need a test for [this](https://github.com/kiegroup/kie-wb-common/pull/2397/files#diff-30cc565a571993b66f4f513c00f0617dR274) too (that, if I'm not mistaken handles the dynamic resizing of the text during a resize)?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresTextDecorator.java
```diff
@@ -244,6 +246,10 @@ public void setTitleStrokeWidth(final double strokeWidth) {
         text.setStrokeWidth(strokeWidth);
     }
 
+    public void setTitleStrokeAlpha(double strokeAlpha) {
```
jomarko: just for consistency, could we use `final double strokeAlpha` ?

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresTextDecorator.java
```diff
@@ -62,14 +62,15 @@
...
     private ViewHandler<TextDoubleClickEvent> textDblClickEventViewHandler;
     private Text text;
-    private TextBoundsWrap textWrapper;
+    private TextTruncateWrapper textWrapper;
     private LayoutContainer.Layout currentTextLayout;
     private double width;
     private double height;
 
-    public WiresTextDecorator(final Supplier<ViewEventHandlerManager> eventHandlerManager) {
+    public WiresTextDecorator(final Supplier<ViewEventHandlerManager> eventHandlerManager,
```
romartin: I would probably, instead of adding a new `BoundinbBox` argument on the constructor, create a new method on this class, somelike like `resize(double width, double height)`. This is because in fact the `boundingBox` instance is only used to obtain the width and height values to resize to, and this is not part of the construction of the instance (the `initialize` method). Also this way you can, at some point later, call the new `resize` method to change the size for the text wrapping, without the need of initializing again. Does that makes sense?

danielzhe:
@romartin Checking the bug. Thanks.

danielzhe:
As I was suspecting, this issue that @manstis reported was not introduced by this PR, so I'm opening a new JIRA for it. @romartin 
text_label_preview_issue

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresTextDecorator.java
danielzhe: I don't see why is a issue the way I did, since the bug that I fixed with this specific change is that the size was not being set to the `textBoundaries` when the instance are created, only when the first resize happens. So the way I did we can guarantee that the size is set and IMHO is part of the `initialize()`, because we have to now the initial boundaries to initialize it otherwise it will not work as expected. Also, I don't see why we would like to resize the node but not update the bounds of the text, since we want the text to use all available space.

danielzhe:

romartin:
Please @danielzhe  can you share the link of the new issue you just reported? It's for double checking all parameters are set properly.. thanks!

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresTextDecorator.java
romartin: ok @danielzhe , agree on the explanation, it makes sense for me :+1:

ibek:
@romartin I agree with that on the tasks - truncating the text would help here. However, I think we shouldn't do it for events and gateways - 1. the labels should be under the nodes and 2. these labels explain the behaviour, so they have to be understood fully ... otherwise it makes no sense to have them at all if the labels don't add any value to the model.

manstis:
@romartin It seems that one fix does not fit all.

This PR is great for DMN however clearly not OK for all of BPMN's nodes (although OK for some).

I think we're going to need a new JIRA to be able to specify what `WiresTextDecorator` is used for different `SVGShapeView` implementations (possibly a different base sub-class, IDK). `SVGShapeView` is instantiated by the `SVGShapeViewBuilder` defined for the _domain_).

danielzhe:
@romartin Here is the JIRA: https://issues.jboss.org/projects/DROOLS/issues/DROOLS-3539

romartin:
Hey @danielzhe @manstis @ibek @wrnkld 

So as this fits for DMN, but not completely for BPMN, IMO there are two ways to go:
* Provide different wrapping strategies for each domain, or even shape (will require some effort...)
* Use the new truncate strategy, but adding the proposed tooltip or something similar, making this way all the labels readable anyway (I think this will be faster/easier than the option above, as it's only about showing some text with some delay on the canvas...)

Well not sure, how you think should we proceed guys? Thx

manstis:
@danielzhe @romartin @ibek @etirelli 

Summary: The PR to make DMN's nodes more readable is not compatible with BPMN's needs.

Proposal: Split this PR into 2: One allowing the Font to have zero stroke width (improves DMN vastly). The other to add support for different `TextWrapper` for different types of node in the Graph. DMN would then get it's new `TextTruncateWrapper` and BPMN could have different _wrappers_ for Events, Gateways and other types of node.

I guess I'm looking for the stakeholders, @ibek @etirelli, to approve.

ibek:
I like your proposal @manstis ... it will enable us to truncate nodes where it makes sense ... and we can finish this PR to improve the font in DMN, that's a win :)

romartin:
that sounds great for me @manstis ! :+1: 
Go ahead guys, let's split this PR and make the text wrapping declaration possible for different domains :+1:

danielzhe:

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/wires/ext/WiresTextDecoratorTest.java
manstis: @danielzhe Waiting for your reply to this.

danielzhe: Oh, sorry, I didn't saw this one. I'm writing the test.

romartin:
Hey @manstis  @danielzhe 

Not sure if I can approve yet... I mean, looking at the code, it will be still truncating the texts for BPMN events and gateways, right? So merging this will result in some BPMN labels not completely legible? 

If that's the case, I don't say we cannot merge this, but would be nice if you can work on the ability to specify text wrapping by bean/domain as for next 7.3 release, so we can ensure that labels for BPMN at least will remain as before.

Does that makes sense? What's the way to follow here?

Thanks!

manstis:
@romartin No, you are wrong. This PR now uses the same wrapper as is on master but simply removes font border stroke width for DMN.

PR #2398, opened 2019-01-15
JBPM-7790 - Process instance diagram zoom controls. Fix for standalone webapps

manstis:
This PR fixed the DMN and Stunner _standalone_ webapps following [this](https://github.com/kiegroup/kie-wb-common/commit/5099ea6c052402cb6eda6f7ba8e69d335f70f554).

The _canvas_ needs a unbreakable chain of `RequiresResize` from AppFormer to the editor. Furthermore the size of the _canvas_ is taken from the pixel dimensions of its parent. The _standalone_ webapps had a break in pixel sizes in the `Widget` inbetween AppFormer's `PageViewImpl` (a tabs content) and Stunner's `SessionPresenterView` (the graph editor view).

Before JBPM-7790's CSS fix was merged DMN had a CSS selector for `.tab-content > div > div > div { height: 100% }` that had the same effect (but was wrong as it clearly broke JBPM console).

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp/src/main/java/org/kie/workbench/common/dmn/showcase/client/screens/SessionScreenView.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp/src/main/java/org/kie/workbench/common/dmn/showcase/client/screens/SessionScreenView.java
@@ -23,11 +23,5 @@ public interface SessionScreenView extends IsWidget {
 
     void showScreenView(final IsWidget view);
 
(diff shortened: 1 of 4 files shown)


manstis:
@karreiro Did you compile all of `kie-wb-common` (or at least "Stunner")?

manstis:
@karreiro I just re-tested (compiling "Stunner" first) and it appears to (still) work for me. Please confirm you compiled the required files and tested.. if it still does not work then I need to scratch my head more!

manstis:
@jomarko Just awaiting your review please.

PR #2403, opened 2019-01-17
refactoring of AbstractToolbarCommand: direct message declaration rep…

treblereel:
…laced

by translation mechanism, it contains only a default value but can be easily
replaced with localized implementation.
Redundant public and static declorations removed from StunnerWidgetsConstants.


kiereleaseuser3:

romartin:
Hey @treblereel , thanks for your first collaboration! :+1: 

Let you know that we use to link our commit messages to tickets on JIRA. As this change is about i18n, there is no need to create a new ticket as we already have the [JBPM-5622](https://issues.jboss.org/browse/JBPM-5622), but should be great if you can reword you commit and add the prefix "JBPM-5622: ...". Also this is apply for future tasks, for example, if you find a bug, just report it (previously check not already reported), and then you can work on it and use the ticket number in your commit messages. 

FYI This is the [Stunner's JIRA agile board](https://issues.jboss.org/secure/RapidBoard.jspa?projectKey=JBPM&rapidView=3838&view=planning), where you can find all the tickets, sprints, etc, plus create new ones.

Thanks!

treblereel:
retest this please

romartin:

hasys:
Jenkins retest please

romartin:
Hey @hasys , merging as it does not hurts.. .but good point, this is the method in the abstraction, so any toolbar button will show it, in case it needs confirm, but I guess the only one using this is now overriding this method...

PR #2405, opened 2019-01-17
JBPM-7613: case modeler - fix toolbar buttons

xiezhang7:


kiereleaseuser3:

xiezhang7:
ok to test

romartin:
ok to test

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/toolbar/command/CaseManagementClearToolbarCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import org.kie.workbench.common.stunner.core.client.i18n.ClientTranslationService;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.ClearSessionCommand;
+import org.kie.workbench.common.stunner.core.util.DefinitionUtils;
+
+@Dependent
+@CaseManagementEditor
+public class CaseManagementClearToolbarCommand extends ClearToolbarCommand {
+
+    @Inject
+    public CaseManagementClearToolbarCommand(DefinitionUtils definitionUtils,
```
romartin: This bean is not really necessary - test deleting it, I think it will work anyway as expected.  Lemme give you some quick explanation:

- The toolbar is actually loading the `ClearToolbarCommand`, so the clear button in all domains will appear (BPMN, CM,etc)
- The implementation for [ClearToolbarCommand](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/toolbar/command/ClearToolbarCommand.java) already takes a `ManagedInstance<ClearSessionCommand>` as an argument, and if you [look at the implementation](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/toolbar/command/AbstractToolbarCommand.java#L61), this is because it obtain the right instance, if any, for each domain, otherwise it uses the default one

So this is why there is not need to override this class only to use a different `ClearSessionCommand`, as if your custom instance is being annotated with the `@CaseManagementEditor`, it will use it.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/session/command/impl/CaseManagementClearSessionCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
+import org.kie.workbench.common.stunner.core.client.command.CanvasCommandFactory;
+import org.kie.workbench.common.stunner.core.client.command.SessionCommandManager;
+import org.kie.workbench.common.stunner.core.client.session.Session;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.ClearSessionCommand;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.ClearSessionCommandExecutedEvent;
+
+@Dependent
+@CaseManagementEditor
+public class CaseManagementClearSessionCommand extends ClearSessionCommand {
```
romartin: mmm if I'm correct, you only need to create this class because of you need to inject a different `CanvasCommandFactory` rather than the default one, right?

In that case, this implementation is fine... but we could try to improve it. Soo looking at my review comment in the `CaseManagementClearToolbarCommand` class, where it explains how at runtime it injects the right (CD)) instance based on each domain, could you please see if it's possible to improve the actual `ClearSessionCommand` (or whatever parent) in order to, instead of injecting a `CanvasCommandFactory`, it injects a `ManagedInstance<CanvasCommandFactory>`, so at runtime it looks for the right instance based on the annotations. 

This way there will be no need to create this class anymore. WDYT?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/CaseManagementClearCommand.java
```diff
@@ -0,0 +1,54 @@
...
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+
+public class CaseManagementClearCommand extends ClearCommand {
+
+    String rootUUID;
+    Shape rootShape;
+
+    @Override
+    protected AbstractCanvasCommand newCanvasCommand(AbstractCanvasHandler context) {
+        rootUUID = context.getDiagram().getMetadata().getCanvasRootUUID();
```
romartin: doing these initialization (`rootUUID`, `rootShape`) on the newCanvasCommand results a bit tricky... isn't it?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/CaseManagementCanvasCommandFactory.java
```diff
@@ -49,6 +49,22 @@ public CaseManagementCanvasCommandFactory(final ManagedInstance<ChildrenTraverse
               viewTraverseProcessors);
     }
 
+    @SuppressWarnings("unchecked")
+    public static int getChildIndex(final Node parent,
```
romartin: well this method has nothing to do with a canvas neither a factory... doesn't make sense to move it somewhere?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/graph/CaseManagementAddChildNodeGraphCommand.java
```diff
@@ -28,14 +28,24 @@
 @SuppressWarnings("unchecked")
 public class CaseManagementAddChildNodeGraphCommand extends org.kie.workbench.common.stunner.core.graph.command.impl.AddChildNodeCommand {
 
-    private final int index;
+    private final OptionalInt index;
```
romartin: The `index` is now optional?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/canvas/CaseManagementDeleteCanvasNodeCommand.java
```diff
@@ -42,4 +43,9 @@ public CaseManagementDeleteCanvasNodeCommand(Node candidate,
     protected AbstractCanvasCommand createUndoCommand(Node parent, Node candidate, String ssid) {
         return new CaseManagementAddChildNodeCanvasCommand(parent, candidate, ssid, index);
     }
+
+    @Override
+    protected void parentApplyElementMutation(AbstractCanvasHandler context) {
+        // no parent applyElementMutation
```
romartin: doesn't makes sense for me... if there no parent, why not setting it to null, and just let the parent class do the job (it seems to already check nullity for parent)?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/toolbar/command/CaseManagementCutToolbarCommand.java
```diff
@@ -0,0 +1,39 @@
...
+import org.kie.workbench.common.stunner.core.client.i18n.ClientTranslationService;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.CutSelectionSessionCommand;
+import org.kie.workbench.common.stunner.core.util.DefinitionUtils;
+
+@Dependent
+@CaseManagementEditor
+public class CaseManagementCutToolbarCommand extends CutToolbarCommand {
+
+    @Inject
+    public CaseManagementCutToolbarCommand(DefinitionUtils definitionUtils,
```
romartin: This class is not necessary - same as comment as for `CaseManagementClearToolbarCommand`

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/toolbar/command/CaseManagementPasteToolbarCommand.java
```diff
@@ -0,0 +1,39 @@
...
+import org.jboss.errai.ioc.client.api.ManagedInstance;
+import org.kie.workbench.common.stunner.client.widgets.toolbar.command.PasteToolbarCommand;
+import org.kie.workbench.common.stunner.cm.qualifiers.CaseManagementEditor;
+import org.kie.workbench.common.stunner.core.client.i18n.ClientTranslationService;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.PasteSelectionSessionCommand;
+import org.kie.workbench.common.stunner.core.util.DefinitionUtils;
+
+@Dependent
+@CaseManagementEditor
+public class CaseManagementPasteToolbarCommand extends PasteToolbarCommand {
```
romartin: This class is not necessary - same as comment as for `CaseManagementClearToolbarCommand`

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/session/command/impl/CaseManagementCutSelectionSessionCommand.java
```diff
@@ -0,0 +1,43 @@
...
+import org.kie.workbench.common.stunner.core.client.command.SessionCommandManager;
+import org.kie.workbench.common.stunner.core.client.session.Session;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.CopySelectionSessionCommand;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.CutSelectionSessionCommand;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.CutSelectionSessionCommandExecutedEvent;
+import org.kie.workbench.common.stunner.core.client.session.command.impl.DeleteSelectionSessionCommand;
+
+@Dependent
+@CaseManagementEditor
+public class CaseManagementCutSelectionSessionCommand extends CutSelectionSessionCommand {
```
romartin: here also we could make this class unnecessary by improving the `CutSelectionSessionCommand`, so same comment as `CaseManagementClearSessionCommand`, instead of injecting `DeleteSelectionSessionCommand` you could inject a  `ManagedInstance<DeleteSelectionSessionCommand>` and do the lookup at runtime... and finally remove this class

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/graph/CaseManagementAddChildNodeGraphCommand.java
xiezhang7: Hi @romartin , the `index` is optional for [CaseManagementCloneNodeCommand](https://github.com/kiegroup/kie-wb-common/blob/6916db21c64ce44d56996981d24d46e3dc1a5cc1/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/graph/CaseManagementCloneNodeCommand.java#L70). If the `index` is not presented, the node will added to the [end](https://github.com/kiegroup/kie-wb-common/blob/540191cbb89061e8520fdcdc3a93336243f1ab1e/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/graph/CaseManagementSetChildNodeGraphCommand.java#L123).

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/toolbar/command/CaseManagementClearToolbarCommand.java
xiezhang7: This bean is deleted.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/session/command/impl/CaseManagementClearSessionCommand.java
xiezhang7: This class is deleted. 
Obtain the correct instance of `CanvasCommandFactory` in [ClearSessionCommand](https://github.com/kiegroup/kie-wb-common/blob/4094fe9b797f9c134622d7ce5c1f545201b9cf72/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/ClearSessionCommand.java#L82).

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/CaseManagementClearCommand.java
xiezhang7: Removed the overriding of `newCanvasCommand` method.
Moved the initialization code to the [execute](https://github.com/kiegroup/kie-wb-common/blob/4094fe9b797f9c134622d7ce5c1f545201b9cf72/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/CaseManagementClearCommand.java#L32) method.

xiezhang7:

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/explorer/tree/TreeExplorerView.java
```diff
@@ -55,7 +55,7 @@ public TreeExplorerView(final Tree<TreeItem> tree) {
                      final HandlerRegistration handlerRegistration
     ) {
         this.presenter = presenter;
-        this.uiBinder = uiBinder;
+        TreeExplorerView.uiBinder = uiBinder;
```
romartin: well I know that's not your fault... but the use of this `static ViewBinder uiBinder` is not the correct one. Doesn't make sense, in the constructor, to update this static field.... I think it's done just for being able to inject a mock instance when testing... anyway, let's fix this once having more time.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/command/impl/ReverseCommand.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.kie.workbench.common.stunner.core.command.impl;
+
+import org.kie.workbench.common.stunner.core.command.Command;
+import org.kie.workbench.common.stunner.core.command.CommandResult;
+
+/**
+ * Reverse the execute and undo operations of the given command
+ */
+public class ReverseCommand<T, V> implements Command<T, V> {
```
romartin: Not sure if really makes sense... I mean, assuming the idea that a `Command` is a state-full object, it becomes usual than a command updates its state on the `execute` method, and the `undo` method restores the state.  So if the command has not been executed, doesn't make sense to undo it, and this command is basically what it does...
What's the goal for this?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/PasteSelectionSessionCommand.java
```diff
@@ -155,10 +164,11 @@ private void handleCtrlV(final Key[] keys) {
             CommandResult<CanvasViolation> finalResult;
             if (wasNodesDeletedFromGraph()) {
                 //in case of a cut command the source elements were deleted from graph, so first undo the command to take node back into canvas
-                clipboardControl.getRollbackCommands().forEach(command -> command.undo(getCanvasHandler()));
-                finalResult = executeCommands(nodesCommandBuilder, processedNodesCountdown);
+                clipboardControl.getRollbackCommands().forEach(command -> nodesCommandBuilder.addFirstCommand(new ReverseCommand(command)));
```
romartin: all these lines result a bit tricky... I mean, you're iterating over the `getRollbackCommands()` twice, and both times adding them into the same `nodesCommandBuilder`... and this is probably because you're wrapping the commands into a `ReverseCommand` as well.... (as commented doesn't makes much sense for me...)
I think that this should be changed as:

1. Instead of wrapping the commands into a `ReverseCommand` and adding into the list using `addFirstCommand` - I would just iterate over these commands and call the `undo` method, as they've already executed isn't it?
2. The second iteration is already fine (line 171), iterate and execute...

romartin:
Also please @LuboTerifaj - take into account on the review that this changes also **affects both BPMN and DMN**, so please be sure to test the toolbar buttons on all editors so we don't break any one... Thanks!!

LuboTerifaj:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/session/command/impl/CaseManagementCutSelectionSessionCommand.java
xiezhang7: Hi @romartin , this class was created to inject an instance of  `DeleteSelectionSessionCommand` which picks the right instance of `CanvasCommandFactory` for case modeler. 
Now that `DeleteSelectionSessionCommand` has been modified to [inject the right instance](https://github.com/kiegroup/kie-wb-common/blob/4094fe9b797f9c134622d7ce5c1f545201b9cf72/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/DeleteSelectionSessionCommand.java#L91) of `CanvasCommandFactory` at runtime, `CaseManagementCutSelectionSessionCommand` is not needed anymore. 
The parent class of `CaseManagementCutSelectionSessionCommand` doesn't use `CanvasCommandFactory` directly, it only uses `DeleteSelectionSessionCommand` which has only one type of instance. So there is no need to change the parent class to use `ManagedInstance`.

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/explorer/tree/TreeExplorerView.java
xiezhang7: I'll fix this in [PR#2420](https://github.com/kiegroup/kie-wb-common/pull/2420).

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/command/impl/ReverseCommand.java
xiezhang7: Hi @romartin , 

I need this class to create a `CompositeCommand`. 
By executing this `CompositeCommand`, I want to 

1. `undo` some sub-commands
2. `execute` other sub-commands

as one operation.

The assumption is the sub-commands in 1 are executed already before executing the `CompositeCommand`.

In order to do 1, I wrap the sub-commands with `ReverseCommand`, and then add these `ReverseCommand` to the `CompositeCommand`. So that, when the `ReverseCommand` is executed, the `undo` of the sub-commands is invoked.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/PasteSelectionSessionCommand.java
xiezhang7: Hi @romartin , 

Both `undo` commands and re-`execute` commands are added to a `CompositeCommand`, and the `CompositeCommand` is executed as one operation. So that the user only need to click the "undo" button once, and the entire "paste" operation will be undone.

If the `undo` commands are iterated and executed separately, the user will have to click the "undo" button multiple times in order to undo the whole "paste" operation. I noticed this kind of problem, so I changed the code to execute one `CompositeCommand` only.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/command/impl/ReverseCommand.java
romartin: Ok I see... in some way the use of the API due to this command results strange but I'm agree on adding this now, it achieves the goal you need and also it's easy to revert at some point, anyway it's a good way to move forward, great job! :+1:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/command/impl/PasteSelectionSessionCommand.java
romartin: ok, as already commented, let's do this way :+1:

xiezhang7:
> Hello @xiezhang7
> 
> I've looked also at Process and DMN modeler, and I have found one regression. If a node is selected and there is also another one right below this node, after paste, the new node is created, but it shouldn't because nodes cannot be covered by each other.
> "Command execution failed. Reason/s: [1] (ERROR) Containment operation not allowed" should be displayed.

Hi @LuboTerifaj , I think this is an existing issue, not a regression. Because it could also be reproduced with a build from the master branch. I don't think the code related to this issue is modified in this pull request. Could you please double check? Thanks.

LuboTerifaj:
Hi @xiezhang7,

I have double checked the issue, and it is not present on master branch. "Containment operation not allowed" message is displayed when trying to paste the node to already occupied position on the canvas. This behaviour is correct and expected.

Thank you.

LuboTerifaj:
Reported issues mentioned above, except last regression, which should be fixed in this PR:

- [JBPM-8178 - Enable paste node between two nodes](https://issues.jboss.org/browse/JBPM-8178)
- [JBPM-8179 - Nodes multi-selection doesn't work](https://issues.jboss.org/browse/JBPM-8179)
- [JBPM-8180 - Esc cancels only node's ghost](https://issues.jboss.org/browse/JBPM-8180)

xiezhang7:
Hi @LuboTerifaj ,

I think "Containment operation not allowed" means a node cannot be added inside another node, and it is not triggered because nodes are covered.

I can get the error message that you mentioned by doing this:

> 1. select a node
> 2. click "copy" button
> 3. **de-select the node (by selecting other nodes or creating new node or doing something else), then re-select the node**
> 4. click "paste" button
> 5. a new node is created and added to the selected node, which is not allowed, and the error message is displayed 

Even if I select another node (which is not allowed to contain the new node) and paste the new node to it, then I will still get the "Containment operation not allowed" error message. 

However, I also noticed this kind of existing behavior when working on this task:

> 1. select a node
> 2. click "copy" button
> 3. **click "paste" button immediately without doing anything else (i.e. without de-selecting the node)**
> 4. a new node is created and added inside **the parent of the selected node**, and the new node is placed below the selected node 

If there is already a node below the selected node, this node will be covered by the new node created in step 4. **This is like the issue you mentioned.**

**The difference between the two cases is whether the new node is pasted to the parent or the selected node.** 
And these two cases are consistent and reproducible with both master branch and this pull request in my local testing environment, which means maybe this kind of behavior is not new.

Please check the video clips I recorded with master branch and this pull request:
**_Master branch:_**
master branch
**_This pull request:_**
This pull request

@LuboTerifaj please let me know if I did the correct steps to reproduce the issue. Thanks!

LuboTerifaj:
Reported issue: [JBPM-8188 - Pasting node covers another one](https://issues.jboss.org/browse/JBPM-8188)

xiezhang7:
Thanks @LuboTerifaj !

romartin:
great job @xiezhang7 @LuboTerifaj , thanks!

PR #2407, opened 2019-01-18
DROOLS-3222 Test duration should not start with zeros

kkufova:
@danielezonca, @Rikkola, could you please take a look?

I know this solution is not exactly perfect, but it cannot be achieved directly using `DateTimeFormat` and it is not possible to use `SimpleDateFormat` because of GWT.

Before:

before

After:

after


kkufova:
@danielezonca, @Rikkola, you're right. Tests added, please take a look.

kkufova:
@manstis, could you please merge this?

PR #2411, opened 2019-01-21
DROOLS-3399: [DMN Designer] Data Types - Disable the global 'Save' button when edit mode is enabled

karreiro:
See: https://issues.jboss.org/browse/DROOLS-3399

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/609
- https://github.com/kiegroup/kie-wb-common/pull/2411

---

There's a corner case on this PR. When the user opens two DTs on edit mode and saves the first one, the "Save" button is enabled. However, since we'll prohibit the parallel edition ([DROOLS-3015](https://issues.jboss.org/browse/DROOLS-3015)), I don't think we need to fix this here.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListItem.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListItem.java
@@ -23,11 +23,13 @@ import java.util.stream.Stream;
 
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.Dependent;
+import javax.enterprise.event.Event;
 import javax.inject.Inject;
 
 import elemental2.dom.HTMLElement;
 import org.kie.workbench.common.dmn.client.editors.types.common.DataType;
 import org.kie.workbench.common.dmn.client.editors.types.common.DataTypeManager;
+import org.kie.workbench.common.dmn.client.editors.types.listview.common.DataTypeEditModeToggleEvent;
 import org.kie.workbench.common.dmn.client.editors.types.listview.common.SmallSwitchComponent;
 import org.kie.workbench.common.dmn.client.editors.types.listview.confirmation.DataTypeConfirmation;
 import org.kie.workbench.common.dmn.client.editors.types.listview.constraint.DataTypeConstraint;
@@ -54,6 +56,8 @@ public class DataTypeListItem {
 
     private final DataTypeConfirmation confirmation;
 
+    private final Event<DataTypeEditModeToggleEvent> editModeToggleEvent;
+
     private DataType dataType;
 
     private int level;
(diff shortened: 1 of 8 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-project-client/src/main/java/org/kie/workbench/common/dmn/project/client/editor/DMNDiagramEditor.java
```diff
@@ -245,6 +250,20 @@ protected String getEditorIdentifier() {
         return EDITOR_ID;
     }
 
+    public void onDataTypeEditModeToggle(final @Observes DataTypeEditModeToggleEvent event) {
+        setTimeout(getOnDataTypeEditModeToggleCallback(event), 250);
```
jomarko: Just curious, why we need to use this `setTimeout` instead of execute `EditModeToggleCallback` immediately when the event is observed?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListItem.java
```diff
@@ -54,6 +56,8 @@
 
     private final DataTypeConfirmation confirmation;
 
+    private final Event<DataTypeEditModeToggleEvent> editModeToggleEvent;
```
jomarko: I think wee need to fire this event from different class, current implementation causes save button is enabled even if some data type entry remains in edit mode:
- Add two new data type entries
- Cancel edit mode of the first entry
- Save button will be enabled and second data type entry will stay in edit mode

screenshot from 2019-01-21 15-32-32

karreiro: @jomarko Yeah, this is the corner case that I mentioned in the PR description, right?

If it is, I don't think we need to fix it right now, since we'll prohibit the parallel edition on [DROOLS-3015](
https://issues.jboss.org/browse/DROOLS-3015).

Thanks.

kie-wb-common-dmn/kie-wb-common-dmn-project-client/src/main/java/org/kie/workbench/common/dmn/project/client/editor/DMNDiagramEditor.java
karreiro: @jomarko When the asset is not dirty and it's updated, an event refreshes the whole widget and enables the button again (so our disabled state is lost).

To avoid the undesired behaviour, we can (a) look at this event and add an extra logic to handle this, or we can (b) just delay our command. I prefer the second alternative to keep the code cleaner.

karreiro:
@jomarko I've added the comment in an individual commit.

PR #2417, opened 2019-01-22
JBPM-7403: Stunner - Add support for BPMNDiagram resolution attribute

romartin:
Hey @hasys 

See https://issues.jboss.org/browse/JBPM-7403

Also sharing the example process I used for testing this stuff, hope it helps :+1: 

[aris-resolution-v2.zip](https://github.com/kiegroup/kie-wb-common/files/2784974/aris-resolution-v2.zip)

Thanks!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/tostunner/properties/BasePropertyReader.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/tostunner/properties/BasePropertyReader.java
@@ -21,8 +21,11 @@ import java.util.List;
 import java.util.Optional;
 
 import org.eclipse.bpmn2.BaseElement;
-import org.eclipse.bpmn2.di.BPMNPlane;
+import org.eclipse.bpmn2.di.BPMNDiagram;
+import org.eclipse.bpmn2.di.BPMNLabel;
+import org.eclipse.bpmn2.di.BPMNLabelStyle;
 import org.eclipse.bpmn2.di.BPMNShape;
+import org.eclipse.dd.dc.Font;
 import org.kie.workbench.common.stunner.bpmn.backend.converters.customproperties.CustomElement;
 import org.kie.workbench.common.stunner.bpmn.definition.property.background.BackgroundSet;
 import org.kie.workbench.common.stunner.bpmn.definition.property.background.BgColor;
@@ -39,16 +42,21 @@ import org.kie.workbench.common.stunner.bpmn.definition.property.font.FontSet;
 import org.kie.workbench.common.stunner.bpmn.definition.property.font.FontSize;
 import org.kie.workbench.common.stunner.core.graph.content.Bounds;
 
-public abstract class BasePropertyReader {
+public class BasePropertyReader {
 
     protected final BaseElement element;
     protected final BPMNShape shape;
-    protected final BPMNPlane plane;
+    protected final BPMNDiagram diagram;
+    protected final double resolutionFactor;
 
-    public BasePropertyReader(BaseElement element, BPMNPlane plane, BPMNShape shape) {
+    public BasePropertyReader(final BaseElement element,
+                              final BPMNDiagram diagram,
+                              final BPMNShape shape,
+                              final double resolutionFactor) {
         this.element = element;
-        this.plane = plane;
+        this.diagram = diagram;
         this.shape = shape;
+        this.resolutionFactor = resolutionFactor;
     }
 
     public String getDocumentation() {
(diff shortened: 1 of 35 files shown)


romartin:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/tostunner/properties/BasePropertyReader.java
```diff
@@ -63,21 +71,50 @@ public String getDescription() {
     }
 
     public FontSet getFontSet() {
```
hasys: Since we have no plans to add Font configuration ability to the Stunner, does it make sense do not parse user font and just use default one?

Right now fonts for imported process looks different for imported nodes and new ones. But it is just until you save/reopen process. After save and reopen - default font will be used.

Before save and reopen:
image

After Save and reopen:
image

romartin: good catch... mmmm

romartin:
Hey @hasys 

Good Qs! I know there are some reasons to apply the transforms you mention, and probably we should go that way, but lemme give you some feedback as well and then let's decide what to do. Anyway good job, thanks!

There are some implications when using fixed sizes for some shapes, so I was trying to see if just relying on the `resolution` attribute (and so applying a certain scale/translate factor) should be enough. These are some reasons for this:

- By fixing sizes for certain shapes we'll broke the whole source diagram layout too
- The sizes for the shapes (gateways, events, etc) are given by the SVG declaration. It means that UX can change these sizes only by updating the svg files. If we need to use fixed sizes now on the bpmn parsing, it probably will imply having to maintain some values `hardcoded` on the code as well...  so changes will become harder
- All the control points and connections in the bpmn process are based on absolute locations, which means that if we apply fixed sizes for certain shapes we also have to consider handling all these CPs and connections, in order to re-position each one, meanwhile by applying a fixed factor for the whole process parsing results easier to maintain everywhere

In fact, in order to make these transformations a bit more "configurable", it's possible to set a VM parameter for it ( `-Dorg.kie.stunner.bpmn.marshaller.resolution` ).

On the other hand I understand these implications you mention, and also IIRC we said to go this way, by fixing certain sizes...

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/tostunner/properties/BasePropertyReader.java
romartin: lemme do a few tries again by skipping this...

romartin:

romartin:
Hey @hasys 

Well finally applied the approach you mention (also proposed on the call...). Now sizes for gateways and events are fixed, also text style is not being parsed anymore. 

So the thing is that now all shapes look same sizes, but I assume that as a task for the migration process, it's necessary to adjust the layout for some shapes, see the result now:

screenshot from 2019-01-24 02-57-52

Also added tests. Lemme know what you think.. thanks!!

hasys:
Hi @romartin,

good points, thank you!

> By fixing sizes for certain shapes we'll broke the whole source diagram layout too

I guess, it is better, because alternative is different size of new and old nodes without any easy way to fix it. Yes, you can fix *.bpmn file manually, but is not an easy way. Also, it is a possibility to create new node instead of old one, but you also should not forget to copy all node properties, and there are much more space to make a type/forget something/make another mistake.

On the other hand, as improvement of this task, I guess it is possible to calculate how to move this event/gateway to keep layout unchanged.

> The sizes for the shapes (gateways, events, etc) are given by the SVG declaration. It means that UX can change these sizes only by updating the svg files. If we need to use fixed sizes now on the bpmn parsing, it probably will imply having to maintain some values hardcoded on the code as well... so changes will become harder

In my opinion default size of nodes shouldn't be changed often, or we will endup with point one of this discussion (we will broke the whole source diagram layout) or new/old nodes will have different sizes.

Also, I guess, it is possible to parse SVG declaration and use values from it as default size. So, no hardcoded values.

> All the control points and connections in the bpmn process are based on absolute locations, which means that if we apply fixed sizes for certain shapes we also have to consider handling all these CPs and connections, in order to re-position each one, meanwhile by applying a fixed factor for the whole process parsing results easier to maintain everywhere

I am not sure here, but, probably, when we will apply enhancement from first point (recalculate position of even/gateway based on it's change) it can be a solution for this as well.

> In fact, in order to make these transformations a bit more "configurable", it's possible to set a VM parameter for it (` -Dorg.kie.stunner.bpmn.marshaller.resolution` ).

If I correct `-Dorg.kie.stunner.bpmn.marshaller.resolution` property is a global one, so you will need to restart you server everytime you want to import process from different tool.

So yes, I guess your fixes are great, thank you!

For now to me it looks like there are failing tests in our marshaller due to different size of nodes in old designer. When tests will be fixed, I guess this PR is ready, thank you!

romartin:
Hey @hasys 

Yeah you're right on all your comments, good points :+1:  

So I'm agree that this can be improved, for example by trying to fix the layout issues (due to fixed sizes), or parsing the SVG to avoid hardcoding values... but I'm not really sure if make sense spending much time on  coding all there, so if you agree, let's report a new ticket for these improvements and let's go ahead with this for now.  As we have a 7.3 demo session soon, I'll get feedback also from others and if we think some improvement should be included, we're still on time.

Sooo trying to fix the compile issues and ready to review again. Thanks!

romartin:

romartin:
Jenkins please retest this

romartin:

romartin:
rebased

romartin:
Jenkins please retest this

romartin:

PR #2419, opened 2019-01-23
DROOLS-3517: [DMN Designer] Data Types - Constraints - Constraint property

danielzhe:


kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/ConstraintType.java
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.api.definition.v1_1;
+
+public enum ConstraintType {
```
jomarko: I think we do not need the private field in this case, see the code below how we could simplify this class:
```java
public enum ConstraintType {
    ENUMERATION,
    EXPRESSION,
    RANGE;

    public static final String CONSTRAINT_KEY = "constraintType";

    public static ConstraintType fromString(final String value) {

        if (ENUMERATION.name().equalsIgnoreCase(value)) {
            return ENUMERATION;
        }

        if (EXPRESSION.name().equalsIgnoreCase(value)) {
            return EXPRESSION;
        }

        if (RANGE.name().equalsIgnoreCase(value)) {
            return RANGE;
        }

        return null;
    }
}
```

kie-wb-common-dmn/kie-wb-common-dmn-api/src/test/java/org/kie/workbench/common/dmn/api/definition/v1_1/ConstraintTypeTest.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.api.definition.v1_1;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConstraintTypeTest {
```
jomarko: Please add also a test that assert `null` is returned in cases like `ConstraintType.fromString("unknownvalue")`

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/InputClauseUnaryTests.java
```diff
@@ -54,15 +54,20 @@
     @FormField(afterElement = "description", labelKey = "text")
     protected Text text;
 
+    protected ConstraintType constraintType;
```
jomarko: Just interested why `protected` access is needed, is it due to an `errai` binding?

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/ConstraintTypeFieldPropertyConverter.java
```diff
@@ -0,0 +1,37 @@
...
+        } else {
+            return ConstraintType.fromString(constraintType);
+        }
+    }
+
+    public static String dmnFromWB(final ConstraintType constraintType) {
+        if (constraintType == null) {
+            return null;
+        }
+        return constraintType.value();
```
jomarko: According to my previous comment, this could be `constraintType.name().toLowerCase()`

jomarko: Or just implement `value()` in this way.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/InputClauseUnaryTestsPropertyConverterTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String CONSTRAINT_TYPE_TEXT = "enumeration";
+
+    @Mock
+    private org.kie.dmn.model.api.UnaryTests dmnUnary;
+
+    @Mock
+    private Map<QName, String> additionalAttributes;
+
+    @Before
+    public void setup() {
```
jomarko: I think we should take better advantage of this `setup()` method, same holds true for `OutputClauseUnaryTestsPropertyConverterTest`.

Currently we have `setup()` designed with too strong presumption , there will be just one test method. if we rewrite the `setup()` method slightly, we can test all constraint types, attaching code in comment below.

jomarko:
```java
@RunWith(MockitoJUnitRunner.class)
public class InputClauseUnaryTestsPropertyConverterTest {

    private static final String ID = "thisId";
    private static final String TEXT = "1,2,3";

    private static final QName key = new QName(DMNModelInstrumentedBase.Namespace.KIE.getUri(),
                                               ConstraintType.CONSTRAINT_KEY,
                                               DMNModelInstrumentedBase.Namespace.KIE.getPrefix());

    @Mock
    private org.kie.dmn.model.api.UnaryTests dmnUnary;

    @Mock
    private Map<QName, String> additionalAttributes;

    @Before
    public void setup() {

        when(dmnUnary.getId()).thenReturn(ID);
        when(dmnUnary.getAdditionalAttributes()).thenReturn(additionalAttributes);
        when(dmnUnary.getText()).thenReturn(TEXT);
    }

    @Test
    public void testWbFromDMNEnumeration() {
        testWbFromDMN(ConstraintType.ENUMERATION);
    }

    @Test
    public void testWbFromDMNExpression() {
        testWbFromDMN(ConstraintType.EXPRESSION);
    }

    @Test
    public void testWbFromDMNRange() {
        testWbFromDMN(ConstraintType.RANGE);
    }

    private void testWbFromDMN(final ConstraintType constraintType) {
        when(additionalAttributes.getOrDefault(key, "")).thenReturn(constraintType.name().toLowerCase());

        final InputClauseUnaryTests inputClause = InputClauseUnaryTestsPropertyConverter.wbFromDMN(dmnUnary);
        assertEquals(inputClause.getConstraintType(), constraintType);
        assertEquals(inputClause.getId().getValue(), ID);
        assertEquals(inputClause.getText().getValue(), TEXT);
    }
}
```

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/UnaryTestsPropertyConverterTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.kie.workbench.common.dmn.api.property.dmn.Id;
+import org.kie.workbench.common.dmn.api.property.dmn.Text;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class UnaryTestsPropertyConverterTest {
```
jomarko: Also here I would expect we test not just `ENUMERTION`

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/DataTypeConstraintModal.java
```diff
@@ -85,17 +88,30 @@ void clearAll() {
...
     void load(final String type,
-              final String value) {
-        constraintValue = value;
-        prepareView(type, value);
+              final String value,
+              final ConstraintType constraintType) {
+
+        this.constraintValue = value;
+
+        if (!Objects.isNull(value) && !value.isEmpty() && constraintType == null) {
```
jomarko: We can shorten this condition:
```java
 (StringUtils.nonEmpty(value) && constraintType == null) 
```

FQN for `StringUtils` is `org.kie.workbench.common.stunner.core.util.StringUtils`

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/common/DataTypeManagerTest.java
```diff
@@ -123,6 +124,16 @@ public void setup() {
         manager = spy(new DataTypeManagerFake());
     }
 
+    @Test
+    public void testWithConstraintType(){
```
jomarko: Just Formatting, missing space between brackets, should be `testWithConstraintType() {`

jomarko:
@danielzhe Hi, I tried manually import asset form external tool, in that case the constraint type seems to be loaded correctly. however when I try save and reopen scenario for the asset created with our tool, the constraint type is lost. Could you please read my scenario and tell if it is valid scenario or it is not in the scope of the current jira.

- Create new DMN
- Click data types tab
- Add one entry **(tNum : number)**
- Specify constraint type for this entry as **Range**
- Save the data type entry
- Save the dmn asset
- Reopen the dmn asset
- Invoke the constraint type dialog for the same data type entry
- There will be **nothing selected**
- I would expect the **Range** will be selected

jomarko:
One more question, If I create decision table like shown, with constrain for input clause column and I download the dmn file, there is stored just the constraint text, not the constraint type. Is that expected/correct?

screenshot from 2019-01-24 11-07-37

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/ConstraintType.java
danielzhe: I'm using the private field to facilitate the serialization, since I'm serializing all lowercase. I could use toLowerCase(), but I think that using .value() is more safe. What do you think?

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/InputClauseUnaryTests.java
danielzhe: I guess it can be private.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/definition/v1_1/InputClauseUnaryTestsPropertyConverterTest.java
danielzhe: Great! Thank you very much!

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/ConstraintType.java
jomarko: I tried tests you prepared with my proposed changes and they passed. Due to this I think both variants are safe. Leaving final decision to you and @manstis

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/range/DataTypeConstraintRangeViewTest.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.client.editors.types.listview.constraint.range;
+
+public class DataTypeConstraintRangeViewTest {
```
manstis: Any tests to add?!?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/enumeration/DataTypeConstraintEnumerationViewTest.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.client.editors.types.listview.constraint.enumeration;
+
+public class DataTypeConstraintEnumerationViewTest {
```
manstis: Any tests to add?!?

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/InputClauseUnaryTests.java
manstis: Should we be exposing the `constraintType` in the Properties Panel too? i.e. add the `@FormField(...)` annotation to this field. That would allow the User to select the type of constraint too for Decision Table `InputClause` columns.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/ConstraintType.java
manstis: @jomarko I think I prefer @danielzhe implementation as it avoids the need to _remember_ to convert `ConstraintType.toString()` or `ConstraintType.name()` to lower case before writing to the XML (or at least, IMO, makes it less error prone in the future).

danielzhe:
@jomarko That's right because the RangeComponent is not implemented yet, so nothing is saved. At this moment, only Expression is save. About the other behaviour, when you create a Decision Table, I think probably we need to open another JIRA to add the option to select the type there.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/v1_1/ConstraintType.java
jomarko: ok, thank you both

jomarko:
@danielzhe thank you, one more question to double check my understanding, is this PR just some preparation on backend side without effect in UI? Asking because reading [DROOLS-3517](https://issues.jboss.org/browse/DROOLS-3517) I would expect if I save file, there is stored constraint type even if I select range or enumeration, however currently it is not.

Am I right that more changes (I am fine if in separate PR)  will come to resolve [DROOLS-3517](https://issues.jboss.org/browse/DROOLS-3517)?

manstis:
Jenkins please retest this.

danielzhe:

danielzhe:
@jomarko It will not save now since the components are not done yet.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListItem.java
```diff
@@ -242,12 +243,13 @@ void discardNewDataType() {
...
-                .withName(getOldName())
-                .withType(getOldType())
-                .withConstraint(getOldConstraint())
-                .asList(getOldIsList())
-                .get();
+            .withDataType(getDataType())
+            .withName(getOldName())
+            .withType(getOldType())
+            .withConstraint(getOldConstraint())
+            .withConstraintType(getConstraintType())
```
karreiro: @danielzhe as we discussed on Slack, we need to use the _old_ value here ;-)

manstis:
@danielzhe Looks good and works well.. 

When the minor fixes mentioned [above](https://github.com/kiegroup/kie-wb-common/pull/2419#pullrequestreview-196711257) are done I'm happy to approve.

PR #2421, opened 2019-01-24
JBPM-8157: Stunner - Documentation page produces an error when a process with a UserTask is being opened

romartin:
Hey @hasys  @wmedvede 

This is a quick fix for [JBPM-8157](https://issues.jboss.org/browse/JBPM-8157). 

I have reassigned the ticket to @tiagodolphine , because I'm not really sure if that's the right fix and also because there are still missing the tests, but at least this commit fixes this bug, which is urgent, as it's being produced lot of times and really on the initial authoring stages... 

So for now, if you agree, let's commit this fix and let @tiagodolphine check and test it properly once he's back. WDYT?  

Thanks!


romartin:

wmedvede:
@hasys @romartin 
I confirm that the issue is produced when a property in the model does not have the label propertly set.
This way to reproduce the issue is
1) to the /home/wmedvede/development/projects/droolsjbpm/kie-wb-common/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/resources/org/kie/workbench/common/stunner/bpmn/resources/i18n/StunnerBPMNConstants.properties  file and comment the xxxx.label for a property

e.g.: org.kie.workbench.common.stunner.bpmn.definition.property.task.SLADueDate.label

2) after rebuilding, etc, if you try to open the process with a node that uses this property, then the error is produced.

So I conclude Roger's fix is good.
+1 from my side

romartin:
Hey @hasys @wmedvede 
Thanks for the review! As you say, it's probably not the best fix, but I'll keep the ticket open to let @tiagodolphine review it again, at least it allows us to go ahead and save/load the process... so merging!
Thanks!

PR #2431, opened 2019-01-30
DROOLS-3581: [DMN Designer] Decision Table: Remove use of TextBox for inline cell editing

manstis:
See https://issues.jboss.org/browse/DROOLS-3581


manstis:
@danielezonca @jomarko This PR seems to have some _extra_ formatting fixes (I format the whole code base automatically when submitting PRs). I'm happy to revert them if you'd prefer to keep this PR _clean_.

manstis:
@jomarko I'll answer both questions with one answer!

The "multi-line function name" in an Invocation issue was not introduced by this PR but I could not find any reference in the DMN specification to suggest that a _name_ of a Function can include line breaks. Therefore to be _safe_ I've changed the `InvocationGrid` to use a `TextBox` for the function name.

This then answers your question about the `TextBoxDOMElement` and `TextBoxSingletonDOMElementFactory` classes... they are now used!

The second commit on this PR also include a few cosmetic tweaks to some of the styling to make the fonts in the HTML widgets (in-cell edit) closer in appearance to the fonts in the Canvas widget (after editing is complete). I believe there is a JIRA to make them _exactly_ the same but I took this opportunity to improve it.

PR #2435, opened 2019-01-31
DROOLS-3586: [DMN Designer] Boxed expression view: Expression type label should include spaces

manstis:
See https://issues.jboss.org/browse/DROOLS-3586


jomarko:
Please keep merged, I will report if I spot something during review of other PRs.

PR #2442, opened 2019-02-05
JBPM-8189: remove edges between tasks, subcases and subprocesses

xiezhang7:


xiezhang7:
Hi @romartin @LuboTerifaj , this pull request is for [JBPM-8189](https://issues.jboss.org/browse/JBPM-8189). Thanks.

Create a diagram with case modeler like this:
screenshot from 2019-02-04 11-51-41

Change the extension of the saved file from `bpmn-cm` to `bpmn` , and open the file in process designer. It should look like this:
screenshot from 2019-02-04 11-52-29

There should not be any edges between user tasks, subcases or subprocesses which are child of stages.

xiezhang7:
ok to test

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-backend/src/main/java/org/kie/workbench/common/stunner/cm/backend/CaseManagementDirectDiagramMarshaller.java
```diff
@@ -65,6 +68,16 @@
 @CaseManagementEditor
 public class CaseManagementDirectDiagramMarshaller extends BaseDirectDiagramMarshaller {
 
+    private static final double GAP = 100.0;
+    private static final double ORIGIN_X = 50.0;
+    private static final double ORIGIN_Y = 50.0;
+    private static final double EVENT_WIDTH = 55.0;
```
romartin: Well so the sizes for the different shapes (stage, event, etc) are in fact declared only on the SVG file itself. I mean, hardcoding here the size values is not really correct.. as if for some reason, from UX, the icon changes, the resulting locations will be not the right ones until we change these values as well, right?
So what's the goal for this? just trying to perform some layouting right?

romartin:
Jenkins please retest this

romartin:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-backend/src/main/java/org/kie/workbench/common/stunner/cm/backend/CaseManagementDirectDiagramMarshaller.java
xiezhang7: Hi @romartin ,

I have [changed the code](https://github.com/kiegroup/kie-wb-common/pull/2442/files#diff-d4d2498a554abe172a89ca2c8e983fe4) to always use the size from SVG file, so the hardcoded values here won't affect the UI.

You are right, these values are used to do layout for two reasons:

1. Since there are not sequence flows between the child nodes inside stages, these nodes are not strictly ordered. But users can order the child nodes in case modeler. So when performing marshalling, different coordinates are assigned to the child nodes and make sure they are arranged in the same order as they are placed in the case modeler. When performing unmarshalling, these coordinates are used to restore the order of the child nodes. 
2. Assign coordinates to the nodes so that they are at valid locations when the file is opened in process designer.

Thanks.

LuboTerifaj:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-backend/src/main/java/org/kie/workbench/common/stunner/cm/backend/CaseManagementDirectDiagramMarshaller.java
romartin: Yes I see [this code](https://github.com/kiegroup/kie-wb-common/pull/2442/files#diff-d4d2498a554abe172a89ca2c8e983fe4) which results in using the size declared in the SVG, but that's the reason why I posted the comment - Lemme try to explain, this is how it works, it does not mean it's the best approach, but here it is:
* As the node "size" (not the location) is being declared on the view side (SVG), once creating the `Element` instance (eg: `Node`) associated, the `Bounds` has not yet been set
* The `Bounds` for a graph `Element` are being set once the shape is being build, because it's the shape declaration (parsed from an SVG file) which contain the size related attributes
* So once "saving" the process, the graph and all its elements will have some well defined Bounds - so you have the location and "size" for each element
* Once "loading" a process, it's also supposed that the Bounds are well defined

So I'm still not sure why you've to create `Bounds` instances by using some hardcoded values (`EVENT_WIDTH`, `EVENT_HEIGHT`, etc), eg [here](https://github.com/kiegroup/kie-wb-common/pull/2442/files#diff-c6f1a67ca776b7b12545c995b12b1d53R120), they're already "ported" from the UI to the server and so available for serializing. 

What am I missing? :)

Thx

LuboTerifaj:

LuboTerifaj:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-backend/src/main/java/org/kie/workbench/common/stunner/cm/backend/CaseManagementDirectDiagramMarshaller.java
xiezhang7: Hi @romartin ,

`Start event` and `end event` are not created in the case modeler UI. Case modeler doesn't support adding `start event` or `end event` in the UI. They are created when saving the case. That's why the `Bounds` for `start event` and `end event` are created by the marshaller with those hardcoded values.

Other elements from case modeler UI do have size from `Bounds`. But in the UI, they are added by index not by location, so there is no location information.

I have changed the code to use the existing size instead of hardcoded values for the elements created in the case modeler UI. 

Thanks.

romartin: @xiezhang7 mmmm good point about start/end events yeah.. and this also may apply to other elements as well - as they're not being added from the UI, the bounds (obtained from the shape) are not well defined. 
TBH this "issue" or missing capability is starting to make noise, as I also hardcoded some size values on the BPMN marshallers side (see eg. https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/tostunner/properties/EventPropertyReader.java#L43). 
So as this is a missing capability on the core side, we cannot fix it now in just a few changes... so I'm agree on declaring the values for event sizes, as you're doing, but please let's try to keep this as much isolated as possible, so we can refactor/fix it at some point later.

PR #2446, opened 2019-02-06
[DO_NOT_MERGE] Fix random MigrationToolTest fails

tomasdavidorg:


jomarko:
@tomasdavidorg Hi, I was just interested about the unstable test fix so I had a look on code. I got additional idea how to stabilize the test, for cloning we use the code [1]. I checked the javadoc of `Git` and `CloneCommand` classes at [2], to me it seems we should close the obtained git resource after `call()` because we do not use it anywhere else.

Please take this just as and idea, I didn't check too deeply, maybe I miss something about closing the `Git` resource.

[1]
```java
Git.cloneRepository()
                .setURI(projectRepoDir)
                .setDirectory(projectDir)
                .call();
```

[2]
https://download.eclipse.org/jgit/site/5.2.1.201812262042-r/apidocs/index.html

PR #2449, opened 2019-02-06
Fix relative path problems in kie-wb-common-compiler-offprocess-classpath

tiagobento:
When building `kie-wb-common` outside its root dir (using `mvn -pl`, `mvn -f`, or even using [kiegroup-all](https://github.com/tiagobento/kiegroup-all)), I was getting a `java.nio.file.NoSuchFileException`.

The fix was to supply the `${project.basedir}` POM built-in property so that the `ClassPathMavenGenerator#main` method can work only with relative dirs.

### DIFF
--- a/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-offprocess-classpath/pom.xml
+++ b/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-offprocess-classpath/pom.xml
@@ -604,7 +604,10 @@
           </execution>
         </executions>
         <configuration>
-          <arguments>${version.org.kie},${project.basedir}</arguments>
+          <arguments>
+            <argument>${version.org.kie}</argument>
+            <argument>${project.basedir}</argument>
+          </arguments>
           <mainClass>org.kie.workbench.common.services.backend.compiler.offprocess.generator.ClassPathMavenGenerator</mainClass>
         </configuration>
       </plugin>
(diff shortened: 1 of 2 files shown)


tiagobento:
@ederign Could you please take a look at this?

ederign:
please jenkins retest this

tiagobento:
please jenkins retest this

tiagobento:
@manstis Do you mind taking a look as well?

manstis:
@tiagobento Happy to review; how do I test this PR? 

Do I need to run tests or just compile `kie-wb-common` outside the repository root?

tiagobento:
@manstis I think compiling `kie-wb-common` outside the repository root will suffice, since this code in ran during compilation.

tiagobento:
@domhanak This code is ran during compilation, if something breaks, compilation breaks.. That's the only reason I can think for this class not having tests.

tiagobento:
@domhanak Can we merge this?

kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-offprocess-classpath/pom.xml
```diff
@@ -604,7 +604,7 @@
           </execution>
         </executions>
         <configuration>
-          <arguments>${version.org.kie}</arguments>
+          <arguments>${version.org.kie},${project.basedir}</arguments>
```
jhrcek: I'd recommend using the form
```xml
<arguments>
   <argument>argument1</argument>
   <argument>argument2</argument>
            ...
</arguments>
```
just to keep the config clearer
See [exec-mave-plugin docs](https://www.mojohaus.org/exec-maven-plugin/usage.html)

tiagobento: Thanks for the suggestion, @jhrcek! Just applied the change on 161c3a0.

tiagobento:
@manstis Can we merge this, please? :)

PR #2453, opened 2019-02-07
JBPM-8169 Stunner - unify and fix properties in Process Documentation

tiagodolphine:
- Inserted Description in process Overview.
- Inserted data assignments in all nodes (the nodes, which are supposed to have them), for now I just used the current serializer to show the assignments
- Inserted ConditionExpression property name (property value is there) in Conditional events.
- Fixed new line characters to be shown on properties 
- Removed  the fixed "Name" properties to avoid duplicating it when all the dynamic properties are filled on the documentation.
- Removed the KPI column on the variables section, since this is not applied on stunner.
- Related to [`JBPM-8157`](https://issues.jboss.org/browse/JBPM-8157) I applied the property name to get fallback, this will avoid error in case the property are not found on the i18n file for any reason. `caption -> label ->  title -> propertyId`
- Filter applied to only show properties that has some value set, empty properties not be shown WDYT ?

Don't forget to test both thickes [ JBPM-8169 ](https://issues.jboss.org/browse/JBPM-8169) / [`JBPM-8157`](https://issues.jboss.org/browse/JBPM-8157)

@romartin @LuboTerifaj


tiagodolphine:

tiagodolphine:
Jenkins retest this

tiagodolphine:

tiagodolphine:
Jenkins retest this

tiagodolphine:

tiagodolphine:
Hi @romartin my plan is it to provide all the tests related to documentation feature on the [JBPM-8171](https://issues.jboss.org/browse/JBPM-8171). 
Anyway I will ckeck if there are any test that could be inserted that not depends direclty on the documentation new classes.
But don't worry I will provide all the tests on the [JBPM-8171](https://issues.jboss.org/browse/JBPM-8171) :) 

One other thing is to don't forget to test both thickes [ JBPM-8169 ](https://issues.jboss.org/browse/JBPM-8169) / [`JBPM-8157`](https://issues.jboss.org/browse/JBPM-8157) because this PR is intended to fix both.

Thanks.

tiagodolphine:
Hi @LuboTerifaj  thanks for the review. One point we need to be aware is that on this first step we want the parity with the old designer, I think we can open JIRAs for improvements, otherwise, this could be a too long task.
Anyway, I tried to fix/implement all of your comments as follow:

* If property contains more then one empty character, there is displayed only one in Documentation. Is it ok?
**Fixed. Old designer does not work.**

* Properties are sorted in different way in Documentation then in Diagram properties.
**Sorting by name (alphabetical order)**

* What is "IsExecutable" property for?
**This is a property of the process. Old designer shows it as well.**

* One minor thing. Not all intermediate events are yellow, but in Documentation they are all under one yellow category.
**We are using the category to group the elements, and the respective category icon (same behavior as the palette)**

* All start Events have additional "isInterrupting" property.
**They are present on the old designer, but if necessary we can remove them.**

* If I change any property value, save close and open process, nodes are listed in different order.
**Sorting by name (alphabetical order)**

* "Default Route" property in Gateways has value: ID of linked node. Is it ok?
**Old designer shows the ID as well. We can exclude the property to be showed or include the name.
This is the ID of the Connector/Edge, for now I'm showing the name of the Target Node from the Connector, if set, otherwise keeping the connector ID.**

* In MI sub-process, there is a property "org.kie.workbench.common.stunner.bpmn.definition.property.task.MITrigger". It shouldn't be there right? At least the name is wrong for sure.
**The ID is showed when there is no Caption/Label/Title configured on the i18n properties file for that property.
I'll exclude this property to be showed, it seems it does not make sense to show.
Old designer does not show subprocesses at all.**

* There are missing process variables in sub-processes.
**Good point, now only process variables are being processed, IMO it is better to open a new JIRA to handle it, can you do this, please?**

* Business rule task has additional properties:
  -- _Namespace, Decision Name, DMN Model Name_
  -- Business Rule task is also not under Decision Tasks category.
**Removed these Decision properties from the documentation, they seems to be related to DMN.**

* In User Task, there is missing "Subject" property.
**I excluded it to not be showed, but I will add it again.**

* "Activity Reference" property in Intermediate Compensation event, has ID as value. Is it ok?
**Same as the default route. We can exclude the property to be showed or include the name of the node, for now I'm showing the name if set, otherwise the ID.
Old designer seem that does not show Intermediate Compensation event at all.**

tiagodolphine:

tiagodolphine:
Jenkins retest this

tiagodolphine:

romartin:
hey @tiagodolphine , any progress on this? Would be nice having it merged as for the release... thanks!

tiagodolphine:
@romartin hey I need to solve the conflicts that have appeared now.
@LuboTerifaj for `Business rule task` I'm following the approach of same categories as the palette, I mean there is no Decision Tasks.  Since the categories and elements associated with them are dynamically included on the documentation any hardcoded thing will complicate things for now. IMO we can open enhancements JIRAs, this will allow the PR to be merged, WDYT ?

romartin:
@tiagodolphine @LuboTerifaj go go go go! :+1:

tiagodolphine:

tiagodolphine:
Thanks @LuboTerifaj !

@romartin le'ts merge :)

LuboTerifaj:
From my [comment](https://github.com/kiegroup/kie-wb-common/pull/2453#pullrequestreview-202110965) above, only one issue is present on master. 
I have reported it as enhancement: [JBPM-8264: Stunner - Different nodes and properties order in documentation](https://issues.jboss.org/browse/JBPM-8264)

PR #2456, opened 2019-02-07
JBPM-8204: Actor configuration in Human Tasks doesn't work with Standalone Central

wmedvede:


wmedvede:
Hello @romartin @hasys  would you mind review this PR?
Thanks.

Quick explanation about how to test:

The issue reported in:
        https://issues.jboss.org/browse/RHPAM-1647
        https://issues.jboss.org/browse/JBPM-8204
has two parts:

1) The issue that the user is reporting is happening because the configuration YAML file he's is using is not correct, and that's why Business Central is not connecting well with the Users System producing the issue. For the configuration side the fix is just to provide the correct configuration YAML file. I'll update the JIRA with a correct configuration, but we don't have to worry about this part.

2) Assuming the User System is not working, we still want Stunner to be able to open a process, with a UserTask with e.g. "roles" configured, and we want to configure, save, and open the task and see the configured stuff properly, etc. So this PR includes proper enhancements for this.
Additionally the fix will show a proper flying notification if the User system is throwing errors and the user will be able to work well, etc.

Note:
In general the error will never happen, i.e., the User System is always working, and different WB components are using it. So for "forcing" a potential error and being able to test the enhancement we can do this:

Set the following property org.uberfire.ext.security.management.wildfly.cli.port to a non existing port, and start the WB.

If I'm not wrong this property can be set by passing JVM parameters so alternative1 is basically to start the WB  with  -Dorg.uberfire.ext.security.management.wildfly.cli.port=12345.

If alternative1 not works, the alternative2 is to find the security-management.properties file for the given WB and set the value manually in this file org.uberfire.ext.security.management.wildfly.cli.port=12345

p.s
Since the User System is called many times during normal work, the solution is doing the best effort to not flood the user with millions of notifications, instead the notification is thrown for the first time the user interacts with the UserTask. If you e.g. go to another node, and return to the UserTask then you'll get the notification again. This is the most reasonable approach I think

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
```diff
@@ -119,18 +127,15 @@ public void deserializeAssignees(String serializedValue) {
...
+                    addAssignee(new Assignee(assigneString));
                 }
             }
         }
     }
 
     private void addAssignee(Assignee assignee) {
         AssigneeListItem listItem = listItems.get();
-        listItem.init(type, assignee, this::doSave, this::removeAssignee);
+        listItem.init(type, assignee, this::doSave, this::removeAssignee, this::onError);
```
romartin: Hey @wmedvede 
Yeah absolutely agree on your comments, by looking at this code it looks like we're performing a server call for each assignee, which is not good at all!! IMO We should improve this ASAP, as this issue will not be visible on our development environment, looks like it's an issue that will appear on a production environment when several processes and tasks and assignee's are involved.. can you please report a bug/improvement for this? agree @hasys ? thx in advance guys

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
```diff
@@ -55,13 +59,17 @@
 
     private int max = -1;
 
+    private int maxSearchErrorNotifications = 1;
```
romartin: well that will work but sounds a bit tricky.. isn't it? hehe :) 
Just a couple of comments:
* I understand the goal for this is just firing a single notification event, even on multiple errors. If that's the case, there is need to use an `int` type, just a `boolean` type should work, right?
* Also I understand you did this because of, in case the users management is not active, for whatever reason, it will fire an error on each call, so this way you only are firing a single notification, is that correct? I say this because this could be improved a bit by:
** Before starting to do several calls to the server for each assignee, you can check the method [ClientUserSystemManager.isActive()](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-client/src/main/java/org/uberfire/ext/security/management/client/ClientUserSystemManager.java#L242), this way there is even no need to perform each call to the server and handle multiple errors
** This also means that there is probably no need for the  `maxSearchErrorNotifications` flag, because if an error is being produced on a call for a concrete assignee, as you previously checked if the users system `isActive`, it makes sense to display the concrete error for that concrete call. Does that makes sense?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
wmedvede: @romartin well, that's the way this widget was left by the initial "owner".

But even so, I'm not sure why you are against performing several searches to the server for querying the available users.
See that the type-ahead usually works this way..., e.g, in general the type-ahead performs calls to the server when the user is e.g. typing for searching and showing the available options, etc. That's why this widget executes several calls to the server. And if you have multiple lines, and you can "select" on each line individually, then it's not possible to avoid having server side searches for each line.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
wmedvede: @romartin Intresting, I didn't know this isActive() method existed.
But even so, I could see that this isActive() calculation is done only one time and cashed in the user session when the ClientUserSystemManager is started. I means that the "connection" with the external user system is "checked" only one time. No retries at all.
I think this is a bit tricky too. Note that the user system is in general "plug-able" an thus we are agnostic from the particular implementation, etc, but also the fact that there was an issue e.g. at the moment initialization (for whatever implementation) e.g.  "a socket error" does not mean that the connection won't work at a latter time. Specially in microservice based architectures, etc., or even in this WF client implementation we have.
So I don't think its safe to trust on this cashed "isActive()" value.

The "integer" instead of a "boolean" is because I was sure, that if I used a "boolean" there will be people asking "why don't you use an integer instead?" so it's more flexible and we can configure the number of notifications we want to allow, etc. "Can you set 2 notifications, instead?"
 Sad, but true.

So to summarize, I'll change the integer for a boolean, and I won't rely on the isActive() method you are suggesting due to the explanations above.

wmedvede:
jenkins execute full downstream build please

wmedvede:
Hi, @romartin @hasys , as usual the second build railed BUT artifacts were generated, so this is ready for QE too. The PR was properly updated with @romartin suggestions and is rebased with master.
Thanks.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
romartin: ah ok, makes sense

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
romartin: ok, good point too, agree

romartin:
thanks @wmedvede , waiting for @hasys review again.

romartin:
ping @hasys plz - urgent

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeLiveSearchService.java
```diff
@@ -112,17 +127,40 @@ public void searchEntry(String key, LiveSearchCallback<String> callback) {
...
-    private boolean processError(LiveSearchCallback<String> callback) {
-        callback.afterSearch(new LiveSearchResults<>());
+    private boolean processSearchEntryError(String key, LiveSearchCallback<String> callback, Throwable throwable) {
+        LiveSearchResults<String> results = new LiveSearchResults<>(1);
+        if (!isEmpty(key)) {
+            addCustomEntry(key);
+            results.add(key, key);
+        }
+        callback.afterSearch(results);
+        processError("It was not possible to get user or group: " + key + " from the users system.", throwable);
```
hasys: Looks like it is a text displayed to users, should we add i18n to it as you did for `assignee.searchError`?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeLiveSearchService.java
```diff
@@ -112,17 +127,40 @@ public void searchEntry(String key, LiveSearchCallback<String> callback) {
...
+        return false;
+    }
+
+    private boolean processSearchError(List<String> filteredCustomEntries, int maxResults, LiveSearchCallback<String> callback, Throwable throwable) {
+        int maxSize = maxResults > filteredCustomEntries.size() ? filteredCustomEntries.size() : maxResults;
+        maxSize = maxSize < 0 ? 0 : maxSize;
+        LiveSearchResults<String> result = new LiveSearchResults<>(maxSize);
+        filteredCustomEntries.subList(0, maxSize).forEach(entry -> result.add(entry, entry));
+        callback.afterSearch(result);
+        processError("It was not possible to execute search on the users system.", throwable);
```
hasys: Looks like it is a text displayed to users, should we add i18n to it as you did for `assignee.searchError`?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeEditorWidget.java
hasys: @wmedvede, I guess it possible to make some cache here, the question is how many requests do we have. Is it really for every single user? Or it is for every user task? Or it is for every touch to user task? Something else? Not clear to me now, sorry.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeLiveSearchService.java
wmedvede: No we shouldn't. See that error messages are "low level" log traces that will only be displayed in the browser development log console, and only if the log level is properly set.
If you look at other places in Stunner were we use this "log level" log traces you'll find that we usually don't waste time and resources adding i18n to this type of messages, and I don't think it should be the case. So this is absolutely not a reason for blocking this PR.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/assigneeEditor/widget/AssigneeLiveSearchService.java
wmedvede: same here, this log trace won't have i18n

wmedvede:
@hasys regarding your query above 
"I guess it possible to make some cache here, the question is how many requests do we have. Is it really for every single user? Or it is for every user task? Or it is for every touch to user task? Something else? Not clear to me now, sorry."

I believe it can be theoretically possible to set-up N different caching mechanisms, but the real question is "does it make sense?" I you look around the product all "live searchs" we have are based on this LiveSearchComponent, and the work more or less the same way. Additionally this live search is executed only for UserTasks, and only in the case where the user is configuring e.g. the Actors, Roles for a task. And this component has worked this way in all Stunner versions and there are zero complains about it.
So my opinion is that "this is too much", or "unrealistic" staring a debate about caching strategies for this component.

hasys:
> I believe it can be theoretically possible to set-up N different caching mechanisms, but the real question is "does it make sense?"

Hi @wmedvede, yes sure, it was actually my question, sorry I wasn't clear enough.

So looks good, thank you. And, as I mentioned, I will test prioritization standalone build as it will be available.

wmedvede:
To summarize:

@romartin @hasys 

1) We wasted time on this JIRA because it was marked as blocker.
2) This PR introduces a reasonably good improvement, and fixes  the blocker
3) The artifacts are generated and the .war file can be tested
4) when you @hasys  it's said "Also, we have no built standalone wars to test so, will test that everything is fine one more time with product build when it will be available." 
See that there are really no difference between testing the .war or the standalone .jar file in this case.

So to be honest I don't see any reasons for blocking this and I think it's a good improvement over what we had.

hasys:
@wmedvede, I already tested scenario described in your PR message, but still will be good to test it with use-case described in original Jira.

So my sum up:
1. Next time we should discuss better what is blocker
2. According to point 1. better prioritize issues for the next release
3. .war already tested with scenario described in PR message, thank you @wmedvede for great explanation!
4. During Product jira verification I will test Standalone jar with scenario from reporter as well.

Thank you @wmedvede, this PR already approved by me and @romartin so it is ready to go.

PR #2460, opened 2019-02-08
DROOLS-3612: [DMN Designer] Java Function is context after reopening

manstis:
See https://issues.jboss.org/browse/DROOLS-3612


jomarko:
Please see my feedback in https://issues.jboss.org/browse/DROOLS-3612

manstis:
@jomarko Thank-you for the swift feedback. I've fixed the issue in a new commit.

PR #2461, opened 2019-02-08
DROOLS-3616: [DMN Designer] Literal expression name and data type popup missing

manstis:
See https://issues.jboss.org/browse/DROOLS-3616


manstis:
Jenkins please retest this.

PR #2463, opened 2019-02-11
Fix failing integration tests - add Kie Server mode

cristianonicolai:


cristianonicolai:
@mswiderski @pefernan it feels that we're breaking behaviour with the kie server mode, shouldn't _org.kie.server.mode=REGULAR_ be the default if you don't specify it?

pefernan:
@cristianonicolai sorry for breaking this! Regarding starting the kie-server on regular mode, we had some conversations about that and development mode option won. The main argument was that our examples are snapshots (development mode only supports deploying snapshots) and they wanted to keep it as default mode to make things easy for customers that are using on their current servers. However I see your point, and we can change that easily.

cristianonicolai:
@pefernan I see the value for someone getting started but I'm afraid that will break existing servers behaviour. If that is something that been considered as part of this decision I guess it is fine but we need to make sure it is communicated to users. 
Another question, with these modes now, we can't really mix -release and -snapshot versions for the containers?

pefernan:
@cristianonicolai, no we won't be able to mix snapshots and releases. Dec mode kie-server only allows to deploy snapshots and regular mode allows releases. Also developed mode provides a more flexible policy on updates allowing to update a container with active process instances and update a container aborting al the active instances (redeploy) if you want we can talk later about this.

PR #2473, opened 2019-02-13
JBPM-8236: Fix java 11 build

hasys:
Hi @romartin, @manstis 

it is a fix for java 11 build, since CORBA is removed from Java 11: https://openjdk.java.net/jeps/320

@romartin, any ideas why we had it? Tests are passing with regular java.lang.Object as well.


hasys:
Hi @romartin, build failed to know unrelated test-fails. It is ready to go.

romartin:
Hey @hasys 
Great catch!!! TBH I'm quite sure this is an error... same happened for me time ago, sometimes when coding fast in IntelliJ and writing `Object`, instead of adding the default import `java.lang.Object` it was using this COBRA one... so it's an error for sure, we should no depend on this at all.
Thanks, good catch!

PR #2479, opened 2019-02-14
DROOLS-3519 / DROOLS-3643: [DMN Designer] Data Types - Constraints - Change the "Placeholder" accordingly to the type / Disable Data Types shortcuts when users open the Constraint dialog

karreiro:
See:
- [DROOLS-3519](https://issues.jboss.org/browse/DROOLS-3519): [DMN Designer] Data Types - Constraints - Change the "Placeholder" accordingly to the type   
- [DROOLS-3643](https://issues.jboss.org/browse/DROOLS-3643): [DMN Designer] Data Types - Constraints - Disable Data Types shortcuts when users open the Constraint dialog

---

---

PS.: It's not visible in the GIF, but in the end I pressed <kbd>Ctrl</kbd> + <kbd>S</kbd>, and the alias was disabled as expected (since the constraints dialog was opened).


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/ConstraintPlaceholderHelper.java
```diff
@@ -0,0 +1,92 @@
...
+        final String name = names[names.length - 1];
+
+        return capitalize(name);
+    }
+
+    private String capitalize(final String name) {
+        return name.substring(0, 1).toUpperCase() + name.substring(1);
+    }
+
+    private Optional<BuiltInType> findBuiltInTypeByName(final String name) {
```
jomarko: At least this one method seems to me as should be placed in `org.kie.workbench.common.dmn.client.editors.types.common.BuilInTypeUtils`, and maybe also `builtInTypeAsKey`

karreiro:
Rebased 🛠

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/ConstraintPlaceholderHelper.java
karreiro: @jomarko, I agree the `findBuiltInTypeByName` should be in the `BuilInTypeUtils`, this fix is applied.

However, the `builtInTypeAsKey` is more i18n specific in this context, so I'll keep it here for now, and maybe in the future we can move it to `BuilInTypeUtils` if necessary.

Thank you.

manstis:
@jomarko "..on my machine _Ctrl+S_ for constraint popup doesn't save the entry on background, however I am offered to save the html page, is it expected" This is the default behaviour of the browser; so in the absence of any _application_ keyboard handler (that calls `Event.preventDefault()`) the browser is offering it's default handling. The same is for _Ctrl+N_ and other shortcuts if the Data Type tab is _not_ selected (or the Data Type tab keyboard handling is not enabled).

jomarko:
@manstis thank you for explanation, will try to remember as I definitely should already know this `preventDefault()` thing.

PR #2483, opened 2019-02-14
DROOLS-3646: [DMN Designer] Updating two files in the editor raises an error

danielzhe:

PR #2487, opened 2019-02-15
DROOLS-3656: [DMN Designer] Select logic type missing entries

manstis:
See https://issues.jboss.org/browse/DROOLS-3656

The issue was much the same as my previous PR.. when multiple DMN files are open each was getting its own `CellEditorControlsView` added to the HTML `body` which lead to Bootstrap binding popovers to the wrong element. This PR ensures all editors share a single instance attached to `body`.

Please note I could not replicate without first merging #2483 as I was unable to have multiple editors simultaneously open. With the referenced PR merged locally I could successfully replicate the issue - and tested my fix.


manstis:
@jomarko I really don't like merging w/o your review.. however on this occasion I think we must.

PR #2498, opened 2019-02-22
JBPM-8267 - Avoid duplicated error popups

cristianonicolai:
Depends on https://github.com/kiegroup/appformer/pull/647


cristianonicolai:
@tiagobento I'm not exactly sure about the impact of not calling the _originalRpcErrorCallback_, would that be fine to be skipped?

tiagobento:
The "original rpc callback" defers to `defaultWorkbenchErrorCallback` eventually, because it's registered on `handleUncaughtException`, so I would keep this call.

The `GenericErrorPopup` stacks error messages when it's showing, in the case multiple errors occur. It does it by concatenating a "|" between the parameters used to call `setup`. Removing lines 126 and 127 on `DefaultWorkbenchEntryPoint` is making the information about the Errai Bus message be erased, so I would change that..

I think to solve that, we could revert back to what it were and only change the `GenericErrorPopup` to concatenate things regardless of its state on `setup` (showing or not), and reset it's content once it's closed. Of course, then we would delete the `show` call on `DefaultWorkbenchEntryPoint#127`. What do you think?

cristianonicolai:
@tiagobento thanks, take a look if I got your idea correctly :)

cristianonicolai:

cristianonicolai:
@tiagobento tests added, thanks

cristianonicolai:

PR #2505, opened 2019-02-27
JBPM-8161: fix the scrollbars which appear even if process does not fit all available space

xiezhang7:


xiezhang7:

xiezhang7:
Hi @romartin @LuboTerifaj , this pull request is for [JBPM-8161](https://issues.jboss.org/browse/JBPM-8161). Thanks.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-api/src/main/java/org/kie/workbench/common/stunner/cm/definition/CaseManagementDiagram.java
```diff
@@ -67,8 +67,11 @@
     public static final String DIAGRAM_SET = "diagramSet";
     public static final String PROCESS_DATA = "processData";
     public static final String CASE_MANAGEMENT_SET = "caseManagementSet";
-    public static final Double WIDTH = 2800d;
-    public static final Double HEIGHT = 1400d;
+
+    // set width and height to 0 to hide the diagram, so that its shape bounding box will not be shown
+    public static final Double WIDTH = 0d;
```
romartin: probably instead of setting width and height to 0, which doesn't make much sense... why not better just not setting it, so you can remove these couple of fields? It looks strange setting size to 0... isn't it?

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/canvas/controls/builder/CaseManagementObserverBuilderControl.java
```diff
@@ -73,12 +73,6 @@ public CaseManagementObserverBuilderControl(final ClientDefinitionManager client
...
     public Node<View<?>, Edge> getParent(final double x,
                                          final double y) {
-        // The default implementation uses GraphBoundsIndexer that finds the top (z-index) Node
-        // that contains the (x, y) location. However Case Management does not update Graph's Node's
-        // locations to represent the visual layout. Consequentially we find the Node by examining
-        // the  actual rendered Canvas.
-        final Optional<Node<View<?>, Edge>> parent = canvasHandler.getElementAt(x,
-                                                                                y);
-        return parent.orElse(null);
+        return CaseManagementBuilderControlHelper.getParent((CaseManagementCanvasHandler) canvasHandler, x, y);
```
romartin: I think there is no need to create a helper class for this... I mean the "original" code here was mostly relying on `canvasHandler.getElementAt(x,y)`, but this method is not returning the right element at the given location. Looking at the signature for `canvasHandler#getElementAt`, it's an `abstract` method in class `AbstractCanvasHandler`, and also exist a case management concrete implementation `CaseManagementCanvasHandler`, so why not just overriding the  `getElementAt` method in the `CaseManagementCanvasHandler` and fix the issue as you're doing in the `CaseManagementBuilderControlHelper`? This way there is no need to change these lines, only fix it by overriding the  getElementAt method in the `CaseManagementCanvasHandler`.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/canvas/controls/builder/util/CaseManagementBuilderControlHelper.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Optional;
+
+import org.kie.workbench.common.stunner.cm.client.canvas.CaseManagementCanvasHandler;
+import org.kie.workbench.common.stunner.core.graph.Edge;
+import org.kie.workbench.common.stunner.core.graph.Node;
+import org.kie.workbench.common.stunner.core.graph.content.view.Point2D;
+import org.kie.workbench.common.stunner.core.graph.content.view.View;
+
+public class CaseManagementBuilderControlHelper {
```
romartin: As commented above, I would just remove this class and apply this fix in the `CaseManagementCanvasHandler#getElementAt`

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/canvas/controls/builder/CaseManagementObserverBuilderControl.java
xiezhang7: Hi @romartin , it makes sense to me. I have changed the code to override [CaseManagementCanvasHandler::getElementAt](https://github.com/kiegroup/kie-wb-common/pull/2505/files#diff-126d1a2d4d4fcd3cdf32538df633eed9). Thanks.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-api/src/main/java/org/kie/workbench/common/stunner/cm/definition/CaseManagementDiagram.java
xiezhang7: Hi @romartin , I agree it is strange to set size to 0 in the diagram, but if not setting it explicitly, values from svg will be used to create the shape. So instead of setting size 0 to the diagram, I modified [CaseManagementSvgDiagramShapeDef](https://github.com/kiegroup/kie-wb-common/pull/2505/files#diff-f70a500db911ec1b427b62fb75ad6991) to set height and width to 0 when creating the shape. I think this makes more sense. Thanks.

xiezhang7:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-api/src/main/java/org/kie/workbench/common/stunner/cm/definition/CaseManagementDiagram.java
romartin: well it's not correct anyway... doesn't make sense to create a rectangle by setting width/height to 0. So it works because the shape for a bpmn/cm diagram it's not being displayed, as the diagram is considered the canvas `root` shape in Stunner (see [Metadata](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/diagram/Metadata.java#L39)), and it will never try to render the rectangle..
Sooo as the shape for the diagram is in fact never created... can you please do a quick try just b removing the shape declaration for the CM diagram, and so you'll be able to remove the associated `ShapeDef`, etc, and so also avoid this "trick" on setting sizes as 0. Looking at the code I think it will work, as it's not calling the `ShapeFactory` for the canvas `root` element, so there should be no need to declare any shape for the CM diagram.
If that works, I'll try the same on BPMN :+1: 
thanks in advance!

romartin: After some talks about it, it looks good to me moving forward now by setting the size 0 on the client side,bpmn is also facing same issue, so we'll fix it later.

PR #2507, opened 2019-02-28
JBPM-6758: Stunner - Adding support for childLaneSet BPMN2.0 element in Red Hat BPMS

wmedvede:


wmedvede:
Hello @romartin @hasys would you mind review this PR?
Thanks!

romartin:
Can you please @wmedvede attach in this PR some process examples which contain nested lanes, so we can test your work? Thanks!

romartin:
ah well in this PR there are already some bpmn files to test it... good :+1:

hasys:

wmedvede:
artifacts properly generated

hasys:

wmedvede:
full build failed in non related tests, but artifacts were generated

romartin:
ok thanks @wmedvede , it's now @hasys turn :)

hasys:
Hi @wmedvede,

code and tests looks good, works great, thank you! Just one minor thing:

* When you export process with nested lanes, you should do some changes and save the process. If you just import and close the process and try to build, you will see Build error `[KBase: defaultKieBase]: unable to parse xml : Exception class org.xml.sax.SAXParseException : <lane> may not be nested`

It is not possible to just save process after import, you need to make some changes (like move any node around). Also when you just import process it already looks valid but in Alerts window you will see the same issue as during build.

My suggestion is to force save with new layout without nested lines. Is it possible?

Thank you!

wmedvede:
> Hi @wmedvede,
> 
> code and tests looks good, works great, thank you! Just one minor thing:
> 
> * When you export process with nested lanes, you should do some changes and save the process. If you just import and close the process and try to build, you will see Build error `[KBase: defaultKieBase]: unable to parse xml : Exception class org.xml.sax.SAXParseException : <lane> may not be nested`
> 
> It is not possible to just save process after import, you need to make some changes (like move any node around). Also when you just import process it already looks valid but in Alerts window you will see the same issue as during build.
> 
> My suggestion is to force save with new layout without nested lines. Is it possible?
> 
> Thank you!

Hi @hasys ,
I don't see where to do this "automatic" saving of the process and to be honest I don't like it unless the user is aware that.

Iv'e asked a similar query during Stunner meeting at the day of starting with this and @ibek and @krisv agreed that initial scenario/implementation assumes that workflow is:
1) import the file into the WB
2) open the file to see the modified version
3) do a silly modification and save again 

By now we know that is a sort of specific "feature" for a particular use case. And in general initial users of this feature are programmers in a specific context that will know the workflow.

On the other hand, given the incremental build of the project, in all scenarios, as soon the imported file is copied into the project repository the incremental build and also the indexing will be triggered, and we'll always have the errors complaining about the nested lanes in the log. So only "nested lanes" message free option I can see here is to try to copy the file to some place not reactive to the project file system events, try to parse and analyse the file at this location, ask the user/or show a message, and then finally copy/move the file to the desired location in the project directory. Or any other variant that not copy the imported file directly into the project directories. I don't see an easy/quick solution.

So from my side I'd say that we should move this ahead and continue fixing the other related import issues/features. Once we get all working we can think about something more elaborated.

Thanks!

romartin:

hasys:

hasys:
Hi @romartin,

this PR is ready to go! :fireworks:

PR #2508, opened 2019-02-28
DROOLS-3664: [DMN Designer] Data Types - Constraints (Range/Enumeration) - Add "Date" component when the type is "Date"

danielzhe:
This PR adds the `DateSelector` for Constraints.

Also, this PR adds the architecture to make possible to select specific component based on the constraint type. Since at this point we don't have a component for every type, I also created a component called `GenericSelector` which basically is the same existing component (a input box) where user can type anything. The `GenericSelector` is temporary and will be removed in the future when we had components to all available types.

Also, [DROOLS-3665](https://issues.jboss.org/projects/DROOLS/issues/DROOLS-3665) (time component) and [DROOLS-3666](https://issues.jboss.org/projects/DROOLS/issues/DROOLS-3666) (Years/Mos component) are being created based on this PR.


jomarko:
I haven't found an issue during manual check. I will continue with code review next week.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/generic/GenericSelectorViewTest.java
```diff
@@ -0,0 +1,96 @@
...
+        final String attribute = "placeholder";
+        final String value = "value";
+
+        selectorView.setPlaceholder(value);
+
+        verify(input).setAttribute(attribute, value);
+    }
+
+    @Test
+    public void select() {
```
karreiro: I've noticed that this test is called `select` and the others start with the `test` prefix. On `kie-wb-common` and `appformer` the convention is starting test methods with the `test` prefix and suffix the classes with the `Test` suffix. So I think we could fix it and always use the `test` prefix for every method ;-)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateSelectorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private DateSelector dateSelector;
+
+    @Before
+    public void setup() {
+        dateSelector = spy(new DateSelector(view, valueFormatter));
+    }
+
+    @Test
+    public void testGetValue() {
```
karreiro: The coverage is not the only important thing in a unit test. Notice that the `getValue` method is `100%` covered, however, I could change it:
```java
    @Override
    public String getValue() {
        view.getValue();
        return "Wrong value!!!";
    }
```
..and the test `testGetValue` still would pass. We need to **really** test the logic of the method. See the following unit test (now, it really tests the method and breaks when the logic changes):

```java
    @Test
    public void testGetValue() {

        final String expectedValue = "value";
        when(view.getValue()).thenReturn(expectedValue);

        final String actualValue = dateSelector.getValue();

        assertEquals(expectedValue, actualValue);
    }
```

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeSelectView.java
```diff
@@ -64,6 +64,7 @@
     private DataTypeSelect presenter;
 
     private String value;
+    private String rawValue;
```
karreiro: I think we could remove this.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/TypedValueSelector.java
```diff
@@ -0,0 +1,41 @@
...
+
+    String getValue();
+
+    void setValue(final String value);
+
+    void setPlaceholder(final String placeholder);
+
+    Element getElement();
+
+    void onValueChanged(final Consumer<Object> onValueChanged);
```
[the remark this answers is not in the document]
danielzhe: I'm not sure, because at the end it will be consumed [here](https://github.com/kiegroup/kie-wb-common/pull/2508/files#diff-b4ba55a038e6f81b9a2af7cb5c7607faR65). Maybe I should change what is passed to `onValueChanged.accept(event);`. Wdyt?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateControl.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.gwt.core.client.JavaScriptObject;
+import elemental2.dom.Element;
+import jsinterop.annotations.JsMethod;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+
+import static jsinterop.annotations.JsPackage.GLOBAL;
+
+@JsType(isNative = true, namespace = JsPackage.GLOBAL)
+public abstract class DateControl {
```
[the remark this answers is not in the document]
danielzhe: Good. Thank you! :)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/generic/GenericSelectorViewTest.java
danielzhe: Well spotted! Thank you

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/TypedValueSelector.java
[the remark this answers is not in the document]
karreiro: @danielzhe I think we could change everywhere from `Object` to `elemental2.dom.Event`:


Am I missing something?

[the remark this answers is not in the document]
danielzhe: No, I was missing. Sorry. I did the change as request. Thank you. :)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateValueFormatterTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    public void testToDisplayInvalidDateString() {
+        testToDisplay("someInvalidDate", "");
+    }
+
+    @Test
+    public void testToDisplayValidDates() {
+        testToDisplay("date(\"2019-1-1\")", "01 " + localizedShortMonths[0] + " 2019");
+        testToDisplay("date(\"2019-6-2\")", "02 " + localizedShortMonths[5] + " 2019");
+        testToDisplay("date(\"2019-12-31\")", "31 " + localizedShortMonths[11] + " 2019");
```
jomarko: Thank you for these tests! They brought me to an idea. What if user put also whitespace, like "date( \"2019-12-31\" )". In that case date is not parsed. I think we could improve it by adding `removePrefixAndSuffix` next to `DateValueFormatter.removePrefixAndSuffix`. Please let me know what do you think. Also if you think the risk is too high, don't implement it.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/generic/GenericSelectorView.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public void setPlaceholder(final String placeholder) {
+        input.setAttribute("placeholder", placeholder);
+    }
+
+    @Override
+    public void onValueChanged(final Consumer<Event> onValueChanged) {
+        input.onkeyup = null;
+        input.onkeyup = event -> {
```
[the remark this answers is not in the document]
danielzhe: Good point. I thought it could leak (some memories from the past... :D), but after your comment I did some tests and it doesn't leak.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/enumeration/item/DataTypeConstraintEnumerationItemView.java
```diff
@@ -201,15 +215,15 @@ private void setText(final String value) {
             valueText.textContent = none();
         } else {
             valueText.classList.remove(NONE_CSS_CLASS);
-            valueText.textContent = value;
+            valueText.textContent = typedValueSelector.toDisplay(value);
         }
     }
 
     private void setInput(final String value) {
         if (isNULL(value)) {
```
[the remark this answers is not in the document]
danielzhe: This custom implementation checks if `value` is the string `""null""`.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateValueFormatterTest.java
danielzhe: For new files, it will not be possible to type the value for this case, since we are generating it based on what user select in the interface. But for older files, may it was the case that user already have some file with the value saved with the white spaces like you mentioned. Is that a possibility? If so, I think we can handle it with a `trim()` right next `removePrefixAndSuffix` as you mentioned. What do you think?

jomarko: I agree, maybe we can put `trim()` directly into `removePrefixAndSuffix`?
```java
String removePrefixAndSuffix(final String rawValue) {
        return rawValue.replace(PREFIX, "").replace(SUFFIX, "").trim();
    }
```

jomarko:
@danielzhe thank you for feedback incorporation, I responded about the `trim` [here](https://github.com/kiegroup/kie-wb-common/pull/2508#discussion_r262372631).

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateValueFormatterTest.java
danielzhe: There is. :)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/TypedValueSelector.java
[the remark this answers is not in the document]
karreiro: Np ;-) Thank-you!

PR #2509, opened 2019-03-01
JBPM-7629: [Stunner] Not possible to move some nodes to/from sub-process

romartin:
Hey @hasys @tiagodolphine 

See [JBPM-7629](https://issues.jboss.org/browse/)

Sorry guys about the commit size but to fix these kind of issues there were necessary to apply both fixes and improvements on the code, and have to be done at once as well. It was not an easy task, fixes were on the core and quite complex... but I'm happy on this, it was necessary at some point to achieve all our requirements and also making the code better and more robust.

So will give you feedback about these changes- which issues are they fixing and also some comments about the code updates, for having an idea of what's going on here...

**Issues & fixes**

Once a couple of nodes are connected by _sequence flows_, some issues appear. So fixed those and a few others as well: 

* It's not possible to move two (or more) nodes, connected by seq flows, from/into some subprocess
Before:
issue-subp-multiple-containment

After:
fix-subp-multiple-containment

* It's possible to move a gateway as a child of some task, when the gateway is connected by some seq flow
Before:
issue-task-containment

After:
fix-task-containment

* Consider an emb. subprocess which contains a task (`task1`) and a lane, which contains another task (`task2`) as well
** It's not possible to connect `task1` to `task2`, even both are in same subprocess
** It's possible to connect a task, which is on the process' root, to task2, and viceversa, and both are in different subprocesses
Before:
issue-subp-lanes-connection
issue-subp-lanes-connection2

After:
fix-subp-lanes-connection

* It's possible to add more than a single incomming connection for an Intermediate Signal Event while authoring, but once validating the process it complains about it. It should just do not allow adding a second incoming connection while authoring. (This is the right fix for [this previous one](https://github.com/kiegroup/kie-wb-common/pull/2451#issuecomment-463458481) we're agree on @hasys ) :+1: 
Before:
issue-connector-cardinality

After:
fix-connector-cardinality

PS @hasys @tiagodolphine  - I've noticed that once adding a lane into a subprocess the marshallers are not working good, but it's already reported in [JBPM-7534](https://issues.jboss.org/browse/JBPM-7534), so well at least this fixes part of the issue and ensure the core and UI are behaving as expected, next step now is to fix JBPM-7534

**Code updates**

* Refactored command related parent-child relationships (`setChild`, `removeChild`, `updateChild`, etc) to support multiple candidates (eg: `SetChildrenCommand`, `UpdateChildrenCommand`, `RemoveChildrenCommand`, etc) 
* Refactored both containment and cardinality rule contexts to support multiple candidates
* Support for multiple rule context evaluations
** The command execution context has been improved by having some evaluation state, which can be shared across commands in order to do composited evaluations for multiple contexts
** Once a command checks the execution context (calling `allow`), whatever rule evaluation (cardinality, containment, etc) it performs is being added into the execution context's state, so whatever other command which used the same execution context will have access to the state and so able to evaluate the rules against the "transient" state, instead of the real graph instance (which is only updated once calling `execute`) 
* Refactored `ContainmentAcceptorControl` in order to evaluate/execute multiple containment commands instead of rely on command composition 
* Reverted some previous quick fix applied [here](https://github.com/kiegroup/kie-wb-common/pull/2451#issuecomment-463458481) - Now cardinality rule violations for edges are of type `ERROR` again
* Fixed (re-implemented) the `ParentTypeMatcher` stuff
* Some other fixes on rule evaluations and commands stuff  
* Finally rather than fixing/adding the right tests, I've been adding some tests that run against "real" graph structures and rules, in order to avoid breaking this next times

Thanks!!
Roger


romartin:

hasys:
Hi @romartin,

great work! I didn't have much time to test it today, but I did some quick test and found some issues listed below. Will provide more feedback including code review on Monday. Thank you!

### Found issues:
1. Select nodes as highlighted on the picture below, move them to Ad-Hoc sub-process. Nodes position will be reverted, witch is fine, but Events borders won't be reset to default "selected" state  (blue border). Tasks won't be highlighted by red color at all when you are trying to do a wrong move.
image
2. Same as 1. is applied also for Sequence flows, but only in case if connection wasn't part of selected group.
image
3. For some cases when you move nodes from one sub-process to another all nodes displays under the new sub-process, but sequence flows not.
screenshot from 2019-03-01 16-33-55
4. When I clicked on Created Embedded Sub-process, browser freezes and I asked to refresh page:
image
image
5. Bendpoints not moved with the rest of the nodes if groups of nodes where selected by Shift+Click (there are no other possibilities to select group of nodes if it's located inside of the parent). If nodes located on the diagram level and selected by group selection, all works fine.
screenshot from 2019-03-01 16-54-01
6. When I append node close to the end of the sub-process, new node will be located somewhere far away from the sub-process, but still will be a child node of the sub-process:
image
7. Model Start->End process, add sub-process to the canvas. Select Start->End nodes and remove it. Undo, select Start->End again, move it inside of the sub-process. You will see copy of Start and End nodes, but they won't be part of a graph. If you select or remove nodes inside of the sub-process, its copies also will be selected/removed. If you select/remove copy of the node, original node will be selected/removed as well.
image

### Unrelated Issues found:
1. When you close Alerts panel and trigger some rules violation, like pictured below, Alerts panel will be opened again.
image

romartin:
Hey @hasys 

Thanks for the review!! I've been working on most of issues you reported [above](https://github.com/kiegroup/kie-wb-common/pull/2509#issuecomment-468717691). 

**Fixed 1, 2, 4, 6 and 7**.

_Note:_ I've fixed 6 by throwing an error when the new node, created from the quick menu, is being added outside the parent. So it works as a few weeks ago now - you've to resize the parent and then click again. I know it's not ideal but at this stuff is something not easy to refactor, and I'm pretty sure we already have some reported issues about it. So let's go ahead with this, if you're agree.

So **missing to fix 3 and 5 and the unrelated issue**. As most of the issue are not regressions, and I've fixed most on the new commit, I think we should be now pragmatic and report new issues, trying to move this stuff forward. There are two reason: it's quite lot of code, so as soon as we merge it we'll get feedback soon too, and because I'm spending more time than expected on bug fixing and I should already be moving to new stuff. So let's try to move the issue found, not related to this task, to our next bug fixing phase, if you agree!!

Thanks!

romartin:

hasys:
Hi @romartin, looks good to me now, great job, thank you!

#### Reported unrelated issues:
* [JBPM-8304](https://issues.jboss.org/browse/JBPM-8304) - events changes its highlight to deselected state after incorrect move
* [JBPM-8303](https://issues.jboss.org/browse/JBPM-8303) - alerts opens every time designer rules are violated
* [JBPM-8299](https://issues.jboss.org/browse/JBPM-8299) - nodes hidden under sub-process during drag and drop
* [JBPM-8300](https://issues.jboss.org/browse/JBPM-8300) - bendpoints doesn't move with nodes when selected by Shift+Click
* [JBPM-8301](https://issues.jboss.org/browse/JBPM-8301) - boundary event remains attached to the node when you move it out
* [JBPM-8302](https://issues.jboss.org/browse/JBPM-8302) - sequence flow displayed despite of node removed

romartin:
great, thanks a lot  @hasys , great job again!! :+1:

PR #2510, opened 2019-03-01
DROOLS-3668 / DROOLS-3667 / DROOLS-3669: Included Models tab

karreiro:
See:
- https://issues.jboss.org/browse/DROOLS-3667
- https://issues.jboss.org/browse/DROOLS-3668
- https://issues.jboss.org/browse/DROOLS-3669

---

Demo:


jomarko:
### Manual check notes
- From my point of view would be better to have just one button for canceling renaming of included model. Please ignore this if those two cancel buttons have different behavior.
- If you rename some model, save the dmn diagram and reopen, you will get this [1]. Using the file you provided.

[1]
```
An error was produced during process diagram parsing. There might be bpm nodes not yet supported by current Process Designer version, or the xml content can contain errors. You can try to open the file with the Legacy Process Designer or repair it manually.
```

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/common/cards/frame/CardFrameComponentTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Test
+    public void testIsToggleTitleEnabledWhenItsEnabled() {
+
+        final CardComponent card = mock(CardComponent.class);
+
+        when(card.onTitleChanged()).thenReturn(newTitle -> true);
+        doReturn(card).when(cardFrame).getCard();
+
+        assertTrue(cardFrame.isToggleTitleEnabled());
```
jomarko: Could you please explain more that `onTitleChanged()` function. I thought that it will cause save button is disabled/grey, until the user really changes the name, however during manual test I found out it is not a case.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/common/cards/CardsGridComponent.java
```diff
@@ -0,0 +1,92 @@
...
+        cards.forEach(this::appendCard);
+        setupEmptyState(cards.isEmpty());
+    }
+
+    public void setEmptyState(final HTMLElement emptyStateElement) {
+        this.emptyStateElement = emptyStateElement;
+    }
+
+    private void setupEmptyState(final boolean isEmptyStateEnabled) {
+        if (isEmptyStateEnabled && getEmptyStateElement().isPresent()) {
```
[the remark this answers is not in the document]
manstis: @jomarko I think this is correct.. it defends against [this](https://github.com/kiegroup/kie-wb-common/pull/2510/files#diff-988f8e729b8a14201547f32c936601a3R57) not being called by the explicit concrete consumer of the `CardsGridComponent`

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/included/imports/persistence/ImportRecordEngine.java
```diff
@@ -0,0 +1,106 @@
...
+    public List<IncludedModel> create(final IncludedModel record) {
+        // TODO: https://issues.jboss.org/browse/DROOLS-3719
+        return singletonList(record);
+    }
+
+    @Override
+    public boolean isValid(final IncludedModel record) {
+        final boolean isUnique = isUnique(record);
+        if (!isUnique) {
+            flashMessageEvent.fire(messageFactory.getNameIsNotUniqueFlashMessage(record));
```
manstis: You could possibly show different messages depending on whether the `name` or (presumably) `path` are unique or not.. obviously it's not possible to test this in the UI yet, at least not until the ability to add new _included models_ is added... but any change here can follow on with that work; if needed.

manstis:
Oh, @karreiro I forgot to add in my review comment a note about just how _reusable_ the generic Card will be (for scesim).. as they'd need to have a dependency on `kie-wb-common-dmn-client` that will obviously bring a whole heap of other stuff too (it won't be a problem in "Business Central" as the WAR will have DMN anyway...). I wonder whether moving the generic Card stuff to, for example, `kie-wb-common/kie-wb-common-widgets/kie-wb-common-ui` is better (most, if not all, editors have a dependency on that module already).

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/common/cards/CardsGridComponent.java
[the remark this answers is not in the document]
karreiro: Precisely (thank you @manstis) :-)

Since it's a generic component, the API user may not set the empty state.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/included/IncludedModelsPageState.java
```diff
@@ -0,0 +1,49 @@
...
+@ApplicationScoped
+public class IncludedModelsPageState {
+
+    private IncludedModelsPageStateProvider pageProvider;
+
+    public void init(final IncludedModelsPageStateProvider pageProvider) {
+        this.pageProvider = pageProvider;
+    }
+
+    public List<IncludedModel> generateIncludedModels() {
```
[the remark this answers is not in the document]
karreiro: Generally, I use static classes for utility methods (something that does not represent state). Since this class holds the page state, I think that's a good idea to keep it in CDI context. Also, considering the direction that this class will evolve, I believe that keeping this as `@ApplicationScoped` makes things more flexible. But, thank you for raising this. It's an interesting observation :-)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/common/cards/frame/CardFrameComponentTest.java
karreiro: An implementation of the `CardComponent` can or cannot have an `onTitleChanged` function. When it has a function, the component will enable the toggle feature for the title (so the user will be able to change the title by clicking on it).

When the user changes the title, we invoke that function, passing the new title as a parameter. The implementation of `CardComponent` decides what happens with the title after that. In our `DMNCardComponent` we're doing this:

```java
    public Function<String, Boolean> onTitleChanged() {
        return newName -> {

            final String oldName = getIncludedModel().getName();

            getIncludedModel().setName(newName);

            if (getIncludedModel().isValid()) {
                getIncludedModel().update();
                getGrid().refresh();
                return true;
            } else {
                getIncludedModel().setName(oldName);
                return false;
            }
        };
    }
```

Notice, if the new title is valid, we're updating the `IncludedModel` `true` is returned, otherwise we return `false` and the card component knows that the new value should be discarded.

[1] https://github.com/karreiro/kie-wb-common/blob/DROOLS-3667-3668-3669/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/common/cards/CardComponent.java#L42

karreiro:
Thank you for the review @manstis and @jomarko :-) Changes applied.

I've introduced a new validation, for blank model names:




Regarding the "X" vs. the input "X" button, I talked to Liz and she proposed to change the behaviour of the input "X" button, now it cleans the input, see:






---

@jomarko,

> If you rename some model, save the dmn diagram and reopen, you will get this [1]. Using the file you provided. Plain xml editor is opened instead of the dmn designer.

The [1] error was not introduced by this PR. It's a known issue and it's my next JIRA (I'm opening a JIRA for that in a few minutes).


> You can also invoke edit/rename for multiple cards at once, but when you click save, all edit modes are canceled and saved just the one where you clicked.

Fixed! Now only one edit mode at time is enabled:




---

@manstis,

Thank you for the suggestion. I've moved the card component to the `kie-wb-common-ui` ;-)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/included/imports/persistence/ImportRecordEngine.java
karreiro: Yeah, Mike.. for sure :-)

I've already slightly changed this method (introducing a new validation). Probably it will be moved to another class when we need even more complex validations with other fields like you mentioned (similar to what we have in Data Types).

karreiro:
@jomarko Here's the JIRA https://issues.jboss.org/browse/DROOLS-3725 related to the issue you mentioned.

manstis:
@gitgabrio @danielezonca FYI... this PR adds support for a "Card View" to `kie-wb-common/kie-wb-common-widgets/kie-wb-common-ui` with the intent that it's reusable by _scesim_ for (presumably) the DMN model selection (IIUC). @karreiro is your "go to" man when you get to the point of needing to re-use it..

gitgabrio:
@karreiro @manstis 
Just to know, what this component is for? I mean, what it does?

manstis:
@gitgabrio In the DMN Editor it allows the User to view _external_ DMN files that have been _imported_. @gitgabrio I was led to believe `scesim` was to use something similar to select a DMN file to test against... of course I can be completely wrong. Let's see what @karreiro has to offer (he's on PTO until Wednesday).

gitgabrio:
@manstis @karreiro 
Well, maybe it is not - that's why I asked. What is needed for scesim is just a "combobox" to _choose_ already existing dmn files (it should look like the "package" selector, but for DMN files). That combobox should then be included in the "new scenario creation" widget.

manstis:
@karreiro I'll let you merge as you see fit. 

IDK if you want to squash the commits (your original PR kept the different JIRA as different commits).

PR #2523, opened 2019-03-08
DROOLS-3725: [DMN Designer] Included Models - Error when a file with included models is saved

karreiro:
See https://issues.jboss.org/browse/DROOLS-3725

---

Before:

After:

---

The error was happening because the property `xmlns:drools` (from the `import` element) was being discarded (PS.: this property was added by a third party editor).
```xml
<semantic:import
  namespace="http://www.treasuretech.com/definitions/789456123459789"
  name="model-test"
  treasure:fileId="123456789123456798"
  treasure:fileName="karreiro/dmn-tests/model-test"
  importType="http://www.omg.org/spec/DMN/20180521/MODEL/"
  drools:modelName="model-test"
  xmlns:drools="http://www.drools.org/kie/dmn/1.1"/>
```


karreiro:
@jomarko I just check the steps above, and I think the PR is working as expected (regarding the warning related to unsaved changes), see:



Am I missing something?

karreiro:
Thank you, @jomarko.

PR #2528, opened 2019-03-08
[7.18.x] RHDM-869: [DMN Designer] [IE11] Empty data type tab

danielzhe:


danielzhe:
@karreiro Changes applied.

PR #2532, opened 2019-03-13
DROOLS-3756 / DROOLS-3757: [DMN Designer] Data Types - Constraints (Range/Enumeration) - Add "Number"/"String" component when the type is "Number"/"String"

karreiro:
See:
- https://issues.jboss.org/browse/DROOLS-3756
- https://issues.jboss.org/browse/DROOLS-3757

---


karreiro:
@jomarko @danielzhe  Thank you for the review :-)

@jomarko Regarding the NumberSelector behaviour, this is stated by the input `type` attribute (for the NumberSelector component the input `type` is `number`, for other inputs, the default is `text`):

Screen Shot 2019-03-14 at 09 59 02

The `getInputType` method returns the input type, since it returns `number`, the browser takes care of the native small widget (to the right) and for blocking some characters, but this is something native from the browser.

From our side, we're testing the `getInputType` method in the `NumberSelectorTest` class :-)

PR #2535, opened 2019-03-14
AF-1875-Add string cut for breadcrumb Links and toolbar title

ohonchar:
Added string cut for Toolbar title to avoid breaking html layout due long names


domhanak:
@manstis Are you ok with this?

PR #2557, opened 2019-03-26
JBPM-6454 Stunner - rethink text wrapping and fonts for nodes

tiagodolphine:
- Text Wrapping for BPMN nodes
- Multi-line support
- New node label layout:
Vertical and Horizontal alignment
Position: outside/inside parent node
Orientation: vertical, horizontal
Margins

- Long text in tasks does not overlap icon of task type
- Events, Gateways, has labels on bottom-middle
- Removed the **bold** from labels

Assembly:
https://github.com/kiegroup/lienzo-core/pull/94
https://github.com/kiegroup/lienzo-tests/pull/73

@romartin 
@LuboTerifaj

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/shape/def/DMNDecisionServiceSVGShapeDefImpl.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/shape/def/DMNDecisionServiceSVGShapeDefImpl.java
@@ -73,8 +73,11 @@ public class DMNDecisionServiceSVGShapeDefImpl implements DMNDecisionServiceSVGS
                 .fontSize(bean -> bean.getFontSet().getFontSize().getValue())
                 .textWrapperStrategy(bean -> TextWrapperStrategy.TRUNCATE)
                 .strokeAlpha(bean -> 0.0d)
-                .position(bean -> HasTitle.Position.TOP)
-                .positionYOffset(bean -> Y_OFFSET)
+                .verticalAlignment(bean -> HasTitle.VerticalAlignment.TOP)
+                .horizontalAlignment(bean -> HasTitle.HorizontalAlignment.CENTER)
+                .referencePosition(bean -> HasTitle.ReferencePosition.INSIDE)
+                .orientation(bean -> HasTitle.Orientation.HORIZONTAL)
+                .margin(HasTitle.VerticalAlignment.TOP, Y_OFFSET)
                 .build();
     }
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/shape/view/handlers/DMNViewHandlers.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/shape/view/handlers/DMNViewHandlers.java
@@ -19,6 +19,7 @@ package org.kie.workbench.common.dmn.client.shape.view.handlers;
 import org.kie.workbench.common.dmn.api.definition.DMNDefinition;
 import org.kie.workbench.common.dmn.api.definition.DMNViewDefinition;
 import org.kie.workbench.common.stunner.core.client.shape.TextWrapperStrategy;
+import org.kie.workbench.common.stunner.core.client.shape.view.HasTitle;
 import org.kie.workbench.common.stunner.core.client.shape.view.ShapeView;
 import org.kie.workbench.common.stunner.core.client.shape.view.handler.FontHandler;
 import org.kie.workbench.common.stunner.core.client.shape.view.handler.SizeHandler;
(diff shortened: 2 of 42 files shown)


tiagodolphine:
Jenkins retest this

tiagodolphine:
Jenkins retest this

tiagodolphine:
Jenkins retest this

tiagodolphine:

tiagodolphine:

romartin:

romartin:
Hey @tiagodolphine 

I think the full build is failing due to some DMN compilation issue... can you please check and re-fire the build once possible?

Also, just to clarify this WIP, when is it ready for review? what's missing? can reviewers start our work here?

Thx

tiagodolphine:
@romartin I'm working on the unit tests, the code itself if ready, I will update it soon, besides I'll check this DMN issue you mentioned.
Thanks.

tiagodolphine:

tiagodolphine:
@romartin this is ok to be reviewed, just missing the **lienzo-core** tests. I will update it soon.
@LuboTerifaj as I mentioned on the call earlier, the PR you reported the issue that broke the editor, is merged on this PR, so now let's keep all changes here, I think it is better for review and that issue is not present anymore (at least I was not able to reproduce).

Thanks !!

tiagodolphine:

tiagodolphine:
Hi @LuboTerifaj , thanks for the review !

The follow isssues regarding the inline editor, I think, they will be handled on the [JBPM-7117](https://issues.jboss.org/browse/JBPM-7117), that @treblereel is working on.
_*To add multiline String to a node using "inline editing popup window" (not Diagram properties), a user must open the window again and again. After every "enter" press, the window is closed.
*When already mentioned window is opened, it is possible to input characters, but after moving with mouse pointer outside this window and then back, the typing is disabled._

About the others: 

_*If a longer String (for example 150 cahracters long) is added to more activities, the whole designer is rapidly slower than before. It is lagging when moving or resizing nodes._

**I'll take a look, can you export a bpmn that you used for tests and attach to the JIRA ?**

_*After adding text to service tasks, resizing point is disappeared. There is a need to deselect node and select it back._

**I'll take a look but this seems to be unrelated with this PR, I need to confirm.**

_*All tasks, which contains icon in upper left corner, don't have its name in the middle of task. I know that it is implemneted so to avoid overlapping icon. If a task is not enlarged, it looks good, but after resizing, it looks a little bit asymetric. Would be it possible to distance text also from the right edge of node tasks? It could be centred if the text is below icon._

**Yes, now it has a margin set on the left side, to avoid overlapping the icon on left, I can decrease the margin that is now with size of 50px. Besides that the text will have the horizontal alighment centered.** 

_*It is possible to resize an activity in a way, that the text will be exactly on the edge of this activity. It should be dispalyed either outside or inside._

**Can you please provide some screenshot or steps to execute to reproduce it?**

_* Lanes don't show any new line characters. It also causes, that if a longer text is used containing new line character (e.g. "123456789123456789123456789\nsgd"), after the size of lane is enlarged, not the whole part before new line character is displayed, even there is a place to show it._

**I let the lanes with the previous config, that has only one line size (if I recall correctly), but I will change it to fit the whole shape.**

_* In Lane, if there is new line character after any word, three dots are displayed after a few characters of the first or more words. It can be zero chars, but also one or more. It depends on the length of used words._

**I will take a look on this issue.**

Thanks again.

tiagodolphine:

tiagodolphine:
Hi @LuboTerifaj updated with fixes for the comments. So ready to review again.

About 
*After adding text to service tasks, resizing point is disappeared. There is a need to deselect node and select it back.
This is not related to this PR, it is better to open a new Jira for it, and in fact, it happens not only with service tasks it happens to all nodes, and any modification on properties make the resize control points to disappear.

Thanks.

LuboTerifaj:
Regarding issues with inline editor, reported issues:

- [Stunner - Pressing enter closes Inline text editor](https://issues.jboss.org/browse/JBPM-8407)
- [Stunner - Cannot input characters in to Inline editor](https://issues.jboss.org/browse/JBPM-8408)

tiagodolphine:
Hi @LuboTerifaj thanks for the review and opening the new JIRAs.
*Task with name (e.g. "Taskvfdsasda") resizing, ....
Mmm yes I agree, I'll take a look, in fact, the size limiting to wrap is based on the width of a given "word", not on the number of chars, but they should look the same in case the text and the node size are the same.
*It is not possible to move with nodes, which are placed over the input text.
I'll take a look on this issue, maybe this issue is already present master.

Thanks.

tiagodolphine:

tiagodolphine:
@LuboTerifaj this is ok to be reviewed I fixed your comments, including the one about child nodes on parent with long title.

Thanks.

LuboTerifaj:
@tiagodolphine 

There is one more thing broken. In Case Management editor, every node has shifted its name. This needs to be fixed.

Overall, It works in different way then in process designer, but this is present also on master, so I can open Jira for that if it is out of scope of this PR.

tiagodolphine:
- About the first comment it seems to be working on this way, 1 char per line:
image
Anyway, IMO this behavior is not wrong, and besides this is an uncommon use case, WDYT to open a new JIRA as an enhancement specifying all the restrictions, examples and the expected behavior?

- The issue regarding the svg on documentation, I was not able to reproduce:
Documentation:
image

Process Diagram:
image

Do you see any blocker on the PR? I mean, we can open enhancements as detected with more details on new JIRAs, it is important to move on to merge this since it is becoming too long with many changes.

Thanks !!

tiagodolphine:
@LuboTerifaj I added the fix to the Case Management

tiagodolphine:

tiagodolphine:
@LuboTerifaj  thanks for the review and opening the JIRAs.

@romartin this is ok to be merged.

PR #2564, opened 2019-04-01
JBPM-6453 - handle error when Node creation is not possible

hasys:
Hi @romartin,

looking at [JBPM-6453](https://issues.jboss.org/browse/JBPM-6453). It is a very old issue, we should resolve it or close. For now I did some refactoring and added test to changes.

Can you navigate me to right direction how error should be handled if something going wrong?


romartin:
Hey @hasys ! I was looking at source code trying to see how to fix it when just noticed this class is part of the old marshallers stuff. They're not being used and as commented on our meetings, the idea now is to remove those "old" classes (see [JBPM-8395](https://issues.jboss.org/browse/JBPM-8395)).

Saying this, why not just forgetting this fix and closing this PR? WDYT?

hasys:
Good point @romartin,

closing this PR.

PR #2571, opened 2019-04-04
Tidy-up and add additional Submarine requirements

manstis:


kie-wb-common-submarine/kie-wb-common-submarine-client/src/main/java/org/kie/workbench/common/submarine/client/editor/BaseSubmarineEditor.java
```diff
@@ -137,14 +131,30 @@ public boolean isReadOnly() {
...
+    /**
+     * Used by Submarine to set the content of the editor.
+     */
+    public abstract void setContent(final String value);
+
+    /**
+     * Used by Submarine to get the content of the editor. This should return a {@link String}
+     * representation of the editors content to persist to an underlying persistent store.
+     */
+    public abstract Promise getContent();
```
jomarko: @manstis Hi, I checked javadoc of `Promise`, shouldn't we use generics? `Promise<String>`. Maybe I would like to know more about returned format, it it json, xml, csv, or short sample?

manstis: @jomarko I originally had `Promise<String>` however @porcelli changed it back to the non-generic form shown here in a PR he submitted to my fork.. IDK if he has good reason.. AFAIK `getContent()` is expected to *always* return a `String`.. personally I'm happy to revert back to `Promise<String>`.. @porcelli can you provide any insight as to why you're preferring the un-generic `Promise` return type (is it so we can integrate the "Java" generic form of `Promise` with TypeScript/JavaScript?

porcelli: It was a quick shortcut in my case to not get APT to use generics :)
We need tidy up this work anyway.. the idea was just a quick prototype

PR #2575, opened 2019-04-08
RHPAM-1892 - Ad Hoc Subprocess: Signal end node is not allowed

Josephblt:
https://issues.jboss.org/browse/RHPAM-1892

Ad Hoc Subprocess Nodes supports End Nodes containment.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/AdHocSubprocess.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/AdHocSubprocess.java
@@ -47,7 +47,7 @@ import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.pr
 
 @Portable
 @Bindable
-@CanContain(roles = {"cm_activity", "cm_stage", "IntermediateEventsMorph", "GatewaysMorph"})
+@CanContain(roles = {"cm_activity", "cm_stage", "IntermediateEventsMorph", "GatewaysMorph", "EndEventsMorph"})
 @CanDock(roles = {"IntermediateEventOnSubprocessBoundary"})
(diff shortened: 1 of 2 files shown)


Josephblt:

Josephblt:
Hello @romartin, @LuboTerifaj. If you guys could please review this, that would be great. Thanks

Josephblt:

Josephblt:
Hey @LuboTerifaj. Could you please review this? Thanks.

Josephblt:

wmedvede:
Hello @romartin @Josephblt 
To add EndEvents in a Stunner Ad-hoc subprocess is currently not enabled because bpmn2 spec says that "addhoc subprocesses" don't contain end events. This is why it's currently not possible to add them, even when it was possible do it in old jBPMDesigner. See the copy paste from BPMN2 spec below.

I'd suggest @Josephblt  to take a deeper look and ask @mswiderski if we have semantic for this type of use case in jbpm engine.


13.2.5 Ad-Hoc Sub-Process
An Ad-Hoc Sub-Process or Process contains a number of embedded inner Activities and is intended to be executed
with a more flexible ordering compared to the typical routing of Processes. Unlike regular Processes, it does not
contain a complete, structured BPMN diagram description—i.e., from Start Event to End Event. Instead the Ad-Hoc
Sub-Process contains only Activities, Sequence Flows, Gateways, and Intermediate Events. An Ad-Hoc
Sub-Process MAY also contain Data Objects and Data Associations. The Activities within the Ad-Hoc Sub-
Process are not REQUIRED to have incoming and outgoing Sequence Flows. However, it is possible to specify
Sequence Flows between some of the contained Activities. When used, Sequence Flows will provide the same
ordering constraints as in a regular Process. To have any meaning, Intermediate Events will have outgoing
Sequence Flows and they can be triggered multiple times while the Ad-Hoc Sub-Process is active.
The contained Activities are executed sequentially or in parallel, they can be executed multiple times in an order that is
only constrained through the specified Sequence Flows, Gateways, and data connections.

wmedvede:
Sorry for the comment above guys @Josephblt @romartin  , I haven't read the thread in the JIRA and it looks like @hasys  raised same question as I, and it was already responded by @mswiderski .
+1 from my side

Josephblt:
Hello @LuboTerifaj. If you have some time could you please take a look at this? Thanks

Josephblt:

PR #2579, opened 2019-04-09
DROOLS-3702: [Stunner] Arrows are incorrectly positioned when save-open diagram

manstis:
See https://issues.jboss.org/browse/DROOLS-3702

This fixes the issue reported by @danielzhe and the preview when authoring.

It improves on the `MagnetConnector.Builder.forTarget(..)` method to be less naive and considers target nodes above or below the source. The old implementation simply checked for left or right that was OK for BPMN as most diagrams run from left to right whereas DMN is top to bottom.

However please note the preview does not appear to honour the "auto" setting on the connection. This is not a regression caused by this PR but is in `master` too. I've included @romartin and @hasys on the review for BPMN.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/MagnetConnection.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/MagnetConnection.java
@@ -33,7 +33,9 @@ import static org.kie.soup.commons.validation.PortablePreconditions.checkNotNull
 public class MagnetConnection extends DiscreteConnection {
 
     public static final int MAGNET_CENTER = 0;
+    public static final int MAGNET_TOP = 1;
     public static final int MAGNET_RIGHT = 2;
+    public static final int MAGNET_BOTTOM = 3;
     public static final int MAGNET_LEFT = 4;
 
     private Point2D location;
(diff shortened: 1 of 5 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/MagnetConnection.java
```diff
@@ -99,6 +101,14 @@ public String toString() {
     @NonPortable
     public static class Builder {
 
+        private enum RelativePosition {
+            ABOVE,
```
romartin: well just minor minor comment, but why not using carindals (eg: east, north, etc) instead of above, below.. just to keep consistency? or you think that words describe better the builder methods?

manstis: Hi @romartin I just chose to keep the terms consistent with what already existed in the [class](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/graph/content/view/MagnetConnection.java). However if you prefer to change these to the cardinal directions then I'm happy to do so.

manstis:
@romartin Regarding your comment about when the User attaches a connector to an explicit Magnet we, in the DMN team, have chosen to use the use of `auto` as a quick fix to the problem presented by @danielzhe in the JIRA. See @etirelli comment https://issues.jboss.org/browse/DROOLS-3702?focusedCommentId=13718294&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13718294.

If a User has chosen an explicit Magnet position then the x, y of this is preserved during marshalling however in operation the connector would appear to be "auto". If (when) Users complain of this we'd need to be a lot smarter - I think we'd shift to not using `auto` all together and have new connections default to `CENTRE` with `auto=false`. The current behaviour is to attach to an _edge_ Magnet and `auto=true`.

PR #2596, opened 2019-04-23
AF-1931: Description field is cleared on New Project pop up

gabriel376:
Fix and Tests for AF-1931

**Bug Details**
1. Click "Add New Project" to show pop up
2. Type some value on Description field
3. Click "Show Advanced Options" to show fields Version, Artifact and Group ID
4. Click "Hide Advanced Options" to hide these fields
5. **BUG: Description field is set to default value (blank)**

**Fix**
1. Removed logic to set Description default value when "Hide Advanced Options" is clicked
2. Kept logic to restore values for fields Version and Group ID

**Tests**
1. Added tests for Presenter to check if logic to restore default values for fields Version and Group ID is correct
2. Added tests for View to check if logic for the click of Show/Hide Advanced Options is correct

**Notes**
1. I personally find strange to have fields on Advanced Options to be reset when I click "Hide", because the label "Hide" just tells me that the panel will be invisible and the values will not be changed. Perhaps we can have two labels, one for hiding the panel and another one to reset the values.


kie-ci9:

ederign:
@paulovmr could you please take a look on this!

gabriel376:
@paulovmr sounds great to me.
I will implement this as soon I get some sort of approval to make this modification to the product. 
We will need to implement for all supported languages too. Who is responsible for translation ?

barboras7:

ederign:

ederign:
@gabriel376 just do it in english (with i18n) and all others translations will be done by other team

gabriel376:
**New Changes:**
1. Changed text "Show/Hide Advanced Options" to "Configure/Restore Advanced Options"
2. Move field "Artifact ID" from "Advanced Options" panel. This field is automatically filled by Project field and is not restored when the anchor is clicked, so it makes more sense to be along with fields Project and Description

barboras7:

barboras7:

barboras7:
Hi @gabriel376,

I have one concern regarding this fix. While I understand why the ArtifactId is not hidden anymore, I am not sure that from the UX point of view it is the best option. 

I believe that the GAV were hidden because some users may not understand the concept and they just want to create a project. Now, every user sees the Artifact ID field when creating a project and many of these users may have no idea what it is (it could be quite confusing).

@ederign, @paulovmr what do you think, guys?
Screenshot from 2019-05-06 12-31-24

gabriel376:
@barboras7 I agree with you, we need approval for both UX and DEV

paulovmr:
@gabriel376 I agree with @barboras7 on this, I think the Artifact ID is an advanced option and should stay inside the panel. With that said, I think the best to do here is to also restore the default artifact id value when "Reset Advanced Options" is clicked.

gabriel376:
Hi @paulovmr, thanks for your suggestion!
I will push another commit for this

barboras7:

ederign:
@paulovmr @barboras7 could you please take a look on this?

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/AddProjectPopUpPresenter.java
```diff
@@ -394,9 +394,10 @@ public void cancel() {
         view.hide();
     }
     
-    public void restoreDefaultAdvancedOptions() {
+    public void restoreAdvancedOptions() {
         libraryPreferences.load(loadedLibraryPreferences -> {
             view.setVersion(loadedLibraryPreferences.getProjectPreferences().getVersion());
+            view.setArtifactId("");
```
paulovmr: Here we should restore the project name without spaces instead of empty string (`view.setArtifactId(view.getName().replaceAll(" ", ""));`), because if the user first fills the name, then clicks on "Configure Advanced Options", then clicks "Restore Advanced Options", the artifact id will be empty.

gabriel376: good catch!

gabriel376: @paulovmr done!

paulovmr:

gabriel376:
I am adjusting some test cases, will commit asap

paulovmr:

paulovmr:
Unrelated selenium test failures.

gabriel376:
thanks!

PR #2611, opened 2019-05-02
Make kie-wb-common-library-spaces-screen module inherit version from its parent

tiagobento:
Using custom tags causes errors on release builds because the version update script doesn't update these custom versions anymore.


tiagobento:
cc @mbiarnes

PR #2612, opened 2019-05-02
JBPM-8432 API for Marshalling with a response to include error/warning messages

tiagodolphine:
- API for Marshalling that contains a response with messages regarding the marshaling process. 
- A simple implementation that for now just return an error message (in case current marshallers throws an exception due to any unsupported element or whatever) or success with the expected result (Graph). 

@romartin 
@wmedvede 
@LuboTerifaj  @hasys

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshaller.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshaller.java
@@ -21,6 +21,10 @@ import java.io.InputStream;
 import org.kie.workbench.common.stunner.core.diagram.Diagram;
 import org.kie.workbench.common.stunner.core.diagram.Metadata;
 import org.kie.workbench.common.stunner.core.graph.Graph;
+import org.kie.workbench.common.stunner.core.marshaller.MarshallingMessage;
+import org.kie.workbench.common.stunner.core.marshaller.MarshallingRequest;
+import org.kie.workbench.common.stunner.core.marshaller.MarshallingResponse;
+import org.kie.workbench.common.stunner.core.validation.Violation;
 
 /**
  * Provides marshalling and unmarshalling services for a Diagram.
@@ -39,6 +43,32 @@ public interface DiagramMarshaller<G extends Graph, M extends Metadata, D extend
      */
     G unmarshall(final M metadata,
                  final InputStream input) throws IOException;
+
+    /**
+     * Default implementation that returns a {@link MarshallingResponse} with the result based on
+     * {@link DiagramMarshaller#unmarshall(Metadata, InputStream)} with an error in case of any exception.
+     * @param request
+     * @return
+     */
+    default MarshallingResponse<G> unmarshallWithValidation(final MarshallingRequest<InputStream, M> request) {
+
+        try {
+            final G result = unmarshall(request.getMetadata(), request.getInput());
+            return MarshallingResponse.builder()
+                    .result(result)
+                    .state(MarshallingResponse.State.SUCCESS)
+                    .build();
+        } catch (Exception e) {
+            final String message = e.getMessage();
+            return MarshallingResponse.builder()
+                    .state(MarshallingResponse.State.ERROR)
+                    .addMessage(MarshallingMessage.builder()
+                                        .type(Violation.Type.ERROR)
+                                        .message(message)
+                                        .build())
+                    .build();
+        }
+    }
 
     /**
      * Serializes a diagram instance of type <code>D</code> as string.
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.marshaller;
+
+import java.util.Objects;
+
+import org.kie.workbench.common.stunner.core.validation.DomainViolation;
+
+public class MarshallingMessage implements DomainViolation {
+
+    private String elementUUID;
+    private int code;
+    private Type type;
+    private String message;
+
+    public String getElementUUID() {
+        return elementUUID;
+    }
+
+    @Override
+    public String getMessage() {
+        return message;
+    }
+
+    @Override
+    public Type getViolationType() {
+        return type;
+    }
+
+    public int getCode() {
+        return code;
+    }
+
+    public MarshallingMessage(String elementUUID, int code, Type type, String message) {
+        this.elementUUID = elementUUID;
+        this.code = code;
+        this.type = type;
+        this.message = message;
+    }
+
+    public static MarshallingMessageBuilder builder(){
+        return new MarshallingMessageBuilder();
+    }
+
+    public static class MarshallingMessageBuilder {
+
+        private String elementUUID;
+        private int code;
+        private Type type = Type.ERROR;
+        private String message;
+
+        public MarshallingMessageBuilder elementUUID(String elementUUID) {
+            this.elementUUID = elementUUID;
+            return this;
+        }
+
+        public MarshallingMessageBuilder code(int code) {
+            this.code = code;
+            return this;
+        }
+
+        public MarshallingMessageBuilder type(Type type) {
+            this.type = type;
+            return this;
+        }
+
+        public MarshallingMessageBuilder message(String message) {
+            this.message = message;
+            return this;
+        }
+
+        public MarshallingMessage build() {
+            return new MarshallingMessage(elementUUID, code, type, message);
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof MarshallingMessage)) {
+            return false;
+        }
+        MarshallingMessage that = (MarshallingMessage) o;
+        return getCode() == that.getCode() &&
+                Objects.equals(getElementUUID(), that.getElementUUID()) &&
+                type == that.type &&
+                Objects.equals(getMessage(), that.getMessage());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getElementUUID(), getCode(), type, getMessage());
+    }
+
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("MarshallingMessage{");
+        sb.append("elementUUID='").append(elementUUID).append('\'');
+        sb.append(", code=").append(code);
+        sb.append(", type=").append(type);
+        sb.append(", message='").append(message).append('\'');
+        sb.append('}');
+        return sb.toString();
+    }
+}
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
@@ -16,6 +16,8 @@
 
 package org.kie.workbench.common.stunner.core.marshaller;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
 import org.jboss.errai.common.client.api.annotations.MapsTo;
@@ -30,6 +32,8 @@ public class MarshallingMessage implements DomainViolation {
     private final int code;
     private final Type type;
     private final String message;
+    private final String messageKey;
+    private final List<?> messageArguments;
 
     public String getElementUUID() {
         return elementUUID;
@@ -49,12 +53,24 @@ public class MarshallingMessage implements DomainViolation {
         return code;
     }
 
+    public String getMessageKey() {
+        return messageKey;
+    }
+
+    public List<?> getMessageArguments() {
+        return messageArguments;
+    }
+
     public MarshallingMessage(@MapsTo("elementUUID") String elementUUID, @MapsTo("code") int code,
-                              @MapsTo("type") Type type, @MapsTo("message") String message) {
+                              @MapsTo("type") Type type, @MapsTo("message") String message,
+                              @MapsTo("messageKey") String messageKey,
+                              @MapsTo("messageArguments") List<?> messageArguments) {
         this.elementUUID = elementUUID;
         this.code = code;
         this.type = type;
         this.message = message;
+        this.messageKey = messageKey;
+        this.messageArguments = messageArguments;
     }
 
     public static MarshallingMessageBuilder builder(){
@@ -67,6 +83,8 @@ public class MarshallingMessage implements DomainViolation {
         private int code;
         private Type type = Type.ERROR;
         private String message;
+        private String messageKey;
+        private List<?> messageArguments = Collections.emptyList();
 
         public MarshallingMessageBuilder elementUUID(String elementUUID) {
             this.elementUUID = elementUUID;
@@ -88,8 +106,18 @@ public class MarshallingMessage implements DomainViolation {
             return this;
         }
 
+        public MarshallingMessageBuilder messageKey(String messageKey) {
+            this.messageKey = messageKey;
+            return this;
+        }
+
+        public MarshallingMessageBuilder messageArguments(List<?> messageArguments) {
+            this.messageArguments = messageArguments;
+            return this;
+        }
+
         public MarshallingMessage build() {
-            return new MarshallingMessage(elementUUID, code, type, message);
+            return new MarshallingMessage(elementUUID, code, type, message, messageKey, messageArguments);
         }
     }
 
@@ -98,30 +126,34 @@ public class MarshallingMessage implements DomainViolation {
         if (this == o) {
             return true;
         }
-        if (!(o instanceof MarshallingMessage)) {
+        if (o == null || getClass() != o.getClass()) {
             return false;
         }
         MarshallingMessage that = (MarshallingMessage) o;
-        return getCode() == that.getCode() &&
-                Objects.equals(getElementUUID(), that.getElementUUID()) &&
+        return code == that.code &&
+                Objects.equals(elementUUID, that.elementUUID) &&
                 type == that.type &&
-                Objects.equals(getMessage(), that.getMessage());
+                Objects.equals(message, that.message) &&
+                Objects.equals(messageKey, that.messageKey) &&
+                Objects.equals(messageArguments, that.messageArguments);
     }
 
     @Override
     public int hashCode() {
         return HashUtil.combineHashCodes(Objects.hashCode(getElementUUID()), Objects.hashCode(getCode()),
-                                         Objects.hashCode(type), Objects.hashCode(getMessage()));
+                                         Objects.hashCode(type), Objects.hashCode(getMessage()),
+                                         Objects.hashCode(getMessageKey()), Objects.hashCode(getMessageArguments()));
     }
 
     @Override
     public String toString() {
-        final StringBuilder sb = new StringBuilder("MarshallingMessage{");
-        sb.append("elementUUID='").append(elementUUID).append('\'');
-        sb.append(", code=").append(code);
-        sb.append(", type=").append(type);
-        sb.append(", message='").append(message).append('\'');
-        sb.append('}');
-        return sb.toString();
+        return "MarshallingMessage{" +
+                "elementUUID='" + elementUUID + '\'' +
+                ", code=" + code +
+                ", type=" + type +
+                ", message='" + message + '\'' +
+                ", messageKey='" + messageKey + '\'' +
+                ", messageArguments=" + messageArguments +
+                '}';
     }
 }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
@@ -18,14 +18,18 @@ package org.kie.workbench.common.stunner.core.marshaller;
 
 import java.util.Objects;
 
+import org.jboss.errai.common.client.api.annotations.MapsTo;
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.kie.workbench.common.stunner.core.util.HashUtil;
 import org.kie.workbench.common.stunner.core.validation.DomainViolation;
 
+@Portable
 public class MarshallingMessage implements DomainViolation {
 
-    private String elementUUID;
-    private int code;
-    private Type type;
-    private String message;
+    private final String elementUUID;
+    private final int code;
+    private final Type type;
+    private final String message;
 
     public String getElementUUID() {
         return elementUUID;
@@ -45,7 +49,8 @@ public class MarshallingMessage implements DomainViolation {
         return code;
     }
 
-    public MarshallingMessage(String elementUUID, int code, Type type, String message) {
+    public MarshallingMessage(@MapsTo("elementUUID") String elementUUID, @MapsTo("code") int code,
+                              @MapsTo("type") Type type, @MapsTo("message") String message) {
         this.elementUUID = elementUUID;
         this.code = code;
         this.type = type;
@@ -105,7 +110,8 @@ public class MarshallingMessage implements DomainViolation {
 
     @Override
     public int hashCode() {
-        return Objects.hash(getElementUUID(), getCode(), type, getMessage());
+        return HashUtil.combineHashCodes(Objects.hashCode(getElementUUID()), Objects.hashCode(getCode()),
+                                         Objects.hashCode(type), Objects.hashCode(getMessage()));
     }
 
     @Override
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.marshaller;
+
+import java.util.Objects;
+
+import org.kie.workbench.common.stunner.core.diagram.Metadata;
+
+public class MarshallingRequest<I, M extends Metadata> {
+
+    /**
+     * Mode
+     * <p>
+     * ERROR:
+     * - return errors when it has unsupported nodes
+     * - error messages
+     * AUTO:
+     * - try to adapt unsupported nodes do generic ones
+     * - warn messages
+     * IGNORE:
+     * - remove unsupported nodes and relationships to them
+     * - warn messages
+     */
+    public enum Mode {
+        ERROR,
+        AUTO,
+        IGNORE
+    }
+
+    private I input;
+    private M metadata;
+    private Mode mode;
+
+    public MarshallingRequest(I input, M metadata, Mode mode) {
+        this.input = input;
+        this.metadata = metadata;
+        this.mode = mode;
+    }
+
+    public I getInput() {
+        return input;
+    }
+
+    public M getMetadata() {
+        return metadata;
+    }
+
+    public Mode getMode() {
+        return mode;
+    }
+
+    public static <I, M extends Metadata> MarshallingRequestBuilder<I, M> builder() {
+        return new MarshallingRequestBuilder<>();
+    }
+
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("MarshallingRequest{");
+        sb.append("input=").append(input);
+        sb.append(", metadata=").append(metadata);
+        sb.append(", mode=").append(mode);
+        sb.append('}');
+        return sb.toString();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof MarshallingRequest)) {
+            return false;
+        }
+        MarshallingRequest<?, ?> that = (MarshallingRequest<?, ?>) o;
+        return Objects.equals(getInput(), that.getInput()) &&
+                Objects.equals(getMetadata(), that.getMetadata()) &&
+                getMode() == that.getMode();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getInput(), getMetadata(), getMode());
+    }
+
+    public static class MarshallingRequestBuilder<I, M extends Metadata> {
+
+        private I input;
+        private M metadata;
+        private Mode mode;
+
+        public MarshallingRequestBuilder input(I input) {
+            this.input = input;
+            return this;
+        }
+
+        public MarshallingRequestBuilder metadata(M metadata) {
+            this.metadata = metadata;
+            return this;
+        }
+
+        public MarshallingRequestBuilder mode(Mode mode) {
+            this.mode = mode;
+            return this;
+        }
+
+        public MarshallingRequest build() {
+            return new MarshallingRequest(input, metadata, mode);
+        }
+    }
+}
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
@@ -24,7 +24,7 @@ import org.kie.workbench.common.stunner.core.diagram.Metadata;
 import org.kie.workbench.common.stunner.core.util.HashUtil;
 
 /**
- * Encapsulates the attriutes required on a marshalling process.
+ * Encapsulates the attributes required on a marshalling process.
  * Includes the {@link Mode} that represents how the marshaller should handle errors and exceptions.
  *
  * @param <I> represents the input type
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
@@ -41,7 +41,7 @@ public class MarshallingRequest<I, M extends Metadata> {
      * - return ERROR messages
      * AUTO:
      * - try to adapt and convert unsupported nodes to generic ones
-     * - retun warn messages
+     * - return warn messages
      * - ignore unsupported nodes that doesn't have options to be converted
      * IGNORE:
      * - remove all unsupported nodes and relationships to them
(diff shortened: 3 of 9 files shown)


tiagodolphine:

tiagodolphine:
Jenkins retest this

tiagodolphine:
I found an issue on the gwt compilation. Fixing now.

manstis:
@tiagodolphine Hello, could you please advise how this fits in with the existing mechanism to return a different payload to the client should marshalling of the XML -> BPMN (or DMN) fail. See [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-backend-common/src/main/java/org/kie/workbench/common/stunner/core/backend/service/AbstractVFSDiagramService.java#L181). In these scenarios the client code (see [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-project/kie-wb-common-stunner-project-client/src/main/java/org/kie/workbench/common/stunner/project/client/editor/AbstractProjectDiagramEditor.java#L708)) falls back to showing a text editor for the XML instead of the full _Stunner_ editor. I am just curious about your changes and how they fit in with the existing support for parsing failures.

tiagodolphine:
Hi @manstis in fact, it is not modifying the existing mechanism yet, the result of the unmarshall is a `Graph`, that is now just encapsulated on the `MarshallingResponse`. On the [JBPM-8432](https://issues.jboss.org/browse/JBPM-8432) there as a previous design, but to summarize we need to get warning or error messages when unmarshalling an input file, for instance, a ._bpmn_ generated by an external editor that has unsupported elements on Stunner, this is a need to @wmedvede on his current task(migrate diagram from jbpm designer to stunner) so we planned the marshaller API to give some feedback other than _result_ or _exception_. Besides this, we have another task that is to have a 
kind of translation or bypass, I mean try to convert an unsupported node to a more generic one (i.e ScriptTask if not supported to a NoneTask) see [BAPL-1153](https://issues.jboss.org/browse/BAPL-1153).
Related to your comment about the fallback flow on the client, that is based on `DiagramParsingException`, this remains the same see https://github.com/kiegroup/kie-wb-common/blob/6f2dd862696be47356ff9caddb99dbd0c627195a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-backend-common/src/main/java/org/kie/workbench/common/stunner/core/backend/service/AbstractVFSDiagramService.java#L180 because a `RunTimeException` will be thrown.
If you have more ideas, please comment on the [JBPM-8432](https://issues.jboss.org/browse/JBPM-8432) as well, this is important since we are planning an API to solve all our needs including DMN. I will be glad to modify this PR.
@wmedvede @romartin this includes you as well :)
Thanks !!

tiagodolphine:

romartin:
@manstis thanks for getting involved into this!! Will keep you updated on this kinda new stuff on next times, 
 even because it'll affect DMN... sorry!! 

@manstis lemme try to give you some **backgound** - we've different goals behind these tasks, some are about interals of the BPMN marshallers, which up to you if you want to get involved, but I think the main task to focus, which affects other editors, is the change on Marshalling API in Stunner. 

The **goal** we're trying to achieve is to improve the feedback (also on the UI) when the user imports, loads or saves a diagram (saying diagram because this change will affect not just processes, also cases and dmn diagrams, am I correct @tiagodolphine ? ). Acutally a single message appears on the console and contains all validation, parsing, etc messages, not really readable. The goal is to provide much more detailed, priorized and granular messages and feedback on the UI for end users, so they are able to understand the result of the model transform and decide if the diagram, once in Stunner, should change in some way as well.

To achieve our goal , the **idea** is to provide some messages as a result  of a un/marshalling operation, decribing "what" really Stunner is doing when "converting", for example, a BPMN file into an Stunner process model. So the plan is to change the Marshalling API, by exposing the internal messages into  some data model that can be ported to the client and consumed by UI components.

romartin:
@manstis @tiagodolphine Once given the above background on this.... let's brainstorm on the actual impl isn't it? :+1: 

Actually marshallers are just throwing exceptions when any "error", no matter if under control or unexpected, is happening. This way, as @manstis properly pointed, the editors are handling errors by caching some exception type and displaying the XML editor, instead of Stunner.

I would vote for still considering both `messages` and `Exception` on the picture. 
- `Exceptions` are the core language feature that we should consider as "unexpected" errors, and proabbly still be managed by just opening the XML editor. 
- On the other hand, we're now adding information about marshalling operations, but that information is about our concrete algorithms, rather than failures due to state, envrionment or external related causes. This way we're introducing `messages` as a result of a un/marhalling process, in order to provide feedback about any domain specific action is being taken and could affect the runtime of the process.

WDYT guys????
Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshaller.java
```diff
@@ -40,6 +44,32 @@
     G unmarshall(final M metadata,
```
romartin: I understand that the new `unmarshallWithValidation` method being created here is for keeping backward compatibility, but finally at some point the API will rely on the `MarshallingResponse`. is that correct? 
In that case, we can probably deprecate the unmarshall method... does that makes sensee?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
```diff
@@ -0,0 +1,121 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.marshaller;
+
+import java.util.Objects;
+
+import org.kie.workbench.common.stunner.core.validation.DomainViolation;
+
+public class MarshallingMessage implements DomainViolation {
```
romartin: This should be `@Portable` isn't it?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
```diff
@@ -0,0 +1,124 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.core.marshaller;
+
+import java.util.Objects;
+
+import org.kie.workbench.common.stunner.core.diagram.Metadata;
+
+public class MarshallingRequest<I, M extends Metadata> {
```
romartin: This should be `@Portable` isn't it?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+package org.kie.workbench.common.stunner.core.marshaller;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+public class MarshallingResponse<T> {
```
romartin: This should be `@Portable` isn't it?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
romartin: also probalby some of these fields can be unmutable ( `final` )...

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
romartin: also probalby some of these fields can be `final`...

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
romartin: also probalby some of these fields can be `final`...

tiagodolphine:
Hi @romartin @manstis, the idea is to deprecate the `unmarshall` method implementation and replace with the `unmarsallWithValidation`, for now I just keep both to compatibility and to not insert a big refactor on the code, just to point, this PR is supposed to be a quick one, to allow @wmedvede to keep working. The main PR will be made with all the refactor necessary on marshallers and on the classes using them.
So to conclude the client code that has the control flow based on the `Exceptions` will be the same nothing has changed, but I'm planning to refactor this on the other PR, I personally prefer to handle the erros based on a given response not on Exceptions, instance of and things like that, but this is for the other PR.
So again, this PR is just the definition of the API to allow @wmedvede to keep working based on it, and a simple implementation that just call the current `unmarshall`, the main point here is to have a good API and all ideas are welcome, as I said before I'll be glad to add modifications to have a greater API, and then keep the real implementation of the API on a future PR.
Thanks guys !

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
tiagodolphine: great point, I'll do it.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
tiagodolphine: great point, I'll do it.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
tiagodolphine: great point, I'll do it.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshaller.java
tiagodolphine: Yes this is right, as I said on the other comment I'm keeping both to not insert a big refactor now, since the PR is supposed to be a quick one.  On a new PR let's keep just one `unmarshall` method that returns a  `MarshallingResponse`.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
tiagodolphine: OK :+1:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
tiagodolphine: OK :+1:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
tiagodolphine: OK :+1:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
```diff
@@ -0,0 +1,121 @@
...
+        MarshallingMessage that = (MarshallingMessage) o;
+        return getCode() == that.getCode() &&
+                Objects.equals(getElementUUID(), that.getElementUUID()) &&
+                type == that.type &&
+                Objects.equals(getMessage(), that.getMessage());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getElementUUID(), getCode(), type, getMessage());
```
wmedvede: Probably HashUtil can be use here

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
```diff
@@ -0,0 +1,121 @@
...
+        return type;
+    }
+
+    public int getCode() {
+        return code;
+    }
+
+    public MarshallingMessage(String elementUUID, int code, Type type, String message) {
+        this.elementUUID = elementUUID;
+        this.code = code;
```
wmedvede: Is this code intended for implementing client side i18n? 
If yes, we might probably need to carry the message parameters for building the i18n message.
If no, I believe we need a way for implementing i18n since this is the first thing we'll be asked for as soon we start showing this messages in the UI

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Mode
+     * <p>
+     * ERROR:
+     * - return errors when it has unsupported nodes
+     * - error messages
+     * AUTO:
+     * - try to adapt unsupported nodes do generic ones
+     * - warn messages
+     * IGNORE:
```
wmedvede: Some queries regarding the semantic comes to my mind now that I'm reviewing this:

This is more or less the semantic (please correct me if I'm wrong)

IGNORE:
1) whatever node/element not yet supported/implemented by Stunner is removed and potential relationships to them too. In the response we get a list WARNING messages with the undertaken actions. 
Right? (I still see this a bit complex, specially for intermediate nodes)

2) If the procedure can't be completed, e.g., the relationship with the unsupported element are too complex, etc, the response status will be ERROR and we might have 1..N ERROR DomainViloations. 
Right?

AUTO:
3) A best effort is done to adapt unsupported nodes to something generic, etc. If the response status is SUCCESS we might have a list of WARNING messages with the undertake actions. Right?
 
4) Same here, if the procedure can't be completed, response status == ERROR and we get a list of 1..N ERROR DomainViloations. 
Right?

I think this is fine, but I believe we might have sort of mixed scenarios:

e.g.:
	
If you look at https://issues.jboss.org/browse/BAPL-1153, the parser is complaining about org.eclipse.bpmn2.impl.DataStoreReferenceImpl@9ff080f, something that is not implemented in Stunner and we'll probably never be implemented. This is an example of element that we'll want to remove, and even when it might have relations with other nodes in the flow, it's a "simple" case of node that we can remove.
Now during the parsing of the same process we might find a ServiveTask element, and we know that ServiceTask elements coming form ARIS will probably never have all the proper information that Stunner expects, e.g. the workitem information. BUT, to remove a ServiceTask from the process is more complex, but also the process can loose semantic. So probably this node is a good example of a node that we don't want to remove, but just import. The resulting imported node can probably be a NoneTask, or a ServiceTask with some missing information.

The example above is something I believe can perfectly be a real use case from processes coming from ARIS. However, users might want to treat this case as a sort of mixed approach.
Because AUTO can't be applied for this use case since we don't have an equivalent node to the DataStoreReferenceImpl.
But IGNORE is too much, since just because of the DataStoreReferenceImpl (that has no semantic), we'll also remove all the potential ServiceTask nodes form the process that are important for the process semantic.

manstis:
@tiagodolphine @romartin @wmedvede It's not for me to meddle with your requirements :-)

IIRC the existing `Exception` handling was added for scenarios when the User tries to open a completely duff, screwy `BPMN` (or `DMN` or `Case Management` file) and, for example, `XStream` fails to even parse the content as XML! This falls back to the _text editor_.

This, as Roger [says](https://github.com/kiegroup/kie-wb-common/pull/2612#issuecomment-488888589),  is quite different to being able to process the XML but finding elements that are unsupported by the editor etc or contain invalid state and "sensible defaults" added automatically. I see this PR as invaluable for this scenario.

I'm more than happy to leave review to you guys (and hence have removed myself from review).

romartin:
yep I'm also agree with @wmedvede , the casuistics are a bit more complex so the API was not enough clear yet to represent the different response outputs and status. So I'm fine with @wmedvede 's proposal here, anyway up to you @tiagodolphine to give your feedback!
Thanks guys!

wmedvede:
> yep I'm also agree with @wmedvede , the casuistics are a bit more complex so the API was not enough clear yet to represent the different response outputs and status. So I'm fine with @wmedvede 's proposal here, anyway up to you @tiagodolphine to give your feedback!
> Thanks guys!

And same mixed use cases we'll happen for the jBPM to Stunner migration.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
tiagodolphine: The message is coming from the **jbmn** validation engine, AFAIK it does not support i18n (see [RuleFlowProcessValidator](https://github.com/kiegroup/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/validation/RuleFlowProcessValidator.java), the other validations made on Stunner, we can work with i18n. Anyway one option is to abstract this and assume the messaging is already internationalized when it is set on the `MarshallingMessage`.
wdyt?
cc @romartin

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
tiagodolphine: ok

tiagodolphine:
Hey @wmedvede so to summarize what is the suggestion? to include a new marshaling request type other than IGNORE, AUTO? 
My first idea was that AUTO basically tries to convert an unknown element to a more generic one and plus ignore the unknown elements (and relationships like the IGNORE type) that have no generic to be converted. 
And the IGNORE will always ignore not trying to convert to more generic elements.

wmedvede:
@tiagodolphine ok, well, probably I misunderstood the comments. But yes, so summarize,
if AUTO do:
1) try to migrate a node to a more generic one
and 
2) unsupported nodes like TextAnnotation, DataObject, DataStore are ignored
then AUTO is good!
Thanks

tiagodolphine:
@wmedvede well we can add more types, for now, this is just the API definition, the implementation will be done on a different PR, if we found that it is necessary a new type for different scenarios we can add, no problem, it is up to us, the only thing that is important is to have well defined what will be done on each type and besides try to keep it as simple as possible.

tiagodolphine:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
wmedvede: Hi @tiagobento , I know "old" messages produce by jbpm components, like the jbpm parser, etc, has no i18n.
But I was referring the new messages what will be produced, this new messages I believe should have i18n , and the ones produced by jbpm components can be left as is.

wmedvede:
Hi @tiagodolphine @romartin In mean time we have the green, etc, I took already the PR contents and using the API types locally. With last discussion about the AUTO mode we had, and solving the i18 I think the API semantic is ok.

I saw also that @tiagodolphine already added the api in the method:
     AbstractVFSDiagramService.public D getDiagramByPath(final org.uberfire.backend.vfs.Path file)
This is fine.

Related to this I believe that for facilitating the migration from jbpm-designer -> Stunner implementation, but also will be needed at some point by @tiagodolphine 's development we'll need a method at the AbstractVFSDiagramService/BaseDiagramService a method that exposures the marshalling request mode and returns the marhalling result messages will be needed. But, yes, I assume this will come up latter as part of the implementation of the API plus the derived changes that well have to be applied in the DiagramEditor, etc, for processing the API outcomes, etc. (we can talk latter about this)

By now in my local for moving my development forward I've added something like this in mean time.
    MarshallingResponse<D> getDiagramByPath(final Path file, MarshallingRequest.Mode mode);

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
tiagodolphine: Hi @wmedvede if you take a look at `CommandNotification.build(...)` it is using the `CoreTranslationMessages`, so the message is already set internationalized. The point is when the message (String) is set on the `MarshallingMessage` is supposed to be already internationalized, no matter if it is coming from **Stunner** side or from the **jbpm**, makes sense?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
```diff
@@ -0,0 +1,120 @@
...
+    private Metadata metadata;
+
+    private MarshallingRequest request;
+
+    @Before
+    public void setUp() {
+
+        tested = spy(new DiagramMarshaller() {
+            @Override
+            public Graph unmarshall(Metadata metadata, InputStream input) throws IOException {
```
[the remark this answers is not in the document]
tiagodolphine: Ok

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Optional;
+
+import org.jboss.errai.common.client.api.annotations.MapsTo;
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.kie.workbench.common.stunner.core.util.HashUtil;
+
+@Portable
+public class MarshallingResponse<T> {
+
+    public enum State {
```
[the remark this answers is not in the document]
tiagodolphine: Ok

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
```diff
@@ -0,0 +1,120 @@
...
+                return graph;
+            }
+
+            @Override
+            public String marshall(Diagram diagram) throws IOException {
```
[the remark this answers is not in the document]
tiagodolphine: Ok

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
```diff
@@ -0,0 +1,120 @@
...
+        verify(tested).unmarshall(metadata, input);
+        assertEquals(response.getResult().orElse(null), graph);
+        assertEquals(response.getState(), MarshallingResponse.State.SUCCESS);
+    }
+
+    @Test
+    public void unmarshallWithValidationError() throws Exception {
+        tested = spy(new DiagramMarshaller() {
+            @Override
+            public Graph unmarshall(Metadata metadata, InputStream input) throws IOException {
```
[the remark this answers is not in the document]
tiagodolphine: Ok

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
```diff
@@ -0,0 +1,120 @@
...
+                throw new RuntimeException(MESSAGE);
+            }
+
+            @Override
+            public String marshall(Diagram diagram) throws IOException {
```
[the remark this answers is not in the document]
tiagodolphine: Ok

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Mock
+    private Metadata metadata;
+
+    private MarshallingRequest request;
+
+    @Before
+    public void setUp() {
+
+        tested = spy(new DiagramMarshaller() {
```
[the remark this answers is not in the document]
tiagodolphine: Ok

tiagodolphine:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
romartin: Well good point guys, so the message have to be i18n ,but on the `MarshallingMessage` not sure if we should set the already internationalized literal... this is because actually those `MarshallingMessage` instances are being created on backend, and sent to the client as they're `@Portable`, but each client can be using different language, so I would say that the i18n resolution have to be done at runtime, on client side.

tiagodolphine: Hi @romartin @wmedvede so let's see what to do here, what are the suggestions? If the i18n is made on the client the MarshallingMessage should a key + arguments... anyway this will couple a lot backend and client but this is an option. On the other hand, the i18n could be made on server if the language is passed as a parameter, I don't know how `Errai` handle this on server side, do you know?  On the default `REST` approach, this is very easy to handle since the browser sends the `Accept-Language` header that can be used on the i18n by the server.

romartin: well not really sure... so looking at the HTTP request's header for taking some decision on the server is fine yese, the problem I see is that by returning the complete literals will also increase the response size, and exponentially, it depends on the number of resulting messages and the literals for each one...
On the other hand we could even consider that at some point the validation messages can be provided by some external service, in which case we don't have control for the literals on client side, we just need the right literal from server, but that's not short a term goal anyway.
Soo I would probably create an interface and differnt kind of message implementations, one type could using the complete string literal, and for the other we can use some concrete message implementation that does not contains the whole strings (for performance reasons), and client side is able to resolve this message literals at runtime. Up to you if using the `key + arguments`... looks quite reasonable for me, this way you can probably use some shared Errai interface that also is used on client side for the i18n bundle, in order not to hardcode the key strings. Another option could be using some custom Enum or code for each message literal, as we typicially did on old `C++` code, isn't it?
WDYT?

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingRequest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Mode of the marshalling process
+     * <p>
+     * ERROR:
+     * - return errors when it has unsupported nodes
+     * - return ERROR messages
+     * AUTO:
+     * - try to adapt and convert unsupported nodes to generic ones
+     * - retun warn messages
```
LuboTerifaj:
```suggestion
     * - return warn messages
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.jboss.errai.common.client.api.annotations.MapsTo;
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.kie.workbench.common.stunner.core.util.HashUtil;
+
+/**
+ * Encapcsulates the response of a marshalling process, containing messages details and the respective {@link State}.
```
LuboTerifaj:
```suggestion
 * Encapsulates the response of a marshalling process, containing messages details and the respective {@link State}.
```

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingResponse.java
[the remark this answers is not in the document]
LuboTerifaj: Javadoc for enum State has not been added.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
```diff
@@ -0,0 +1,119 @@
...
+                return null;
+            }
+
+            @Override
+            public DiagramMetadataMarshaller getMetadataMarshaller() {
+                return null;
+            }
+        });
+
+        request = MarshallingRequest.builder()
```
LuboTerifaj: This could stay in setUp method, since it is used in both tests.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
wmedvede: Hello @tiagodolphine @romartin 
I've been doing a quick research and I looks like it's not possible to know the GWT application navigation language at the time the Errai RPC request is being processed at server side. It's a bit strange that Errai doesn't have this ability I have to say.

The most proximate that I've could find is the errai RpcContext class, by using it we can have access to the ServletRequest and we could do something like this:

    RpcContext.getServletRequest().getLocale()

(In fact there are some code around the WB that is trying to use this RpcContext.getServletRequest().getLocale() to do server side i18n)

BUT unfortunately this locale is not updated when the business-central/GWT application navigation language is changed, instead it always returns the locale of the browser local machine. So it's not 100% accurate.

On the other hand I have to say that in general if you look around the WB, etc, there are no consensus/standard for managing this sort of server side generated messages/errors/etc. In general exceptions are thrown with no i18n, or server side messages are produced by just using hard coded messages like "here is my message" etc.

To summarize:

I agree that a so "silly" problem of messages internationalization can have many approaches, and discussion around if it should be done at client side, or server side, etc, can be long, and probably depends also on the use case, etc.

After re-thinking I'd vote for this approach:

1) we have a service that basically returns a list of messages. 

2) the service has the ability of returning this messages in different languages

3) we pass the preferred language to the server and the server resolves the i18n the best as possible with the standard ResourceBundle approach.


So finally the thing is how to pass the preferred language automatically.
Given that it seems like there's no way of getting this at server side from Errai (probalby Tiago Bentos knows a trick) I think that by now we can just pass the language as part of the MarshallingRequest by adding a field "locale".
And we can have a ClientMarshallingRequestBuilder that internally gets the current GWT app navigation language and set this value in the MarshallingRequest being built.
We can use a similar approach used by the TranslationService implementation for getting the locale, see here:
 https://github.com/errai/errai/blob/fa095d8a51d88e2625a255270e6234d30f2a1eb8/errai-ui/src/main/java/org/jboss/errai/ui/client/local/spi/TranslationService.java#L184

4) Finally I don't see as a problem to return the message literals instead of the messageKey + parameters, I mean, I don't believe we'll have performance issues because of this.
See that even nowadays when we parse a Diagram and returns the Portable objects structure we are already sending a lot of information, etc.

romartin: Hey @wmedvede @tiagodolphine 
Well the only thinkg it's important to consider is that for submarine wer're moving everything to client side, including parsers and validations, so as soon we have all this on client, more work will be aready done when moving to submarine. Anyway let's better discuss this on the call that hopefully @wmedvede has scheduled guys.
Thx

tiagodolphine:
@wmedvede @romartin @LuboTerifaj I'll update the PR ASAP with this modification for i18n + Lubo's suggestions.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/test/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshallerTest.java
tiagodolphine: done

tiagodolphine:

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/marshaller/MarshallingMessage.java
```diff
@@ -0,0 +1,159 @@
...
+import org.kie.workbench.common.stunner.core.util.HashUtil;
+import org.kie.workbench.common.stunner.core.validation.DomainViolation;
+
+@Portable
+public class MarshallingMessage implements DomainViolation {
+
+    private final String elementUUID;
+    private final int code;
+    private final Type type;
+    private final String message;
```
wmedvede: is this message field still valid?

wmedvede: Is this message field still valid?

tiagodolphine: @wmedvede I think this field can be used to send a default message in English, this may be helpful. Anyway, I think this is not a big deal we can remove if not necessary during the real implementation, on the roundtrip issue.

wmedvede: ok, yes, I supposed it was for the by default message. For me it's fine to keep it, no problem.

tiagodolphine:
Hi @romartin you approve the last changes it is ok to be merged.
Thanks.

PR #2615, opened 2019-05-03
DROOLS-3895: [DMN Designer] TextAreas can no longer be multi-line

manstis:
See https://issues.jboss.org/browse/DROOLS-3895

Part of an ensemble:

- https://github.com/kiegroup/appformer/pull/684
- https://github.com/kiegroup/kie-wb-common/pull/2615


jomarko:

karreiro:
Jenkins execute full downstream build.

manstis:
Yay @karreiro the fix works then :-)

Without these PRs you CANNOT have multi-linr text in a `TextArea` as pressing `<Enter>` flushes the content and exits _edit_ mode.

manstis:
Hi @jomarko Good spot regarding the Decision Table's `Description` column.

The reason was that it did not use a "line-break" word wrapper. I have however changed the `Description` column to use the same formatting as the both the `InputClause` and `OutputClause` columns. Which is also now consistent with every other place we use a `TextAreaSingletonDOMElementFactory` (i.e. "in-cell `TextArea` editor).

jomarko:
@manstis thank you for changes, I am not going to re-review.

PR #2624, opened 2019-05-06
DROOLS-3974 : Guided Decision Table is not getting updated

Rikkola:
https://issues.jboss.org/browse/DROOLS-3974

Without this the dropdown just shows the code that should fetch the enumeration, while the documentation states that the use of ' quotes should let you use the code in the way this PR enables.

The fix changes 
'Fact.field' : '(new org.company.DataHelper()).getList()'
to
'Fact.field[field]' : '(new org.company.DataHelper()).getList()'

This is the most simple way of getting this to work. I don't really want to touch the areas on the client side code since the unit test coverage is weak.

@jomarko @manstis I think you two are again the best to review. Thank you. Or of course happy to ping anyone else who might be able to review this.

### DIFF
--- a/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/DataEnumLoader.java
+++ b/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/DataEnumLoader.java
@@ -112,6 +112,12 @@ public class DataEnumLoader {
                 }
                 return Collections.emptyMap();
             } else if (list instanceof String) {
+
+                if (mvelSource.contains("'" + entry.getValue() + "'") && !key.contains("[")) {
+                    final String field = key.substring(key.indexOf("#") + 1);
+                    key = key + "[" + field + "]";
+                }
+
                 newMap.put(key, new String[]{(String) list});
             } else {
                 List<?> items = (List<?>) list;
--- a/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/test/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/DataEnumLoaderTest.java
+++ b/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/test/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/DataEnumLoaderTest.java
@@ -15,12 +15,15 @@
  */
 package org.kie.workbench.common.services.datamodel.backend.server.builder.util;
 
+import java.util.Map;
+
 import org.junit.Test;
 import org.kie.soup.project.datamodel.commons.util.RawMVELEvaluator;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.jgroups.util.Util.assertEquals;
 import static org.jgroups.util.Util.assertFalse;
 import static org.jgroups.util.Util.assertTrue;
-import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * DataEnumLoader tests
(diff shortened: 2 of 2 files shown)


manstis:
It looks like the _workaround_ is yet to be approved by the Customer. Is this to be merged regardless? 

Also, you may want to update the documentation to include this _advanced_ case?

manstis:
@Rikkola I'd like to understand better _how_ this works!

I can confirm it _does_ work; but looking through the source IDK _how_!

Rikkola:
@manstis the documentation actually describes this case, but this did not work. I'll comment in code what happens.

Rikkola:
Actually my mistake this exact situation is not in the docs, but we do state:

> It is also possible to load dependent enumeration definitions dynamically from a helper class by enclosing the call to the helper class within quotation marks

The example then shows: 
> 'Country.region[countryCode]' : '(new org.yourco.DataHelper()).getListOfRegions("@{countryCode}")'

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/DataEnumLoader.java
```diff
@@ -112,6 +112,12 @@ public DataEnumLoader(final String mvelSource,
                 }
                 return Collections.emptyMap();
             } else if (list instanceof String) {
+
+                if (mvelSource.contains("'" + entry.getValue() + "'") && !key.contains("[")) {
+                    final String field = key.substring(key.indexOf("#") + 1);
+                    key = key + "[" + field + "]";
+                }
+
```
Rikkola: Now this fix abuses the mechanism by depending on the same field the enum is for. The field has to exist, but we do not use it. 
In the client side there is a check for if there are dependent enums. If there are get the enums from the backend. This is why it updates on each call.
The other option is to also change the client side, possibly in several locations since the behaviour is used on the guided editor. Since the test coverage is weak there this approach felt safer.

jomarko:
@Rikkola some cleanup at first https://github.com/Rikkola/kie-wb-common/pull/4

Rikkola:
Thanks @jomarko I have had this issue with few PRs and when I tried to import the real assert method Idea did not find the method. So I suspect I need to click something somewhere in my Idea settings, or just update it.

jomarko:
Can not approve for now, see https://drive.google.com/file/d/1engqgCZfwayBA21TweWCHIm2Lu_i0sgu/view?usp=sharing

Rikkola:
Taking a look.

Rikkola:
Need to double check. I did my checks with Guided Rule Editors, perhaps DTables has some issues.

Rikkola:
@jomarko Ok. So tried with the provided test project and it was missing ' quotes, this marks the enum for dynamic load.

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/test/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/DataEnumLoaderTest.java
```diff
@@ -149,6 +151,16 @@ public void testInvalidDependentEnum_AdvancedEnum2() {
         assertFalse(loader.hasErrors());
     }
 
+    @Test
+    public void testInvalidDependentEnum_AdvancedEnumForDynamic() {
```
jomarko: I hope this is valid, so propose: `testValidDependentEnum_AdvancedEnumForDynamic`

Rikkola: Right, it actually is.

Rikkola: Fixed.

Rikkola:
It should use the same reload mechanism with both, is the `getList(@{control})` just the same as getList()? So that it returns a random list?

jomarko:
@Rikkola I use slightly updated project, see https://issues.jboss.org/browse/DROOLS-3974

jomarko:
I think my question is answered here https://issues.jboss.org/browse/DROOLS-2867, approving

Rikkola:
@jomarko sounds like that is it.

@kiegroup/gatekeepers Please merge this.

PR #2638, opened 2019-05-14
DROOLS-3994: [DMN Designer] Connectors to do align with Node boundaries

manstis:
See https://issues.jboss.org/browse/DROOLS-3994

**Tests**

Create a diagram
Add a `Decision` from the palette
Add another `Decision` from the first using the toolbox button
Save, close and re-open.

Create a diagram
Add a `Decision` from the palette
Add another `Decision` from the palette
Connect using the toolbox icon on the first `Decision`
Save, close and re-open.

Create a diagram
Add a `Decision` from the palette
Add another `Decision` from the palette
Connect using the toolbox icon on the first `Decision`
Re-position connector endpoints into the centre of each `Decision`
Save, close and re-open.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
@@ -127,6 +127,8 @@ import static org.kie.workbench.common.dmn.backend.definition.v1_1.dd.PointUtils
 @ApplicationScoped
 public class DMNMarshaller implements DiagramMarshaller<Graph, Metadata, Diagram<Graph, Metadata>> {
 
+    private static final double CENTRE_TOLERANCE = 1.0;
+
     private XMLEncoderDiagramMetadataMarshaller diagramMetadataMarshaller;
     private FactoryManager factoryManager;
     private InputDataConverter inputDataConverter;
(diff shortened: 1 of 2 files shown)


jomarko:
@manstis Hi, I can confirm, the connections are now tightly connected to the nodes. however The touched code is about magnets, I think there is some issue (not saying wasn't previously) Could you please double check? Pasting pairs of diagrams before and after reopening.

###  Pair 1
- Before
  - 
- After
  - 

### Pair 2
- Before
  - 
- After
  - 


### Pair 3
- Before
  - 
- After
  -

manstis:
@jomarko Hello, the reported problem pre-exists.

I could only replicate it by re-positioning the connector end-points on the magnets on the edge of the node (e.g. left edge, right edge). If this is different to how you got the differences please say.

The issue is that connectors have a mystical "auto" property that means Stunner re-positions them at runtime to the closest edge. The actual position of the end-point (as stored in the model/file does not change). You can see this when you add two nodes using the toolbox and move them around.. the connector will be seen to "jump" from one edge to another. This is different to when you attach a connector end-point to the middle magnet. In this case the connector can be seen to always lead to the centre of the node no matter how/where you move the nodes.

We do not store the "auto" property of a connector in the DMN file only it's position. We used to always set connectors to "auto" when re-opening a file but this caused the issue reported in the JIRA for which this PR was made (for when a connector is attached to the "centre"). We now check if the position of the end-point was (close to) the centre of the node in which case we do not set it's "type" to "auto"). However connectors that were manually set to an edge are also re-opened as "auto" and you get the visual differences you point out.

This [JIRA](https://issues.jboss.org/browse/DROOLS-3702) holds a discussion about how Stunner works, what DMN does and the conclusion that we should do (short term) what is [quickest](https://issues.jboss.org/browse/DROOLS-3702?focusedCommentId=13718294&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13718294).

I think we're in a "good place" for the purpose of this PR however the need to store Stunners "auto" flag in the DMN file is the real requirement and should not be lost.. I've reported the need [here](https://issues.jboss.org/browse/DROOLS-4033).

danielzhe:
Very good. I didn't found any issue in the manual tests.

manstis:
@jomarko This is ready for re-consideration. See my [reply](https://github.com/kiegroup/kie-wb-common/pull/2638#issuecomment-492657959).

PR #2639, opened 2019-05-14
JBPM-8303 Stunner - Alerts panel opens everytime designer rules are violated

tiagodolphine:
Now the alerts panel remains closed/hidden when validation is triggered.

@romartin 
@LuboTerifaj

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-project/kie-wb-common-stunner-project-client/src/main/java/org/kie/workbench/common/stunner/project/client/screens/ProjectMessagesListener.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-project/kie-wb-common-stunner-project-client/src/main/java/org/kie/workbench/common/stunner/project/client/screens/ProjectMessagesListener.java
@@ -79,6 +79,7 @@ public class ProjectMessagesListener {
         systemMessage.setPath(path);
         messagesList.add(systemMessage);
         PublishMessagesEvent messages = new PublishMessagesEvent();
+        messages.setShowSystemConsole(false);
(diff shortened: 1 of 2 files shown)


tiagodolphine:

tiagodolphine:
Hi @LuboTerifaj the save, rename, copy actions are handled by the workbench, not by stunner, so the alerts panel are being shown because the workbench is doing this. To test it you can try to open other editors: Data Object, Global Variables, Work item definition, etc and test the same scenario. With that said, we have some possibilities:
- Try to make Stunner force hiding the alerts panel on those actions, don't know if this is possible, we need to investigate.
- Modify the workbench, but this will impact the whole workbench, all the editors, this change is probably up to the foundation team, then it is necessary to speak to the guys there and open a JIRA.
@romartin @LuboTerifaj  WDYT?

romartin:
Hey @LuboTerifaj , thanks for pointing this out, good point! Anyway agree with @tiagodolphine , that's a generic feature, so we should report a new issue for the fundation team about it. 

I've tested this fixed WAR and at least the error that @hasys reported on the ticket is not present, so I'm fine with this PR :+1:

hasys:
Hi @tiagobento, @romartin, @LuboTerifaj,

I am also voting for the second possibility: create new issue for the foundation team. Since it is more generic task and will affect other editors. I think we should avoid workarounds as described in the first point: force Stunner to workaround Appformer issues.

tiagodolphine:
So @romartin  @LuboTerifaj I think we can merge this PR, just need the approval, the next step is to open a new JIRA to foundation regarding this issue, can you do it please @LuboTerifaj, since you found the issue, it is better to explain.
Thanks.

romartin:
yep please @LuboTerifaj , let's report the new issue and try to move this PR forward. I'll merge once having  the approval. 
Thanks in advance! :+1:

PR #2649, opened 2019-05-20
DROOLS-4049: PMML files are not listed in Library only in Project Explorer

manstis:
See https://issues.jboss.org/browse/DROOLS-4049

This is a temporary fix to ensure PMML files are listed in Library and the _default_ editor used.


kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/resource/PMMLResourceTypeDefinition.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public String getShortName() {
+        return "PMML";
+    }
+
+    @Override
+    public String getDescription() {
+        return "PMML";
```
danielezonca: What about "Predictive Model Markup Language" instead of the acronym?

manstis:
Superseded...

PR #2652, opened 2019-05-21
[Stunner] Fix a core dependency issue (which was causing the project showcase not running properly on IntelliJ)

romartin:
Hey @tiagodolphine @Josephblt @wmedvede @treblereel @hasys @LuboTerifaj 

Fixed the issue you're also facing with, about running the project showcase inside IntelliJ :+1

**Some feedback**
Will give you some feedback about how I did it, if you're instered on, so next times you can also consider this "approach" for debugging this.

I understand we're all getting stuck, inside IntelliJ, when running the project showcase at this point:

`[INFO] 2019-05-21 03:19:31,809 [pool-4-thread-6] INFO  Reachability strategy set to Annotated`

So I just run in debug mode, and once getting into that point paused the debugging, then on the same IntelliJ debug console just navigated to the frame "`pool-4-thread-6`" (it can change on each execution) and noticed it was in some kinda infitie loop when trying to resolve some injection point. Finally by looking on scope variables at this point, noticed some reference to a "`clientCommandRegistry`" type.

At that point, two fixes came into my mind:
1.- Quickly fix the project showcase by excluding from the server side Ioc (beans.xml) that concrete conflicting classes
2.- Fix the rigth dependencies on the code and injection points

Finally as solution 1) was faster, but dirty and not much reliable, just applid the solution 2) for the fix. See next section.

**Cause & fix**

The root cause was due to an invalid dependency from the (shared) core commons package to the client API, in fact there was an old TODO [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/pom.xml#L44), so that can easily lead to ambiguous or unsatisfied injection points, to non desired situations, and in this case, it ends up on some classpath issue as well on IntelliJ, ratherh than when running the showcase from command line.

So the fix has been just doing things well done - removing the invalid dependency and refactoring a few classes for it. It looks a safe change for me. Tested and everything seems to be working as before, on all shwocases.

PS: Notice also updated the README.md for the showcase project, changed a bit the gwt run configuration parameters, please check you've same ones in your IntelliJ configs. I also invalidated cache and restarted IntelliJ to be sure last changes were being applied...

Can you please review ASAP so we can all benefit from this running again?

Thanks!!

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/session/DMNEditorSession.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/session/DMNEditorSession.java
@@ -54,11 +54,11 @@ import org.kie.workbench.common.stunner.core.client.command.CanvasCommandManager
 import org.kie.workbench.common.stunner.core.client.command.CanvasViolation;
 import org.kie.workbench.common.stunner.core.client.command.Request;
 import org.kie.workbench.common.stunner.core.client.command.SessionCommandManager;
+import org.kie.workbench.common.stunner.core.client.registry.impl.ClientCommandRegistry;
 import org.kie.workbench.common.stunner.core.client.session.Session;
 import org.kie.workbench.common.stunner.core.client.session.impl.DefaultEditorSession;
 import org.kie.workbench.common.stunner.core.client.session.impl.ManagedSession;
 import org.kie.workbench.common.stunner.core.diagram.Metadata;
-import org.kie.workbench.common.stunner.core.registry.impl.ClientCommandRegistry;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.pinning.impl.RestrictedMousePanMediator;
 import org.uberfire.mvp.Command;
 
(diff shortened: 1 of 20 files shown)


romartin:

romartin:
Jenkins please retest this

romartin:

romartin:
ok thanks @hasys , waiting for @tiagodolphine or @wmedvede 's feedback :+1:

tiagodolphine:
Hi @romartin I took a look and the code looks good! It was a great catch by the way :)
As soon as I test on IntelliJ I'll approve it.
Thanks.

romartin:
great @tiagodolphine  thanks, would like to merge it to make our life eaiser these weeks... 
thanks a lot for resporting and testing guys!

romartin:
oops @hasys forgot to mention, although this PR should not have any impact, can you anyway on tomorrow's CI, for example, that the last undo/redo issues, when between editors, are still fixed plz? ( so just double checking I'm not breaking last fixes from @treblereel )
Thanks!

hasys:
Hi @romartin,

in today's build @treblereel 's changes works good. Thank you @romartin, @treblereel !

romartin:
great, thanks for testing @hasys !! :+1:

PR #2655, opened 2019-05-21
DROOLS-3997: [DMN Designer] Included Model Name marshalling issues depending on imported model name

danielzhe:
Each imported model has its own namespace (example: https://kiegroup.org/dmn/_6C3D72C4-001F-446B-8673-8075AC5E927C").

When we import a `model X` in to another `model Y`, the imported `model X` will have an alias inside the `model Y` (example: "included1").

Before this PR, we was using the name of the imported model for create this alias, which may lead to errors in the XML because you can't have alias with spaces.

This PR now introduces a new strategy: each import has a generated alias name, that is the word "included" followed by a sequential number ("included1", "included2", and so on).

If some import are removed, the number are re-used.

In this PR there are some mechanisms to handle the fact that in some points we don't know what is the alias current being used, so we have to search through the `namespace context` in the current DMN File looking for the namespace for then get the alias being used or the `import` nodes collection.

PLUS: You can't add the same imported node twice in diagram. There was a bug in the code that verify this. This is fixed in this PR too.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/common/DMNMarshallerImportsHelperImpl.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/common/DMNMarshallerImportsHelperImpl.java
@@ -64,6 +64,8 @@ public class DMNMarshallerImportsHelperImpl implements DMNMarshallerImportsHelpe
 
     private final DMNMarshaller marshaller;
 
+    static final QName NAMESPACE = new QName("Namespace");
+
     public DMNMarshallerImportsHelperImpl() {
         this(null, null, null, null);
     }
@@ -167,6 +169,7 @@ public class DMNMarshallerImportsHelperImpl implements DMNMarshallerImportsHelpe
 
         final String namespace = anImport.getName();
 
+        drgElement.getAdditionalAttributes().put(NAMESPACE, anImport.getNamespace());
         drgElement.setId(namespace + ":" + drgElement.getId());
         drgElement.setName(namespace + "." + drgElement.getName());
         updateInformationItem(namespace, drgElement);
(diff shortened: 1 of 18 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/common/DMNMarshallerImportsHelperImpl.java
```diff
@@ -64,6 +64,8 @@
 
     private final DMNMarshaller marshaller;
 
+    static final QName NAMESPACE = new QName("Namespace");
```
jomarko: Shouldn't we reuse this constant in `DMNMarshaller` line 456?

jomarko:
### Manual review
I struggled using the imported node in the diagram, see my [comment](https://issues.jboss.org/browse/DROOLS-3997?focusedCommentId=13737610&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13737610) and please let me know if you think it is related to this PR.

danielzhe:
@jomarko I noticed there is some unity tests failing, but I'm working on it now. It shouldn't interfere in the manual tests.

danielzhe:
Tests fixed!

danielzhe:
Thank you, @jomarko .
Ready for you review, @manstis . :)

manstis:
@danielzhe Further to my PR comment I tried locally removing `DRGElement.namespace` and adding the following to `DMNModelInstrumentedBase`. It appears to work the same (a couple of tests needed changing and one needed more data to be mocked for use of `nsContext`).
```
    public String getDefaultNamespace() {
        if (nsContext != null && nsContext.containsKey("")) {
            return nsContext.entrySet().stream()
                    .filter(kv -> Objects.equals(kv.getKey(), ""))
                    .findFirst()
                    .map(Map.Entry::getValue)
                    .get();
        }
        if (this.parent != null) {
            return parent.getDefaultNamespace();
        }
        return "";
    }
```

danielzhe:
@manstis Changes applied.

PR #2666, opened 2019-05-24
[AF-2006] Fix broken test 

desmax74:
Removed checks on getRulesByPackage

### DIFF
--- a/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/test/java/org/kie/workbench/common/services/backend/compiler/ClassLoaderProviderTest.java
+++ b/kie-wb-common-services/kie-wb-common-compiler/kie-wb-common-compiler-core/src/test/java/org/kie/workbench/common/services/backend/compiler/ClassLoaderProviderTest.java
@@ -192,9 +192,6 @@ public class ClassLoaderProviderTest {
         KieModuleMetaInfo kieModuleMetaInfo = metaDataOptional.get();
         assertThat(kieModuleMetaInfo).isNotNull();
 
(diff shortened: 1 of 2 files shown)


desmax74:
jenkins test

wmedvede:

wmedvede:
If someone else can take a look at this, review, etc, it'll more than great since we have PRs bolcked because of this.

I really don't know who are the best suited person for looking into this, probably?
 @manstis  @danielezonca

domhanak:
Could you please comment on why were the checks removed?

Rikkola:
Not against the change, but does anyone know why the results changed?

wmedvede:
@danielezonca sure, I've added you to the thread off the top of my head, but +1 to try to move this forward and open a ticket for fine grained investigation.

Thanks!

domhanak:
> Could you please comment why were the checks removed?

Seriously, we are removing failed asserts from the tests, without knowing the cause of these failures. This is not ok.

mareknovotny:
@desmax74 could you please share your reasoning behind removed test assertion?

There are multiple questions not answered here so I would really love to shed some light on the reasoning.

desmax74:
@mareknovotny the assertions was based on the objects exported two years ago by the compiler, every change on pmml (isn't the first time) who change the  numbers breaks this test and the only solution here is to adapt to the new changes in other modules, or when an author of changes far from this code breaks this tests at the same time must adapt the new requirements in these assertions, or change this assertion based on something useful from pmml cc @lanceleverich

mareknovotny:
@desmax74  I would expect that PMML changes and following Full downstream build run should show the failing tests in other repositories.
Right now the tests are not trustworthy at all as the numbers are always customized to the results.

PR #2672, opened 2019-05-29
DROOLS-3727: [DMN Designer] kogito: Minimal standalone preparation

manstis:
A couple of changes following feedback from @gitgabrio.


manstis:
@romartin FYI, this changes DMNs `DMNDiagramSubmarineWrapper` that _I think_ may be copied and pasted into BPMN's `kogito` editor.. it just makes some things a bit cleaner/more obvious.

PR #2675, opened 2019-05-30
JBPM-8525: Stunner - less and greater signs in data-types breaks the process

romartin:
Hey @hasys 

Here is the fix for this **BLOCKER** - [JBPM-8525](https://issues.jboss.org/browse/JBPM-8525)

PS: TBH I just applied same solution as in the `name` field in the process variables editor. I don't really like it, but at this point we don't have too much time for improving the code. So used same approach and also added concrete tests for it.

Thanks!


romartin:

romartin:

hasys:
Hi @romartin,

it is still possible to use less and greater signs in data types in some cases:
image

Another minor issue: it is possible to put **-** sign in the data type field. It is minor since when you build the project you will see an error and it will be possible to open the process and remove the symbol.

It is not blocking this PR but blocking initial Jira issue. So, we can merge it since this PR covers some cases, but Jira issue should remains open and finished with another PR.

romartin:
ok @hasys I'm agree, let's merge this but keep the ticket open. Can you please update it with the right issues still to fix? Thanks in advance!

PR #2677, opened 2019-05-30
DROOLS-4093: [DMN Designer] Sometimes shortcuts does not work when editing DataType items

danielzhe:
This PR fix the following issues, that causes the shortcuts stop working when you are editing and/or creating a new Data Type:
1. When you click on a DataType row, the selection was not changed, staying on the previous Data Type. With this PR, clicking on a row also select it.

2. The focus and selection for the DataType was lost after you selected a type from the drop down menu. With this PR, now the focus gets back to the selected DataType.

Notice the two examples bellow, one showing the old behavior and other showing the new:

**OLD BEHAVIOR**:
Click doesn't select the row (notice the "light darker blue"), so when you are editing a row that is not selected, the CTRL + S doesn't work. Also, notice that the selection is lost when you select atype in the drop down menu:
old

**NEW BEHAVIOR:**
Now when you click with the mouse, the selection also changes (notice a "light darker blue"). Also the CTRL + S (save) key and CTRL + E (edit) is working. The selection is not lost when you select a type from the drop down:
new

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeListShortcuts.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/shortcuts/DataTypeListShortcuts.java
@@ -122,6 +122,10 @@ public class DataTypeListShortcuts {
         view.focusIn();
     }
 
+    public void highlight(final Element dataTypeElement) {
+        view.highlight(dataTypeElement);
+    }
+
     public void reset() {
         view.reset();
     }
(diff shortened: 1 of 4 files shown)


danielzhe:
@manstis Since @jomarko is not available today, @etirelli asked if we can merge only with your review (if it is everything working in your test), since it's a small change with big impact in usability.

manstis:
Oh @etirelli you are cad!

I'll check first thing tomorrow.

manstis:
https://www.urbandictionary.com/define.php?term=cad

manstis:
@danielzhe I added a new commit to this PR that fixed an issue I found when testing; if you're editing a row and press `<Escape>` the row selection was removed. If you pressed `<Ctrl>`+`<S>` the row remained selected. Could you please double check this works OK and then this can be merged.

danielzhe:
@manstis Thank you!! It worked without any issue.

PR #2685, opened 2019-06-04
DROOLS-4115: [DMN Designer] Included Models - Fix typo in the Included Models tab

karreiro:
See https://issues.jboss.org/browse/DROOLS-4115

Demo:
Screen Shot 2019-06-04 at 10 58 20


karreiro:
ok to test.

PR #2686, opened 2019-06-05
DROOLS-4124: Decision Service can not be larger than 500 (width) x 200 (height)

manstis:
See https://issues.jboss.org/browse/DROOLS-4124

This PR simply removes the maximum limitation and removes the dimension properties from the Properties Panel (for Decision Services). Other elements still have dimension properties on the Properties Panel however see this [comment](https://issues.jboss.org/browse/DROOLS-4124?focusedCommentId=13742697&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13742697)... depending on the response this PR could change a little!


manstis:
@jomarko @danielzhe The common opinion (on the JIRA) was to remove the dimension properties from the Properties Panel for all elements. The second commit on this PR performs it. This PR is therefore complete, ready for final review and merge.

PR #2690, opened 2019-06-07
DROOLS-3755: [DMN Designer] Data Types - Constraints (Range/Enumeration) - Add "Date/Time" component when the type is "Date/Time"

danielzhe:


karreiro:
ok to test

danielzhe:
@karreiro @jomarko Wait before the review, because I'll have to do a change. In this PR the Date and Time are mandatory, but Time shouldn't be.

danielzhe:
@jomarko @karreiro ready.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/time/DateTimeSelectorView.less
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+[data-i18n-prefix="DateTimeSelectorView."] {
+  .date-input-field {
+    width: 110px !important;
```
karreiro: @danielzhe We _**really**_ need to avoid the use of `!important`. We could be more specific here by using extra classes (`.date-input-field.my-css-class`), or the HTML element you're applying it (`div.my-ss-class`), or the hierarchical tree (`.my-class .some-inner-class`, `.my-class > .some-inner-class`), or even the three previous approaches together.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/time/DateTimeSelectorView.less
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  .time-input-field {
+    width: 80px !important;
```
karreiro: ..the same here.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/time/DateTimeSelectorView.less
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  .selector-container {
+    width: 110px !important;
```
karreiro: ..and the same here.

danielzhe:
@karreiro Changes applied. Also there's a fix for the display value with an "," at the end showing when time is not set.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/time/DateTimeValueTest.java
```diff
@@ -0,0 +1,113 @@
...
+    public void testHasDate() {
+
+        final DateTimeValue value = new DateTimeValue();
+        value.setDate("date");
+
+        final boolean actual = value.hasDate();
+
+        assertTrue(actual);
+    }
+//
```
karreiro: @danielzhe Can we remove these commented tests?

danielzhe:
@jomarko About the issue with the toggle, it was not introduced by this PR. In fact only the "label" of the drop down that is changed from "Nothing selected" to "None" and should be always "None" or "Nothing Selected" (I'm not sure) when there's no item selected.
About the issues with the Firefox, I wasn't able to reproduce here. Everything is good here in Firefox. I'm using Firefox 67.0.2. Which version are you using? I did some changes here that may help you, but I'm not sure. It's pushed already.

danielzhe:
@jomarko Ok, I've found that the issue with the position of the elements happens in Firefox in Linux. Working on to fix it. Thanks.

danielzhe:
@jomarko Ready.

cristianonicolai:
jenkins please retest this

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateSelector.java
```diff
@@ -97,5 +101,7 @@ public String toDisplay(final String rawValue) {
         void select();
 
         void onValueInputBlur(final Consumer<BlurEvent> blurEvent);
+
+        boolean isChildOfView(final Object element);
```
jomarko: As in the implementation of `isChildOfView` we cast `Object` to `Element` without any check, wouldn't be better to use `Element` directly in method signature in this interface?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/time/DateTimeValueConverter.java
```diff
@@ -0,0 +1,110 @@
...
+    String extractTime(final String value) {
+
+        if (value.length() <= 11) {
+            return "";
+        }
+        return value.substring(11);
+    }
+
+    String extractDate(final String value) {
+        return value.substring(0, 10);
```
jomarko: Could we please introduce a constant `DATE_PORTION_LENGTH` and then use `DATE_PORTION_LENGTH + 1` in `extractTime`? Feel free to use any name of the constant but I think it will make code easier to understand.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateSelector.java
danielzhe: I don't see why, since we will have to cast anyway. We'll just move the cast to "upper level" in the call chain. Wdyt?

jomarko: From my point of view "automatic" casting `Object` to `Element` is sign we have inappropriate api somewhere. However it is probably caused by some previous API design. So keep as it is in scope of this PR.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/time/DateTimeValueConverter.java
danielzhe: Change applied.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/date/DateSelector.java
karreiro: +1 @jomarko

PR #2706, opened 2019-06-14
DROOLS-4144: [DMN Designer] Imported model is broken after saving changes

danielzhe:
There was 2 issues that was causing the error that this PR fixes:
1. The `xmlns` was duplicated, since `XStreamMarshaller` looks for the DMN URI in the namespaces. If he doesn't found, it adds it as default. Since the default `xmlns` in the file attached to the JIRA was not the default DMN URI, it causes duplicated `xmlns`.

2. After the step 1 applied, there was another issue with `typeref`.  In the provided file, the `namespace` was the same as `kie:xmlns`, but after the conversion of file from DMN 1.1 to 1.2, the `kie:xmlns` is updated. It leads to some references not being found, since the variables was declared in the default namespace that is not the same as `kie:xmlns` anymore. With this PR, if `kie:xmlns` is the same as the default `namespace`, the prefix `kie` is removed from `typerefs`.


cristianonicolai:
jenkins please retest this

PR #2719, opened 2019-06-24
AF-2080: The method defined in a parent data object class is not visible when editing guided decision table

manstis:
See https://issues.jboss.org/browse/AF-2080

### DIFF
--- a/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/BlackLists.java
+++ b/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/BlackLists.java
@@ -29,32 +29,32 @@ public class BlackLists {
      * @param methodName
      * @return
      */
-    public static boolean isClassMethodBlackListed( final Class<?> clazz,
-                                                    final String methodName ) {
-        if ( "class".equals( methodName ) || "hashCode".equals( methodName ) || "equals".equals( methodName ) || "toString".equals( methodName ) ) {
+    public static boolean isClassMethodBlackListed(final Class<?> clazz,
+                                                   final String methodName) {
+        if (isInObjectMethodsBlackList(methodName)) {
             return true;
         }
 
-        if ( Collection.class.isAssignableFrom( clazz ) ) {
-            if ( isInCollectionMethodsBlackList( methodName ) ) {
+        if (Collection.class.isAssignableFrom(clazz)) {
+            if (isInCollectionMethodsBlackList(methodName)) {
                 return true;
             }
         }
 
-        if ( Set.class.isAssignableFrom( clazz ) ) {
-            if ( isInSetMethodsBlackList( methodName ) ) {
+        if (Set.class.isAssignableFrom(clazz)) {
+            if (isInSetMethodsBlackList(methodName)) {
                 return true;
             }
         }
 
-        if ( List.class.isAssignableFrom( clazz ) ) {
-            if ( isInListMethodsBlackList( methodName ) ) {
+        if (List.class.isAssignableFrom(clazz)) {
+            if (isInListMethodsBlackList(methodName)) {
                 return true;
             }
         }
 
-        if ( Map.class.isAssignableFrom( clazz ) ) {
-            if ( isInMapMethodsBlackList( methodName ) ) {
+        if (Map.class.isAssignableFrom(clazz)) {
+            if (isInMapMethodsBlackList(methodName)) {
                 return true;
             }
         }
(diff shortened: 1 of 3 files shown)


manstis:
Jenkins please retest this.

ok to test.

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/BlackLists.java
```diff
@@ -29,81 +29,91 @@
...
-    private static boolean isInListMethodsBlackList( final String methodName ) {
-        return ( "addAll".equals( methodName )
-                || "containsAll".equals( methodName )
-                || "iterator".equals( methodName )
-                || "listIterator".equals( methodName )
-                || "removeAll".equals( methodName )
-                || "retainAll".equals( methodName )
-                || "subList".equals( methodName )
-                || "toArray".equals( methodName )
+    private static boolean isInSetMethodsBlackList(final String methodName) {
```
jomarko: Looking on `isInCollectionMethodsBlackList` and `isInSetMethodsBlackList` I think we should invoke `isInCollectionMethodsBlackList` from `isInSetMethodsBlackList`.

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/BlackLists.java
```diff
@@ -29,81 +29,91 @@
...
+        return ("addAll".equals(methodName)
+                || "containsAll".equals(methodName)
+                || "iterator".equals(methodName)
+                || "removeAll".equals(methodName)
+                || "retainAll".equals(methodName)
+                || "toArray".equals(methodName));
+    }
+
+    private static boolean isInListMethodsBlackList(final String methodName) {
```
jomarko: Similar comment as above, could we reuse somehow `isInCollectionMethodsBlackList` in implementation of `isInListMethodsBlackList` ?

jomarko:
### Manual review

It works fine. Just FYI, I found a similar issue for scenarios [DROOLS-4259](https://issues.jboss.org/browse/DROOLS-4259)

And last note/question, would be possible to add some unit test please?

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/BlackLists.java
manstis: This was just a formatting change, but happy to improve as suggested.

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/util/BlackLists.java
manstis: This was just a formatting change, but happy to improve as suggested.

manstis:
@jomarko Updates and test added in new commit.

Interestingly _existing_ tests failed following my first change and I had to add the `BlackLists.isInObjectMethodsBlackList(..)` method! The value of tests :-)

kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/test/java/org/kie/workbench/common/services/datamodel/backend/server/builder/projects/ClassMethodInspectorTest.java
```diff
@@ -53,6 +55,23 @@ public void testSimpleMethods() throws Exception {
...
     }
 
+    @Test
+    public void testInheritedPublicMethods() {
+        final ClassMethodInspector ext = new ClassMethodInspector(ChildClass.class,
+                                                                  new JavaTypeSystemTranslator());
+
+        for (String s : ext.getMethodNames()) {
+            assertFalse("Method " + s + " is not allowed.",
+                        checkBlackList(s));
```
jomarko: Thank you for tests @manstis, I believe I have lats question. From my point of view `ClassMethodInspectorTest.checkBlackList` is redundant logic that "makes tests greeen". What do you think about its replacement? For this case:
```
assertFalse("Method " + s + " is not allowed.", BlackLists.isClassMethodBlackListed(ChildClass.class, s));
```

manstis: Hi @jomarko I'm less sure about this proposal. Using `BlackLists.isClassMethodBlackListed(..)` in the test assertion would not highlight an error in `BlackLists.isClassMethodBlackListed(..)` itself. For example, what if that method failed to filter out `hashCode`? The tests would then continue to pass although `hashCode` should be excluded.

So I think I believe the reverse logic of what you state; IMO `ClassMethodInspectorTest.checkBlackList(..)` is in essence the test for `BlackLists.isClassMethodBlackListed(..)`.

PR #2720, opened 2019-06-24
JBPM-8576: add Multiple Instance and its related properties

xiezhang7:


xiezhang7:

xiezhang7:
Hi @romartin @LuboTerifaj , 

This is the pull request for [JBPM-8576](https://issues.jboss.org/browse/JBPM-8576). 

I have added the missing Multiple Instance and its related properties. Other properties are missing due to task [JBPM-8085](https://issues.jboss.org/browse/JBPM-8035), so please [confirm if we need to undo JBPM-8085](https://issues.jboss.org/browse/JBPM-8576?focusedCommentId=13750902&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13750902). 

Thanks.

romartin:

romartin:
@LuboTerifaj @romartin please check

romartin:

romartin:
can you please @wmedvede review this? thx in advance!

romartin:
CC @LuboTerifaj missing review here too plz

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/CaseManagementSetChildCommand.java
```diff
@@ -91,16 +91,20 @@ public Node getCandidate() {
...
             }
         }).orElseGet(() -> {
-            if (canvasIndex.isPresent() && canvasIndex.getAsInt() == 0 && isStage(parent, getCandidate())) {
-                // If add a Stage to the start, find the StartNoneEvent
-                List<Node<View<?>, Edge>> childNodes = ((Node<View<?>, Edge>) parent).getOutEdges().stream()
-                        .map(e -> (Node<View<?>, Edge>) e.getTargetNode()).collect(Collectors.toList());
-                for (int i = 0, n = childNodes.size(); i < n; i++) {
-                    if (childNodes.get(i).getContent().getDefinition() instanceof StartNoneEvent) {
-                        return OptionalInt.of(++i);
+            if (canvasIndex.isPresent() && canvasIndex.getAsInt() == 0 ) {
```
wmedvede: I'm not familiar with this code here, so I'm not sure if it's an additional fix or it's related with the MI properties.
I leave this to @LuboTerifaj that has more knowledge in the field.

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/graph/CaseManagementSetChildNodeGraphCommand.java
```diff
@@ -354,48 +356,52 @@ private void removeRelationship(final Edge edge,
...
-            double parentOH = pBounds.getHeight();
-            double childH = childOY + CHILD_HEIGHT + STAGE_GAP;
-            double parentH = parentOH > childH ? parentOH : childH;
+                final double cx = STAGE_GAP;
+                final double cy = STAGE_GAP + (STAGE_GAP + CHILD_HEIGHT) * i;
 
-            double parentOW = pBounds.getWidth();
-            double childW = CHILD_WIDTH + STAGE_GAP * 2;
-            double parentW = parentOW > childW ? parentOW : childW;
+                // set stage child bounds
```
wmedvede: I'm not familiar with this code here, so I'm not sure if it's an additional fix or it's related with the MI properties.
I leave this to @LuboTerifaj that has more knowledge in the field.

xiezhang7:

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/CaseManagementSetChildCommand.java
xiezhang7: Thanks @wmedvede. It is not related to MI. I have moved this code to a separate [commit](https://github.com/kiegroup/kie-wb-common/pull/2720/commits/677909af5ddfbc765b4852e036fe820ab49b0b9d).

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-case-mgmt/kie-wb-common-stunner-case-mgmt-client/src/main/java/org/kie/workbench/common/stunner/cm/client/command/graph/CaseManagementSetChildNodeGraphCommand.java
xiezhang7: Thanks @wmedvede. It is not related to MI. I have moved this code to a separate [commit](https://github.com/kiegroup/kie-wb-common/pull/2720/commits/677909af5ddfbc765b4852e036fe820ab49b0b9d).

xiezhang7:
> Then regarding execution, I could find a couple of issues that I'll try to explain here.

Thanks, @wmedvede . I have fixed the MI variable issues in this [commit](https://github.com/kiegroup/kie-wb-common/pull/2720/commits/f44dbcba2f790f67f684d19576c1bd82f5e5d53f).

wmedvede:
jenkins retest this

wmedvede:
Hi @xiezhang7 I've built and tested your last changes locally and it seems like the case file variables are not yet included in the check.
If I create this variables and the process bellow I can still modify/delete the variables.

CaseFileVariables

CaseFileVariables2

LuboTerifaj:

xiezhang7:
> Hi @xiezhang7 I've built and tested your last changes locally and it seems like the case file variables are not yet included in the check.
> If I create this variables and the process bellow I can still modify/delete the variables.
> 
@romartin @wmedvede I am not sure how to fix this, because case file variables are implicitly renamed only when they are assigned. As a result, it is valid to create process variable and case file variable with the same name and data type. 

But the [variable widget code](https://github.com/kiegroup/kie-wb-common/blob/f1c194f9f3ffd24a45a6efba480edc35f4848d37/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/variablesEditor/VariableListItemWidgetViewImpl.java#L160) only has access to the variable name and data type, the context of the variable is not available there (i.e. whether it is a process variable or case file variable). So it is not possible to decide whether the variable should be renamed to match its assigned name or not. 

Please advise. Thanks. 

Screenshot from 2019-07-22 17-19-40
Screenshot from 2019-07-22 17-54-54

romartin:
Hey @xiezhang7 

Well I don't really have all the detail for this now, but I think I understand the Q... so lemme add a few comments:
* first thing to say is that the "[variable widget code](https://github.com/kiegroup/kie-wb-common/blob/f1c194f9f3ffd24a45a6efba480edc35f4848d37/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/variablesEditor/VariableListItemWidgetViewImpl.java#L160)" you mention is a class for the view side, it means we should not add logic here... so please, if any refactor is needed here, consider doing it on the `Presenter` instance, instead of on the `View`
* IMO the `variable widget` you mention should never has access to that kind of detail, it's better to make it "configurable" in some way to achieve your needs. Then, from the presenter, you can "tune" the widget configuration. Also, in the presenter, you can inject the SessionManager type and obtain the current session, this way you'll also get access to the diagram instance and all the beans and states (the model), so you can evaluate here whatever predicate you need against the model and update the widget state or configuration

Anyway @wmedvede has also good knowledge on that part, is not first time we need to do this kinda checks, so he probably can help us on some solution for this.

Thanks in advance guys!

wmedvede:
> > Hi @xiezhang7 I've built and tested your last changes locally and it seems like the case file variables are not yet included in the check.
> > If I create this variables and the process bellow I can still modify/delete the variables.
> 
> @romartin @wmedvede I am not sure how to fix this, because case file variables are implicitly renamed only when they are assigned. As a result, it is valid to create process variable and case file variable with the same name and data type.
> 
> But the [variable widget code](https://github.com/kiegroup/kie-wb-common/blob/f1c194f9f3ffd24a45a6efba480edc35f4848d37/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/variablesEditor/VariableListItemWidgetViewImpl.java#L160) only has access to the variable name and data type, the context of the variable is not available there (i.e. whether it is a process variable or case file variable). So it is not possible to decide whether the variable should be renamed to match its assigned name or not.
> 
> Please advise. Thanks.
> 
> Screenshot from 2019-07-22 17-19-40
> Screenshot from 2019-07-22 17-54-54

Hi @xiezhang7  @romartin 

I believe that the problem here is that the "case file variables" are prefixed in different places around the code, while in my opinion they should be prefixed only in one place and basically at the time they are created in the "VariablesEditorWidget/Presenter".

This is what we have right now:

1) The VariablesEditorWidget/Presenter doesn't know if it's creating a "normal process variable" or a "case file varialbe", so when the user tries to create a variable "myVariable", the widget always creates a variable with this name. And this name is internally prefixed at save time with the "caseFile_" prefix, so the real name for the variable is "caseFile_myVariable".

See here:

https://github.com/kiegroup/kie-wb-common/blob/c1ebbeace5e6ec4ac28fe578033ffa1c0aed48dd/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/fromstunner/properties/ProcessPropertyWriter.java#L201

I don't think this is a good strategy.

2) On the other hand, when other widgets like the "AssignmentsWidget" needs to get the "case file varialbes" the following method is used during the invocations:

https://github.com/kiegroup/kie-wb-common/blob/5d4c7e2e88e698091584f66f55e67c6d8d076363/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/cm/CaseFileVariables.java#L69

If you see a sort of prefixing is being done also there.

3) On the other hand, when the process is being read, the prefix of the variable is striped. See here:

https://github.com/kiegroup/kie-wb-common/blob/5d4c7e2e88e698091584f66f55e67c6d8d076363/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/backend/converters/tostunner/properties/CaseFileVariableReader.java#L37

So we have that the prefixing/un-prefixing of the variable is being spread around the code in different moments of time, when the variable name is really one name "caseFile_myVariable".

The "caseFile_" prefix is needed because in the end, the variable defined in the "case file variables" will end-up in a property element in the generated process, and collisions of names are not wanted with the names of the "normal process variables". Also at some point it's required to know, if a given property element is for the "case file variables" or for the "normal process variables". So let's say that this "caseFile_" prefix is ok, and required also for backward compatibility.

This is what is created in the generated bpmn2 file:

<bpmn2:property id="aNormalProcessVariable" itemSubjectRef="_aNormalProcessVariable" name="aNormalProcessVariable"/>

<bpmn2:property id="caseFile_myVariabe" itemSubjectRef="_caseFile_myVariable" name="caseFile_myVariable"/>

So to summarize, in my opinion it should work like this.

1) The VariablesEditorWidget/Presenter should have a configuration parameter to know if it's managing "normal process variables" or "case file variables"

2) If "case file variables" are configured, then, if you type "myVariable" the widget will internally create "caseFile_myVariable".  And to remove prefix "caseFile_" is just a matter of visualization.

3) If variables are created well by the VariablesEditorWidget/Presenter, then, there's probably no need to prefixing with "caseFile_" when the process is being written. (see my code pointer above to ProcessPropertyWriter file)

4) And there's no need also to do this sort of calculation in the CaseFileVariables.java#L69

5) At read time the prefix filtering is still needed to know to which set of variables (the "normal process variables" or the "case file variables" a given property must be assigned. But this ok, since there's no other way to know that. Since in the end, in the bpmn2 file what you'll see is things like this. So you need to have something to filter.

<bpmn2:property id="aNormalProcessVariable" itemSubjectRef="_aNormalProcessVariable" name="aNormalProcessVariable"/>
<bpmn2:property id="caseFile_myVariabe" itemSubjectRef="_caseFile_myVariable" name="caseFile_myVariable"/>

6) Since the "VariblesEditorWidget/Presener" is instantiated/configured in two different places, for the "normal process variables" and for the "case file process variables" the widget can be configuration with a parameter e.g. "VARIABLE_TYPE" (PROCESS, CASE_FILE) that indicates to the widget the type of variables that are being managed. So this is not a problem. In fact we are currently passing parameters to other field type editor widgets, see here https://github.com/kiegroup/kie-wb-common/blob/b5e1f526f0b98e47adbd8eb6670d35981c691574/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/connectors/SequenceFlowExecutionSet.java#L52

I believe all this prefixing comes from the fact that this VariablesEditorWidget/Presenter comes form old jBPM designer, and also, different persons, at different moments of time has worked in different parts of code, etc. 

Finally I think there are two paths here:

1) Try to refactor to some variant of the reasoning above 

2) Try the following quick solution.

When the VariablesEditor widget needs to check if "myVariable" can be deleted or modified, execute the validation twice, and validate both 	"myVariable" and "caseFile_myVariable". If one of this two checks fails, then the variable can't be modified/removed. I believe this could cover both scenarios, by I haven't try it. But is a quick thing to test I guess.

3) or any variant of 2)

I hope this helps.

LuboTerifaj:

LuboTerifaj:
Hi @xiezhang7 , @romartin ,

I am looking at PR but I am still not sure, what is the scope of Jira. See my [comment](https://issues.jboss.org/browse/JBPM-8576?focusedCommentId=13762340&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13762340).

I will finish my review, once it will be clarified.

Thanks!

xiezhang7:

xiezhang7:
Thanks @wmedvede @romartin . I fixed the case file variable issue in [this commit](https://github.com/kiegroup/kie-wb-common/pull/2720/commits/b4cfb557e0c74c7e9d904bded7c8e8a35fc2063a) by using a variant of option 2. 
Like `SequenceFlowExecutionSet`, case file variable is declared in the [annotation](https://github.com/kiegroup/kie-wb-common/blob/b4cfb557e0c74c7e9d904bded7c8e8a35fc2063a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/cm/CaseManagementSet.java#L62). And `VariablesEditorFieldRenderer` requests `VariableUtils` to add case file variable prefix when [finding the usage](https://github.com/kiegroup/kie-wb-common/blob/b4cfb557e0c74c7e9d904bded7c8e8a35fc2063a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/variablesEditor/VariablesEditorFieldRenderer.java#L231).

wmedvede:
Hi @xiezhang7 I tested it locally and works ok, thanks!

Just one last not form my side. Since the adding of the flag FindVariableUsagesFlag affects the search in all the node types and not only the MI Subprocess node, I think the VariableUtilsTest should be enriched to consider this flag in all the nodes that can be visited by the search.
I can see you added already this methods testFindVariableUsagesForMultipleInstanceSubProcessWithOnlyInput_caseFileVariable but I believe this only covers the MultipleInstanceSubprocess node.

Thanks

xiezhang7:

xiezhang7:
Hi @wmedvede @LuboTerifaj , I have added tests for `VariableUtils`. 
As requested, I also added annotation `SkipFormField` to hide the multiple instance properties. Thanks.

romartin:
Hey guys,

As commented on today's meeting, please let's do what @wmedvede proposed - let's include the capabilities for all this fields but let's not make them all exposed on the UI (forms) :+1: 

Thanks!

PR #2727, opened 2019-07-01
[7.23.x] AF-2080 (RHDM-988): The method defined in a parent data object class is not visible when editing guided decision table (#2719)

manstis:
See https://issues.jboss.org/browse/AF-2080

(cherry picked from commit 85a22ef5fbb138694803aa1b101456c9e6999f26)


jomarko:
@Rikkola I updated the test slightly here https://github.com/kiegroup/kie-wb-common/pull/2732 may I ask you for review?

PR #2735, opened 2019-07-03
DROOLS-3917 : The 'run all' panel should automatically close after leaving the project's library

Rikkola:
https://issues.jboss.org/browse/DROOLS-3917

@kkufova @danielezonca @gitgabrio Here you are.

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/test/TestRunnerReportingScreen.java
+++ b/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/test/TestRunnerReportingScreen.java
@@ -16,6 +16,8 @@
 
 package org.kie.workbench.common.workbench.client.test;
 
+import java.util.Objects;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/test/TestRunnerReportingScreen.java
```diff
@@ -55,6 +63,12 @@ public Widget asWidget() {
         return panel.asWidget();
     }
 
+    public void onPlaceGainFocusEvent(final @Observes PlaceGainFocusEvent placeGainFocusEvent) {
+        if (!placeGainFocusEvent.getPlace().getIdentifier().equals(IDENTIFIER)) {
```
danielezonca: Are you sure `placeGainFocusEvent.getPlace().getIdentifier()` is always different from `null`?

Rikkola: Perhaps it is best to not trust any of these. I'll add checks. Place and identifier might be null.

Rikkola: PlaceRequest can't be null, but identifier could depending on what the implementation was.

danielezonca:
jenkins retest this please

Rikkola:
@gitgabrio The thing is, when do we know we left? This was the event that fits most cases.
For example I tried closing after we move to outside of Library perspective, but project list for example is still Library. I tried closing after the asset list is closed, but then opening an asset does not close it.

So I had a setup where the screen would hide when closing the asset list or when an asset is opened, but this breaks if you move to anywhere else, for example if we add settings screens or anything like that.

This now closes when ever focus changes and it does have problems, but any screen that would get focus is likely to close this screen.

gitgabrio:
@Rikkola 
Hi, I understand that was the underlying issue. Moreover, I do not know exactly the context where this happens. Just an idea (may be wrong, of course): if "project library" is a widget/view/screen/whatever, moving to another widget *should* fire a **lost** focus event for it. So, would not possible to intercept and react on this ?

Rikkola:
@gitgabrio Yes, but the library loses the focus when the test results gains it. And the test results only loses the focus if the user moves out of the list, either to project list or into an asset.

gitgabrio:
@Rikkola 
OK, ok, no way, you know better! :smile:

Rikkola:
@gitgabrio Not necessary, if there are better ways I'm open to change, but due to the workbench structure I can't think of any better ways.

kkufova:
@gitgabrio, could you please approve this?

PR #2757, opened 2019-07-15
DROOLS-4297 : Guided Rule Template removes quotation marks in concatenated String

Rikkola:
https://issues.jboss.org/browse/DROOLS-4297

https://github.com/kiegroup/drools/pull/2448
https://github.com/kiegroup/kie-wb-common/pull/2757

Checking if this breaks tests.


Rikkola:
@manstis Hi, here is the better fix for https://github.com/kiegroup/drools/pull/2439 . I could not find any hints that the DataType.THIS would have any special setup when set as the field type.

manstis:
Thanks @Rikkola. Please let me know when this is ready for review.

Rikkola:
@manstis ready for review. The existing unit tests should reflect the change.

manstis:
@Rikkola Do you have anybody in QE in mind, as @jomarko is on PTO.

Rikkola:
@manstis I asked Johnny Fomin, not sure what his github name is.

PR #2787, opened 2019-07-29
DROOLS-4314: [DMN Designer] Rename 1.1 package to a generic name (since it covers DMN 1.2 too)

karreiro:
See https://issues.jboss.org/browse/DROOLS-4314


jomarko:
@Bonuseto may I ask you for a review of this PR? I wrote down a acceptance test we should do in jira [DROOLS-4314](https://issues.jboss.org/browse/DROOLS-4314)

karreiro:
@danielzhe Could you please take a look on this when you get a chance? :-)

karreiro:
@danielzhe Could you please take a look when you get a moment free. Thanks.

PR #2791, opened 2019-07-31
DROOLS-4321: [DMN Designer] Search feature - DMN Data Types must be searchable elements

karreiro:
See: https://issues.jboss.org/browse/DROOLS-4321

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/762
- https://github.com/kiegroup/kie-wb-common/pull/2791
- https://github.com/kiegroup/drools-wb/pull/1207

PR #2797, opened 2019-08-02
DROOLS-4396: Focus expression grid after search is closed

jomarko:
For more details see the jira https://issues.jboss.org/browse/DROOLS-4396

The PR is dependent on #2791


jomarko:
@karreiro I would like to ask you for review. However as it is based on you PR, I created just draft PR without test, just to find out what do you think about my proposed approach to solve this.

karreiro:
Thank you, @jomarko. The proposed approach makes sense. Great job adding the default interface ;-)

Unfortunately, I'm going to commit changes in my PR that will generate some conflicts here (nothing that invalidates your approach). So, I think you could wait a little bit, before writing the unit tests to avoid new conflicts hehe (I will commit my changes by tomorrow until the EOD).

Anyhow, the draft is ✅ approved for now ;-)

jomarko:
@karreiro thank you for review. I will wait for your changes and then resolve conflicts and add unit tests.

jomarko:
@karreiro may I ask you for a re-review?

PR #2798, opened 2019-08-03
JBPM-8124: Remove legacy process designer from the asset list. 

inodeman:
@romartin can you check this PR?

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/EditorIds.java
+++ b/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/EditorIds.java
@@ -22,6 +22,7 @@ public interface EditorIds {
     String GUIDED_SCORE_CARD = "GuidedScoreCardEditor";
     String XLS_SCORE_CARD = "ScoreCardXLSEditor";
     String STUNNER_DESIGNER = "BPMNDiagramEditor";
+    String LEGACY_DESIGNER = "jbpm.designer";
     String CASE_MODELLER = "CaseManagementDiagramEditor";
     String SCENARIO_SIMULATION_DESIGNER = "ScenarioSimulationEditor";
 }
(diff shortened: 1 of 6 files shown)


inodeman:
This changes adds Editor Permissions/Entries for the Legacy Designer. This PR relates to 2 additional PRs. (https://github.com/kiegroup/kie-wb-distributions/pull/950, https://github.com/kiegroup/jbpm-designer/pull/858)

romartin:

inodeman:
@romartin Can you re-check this PR, now the New Designer is removed from Editors List

romartin:
missing @hasys review

hasys:

lazarotti:
I would vote for 7.5 release make this message more clear to alert users that they need to enable the legacy designer to open these files instead of this generic message. The other options are also good but it can put the cutoff on risk. Is that fine for PM @ibek ?

ibek:
Is it possible to remove the asset from new asset list but keep the legacy process designer active so users can open legacy processes? Users wouldn't be able to migrate the processes without activating the legacy Process designer.

Alternatively the dialog @lazarotti is talking about could actually have a button to migrate the process if needed.

romartin:
hey @ibek 

TBH I'm not really sure, but I think it's not possible to disable the asset (legacy designer) only when creating, but making it available for using with already existing files... you know @ederign @porcelli @krisv guys if that's possible??

If not, I'm also agree with @lazarotti on just improving the error message popup that appears, which indicates that the legacy desginer must be enabled in order to edit existing files. Do you agree on this otherwise???

Thanks!
Roger

porcelli:
guys... you're definitely trying something that is not supported, and not even sure if we should change stable code that has been working the last ~7 years to make this idea work.

use permissions to enable/disable is also a stretch (this was never meant to do what you're trying). what's close what you need is provided by the experimental framework. [1]

[1] - https://docs.jboss.org/drools/release/7.25.0.Final/drools-docs/html_single/index.html#_wb.experimentalFeatures

hasys:
@ibek, @lazarotti, @krisv, @romartin,

any resolution here? Should we postpone it to 7.6 and provide better experience to users rather than make a decision / do implementation in a rush?

lazarotti:
honestly, the main issue is the not clear message, independent of permission or experimental feature the user will not understand what that message means. I would try first change that message for this use case. If that is done on time, we can evaluate as next step if the task could be changed to experimental framework and deliver to 7.6 or have as it is to 7.5.

porcelli:
@lazarotti that's the point, I don't think we can easily allow changing such message. and I'm not even sure if we should.

romartin:
yeah guys I'm also not much convinced about this, let's discuss that on tomorrow's checkpoint meeting...

PR #2810, opened 2019-08-12
DROOLS-4178: [DMN Designer] Selected Node Is selected also in Documentation

karreiro:
See https://issues.jboss.org/browse/DROOLS-4178

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/DMNDocumentationView.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/DMNDocumentationView.java
@@ -20,6 +20,7 @@ import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
 import com.google.gwt.event.dom.client.ClickEvent;
+import elemental2.dom.DomGlobal;
 import elemental2.dom.HTMLButtonElement;
 import elemental2.dom.HTMLDivElement;
 import org.jboss.errai.ui.shared.api.annotations.DataField;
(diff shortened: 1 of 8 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/DMNDocumentationView.java
```diff
@@ -80,7 +81,8 @@ public DMNDocumentationView(final HTMLDivElement documentationPanel,
 
     @Override
     public DocumentationView<Diagram> refresh() {
-        documentationContent.innerHTML = getDocumentationHTML();
+        refreshDocumentationHTML();
+        refreshDocumentationHTMLAfter200ms();
```
jomarko: Could you please explain why we need to refresh both immediately and after 200 ms?

jomarko:
### Manual review notes

#### Note 1
If you select multiple nodes (draw rectangle around them) such rectangle is also in docuemntation
Screenshot from 2019-08-12 14-15-14
Screenshot from 2019-08-12 14-15-10


#### Note 2
If you select a node in diagram, then switch to the documentation tab, the node is unselected, however if you return back to the model tab, the node is unselected also there, is it demanded?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/DMNDocumentationView.java
karreiro: Canvas has a delay in applying all async operations that clear selections. Some of them are event-based, but coupling the documentation refresh with these events would generate a considerable amount of undesirable refreshes (and would impact on the performance of the editor). So, I prefer to just wait for a large amount to milliseconds (200ms) to ensure that the canvas will be updated. I don't like delays like this one too, but it seems like a fair trade-off.

karreiro:
Thank you for the review, @jomarko. Changes applied.

Note 1: Fixed.

Note 2: When users click away from the node into the canvas, the node is unselected. As a user, I would expect the same when I click on the "Documentation" tab. Thus, I don't see it as an issue, but if you still see it as an unexpected behavior, we can handle it into a separated JIRA and check UX team opinion there.

Thanks again :-)

karreiro:
The broken test is not related to this PR.

PR #2812, opened 2019-08-13
RHDM-886: specify the Maven compiler plugin version

almope:
RHDM-886: specify the Maven compiler plugin version with version.compiler.plugin


almope:
Replaced with https://github.com/kiegroup/kie-wb-common/pull/2814

PR #2817, opened 2019-08-15
DROOLS-4168: [DMN Designer] Documentation - Implement styles and presentation enhancements

karreiro:
See https://issues.jboss.org/browse/DROOLS-4168

**Notice:** Some browsers do not support background colors in the print mode (PDF). Currently, only Chrome shows the blue background color in the Data Types list and the gray color in the cover ([Read more](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-print-color-adjust)). All other styles could be successfully applied.


karreiro:
@jomarko here's the [.war](https://www.dropbox.com/s/mmloj5rv8iytcya/kie-wb-common-dmn-webapp-DROOLS-4168.war?dl=0) for this PR.

karreiro:
@jomarko you're right. Background colors are not supported in some browsers (for the print version). Chrome has the `-webkit-print-color-adjust` but it's not a standard. This CSS property works only on some versions of Chrome and Safari, but I'm enabling it on this PR to make the background works in these browsers.

I don't think we need a new issue for that since it has a minimal impact. Also, considering our recent scrum meetings, the PDF will be created in the server-side in a near the future (not in the browser). So, probably, this detail won't be visible anymore.

jomarko:
@karreiro thank you. I agree no new issue will be reported. Merge this once @danielzhe has a look.

karreiro:
Rebasing the branch.. 👀

PR #2822, opened 2019-08-17
JBPM-8292:  SLA due date property is missing check formatting

inodeman:
@romartin can you check this PR, I am about 80% done, just wanted to send this as a check point. I got the functionality to work which is to validate SLA throughout the app, I still get errors on a couple of tests on the validator, but wanted your input on this. Thanks


romartin:
@inodeman plz fix the conflicts and re-fire the builds. Thx!

PR #2828, opened 2019-08-21
DROOLS-4450: [DMN Designer] DMN Graph - Tooltips are not showing the text properly

karreiro:
See https://issues.jboss.org/browse/DROOLS-4450

Before:
before

After:
after

The rename of the `org.kie.workbench.common.dmn.api.1_1` package (https://github.com/kiegroup/kie-wb-common/pull/2787) caused this issue. Now the labels are updated, and the issue is fixed.


karreiro:
@jomarko I do agree we need to rename the package in the backend as well. But, since we're in the software freeze, I think we can postpone this change to the next Sprint: https://issues.jboss.org/browse/DROOLS-4451.

PR #2832, opened 2019-08-22
AF-2064: Unexpected error is displayed after project is created on OpenShift [Cat1]

paulovmr:
Now if any problem happens when trying to resolve project just after creation, the project will be created normally, the popup will be closed and the success message will appear, the only thing that might happen is the project not being automatically opened.


paulovmr:

PR #2836, opened 2019-08-23
AF-2161: Can't import a sample with the same name

jesuino:
The issue is that this cleanup method[1] tries to start a batch update on system config FS, which was already started by the examples service, the start batch fails, then it would go into finally block, which would wrongly close the batch;
 The solution was to use a call to spaceConfigStorageRegistry instead starting the batch manually., because spaceConfigStorageRegistry makes the correct file lock handling;

[1] https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-backend/src/main/java/org/kie/workbench/common/screens/explorer/backend/server/ExplorerServiceImpl.java#L334


jesuino:
would you please review, @ederign @domhanak

paulovmr:

PR #2842, opened 2019-08-28
Fix kie-wb-common-stunner-showcase

tiagodolphine:
@manstis here is the fix


kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/pom.xml
```diff
@@ -39,6 +39,7 @@
     <module>kie-wb-common-stunner-bpmn-project-api</module>
     <module>kie-wb-common-stunner-bpmn-project-client</module>
     <module>kie-wb-common-stunner-bpmn-project-backend</module>
+    <module>kie-wb-common-stunner-bpmn-kogito-runtime</module>
```
manstis: Don't you only want to add this module for the `kogito` profile?

manstis: OK, thanks for the explanation.

I use the profile name `kogito` in DMN so it'd be good to be consistent - when you get around to doing as you suggest.. we can't have `kogito` representing two different things!

PR #2854, opened 2019-08-30
JBPM-8758: [Stunner] BPMN process can't be saved if opened in parallel with DMN

karreiro:
See https://issues.jboss.org/browse/JBPM-8758


manstis:
As far as a fix for the blocker JIRA goes I guess this is plausible however it seems a little bit of a hack. Ideally I guess the index needs to respond to on focus and loose focus events? Please advise if this is considered interim or final?

manstis:
OK, recompiled everything and I can still get the above error.

Steps I followed:-

1. New, clean installation of WF
2. New, clean deployment of Business Central (build with this PR applied to `kie-wb-common`)
3. Incognito browser session
4. Design-> MySpace -> import Mortgages example
5. New DMN file
6. Add Decision
7. Perform search for "De" (first Decision added)
8. Do not close editor
9. Go back to Mortgages using bread crumbs
10. New Process
11. Error appears

karreiro:
Thank you, @manstis. I could reproduce the scenario above here. Working in the fix right now.

karreiro:
@danielzhe Could you please take a look?

manstis:
@danielzhe Don't you dare find an issue! 

I merged the 7.26.x PR this morning w/o your approval as it was needed for the community build...

PR #2866, opened 2019-09-04
DROOLS-4395: Change DMNMarshallerImportsHelper.loadPath(..) to use org.uberfire.backend.vfs.Path

manstis:
See https://github.com/kiegroup/kie-wb-common/commit/efdad75c8ee65b82185c1385551ca2b231a22ed3#r34944714


manstis:
I've tested with Business Central and confirm PMML imports are still correctly resolved :-)

PR #2874, opened 2019-09-06
DROOLS-4478: [DMN Designer] Decision Service - Models from other vendors are being imported with extra decision services

danielzhe:
I added some comments in my changes because `DMNMarshaller` already have those kind of comments, but I don't know if it is really required.  I think it may be useful to help people from the future to understand why we are removing elements that doesn't have shape.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
@@ -228,6 +228,9 @@ public class DMNMarshaller implements DiagramMarshaller<Graph, Metadata, Diagram
         drgElements.addAll(diagramDrgElements);
         drgElements.addAll(importedDrgElements);
 
+        // Remove DRGElements that doesn't have any local or imported shape.
+        removeDrgElementsWithoutShape(drgElements, dmnShapes);
+
         Map<String, Entry<org.kie.dmn.model.api.DRGElement, Node>> elems = drgElements.stream().collect(toMap(org.kie.dmn.model.api.DRGElement::getId,
                                                                                                               dmn -> new SimpleEntry<>(dmn,
                                                                                                                                        dmnToStunner(dmn, hasComponentWidthsConsumer, importedDrgElements))));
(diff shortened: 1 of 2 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
```diff
@@ -457,6 +460,28 @@ public Graph unmarshall(final Metadata metadata,
...
+    void removeDrgElementsWithoutShape(
+            final List<org.kie.dmn.model.api.DRGElement> drgElements,
+            final List<DMNShape> dmnShapes) {
+
+        if (dmnShapes.size() == 0) {
+            // DMN 1.1 doesn't have DMNShape, so we include all DRGElements and create all the shapes.
+            return;
+        }
+
+        for (int i = drgElements.size() - 1; i >= 0; i--) {
```
manstis: Any reason to iterate downwards?

danielzhe: Because I'm removing items, so I don't have to handle with the index when I remove an item.
Let's suppose this scenario:
drgElements = [a, b, c]
drgElements.size = 3
i = 1

So I removed the "b".

Now, in the next iteration of the loop:
drgElements = [a, c]
drgElements.size = 2
i = 2
And we'll get OutOfBoundsException.
So we'll have to handle with the index, doing i-- after we remove and item.
And I don't want to do that.

But doing it downwards this doesn't happen because the next index will never be changed:
drgElements = [a, b, c]
drgElements.size = 3
i = 1

So I removed the "b".
Now, in the next iteration of the loop:
drgElements = [a, c]
drgElements.size = 2
i = 0

The "a" is evaluated and I don't have to handle with the index.

jomarko:
I will check also.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
jomarko: @manstis @danielzhe I think we can replace this `for` loop with:
```java
drgElements.removeIf(element ->
            !dmnShapes.stream()
                    .filter(s -> s.getDmnElementRef().getLocalPart().equals(element.getId()))
                    .findFirst()
                    .isPresent()
        );
```

jomarko:
### Manual review
Passed

danielzhe:
@jomarko Changes applied! Thank you!

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshaller.java
danielzhe: Good. Thank you!

danielzhe:
@manstis Merge, please.

PR #2942, opened 2019-10-11
JBPM-8835 : Improve performance of response when user performs a click and Drag

inodeman:
This changes accomplishes 3 things
1) Improves Form loading and rendering by checking if the previous form was already rendered if so, it will skil rendering
2) Improves Draging time, eliminates the 800 Millisecond delay when clicking and dragging a node, instead of calling form rendering once at the begining and one at the end, it does only at the end when the mouse is released
3) Improves Node click time to response, eliminates 800 MS delay to show toolbox, instead it shows it and the calls the Form Render at the end instead of at the begining

*** As a side note and as mentioned on call, Every once in a while a click event gets dropped, this is observed both before these improvements and after the improvements. Will research later


inodeman:
@hasys @romartin can you review this PR?

inodeman:

romartin:
Full build failed but WARs are ready to test, thanks!

inodeman:
This PR will be closed as changes from JBPM-8836 Will also be included in the PR sent

PR #2956, opened 2019-10-21
[DROOLS-4541] Adding handler managing for onFocus event of ImportsTab

yesamer:


kie-wb-common-widgets/kie-wb-metadata-widget/src/main/java/org/kie/workbench/common/widgets/metadata/client/KieEditorWrapperView.java
```diff
@@ -35,6 +35,8 @@
         void onEditTabUnselected();
 
         void onOverviewSelected();
+
+        void onImportsTabSelected();
```
manstis: Can we not have a `default` _NOP_ implementation on the interface to avoid adding the empty method on sub-classes? That said.. since the implementation of the new method follows that of the other existing methods I guess it does no harm.

yesamer:
> I guess we'll need a unit test before..... but the change looks good to me.

@manstis Any hint? `addPage()` can be verified with mockito, because it's a superclass method. Can't event verify if that specific page is added, considering current API present. I thin can only test if at least a page is added (maybe).

manstis:
@yesamer Have a look at `MultiPageEditorViewImplTest.testGetTabShownHandler()` in AppFormer.

You can (hopefully!) use this as the starting point for testing the `KieEditorWrapperViewImpl.addImportsTab(..)` method and verifying that when the tab is activated the `Page.onFocus(..)` method is invoked and `presenter.onImportsTabSelected()` called.

yesamer:
No longer necessary

PR #2957, opened 2019-10-21
[AF-2269]: Add indexing thread limit

adrielparedes:
Added indexing theads limit

Related to: 
* https://github.com/kiegroup/appformer/pull/822
* https://github.com/kiegroup/kie-wb-common/pull/2957


kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-backend/src/main/java/org/kie/workbench/screens/workbench/backend/impl/DefaultApplicationScopedProducer.java
```diff
@@ -92,7 +93,7 @@ public DefaultApplicationScopedProducer(@Named("luceneConfig") MetadataConfig co
                                             DefaultIndexEngineObserver defaultIndexEngineObserver,
                                             IndexersFactory indexersFactory,
                                             Event<BatchIndexEvent> batchIndexEvent,
-                                            @Unmanaged ExecutorService executorService,
+                                            @Indexing ExecutorService executorService,
```
porcelli: what exact source of exec service is used here? is the one provided by the container (WildFly)?

adrielparedes: managed or unmanaged we are using the ExecutorServiceProducer, and actually is the same code for both. You can check that class in Appformer to see that we are using the same code, but a new ExecutorService for those services

PR #2969, opened 2019-10-23
KOGITO-474: [DMN Designer] Remove Thorntail profile if no longer needed

manstis:
See https://issues.jboss.org/browse/KOGITO-474


manstis:
@jomarko This PR is good to review.

PR #2990, opened 2019-10-31
KOGITO-523: [DMN Designer] Cannot define Time Data Type Range constraint

manstis:
See https://issues.jboss.org/browse/BAPL-1376

(Although manifested itself in https://issues.jboss.org/browse/KOGITO-523).

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/858
- https://github.com/kiegroup/kie-wb-common/pull/2990

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/DMNKogitoCommonEntryPoint.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/DMNKogitoCommonEntryPoint.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.workbench.common.dmn.webapp.kogito.common.client;
+
+import javax.annotation.PostConstruct;
+
+import org.jboss.errai.ioc.client.api.EntryPoint;
+import org.uberfire.client.views.pfly.sys.PatternFlyBootstrapper;
+
+@EntryPoint
+public class DMNKogitoCommonEntryPoint {
+
+    @PostConstruct
+    public void init() {
+        PatternFlyBootstrapper.ensureMomentTimeZoneIsAvailable();
+    }
+}
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZonesProvider.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZonesProvider.java
@@ -44,10 +44,10 @@ public class TimeZonesProvider {
 
         for (String name : names) {
             final double offset = getOffset(name);
-            final String format = getOffsetString(name);
+            final String offsetString = getOffsetString(name);
             timeZones.add(new DMNSimpleTimeZone(name,
                                                 offset,
-                                                format));
+                                                offsetString));
         }
         return timeZones;
     }
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZonesProvider.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZonesProvider.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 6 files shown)


danielzhe:
@manstis My only concern with this approach is localization to Japanese. I don't know if they use "our alphabet" to timezones or their alphabet, otherwise I think that is ok because is universal and I don't think that create a new timezone is something common.

manstis:
@danielzhe Could you please explain your concern a little more?

Are you saying the server side approach returns i18n'ed texts? I didn't see anywhere the User's locale we considered when extracting the time zone information!? Perhaps I misunderstand.

manstis:
Jenkins please retest this.

manstis:
@danielzhe @karreiro @jomarko This PR is about as good as it's going to get.. ready for review!

manstis:
Jenkins please retest this.

manstis:
Much of the code is generated..... I'm not going to make the generation accommodate repeated literals.

manstis:
Hi @jomarko let me try to explain this PR a bit better.

The Business Central editor has a _real_ Java backend and this uses `TimeZoneServiceImpl` to retrieve a list of time zones from the Java Runtime Environment (from a `tzdb.dat` in your PC's `jre/lib` folder). This is not available to kogito that runs 100% in the browser.

So, class `TimeZoneDumper` can be ran from the CLI (or in an IDE etc) and this simply outputs the results of `TimeZoneServiceImpl` to _stdio_ (`System.out`) as a `String`. I then simply (and manually) copy this `String` and paste it into the new class used by kogito namely [TimeZones](https://github.com/kiegroup/kie-wb-common/blob/02f10ce17a307821e053ebae91078026b6e80724/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZones.java).

Time zones are pretty static and I think this approach is _good enough_. An alternative approach would be to generate the content of the `TimeZones` class at GWT compilation time but to be honest I felt it was over engineered for what is in essence a static list of time zones.

Does this help explain things?

jomarko:
Ok, I understand now probably that `TimeZoneDumper` is helper class that prints content of the `TimeZone.getTimeZones` however I do not see added value in tracking that file.

It requires additional configuration for maven plugins to not have it in jars. Even if Java.FUTURE will provide more timezones, our `Dumper` is not started automatically, its output is compared in any test. 

Maybe I am missing something, but I do not see the reason why to track this file.

manstis:
@jomarko I understand your comment and will happily remove.

However before doing so would like to hear @karreiro and @danielzhe opinions too.

manstis:
@karreiro I'll take a look at Moment Timezone :-)

manstis:
@jomarko @danielzhe This now uses https://momentjs.com/timezone to get client-side time-zones.

manstis:
Jenkins please retest this.

manstis:

manstis:
Jenkins please retest this.

manstis:

manstis:
Jenkins please retest this.

manstis:

jomarko:

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/DMNKogitoCommonEntryPoint.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.jboss.errai.ioc.client.api.EntryPoint;
+import org.uberfire.client.views.pfly.sys.PatternFlyBootstrapper;
+
+@EntryPoint
+public class DMNKogitoCommonEntryPoint {
+
+    @PostConstruct
+    public void init() {
+        PatternFlyBootstrapper.ensureMomentTimeZoneIsAvailable();
```
jomarko: @manstis sorry for my question, could you please explain/remind me, why we load moment time zone in this class, and moment just in `kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/DMNClientEntryPoint` I mean in totally different module.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZonesProvider.java
```diff
@@ -0,0 +1,66 @@
...
+        return TIME_ZONES;
+    }
+
+    private List<DMNSimpleTimeZone> buildTimeZones() {
+        final List<DMNSimpleTimeZone> timeZones = new ArrayList<>();
+        final String[] names = getNames();
+
+        for (String name : names) {
+            final double offset = getOffset(name);
+            final String format = getOffsetString(name);
```
jomarko: Personal opinion, it is confusing to mix `format` and `offsetString`, I mean third parameter of `DMNSimpleTimeZone` 

feel free to ignore if you think it is fine

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/DMNKogitoCommonEntryPoint.java
manstis: Hi @jomarko moment time zone is only used by the _kogito_ editors (to retrieve a list of time-zones in the _time_ related data-type constraint editor popups). moment is used by both Business Central and _kogito_ to handle _date_.

manstis: This _kogito_ related module depends on `kie-wb-common-dmn-client` and hence for _kogito_ editors both moment and moment time zone are loaded/injected into the HTML page.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/services/TimeZonesProvider.java
manstis: Updated... I orginally had a method called `getFormat(..)` that is consistent with moment time zone but then I saw `DMNSimpleTimeZone` called it `offsetString` so I changed the method name for consistency; forgetting to change the variable name.

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

PR #2996, opened 2019-11-05
[AF-2324] Blocking only the project explorer while it is loading

caponetto:
Task: [AF-2324](https://issues.jboss.org/browse/AF-2324)

As requested, only the project explorer will be blocked while it is loading, not the entire workbench.
Currently, users must wait until the loading finishes to continue using the editor when the project explorer is opened. Here's a demo showing this behavior:

2_Master

After this PR, users can interact with the editor while the project explorer is being loaded.
Here's a demo showing this behavior:

3_LoadingChanged_v1
One downside is that users might notice a UI hang while the project explorer is being loaded.
This can be seen when the "select/unselect" flow stops for a while.
For more details, check [RHPAM-2359](https://issues.jboss.org/browse/RHPAM-2359).

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/resources/css/StylesCss.java
+++ b/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/resources/css/StylesCss.java
@@ -53,4 +53,6 @@ public interface StylesCss
     @ClassName("directory-name")
     String directory();
 
+    @ClassName("busy-indicator")
+    String busyIndicator();
 }
--- a/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/BaseViewImpl.java
+++ b/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/main/java/org/kie/workbench/common/screens/explorer/client/widgets/BaseViewImpl.java
@@ -17,7 +17,6 @@ package org.kie.workbench.common.screens.explorer.client.widgets;
 
 import javax.inject.Inject;
 
-import com.google.gwt.core.client.GWT;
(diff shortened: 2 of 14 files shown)


paulovmr:

paulovmr:

ederign:
@porcelli please review here. @caponetto will provide more details on the JIRA.

porcelli:
as I mentioned in the JIRA, I think we should evolve this a bit more and try to get this proposal at a similar load time we have today. Otherwise, we have the risk to this be considered a "regression"

caponetto:
@porcelli I've noticed that there are differences in the loading time when I'm recording or not. 
Looks like it is influencing the numbers. See:

[**When recording**]
Full-screen loading: ~30s
Only PE loading: ~45s

[**When not recording**]
Full-screen loading: ~23s
Only PE loading: ~23-25s

@tomasdavidorg would you mind doing these comparisons?

porcelli:
It would be good to have confirmation from QE, if the difference is at this level of a few seconds... then we'd be good.

tomasdavidorg:
Hi, what is the Full-screen loading vs Only PE loading. I take this improvement as separation of the loading popups. I don't understand what I should do with the measurement. Could you please take a look on my [comment](https://issues.jboss.org/browse/RHPAM-2359?focusedCommentId=13810544&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13810544)? We are mixing several things here.

tomasdavidorg:
I did some measurement and the values are similar when opening asset after indexing (~8s) and also when opening PE (~25s).

caponetto:
Hi, @tomasdavidorg, thanks for sharing your concerns.
The goal for this PR is to move the loading popup from the entire window to the project explorer (PE) area when actions in the PE require some loading, e.g. when opening the PE.

If you take a look at the gifs above (before and after the changes), you'll see that the first gif takes ~30s to complete the loading process, wheres the second gif takes ~45s. So, the concern is that applying these changes could make things worse.

However, I've noticed that this difference between loading time only occurs when I'm recording.
It could be something related to my environment, and that's why it is important to have someone's else opinion.

If the loading times are equivalent to you as well, then I'd say we can go ahead with these changes.

tomasdavidorg:
Thanks for the explanation. As I already wrote, the time was similar for me ~25s.

PR #3001, opened 2019-11-07
DROOLS-4715: [DMN Designer] Structure is lost when user doesn't drag it enough

karreiro:
See: https://issues.jboss.org/browse/DROOLS-4715

**Before:**

**After:**

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/draganddrop/DNDDataTypesHandlerContextTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/draganddrop/DNDDataTypesHandlerContextTest.java
@@ -19,6 +19,7 @@ package org.kie.workbench.common.dmn.client.editors.types.listview.draganddrop;
 import java.util.Optional;
 
 import com.google.gwtmockito.GwtMockitoTestRunner;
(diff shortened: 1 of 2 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/types/listview/draganddrop/DNDDataTypesHandlerContextTest.java
```diff
@@ -130,24 +132,30 @@ public void testGetReferenceWhenCurrentDataTypeIsPresent() {
 
         final Element firstElement = mock(Element.class);
         final Element hiddenElement = mock(Element.class);
+        final Element wrongFirstElement = mock(Element.class);
```
jomarko: We should add a test, where an element pass the condition `elementY > HIDDEN_Y_POSITION && elementX == 0 && !element.classList.contains(DRAGGING)`

karreiro: Thank you for the review, @jomarko. The `firstElement` variable represents the scenario you've mentioned above :-) (the `wrongFirstElement` is just like the `firstElement`, but `wrongFirstElement` has the  `DRAGGING` CSS class, in other words, it's being dragged).

PR #3004, opened 2019-11-12
DROOLS-4735: [DMN Designer] Grid performance is dire

manstis:
See https://issues.jboss.org/browse/DROOLS-4735

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/843
- https://github.com/kiegroup/kie-wb-common/pull/3004
- https://github.com/kiegroup/drools-wb/pull/1264

This PR moves calculation of a row's height from "on demand" calculation of cells' heights to "on value set" i.e. the height of a cell is now calculated when it's value is set hence moving much of the expensive calculation to a far less frequently invoked point in the codebase.

This PR also ensures DMN's `GridLayer` is not instantiated (by Stunner) multiple times and subsequent event handlers attached to each instance (thus leading to multiple re-handling of events that slowed things down).

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
@@ -38,7 +38,6 @@ import org.kie.workbench.common.dmn.client.commands.factory.DefaultCanvasCommand
 import org.kie.workbench.common.dmn.client.editors.expressions.types.ExpressionEditorDefinitions;
 import org.kie.workbench.common.dmn.client.resources.i18n.DMNEditorConstants;
 import org.kie.workbench.common.dmn.client.session.DMNSession;
-import org.kie.workbench.common.dmn.client.widgets.grid.BaseExpressionGrid;
 import org.kie.workbench.common.dmn.client.widgets.grid.BoundaryTransformMediator;
 import org.kie.workbench.common.dmn.client.widgets.grid.ExpressionGridCache;
 import org.kie.workbench.common.dmn.client.widgets.grid.controls.container.CellEditorControlsView;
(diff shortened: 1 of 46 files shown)


jomarko:
### Manual Review Note 1
When I compare files produced by **kie-wb-distributions/business-central-parent/business-central-webapp** and **kie-wb-common/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testeting** the second one doesn't produce new line character in the source file. Is it expected? Not sure if caused by this PR.

manstis:
Hi @jomarko thus PR does not affect the UI to XML process at all so if client-side marshalling (`kie-wb-common-dmn-webapp-kogito-testing`) gives different text compared to server-side marshalling (`kie-wb-distributions`) it is an issue with one of the marshaller. These PRs just affects rendering of the grid on the screen.

jomarko:
@manstis due to memory usage, should we keep client-site-marshalling this way (without new lines) or should I report a KOGITO issue?

manstis:
Hi @jomarko I doubt the inclusion of (even 1000's of) new line characters will affect memory consumption unduly. It's more than likely the XML library we use client-side is not adding the characters (and having them does make the XML more _human readable_). Feel free to make a KOGITO JIRA (IIRC @tiagobento formats the XML in VSCode so we might be able to use the same function to format the output of the client-side marshaller before writing to file).

jomarko:
@manstis reported as https://issues.jboss.org/browse/KOGITO-616

jomarko:
@manstis see my manual acceptance test results at https://issues.jboss.org/browse/DROOLS-4735

Share you opinion, are the red crosses I put caused by wrong dmn table, or is that dmn too big already?

manstis:
@jomarko Hello, I left a comment on the JIRA regarding _editing_ (i.e. I think a new JIRA is in order.. you can get the same result before this PR with larger tables). I think _saving_ should also be a new JIRA as we'd need to see where the bottle neck is.. marshalling UI->XML or even XML->server.

tiagobento:
@manstis @jomarko I'm using native XSLT support on browsers to format the output from both DMN and BPMN editors on kogito-tooling extensions. You can see the rules we use here: https://github.com/kiegroup/kogito-tooling/blob/master/packages/kie-bc-editors/src/DefaultXmlFormatter.ts

jomarko:
Reported remaining issues:
- https://issues.jboss.org/browse/DROOLS-4790
- https://issues.jboss.org/browse/DROOLS-4791

manstis:
@danielezonca All issues resolved/commented.

Please re-review at your pleasure/leisure!

manstis:

PR #3011, opened 2019-11-13
JBPM-8875 : Stunner - Re-usable subprocess doesn't show options for Called Element properties (DROOLS-3727 merge)

inodeman:
@hasys @romartin Can you review this PR?
The problem was that CaseCalledElement and CalledElement depended on ProcessesDataService instance which was defined as an ApplicationScoped singleton, so CaseCalled element when setting up the query and named changed the one that CalledElement depended on

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-project-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/project/backend/service/ProcessesDataService.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-project-backend/src/main/java/org/kie/workbench/common/stunner/bpmn/project/backend/service/ProcessesDataService.java
@@ -22,7 +22,7 @@ import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
-import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Dependent;
 import javax.enterprise.event.Event;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


inodeman:
@hasys can you check this PR with Case Management project, my experience is limited here, so let me know if any issues

inodeman:
@hasys these two files seemed suspicious to mee as they both specify the Same Provider Files(CaseReusableSubprocessTaskExecutionSet, ProcessReusableSubprocessTaskExecutionSet)
Does this seem right? Just wanted to mention it
image
image

inodeman:

hasys:
Hi @inodeman,

yes I already asked @xiezhang7 the same question, if I understood correctly it is not possible to use inheritance here for some reasons related to annotations. But if you will find a way how to reuse code, I will vote for it :)

Going to test this PR with processes and cases now, thank you @inodeman!

hasys:
Hi @inodeman,

thank you, works good, just one question: looks like your fix will have one downside - it will create instance of ProcessesDataService for each opened process/case, is it correct? I guess we can report a future improvement to use just one for Processes and one for Cases. (To get rid of duplications if it's possible also can be a good improvement).

### Check list
- [x] ACKs are set
- [x] Affected version is set property
- [x] Is realization aligned to jira description?
- [x] Are BPMN and CM verified?
  - [x] BPMN
  - [x] CM
- [x] Both Jenkins jobs are green

### Code review
- [x] Code coverage
- [x] License headers
- [x] Static analyser gates
- [x] Code Style

### Acceptance Criteria
> * Steps to reproduce from the [jira issue](https://issues.jboss.org/browse/JBPM-8875)
> * New Project with two processes, open both of them at the same time, check Called Element, should be listed both process from the new project and process from steps to reproduce from previous project (check process from another project also listed)
> * Create new Case Project and two Cases, open case and check that in Called Element you can see both cases

- [x] Tested with full downstream build
  - [x] Firefox
  - [x] Chrome

### Found issues
>  **Blocker**, **Critical** and **Major** bugs and **Blocker** and **Critical** enhancements should be fixed before merge.

#### One minor issue reported
* [JBPM-8899](https://issues.jboss.org/browse/JBPM-8899) - Called Element of re-usable sub-process in cases contains null as option

### Used resources
Processes for quick test can be found here: use **Mortgage_Process** sample project

romartin:
Jenkins please retest this

inodeman:
Jenkins please retest this

PR #3032, opened 2019-11-21
BAQE-1171: Fix dmn data types list view tests on jdk11

jomarko:
For more details see https://issues.jboss.org/browse/BAQE-1171


jomarko:
@manstis I asked you for review even it is @karreiro 's component but his long PTO. It is just change in tests.

PR #3033, opened 2019-11-21
DROOLS-4711: [DMN Designer] User is able to drag single data type entry horizontally

danielzhe:
This PR is based on a branch started by @karreiro 
It address this two JIRAS:
[DROOLS-4711](https://issues.jboss.org/browse/DROOLS-4711): [DMN Designer] User is able to drag single data type entry horizontally
[DROOLS-4714](https://issues.jboss.org/browse/DROOLS-4714): [DMN Designer] Filtering doesn't show results properly

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/search/DataTypeSearchBar.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/search/DataTypeSearchBar.java
@@ -74,7 +74,7 @@ public class DataTypeSearchBar {
 
     public void reset() {
         setCurrentSearch("");
-        dataTypeList.showListItems();
+        getDataTypeList().showListItems();
         view.resetSearchBar();
         restoreDataTypeListPositions();
     }
@@ -95,7 +95,7 @@ public class DataTypeSearchBar {
     }
 
     HTMLElement getResultsContainer() {
-        return dataTypeList.getElement();
+        return getDataTypeList().getElement();
     }
 
     String getCurrentSearch() {
@@ -107,7 +107,7 @@ public class DataTypeSearchBar {
     }
 
     DNDListComponent getDNDListComponent() {
-        return dataTypeList.getDNDListComponent();
+        return getDataTypeList().getDNDListComponent();
     }
 
     void restoreDataTypeListPositions() {
@@ -123,7 +123,6 @@ public class DataTypeSearchBar {
             final Integer positionY = getDataTypeListPositionsStore().get(uuid);
 
             getDNDListComponent().setPositionY(element, positionY);
-
             if (positionY > -1) {
                 HiddenHelper.show(element);
             } else {
@@ -131,6 +130,7 @@ public class DataTypeSearchBar {
             }
         });
 
+        getDataTypeList().collapseAll();
         getDNDListComponent().refreshItemsPosition();
         getDataTypeListPositionsStore().clear();
     }
@@ -159,14 +159,14 @@ public class DataTypeSearchBar {
     }
 
     private List<DataTypeListItem> getDataTypeListItems() {
-        return dataTypeList.getItems();
+        return getDataTypeList().getItems();
     }
 
     private void showEmptyView(final boolean show) {
         if (show) {
-            dataTypeList.showNoDataTypesFound();
+            getDataTypeList().showNoDataTypesFound();
         } else {
-            dataTypeList.showListItems();
+            getDataTypeList().showListItems();
         }
     }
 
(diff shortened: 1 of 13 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/search/DataTypeSearchBar.java
```diff
@@ -181,6 +181,10 @@ public boolean isEnabled() {
         return !isEmpty(getCurrentSearch());
     }
 
+    private DataTypeList getDataTypeList() {
```
manstis: What is the benefit of this method? I was expecting it to be _default_ access and mocked in a test but since it's `private` IDK the reasoning behind the move to using a method?

manstis:
OK, further to my previous comment here's some more concrete examples:-

- New nested field row is hidden when creating a structure. Should it not be visible? It used to be.
- When adding any new data-type existing expanded rows are collapsed. Should they not stay expanded?
- Add two structures each with a nested field. Expand them all. Add a third structure and the UI appears as follows:
Screenshot from 2019-11-25 11-25-13

danielzhe:
@manstis Ouch! I'm checking the issues. As you mentioned in the e-mails, seems like it's related to this PR. Thank you for the review. :)

jomarko:
@danielzhe please check if your PR solves https://issues.jboss.org/browse/DROOLS-4803, thank you

danielzhe:
@manstis Ready for your re-review.

danielzhe:
@manstis Let me know if you want further explanation about why I'm doing something like [this](https://github.com/kiegroup/kie-wb-common/pull/3033/files#diff-bb346bddc3549af7211124daef174ebcR159).

danielzhe:
@jomarko I was checking the issue you reported (https://issues.jboss.org/browse/DROOLS-4803) and it's not addressed in this PR. I thought the changes may work, but it didn't.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/search/DataTypeSearchBarView.java
```diff
@@ -136,6 +147,77 @@ public void showSearchResults(final List<DataType> results) {
...
+
+            final Optional<DataTypeListItem> parentElement = allElements.stream()
+                    .filter(element -> Objects.equals(element.getDragAndDropElement().getAttribute(UUID_ATTR), parentElementId))
+                    .findFirst();
+
+            if (!isParentElementOnList(groupedElements, parentElementId)) {
+                parentElement.ifPresent(p -> groupElementWithItsParent(groupedElements, allElements, p));
+                groupedElements.add(item);
+            } else {
+                int index = getIndexOfParentOrLastElementInGroup(groupedElements, parentElement.get());
```
manstis: Sonar has a good point about this block. Perhaps consider:-
```
parentElement.isPresent(p -> {
                                  int index = getIndexOfParentOrLastElementInGroup(groupedElements, p);
                                  groupedElements.add(index, item);
                              }
```

danielzhe:
@manstis The proposed fix for https://issues.jboss.org/browse/DROOLS-4711 was proposed by @karreiro and I agree with the solution. Despite the fact that is not allowed when you have only one item, I don't think we need to block this because user will only drag it horizontally in this case by mistake and since we are correcting the mistake for the user it's ok. But I guess is just a matter of taste. For me it "feels" more natural the proposed behavior. Maybe we can also change the cursor when user starts the drag and drop for the "forbidden" cursor in this case. I'm not sure. Wdyt?

About the other issue, I'm checking again. I got others issues related to what you reported but apparently not all issues. ;) 
Thank you!

jomarko:
Reading conversation, not sure about the state, seems still some changes will land in, should I postpone my review?

danielzhe:
@jomarko Yes. Thank you.

danielzhe:
@manstis  The issue with order is fixed. I also fixed another issue with collapsed items.

danielzhe:
@manstis Thank you! Ready for your review, @jomarko

jomarko:
@danielzhe there is one issue when user filter etries and navigate trough them with arrow keys, the entries are reordered non intuitively.  https://issues.jboss.org/secure/attachment/12458171/12458171_filter-navigate-trough.webm

manstis:
@jomarko Your PR https://github.com/danielzhe/kie-wb-common/pull/4 sure makes the code nicer.

Since I am not the author of this PR I'll let @danielzhe decide (and hopefully merge in to this!)

Please let me know if my review is required again.

danielzhe:
Thank you, @jomarko ! Merged your PR!

jomarko:
jenkins please retest this

danielzhe:
@manstis Merge, please. Thank you. :)

danielzhe:
@jomarko Thank you!

PR #3040, opened 2019-11-27
Fix issue with inheriting junit-vintage-engine transitive dependency

manstis:
There might be other occurrences...

However I still question why `kie-server-api` should have a dependency on `optaplanner-core`..


manstis:

manstis:
Closing in preference of https://github.com/kiegroup/optaplanner/pull/621

PR #3041, opened 2019-11-27
[KOGITO-670] Removing stunner dependencies fromm dmn marshall

gitgabrio:
@manstis  @jomarko 

See https://issues.jboss.org/browse/KOGITO-670
Basically, I just moved classes from dmn marshaller to dmn webapp kogito common


manstis:
Jenkins please retest this.

manstis:
Tests are to be provided by https://issues.jboss.org/browse/KOGITO-406

This JIRA is in progress but will need refactoring and moving to `kie-wb-common-webapp-kogito-common` (from `kie-wb-common-webapp-kogito-marshalling`).

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-marshaller/src/main/resources/org/kie/workbench/common/dmn/webapp/kogito/marshaller/jre/javax/xml/namespace/QName.java
```diff
@@ -84,7 +82,8 @@ public QName(final String namespaceURI,
 
         //jsonix JSON properties
         setKey(toString());
-        final String usedPrefix = !StringUtils.isEmpty(getPrefix()) ? getPrefix() + ":" : "";
+        String retrievedPrefix = getPrefix();
+        final String usedPrefix = (retrievedPrefix == null || retrievedPrefix.isEmpty()) ?  "" : retrievedPrefix + ":";
```
jomarko: this is not a correct replacement, the implementation of the `StringUtils.isEmpty` is:
```java
null == s || s.trim().length() == 0;
```

This is why I have been asking contributors to use `StringUtils` now my concern about inconsistency is back, somewhere we will check just string, somewhere trimmed string.

I understand we do not wan't a dependency for whole stunner just because of single method, but we shouldn't make exceptions and copy code. Because then it is a habit not and exception. however if you think there is no existing util class available  in this module and not worth to add it there. I will not block merging.

manstis: I think @gitgabrio has reversed the check from `!StringUtils.isEmpty(..)?a:b` to the equivalent of `StringUtils.isEmpty()?b:a`... note the not operator in the original line (I had to double-check). @gitgabrio is not using `StringUtils` as it forces a dependency on Stunner and he wants to keep the DMN marshaller free from Stunner dependencies as it will be re-used by _scesim_ when opening DMN files.

jomarko: @manstis sorry I miss something, I am saying stunner stirng utils usses also `trim()`, new code doesn't

manstis: @jomarko Ah, OK. I missed that!

gitgabrio: @jomarko @manstis 
It is not an issue - I simply missed that "trim" part - I'll fix.

PR #3046, opened 2019-11-28
DROOLS-4795: [DMN Designer] User is unable to save model when structure is removed

danielzhe:


danielzhe:
@jomarko I have a clue about why we disable the save button: suppose the user is editing a DataType name, but didn't saved it. So when the user clicks on the "Save" button at the top, he may expect that the changes that he made in the Data Type was applied too. Maybe @manstis can confirm if is that the case.

manstis:
@danielzhe Your explanation seems plausible.

Unfortunately the [JIRA](https://issues.jboss.org/browse/DROOLS-3399) that added the disabling has little description to understand the reasoning.

I think this fix needs to be a little smarter and only fire the `DataTypeEditModeToggleEvent` event **if** there are no items in _edit_ mode.

Personally I'd like to see the disabling removed too (as a new JIRA) however we'd need to refactor the current global save. Disabling it prevents the User from saving invalid state - when the tick/commit/save icon in the Data Types tab is clicked the name is validated to ensure uniqueness. This check would need to move to the global save and prevent saving if the Data Types tab had validation errors.

danielzhe:
@manstis Ok, but if we fire only when there are no items in edit mode, then we will have the current bugged behavior. What I'm thinking is when user press the "delete" button, disable edit mode for all items that are in edit mode discarding all existing changes.  So the "save" button will be enabled. Wdyt?

jomarko:
Hi @danielzhe @manstis , I think Michael proposes this
1. check number of remaining entries in edit mode
2. handle deletion of a data type entry in the same way as till now
3. then check again number of remaining entries in edit mode
4. then compare numbers from 1. and 3. if one of them is zero and second non zero, we need to fire `DataTypeEditModeToggleEvent`

manstis:
@jomarko @danielzhe I think if we changed [this](https://github.com/kiegroup/kie-wb-common/blob/07257ddb9ac26d099b1c915a5fe0cde9587efc23/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeListItem.java#L377) to call back to `dataTypeList` that an item has been removed (in the `confirmation.ifIsNotReferencedDataType(..)` command) that class could check if there are any `DataTypeListItem` in _edit_ mode and if zero fire the `DataTypeEditModeToggleEvent` event.

I think this approximates to your steps 3 and 4.

jomarko:
Don't we use `DataTypeEditModeToggleEvent` aslo for disabling? That is why I think we should compare number of entries in edit mode before and after delete operation.

jomarko:
I am just going trough backlog, the save mechanism should be really discussed on some meeting, there is also issue if user switch dmn assets https://issues.jboss.org/browse/DROOLS-3952

danielzhe:
@manstis I agree, I think this is the best approach. 
@jomarko I don't think it's necessary to compare number of entries in edit mode, because we can have only one or zero at time.

danielzhe:
@manstis But wait, If we have a item in edit mode and the item is a child of parent deleted node, should we not delete the item?

manstis:
@danielzhe Yes, I think you are correct. If an item is in edit mode and it's parent is deleted then all children should be deleted (and... there are then no items in edit mode and the global save can be re-enabled!)

jomarko:
Please let me know when this is ready for a re-review.

danielzhe:
@manstis @jomarko Changes applied!

jomarko:
jenkins please retest this

danielzhe:
@manstis All green. Merge, please.

danielzhe:
@jomarko Thank you!

PR #3054, opened 2019-12-03
Fix JavaScript for use in GWTTestCase

manstis:
This is in preparation for my ongoing work for https://issues.jboss.org/browse/KOGITO-406

When running `GwtTestCase` Unit Tests for the client-side marshaller I get JS compilation issues in `MainJS.js` at runtime when the JS are deployed to `com.gargoylesoftware.htmlunit.WebClient` that is the headless browser engine used for tests. The failures relate to use of `const` (presumably the `WebClient` does not support the JavaScript version introducing `const`). Use of `var` works fine.

I also added missing semi-colons (as IDEA keeps highlighting the file saying they're missing - although this is not true with JavaScript as their use is optional).


manstis:
When running the tests for KOGITO-406 I need to compile the marshaller module before running the tests as the old (current) `MainJS.js` file is used. My local work for KOGITO-406 includes the change in this PR.. but having it in `master` now means I don't need to keep re-compiling the marshaller module too. So this PR is entirely for convenience.

jomarko:
jenkins please retest this

PR #3065, opened 2019-12-10
[KOGITO-727] Remove dmn-api reference from dmn marshaller gwt module

gitgabrio:
@manstis 

See https://issues.redhat.com/browse/KOGITO-727


gitgabrio:
@manstis could you please merge this, squashing commits (well, the commit... )? Thx!

manstis:
@gitgabrio I tried to squash the single commit to zero.. but it caused a localised black hole in the UK so I've had to roll back to a single commit. I hope that is OK.

gitgabrio:
@manstis 
I'll be strong and keep going on with the single commit... thanks anyway....

PR #3055, opened 2019-12-03
JBPM-8914 Stunner - User task throws exception when you try to move it

treblereel:
JIRA: https://issues.jboss.org/browse/JBPM-8914

Assigning null to localSearchService is redundant

PR #3078, opened 2019-12-18
JBPM-8914: Stunner - User task throws exception when you try to move it.

romartin:
Hey @hasys @treblereel @LuboTerifaj 

I'd been doing some deep analysis of [JBPM-8914](https://issues.redhat.com/browse/JBPM-8914) and found some conclusions:
* Forms (and all fields) are being rendered (created&destroyed) twice, so although not being strictly a bug, it has some implications on this bug - See reported [JBPM-8954](https://issues.redhat.com/browse/JBPM-8954)
* The NPE happens because, in this case, the form field is sending a request to the backend (to fetch users/groups), but on the meanwhile, the form field is being destroyed (due to above point), so once the response is being sent, the form field's state has been "destroyed"

I thin the best way to fix this situation, rather than getting into Errai's bus stuff and how it works, it's jsut assuming that the request can take long, and also that the component state may be destroyed on the meanwhile, so it's just a matter of checking if instance's state is still valid, once processing the reponse, otherwise just DO NOTHING. Also this fix keeps nullifying the class' fields, because I think ti's better to have the state clean (for further GC), and just do nothing if state is nullified.

So this PR replaces https://github.com/kiegroup/kie-wb-common/pull/3055, as IMO the fix is slightly better (it keeps destroying the state), and also because @treblereel is off, so we can manage it easily for now.

Thanks!!


romartin:
@hasys @LuboTerifaj assigning to both as I'm not really sure who will take it... but please, can you give it TOP priority? It's an urgent BLOCKER... 

Thanks in advance!

romartin:

romartin:

romartin:
@hasys @LuboTerifaj 

Closing this PR for now, as it's being replaced by https://github.com/kiegroup/kie-wb-common/pull/3078. I'll cherry-pick again to master once closing it.

Thanks!

PR #3079, opened 2019-12-18
Remove unused dependencies

jomarko:


jomarko:

jomarko:

jomarko:
@hasys hello, I am using just `mvn dependency:analyze`, let me know if you have more questions.

hasys:
@jomarko do you use ignore non compile as well? Like `mvn dependency:analyze -DignoreNonCompile` because if not it can cause issues with runtime dependencies (some runtime dependencies can be incorrectly marked as unused).

jomarko:
@hasys I was not using that option, thank you

PR #3081, opened 2019-12-18
KOGITO-780: [DMN Designer] Ensure -runtime WAR is created with 'production' parameters

manstis:
See https://issues.redhat.com/browse/KOGITO-780..

Basically I removed the `draftCompile` flag and a duplicate `logLevel` entry.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/main/resources/org/kie/workbench/common/dmn/showcase/DMNKogitoRuntimeWebapp.gwt.xml
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/main/resources/org/kie/workbench/common/dmn/showcase/DMNKogitoRuntimeWebapp.gwt.xml
@@ -20,9 +20,8 @@
 
   <inherits name="org.kie.workbench.common.dmn.webapp.kogito.common.DMNWebappKogitoCommon"/>
 
-  <!-- We don't need to support IE10 or older -->
-  <!-- There is no "ie11" permutation. IE11 uses the Firefox one (gecko1_8) -->
-  <set-property name="user.agent" value="gecko1_8,safari"/>
+  <!-- We only need Chrome -->
+  <set-property name="user.agent" value="safari"/>
 
   <!-- To change the default logLevel -->
   <set-property name="gwt.logging.logLevel" value="SEVERE"/>
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/pom.xml
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/pom.xml
@@ -508,10 +508,9 @@
               <strict>true</strict>
               <localWorkers>${gwt.compiler.localWorkers}</localWorkers>
               <deploy>${project.build.directory}/gwt-symbols-deploy</deploy>
-              <draftCompile>true</draftCompile>
-              <extraJvmArgs>-Xmx4096m -Xms1024m -Xss1M -XX:CompileThreshold=7000 -Derrai.dynamic_validation.enabled=true -Derrai.jboss.home=${errai.jboss.home}</extraJvmArgs>
               <module>org.kie.workbench.common.dmn.showcase.DMNKogitoRuntimeWebapp</module>
-              <logLevel>INFO</logLevel>
+              <extraJvmArgs>-Xmx4096m -Xms1024m -Xss1M -XX:CompileThreshold=7000 -Derrai.dynamic_validation.enabled=true -Derrai.jboss.home=${errai.jboss.home}</extraJvmArgs>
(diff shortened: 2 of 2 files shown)


tiagobento:
@manstis I think we should include `<style>OBFUSCATED</style>` and probably `<optimizationLevel>9</optimizationLevel>` as well.

manstis:
@tiagobento Updates made as proposed.

manstis:
@yesamer @romartin FYI these are the settings we're using for DMN... The `-runtime` WAR is ~64.6MB however `WEB-INF` is ~64.2MB (and excluded from the plugin) meaning the plugin itself is 0.4MB or ~400kb!

manstis:
@yesamer @romartin Ignore the sizes I gave above... I was comparing oranges to apples.

tiagobento:
@manstis on previous builds, DMN editor resources are ~26MB

manstis:
@tiagobento What are "DMN editor _resources_"? The WAR? The exploded WAR less WEB-INF?

manstis:
@tiagobento Do you have a tag/label that I can checkout to investigate? Nothing of any significance has been added explicitly to DMN to warrant a x3 size increase (`-runtime` WAR is now 66MB).

tiagobento:
@manstis That's the exploded WAR less WEB-INF :)

tiagobento:
@manstis https://github.com/kiegroup/kogito-online/tree/gh-pages/editors/0.2.6/dmn

manstis:
@tiagobento It's 36MB with this PR...

tiagobento:
@manstis I guess that's because there's `gecko` on the compilation targets? We only need chrome..

manstis:
@tiagobento Correct... I finally found the difference (9MB for another permutation).

I'll update this PR to just compile Chrome.

manstis:
@tiagobento ~26MB now :-)

tiagobento:
@manstis Thx! Looks perfect now.

manstis:
@jomarko New Year reminder to please review this PR :-)

manstis:
Jenkins please retest this.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/main/resources/org/kie/workbench/common/dmn/showcase/DMNKogitoRuntimeWebapp.gwt.xml
```diff
@@ -20,9 +20,8 @@
 
   <inherits name="org.kie.workbench.common.dmn.webapp.kogito.common.DMNWebappKogitoCommon"/>
 
-  <!-- We don't need to support IE10 or older -->
-  <!-- There is no "ie11" permutation. IE11 uses the Firefox one (gecko1_8) -->
-  <set-property name="user.agent" value="gecko1_8,safari"/>
+  <!-- We only need Chrome -->
+  <set-property name="user.agent" value="safari"/>
```
romartin: Just for my own reasoning - is that really true guys @manstis @ederign @porcelli ? For kogito webapps can we only expect support for `safari` engines? Is because our kogito channels are using only this implemention?

porcelli:
good catch @romartin, I think we need both.

manstis:
@romartin @porcelli, 

@tiagobento said we only [needed](https://github.com/kiegroup/kie-wb-common/pull/3081#issuecomment-567975853) Chrome (when we were discussing WAR sized).

porcelli:
@tiagobento is this true for the online editors? I mean... by using safari alone would still work in all browsers?

I know that it's true for VSCode, GitHub Extensions and Desktop

tiagobento:
@porcelli Well, you're right. I don't remember if at some point we decided to only support Chrome/Safari but we can revisit that. Keep in mind that supporting Firefox will increase the size of VSCode extension and Desktop bundles.

porcelli:
for VSCode, GitHub Extension and Desktop all we need is chrome support.

however.. online editor is a different thing.

manstis:
@porcelli Sooooooo, it's your call..

We can add an additional profile to compile a WAR containing more permutations for _online_ however that's the tip of the iceberg... How it's then compiled as part of the community build and how you consume it in `kogito-tooling` for the different channels would remain undetermined.

Please be sure to keep @romartin, @yesamer and myself informed of your decisions.

yesamer:
@manstis I had the same idea, we can use different profiles to handle different versions of `gwt.xml` with different `user-agent`. Undoubtedly, understand how `kogito-tooling` can take the correct version based on the channel is a big question. Anyway, I'm available to discuss more, at the moment I have both enabled for scesim, but I planned to remove gecko engine soon. Thanks!

romartin:
ok so it looks still unclear... :) 

I'll use `safari` for now in BPMN as well, but please let us know guys if that should change, in some way, at some point... :)

Thanks!

PR #3042, opened 2019-11-27
Fix test assert method

cuijulian:
Required for https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1100

PR #3083, opened 2019-12-18
Clean up assertions in CLI tests

yurloc:
When I was reviewing #3042 I noticed a test that uses very strange style of asserting object's properties using the `hasFieldOrPropertyWithValue()` assertion. This is an anti-pattern. I think we need to clean this up otherwise the tests will be difficult to maintain and refactor in the future as it has already happened (see discussion on #3042).

@pefernan Correct me if there is a good reason for which the tests were written this way and should stay like that.

Types of changes:
- Ban string based object inspection.
- Remove redundant assertions (e.g. isNotEmpty before hasSize).
- Don't mix JUnit assertions with AssertJ in one file.
- Simplify Map entry assertions.


yurloc:
@pefernan ping

domhanak:
@yurloc any updates?

yurloc:
Closing for inactivity.

PR #3088, opened 2019-12-19
DROOLS-3048: [DMN Designer] CellEditorControl needs to issue a new instance per use

manstis:
See https://issues.redhat.com/browse/DROOLS-3048

Details of the issue and manual testing acceptance criteria are on JIRA.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinition.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinition.java
@@ -23,6 +23,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
 import javax.inject.Inject;
 
+import org.jboss.errai.ioc.client.api.ManagedInstance;
 import org.jboss.errai.ui.client.local.spi.TranslationService;
 import org.kie.workbench.common.dmn.api.definition.HasExpression;
 import org.kie.workbench.common.dmn.api.definition.HasName;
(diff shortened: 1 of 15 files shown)


manstis:
@jomarko @karreiro New Year reminder to please review this PR :-)

manstis:
Jenkins please retest this.

manstis:
Thank-you for your feedback @jomarko 

I suspect the issue you report is the same as https://issues.redhat.com/browse/RHDM-1181 and not caused by this PR however I will investigate and report back.

manstis:
@jomarko OK, the issue was similar but not identical so I've added a fix here (in the new commit).

The problem was where the `NameAndDataTypePopoverViewImpl` popup is re-used for different **rows** the state was not being correctly set when shown and the `typeRef` value remained the same value as whatever it had been when previously shown. This is now fixed and tests added.

The problem this PR fixes was for **columns** re-using the popovers.

manstis:
@jomarko Please feel free to re-test this.

manstis:
@jomarko Thank-you for your approval, but to answer your question as I prefer you to understand the reasoning vs just going along with what I've submitted; both Decision Table and Relation can have multiple columns with a "Name and Data-type" popover. The issue was when re-using the same "Name and Data-type" popover for all columns when one such column was not visible it's associated popover was destroyed. Since all columns shared the same popover it meant the popover was effectively destroyed for the visible columns too and therefore when the User double clicked a column header to edit the Name/Data-type the popover was being (very quickly) shown and then destroyed.

Moving these two editors to have different instances of the "Name and Data-type" popover per column means only a single un-shared instance is destroyed per non-visible column and visible columns' still have a working instance to show for the User. 

Other grids do not have multiple columns with a shared "Name and Data-type" popover instance and therefore do not need to use `ManagedInstance`.

**However** IF you think it will make the code more consistent and easier understood in the future it does no harm to make ALL grids use `ManagedInstance` for their popover editors and I can make the change in another commit.

Please let me know what you think.

manstis:
I've put this PR into "DO NOT MERGE" until I receive your feedback.

jomarko:
@manstis thank you for your patience.

Your words explained that it is different reusing `CellControlView` for columns vs. rows. I will definitely leave the decision up to you as more experienced developer. 

The only reason for using `ManagedInstance` consistently would be performance, however I think that is not the case.

PR #3089, opened 2019-12-19
DROOLS-4846: [DMN Designer] User is unable to save model when multilevel structure is removed

danielzhe:
structure


danielzhe:
@karreiro @manstis @jomarko Wtb a reviewer here. ;D

PR #3099, opened 2020-01-03
KOGITO-652: Updating setContent method

jesuino:
Update Kogito editors so they can accept the path parameter in `setContent` method.

Also pass the content path when calling `setContent`.

Part of an ensemble:

https://github.com/kiegroup/kogito-tooling/pull/47
https://github.com/kiegroup/kie-wb-common/pull/3099

### DIFF
--- a/kie-wb-common-kogito/kie-wb-common-kogito-client/src/main/java/org/kie/workbench/common/kogito/client/editor/BaseKogitoEditor.java
+++ b/kie-wb-common-kogito/kie-wb-common-kogito-client/src/main/java/org/kie/workbench/common/kogito/client/editor/BaseKogitoEditor.java
@@ -141,7 +141,7 @@ public abstract class BaseKogitoEditor<CONTENT> {
     /**
      * Used by Kogito to set the XML content of the editor.
      */
-    public abstract void setContent(final String value);
+    public abstract void setContent(final String path, final String value);
 
     /**
      * Used by Kogito to get the XML content of the editor. This should return a {@link String}
--- a/kie-wb-common-kogito/kie-wb-common-kogito-client/src/main/java/org/kie/workbench/common/kogito/client/editor/BaseKogitoEditor.java
+++ b/kie-wb-common-kogito/kie-wb-common-kogito-client/src/main/java/org/kie/workbench/common/kogito/client/editor/BaseKogitoEditor.java
@@ -140,6 +140,10 @@ public abstract class BaseKogitoEditor<CONTENT> {
 
     /**
      * Used by Kogito to set the XML content of the editor.
+     * @param path
+     * Relative path to the content resource
+     * @param value
+     * Editor's content
      */
     public abstract void setContent(final String path, final String value);
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editor/AbstractDMNDiagramEditor.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editor/AbstractDMNDiagramEditor.java
@@ -391,7 +391,7 @@ public abstract class AbstractDMNDiagramEditor extends AbstractDiagramEditor {
 
     @Override
     @SetContent
-    public void setContent(final String value) {
+    public void setContent(final String path, final String value) {
         diagramServices.transform(value,
                                   new ServiceCallback<Diagram>() {
 
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/test/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editor/AbstractDMNDiagramEditorTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/test/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editor/AbstractDMNDiagramEditorTest.java
@@ -428,7 +428,7 @@ public abstract class AbstractDMNDiagramEditorTest {
 
     @Test
     public void testSetContentSuccess() {
-        editor.setContent(CONTENT);
+        editor.setContent("", CONTENT);
 
         verify(clientDiagramService).transform(eq(CONTENT), serviceCallbackArgumentCaptor.capture());
 
@@ -441,7 +441,7 @@ public abstract class AbstractDMNDiagramEditorTest {
 
     @Test
     public void testSetContentFailure() {
-        editor.setContent(CONTENT);
+        editor.setContent("", CONTENT);
 
         verify(clientDiagramService).transform(eq(CONTENT), serviceCallbackArgumentCaptor.capture());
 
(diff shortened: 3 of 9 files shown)


manstis:
@jomarko @karreiro @danielzhe @yesamer @gitgabrio @danielezonca FYI only.

When this is merged, in order to test _kogito_ editors from our `-kogito-testing` webapps you'll need to invoke `gwtEditorBeans.get("DMNDiagramEditor").get().setContent("", "");`

kie-wb-common-kogito/kie-wb-common-kogito-client/src/main/java/org/kie/workbench/common/kogito/client/editor/BaseKogitoEditor.java
```diff
@@ -141,7 +141,7 @@ public boolean isDirty() {
     /**
      * Used by Kogito to set the XML content of the editor.
      */
-    public abstract void setContent(final String value);
+    public abstract void setContent(final String path, final String value);
```
jomarko: Please document the new parameter

gitgabrio:
> @jomarko @karreiro @danielzhe @yesamer @gitgabrio @danielezonca FYI only.
> 
> When this is merged, in order to test _kogito_ editors from our `-kogito-testing` webapps you'll need to invoke `gwtEditorBeans.get("DMNDiagramEditor").get().setContent("", "");`

@jesuino Hi!. What this "path" is for? What is the expected format? I agree with @jomarko about the need for proper documentation in the code.

jesuino:
@gitgabrio @jomarko I am adding a description for the path. It is a relative path for the resource where the content was extracted from. (e.g. `src/main/resources/org/acme/process.bpmn`)

jomarko:
@jesuino thank you, I am still not sure what is the difference in behavior when I use empty _path_ and non empty _path_ parameter. Could you please clarify more?

Aren't path and content exclusive? SHouldn't user use just one? Either read content from `path` or set content by `content` parameter?

If that is the case. I think we shouldn't add new parameter to existing public mehtod, but crate new API method. WDYT?

jesuino:
Hello @jomarko ,

Thanks for your great questions. 

>> @jesuino thank you, I am still not sure what is the difference in behavior when I use empty path and non empty path parameter. Could you please clarify more?

The path should not be empty if the origin of the content is known, which means most of the cases. If it is empty, there's no resource (file) origin.

>> Aren't path and content exclusive? SHouldn't user use just one? Either read content from path or set content by content parameter?
In Kogito Tooling the content is set when a user create or open a new file we call editor.setContent(resourcePath, resourceContent). There are a few things to bear in mind:

* We may have an open editor and change the content dynamically - hence the editor is some sort of stateless, editors may have the content changed while already running in the page, it is not coupled to a specific content;
* we may have multiple editors in the same window -> that's the reason that a shared path object won't work;
* Even if the content does not change, the file may be renamed, and set content will be called with the new path and the current content (when rename is supported) - without having to reopen the whole editor. 

In another word, setContent is the entry point for refreshing editors state and it is what is called from kogito tooling side. That's why we pass the path along with the content.

Please let me know if you have any question

yesamer:
@jesuino Hi William, As I said I'm trying to understand better Kogito project, I would like to have some additional explanations about these changes.
At first glance, @jomarko proposal to provide a separate methods to update the path only make sense and could be considered. 
In your answer, you said the editor is "stateless" and this is the reason why you need to call setContent() with both path and content. I don't understand totally this part, I think an editor should have knowledge of its own content (= it's current status)). What I'm missing here?

Thank you to you or to who will reply :)

jesuino:
@yesamer setContent is what kogito tooling calls on editor side. It can load an editor and later change the content called setContent again. In fact, in some cases setContent("") is called before passing the actual content. So the editor is loaded we can change the content that is displayed, without having to reload the editor. In another words, the editor is not bound to a specific content. This is how it works in Kogito work, not business central.

What changes now is that along the content is also passed the path to that content. If you want to retrieve any information from the file, like: the file name or the file path, you know where you find it.

It does not make sense to have a setPath method, the call for it would have to be followed by a call to setContent. Adding a new method means: create new annotation, update annotation processor for that new annotation, new tests and all the editors will have to implement a method and know that it is called and then setContent.

If you don't like the path you simply ignore it and use the content parameter, it should not impact anything in editors, it is just a new information we are adding.

We thought about making PATH part of the new environment API, however, we noticed that that are cases where multiple editors are in the same window (currently, for example, in PR review), so path could not be a shared object and it is tight to the content, so pass a content and the path is the way to go.

yesamer:
@jesuino Thank you for your explanation, I think what I'm missing here it's to understand in details how `setContent("")` works and when it could be called. I think the knowledge of this part it's crucial for me, in order to integrate SCESIM in the correct way. Can you, or someone of the team, present how it works to me? Thanks!

jomarko:
@yesamer I know only how to use it in `dmn-webapp-kogito-runtime` envirnomnet, but not much details about its implementation. usage is when running given module, put into browser console code `gwtEditorBeans.get("DMNDiagramEditor").get().setContent("")`

ederign:
please jenkins execute full downstream build

tiagobento:
From the console logs, looks like there's a memory issue. Given [this link](https://stackoverflow.com/questions/6773726/command-mvn-package-fails-with-status-137) and our history with GWT memory consumption, I'd say there's a good chance that's the problem happening in the build.

tiagobento:
please jenkins execute full downstream build

mbiarnes:

mbiarnes:

mbiarnes:

tiagobento:
@mbiarnes Thx for re-triggering the build!

From the logs what I can see is that there's insufficient memory at some point during the build causing the failures.

```
[INFO] #
[INFO] # There is insufficient memory for the Java Runtime Environment to continue.
[INFO] # Native memory allocation (mmap) failed to map 738197504 bytes for committing reserved memory.
[INFO] # An error report file with more information is saved as:
[INFO] # /home/jenkins/workspace/KIE/master/pullrequest/kie-wb-common-downstream-pullrequests/hs_err_pid12562.log
[ERROR] Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x0000000415580000, 738197504, 0) failed; error='Cannot allocate memory' (errno=12)
```

I'm going to try and build everything locally here to see if there's a problem with the build or increasing the memory would solve the problem.

tiagobento:

mbiarnes:

mbiarnes:
@romartin @tiagobento @ederign 
after increasing to 8G I get status 1: https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/kie-wb-common-downstream-pullrequests/714/console

mbiarnes:

mbiarnes:
increased to 16G

mbiarnes:
Jenkins execute full downstream build with 
-X

tiagobento:

tiagobento:
Good news! It built! 2 tests failed though, but we're better than before.

ederign:
org.jbpm.process.workitem.camel.karaf.itests.CamelWorkitemIntegrationTest.org.jbpm.process.workitem.camel.karaf.itests.CamelWorkitemIntegrationTest
org.apache.maven.surefire.junit4.JUnit4Provider.org.jbpm.process.workitem.camel.karaf.itests.CamelWorkitemIntegrationTest

PR #3114, opened 2020-01-16
KOGITO-257 : Stunner - Condition Expression should show only "Expression" option

inodeman:
Hi @romartin can you review this PR?

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidget.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidget.java
@@ -175,8 +175,19 @@ public class FormPropertiesWidget implements IsElement,
     }
 
     public boolean areLastPositionsForSameElementSame(final Element element) {
-        final Point2D computedPosition = GraphUtils.getComputedPosition((Node<?, ? extends Edge>) element);
+        final Point2D computedPosition;
+
+        if (isNode(element)) {
+            computedPosition = GraphUtils.getComputedPosition((Node<?, ? extends Edge>) element);
+        } else {
+          return false;
+        }
+
         return (lastElement != null && lastElement.getUUID().equals(element.getUUID()) && computedPosition.equals(lastPosition));
+    }
+
+    private static boolean isNode(final Element<? extends Definition<?>> element) {
+        return element instanceof Node;
     }
 
     private void show(final String graphUuid,
@@ -184,7 +195,7 @@ public class FormPropertiesWidget implements IsElement,
                       final Command callback) {
         if (element != null) {
 
-            if (lastElement != null && lastElement.getUUID().equals(element.getUUID())) {
+            if (lastElement != null && lastElement.getUUID().equals(element.getUUID()) && isNode(element)) {
                 lastPosition = GraphUtils.getComputedPosition((Node<?, ? extends Edge>) element);
                 return;
             }
@@ -228,7 +239,9 @@ public class FormPropertiesWidget implements IsElement,
             final String name = definitionUtils.getName(definition);
             propertiesOpenedEvent.fire(new FormPropertiesOpened(formSessionHandler.getSession(), uuid, name));
             lastElement = element;
-            lastPosition = GraphUtils.getComputedPosition((Node<?, ? extends Edge>) element);
+            if (isNode(element)) {
+                lastPosition = GraphUtils.getComputedPosition((Node<?, ? extends Edge>) element);
+            }
         }
     }
 
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/test/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidgetTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/test/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidgetTest.java
@@ -41,6 +41,7 @@ import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
 import org.kie.workbench.common.stunner.core.graph.Element;
 import org.kie.workbench.common.stunner.core.graph.Graph;
 import org.kie.workbench.common.stunner.core.graph.content.definition.Definition;
+import org.kie.workbench.common.stunner.core.graph.impl.EdgeImpl;
 import org.kie.workbench.common.stunner.core.graph.impl.NodeImpl;
 import org.kie.workbench.common.stunner.core.graph.processing.index.Index;
 import org.kie.workbench.common.stunner.core.util.DefinitionUtils;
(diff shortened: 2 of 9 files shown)


inodeman:

kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidget.java
```diff
@@ -227,7 +227,9 @@ private void show(final String graphUuid,
             final String name = definitionUtils.getName(definition);
             propertiesOpenedEvent.fire(new FormPropertiesOpened(formSessionHandler.getSession(), uuid, name));
             lastElement = element;
-            lastPosition = GraphUtils.getComputedPosition((Node<?, ? extends Edge>) element);
+            if (element instanceof  Node) {
```
romartin: I personally prefer to use null != element.asNode() instead of `element instanceof  Node`, having in mind that on javascript the inheritance is different and for a pure client side developer maybe results more familiar using the null (`undefined`).. anyway just a preference.
Please also create a private static method on the class and re-use it, instead of duplicating the `element instanceof  Node`, it's alway good to re-use code :)

kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/test/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidgetTest.java
```diff
@@ -150,6 +154,9 @@ public void setup() throws Exception {
         when(nodeContent.getDefinition()).thenReturn(nodeDefObject);
         when(node2.getUUID()).thenReturn(NODE2_UUID);
         when(node2.getContent()).thenReturn(nodeContent);
+        when(edge.getUUID()).thenReturn(EDGE_UUID);
```
romartin: I know we'd been mocking nodes/edges accross some test classes, me included, but that doesn't makes much sense and it's not a good practice at all. 

In the testing scope most common way to create domain objects, like a node or an edge, which are basically holding some state, but not  providing functionality, is just by creating new instances, for the "real" class, no need for mocking. Mocking domain / data holder objects just adds complexity, you've to re-define the behavior for the methods being used everywhere, on each test class/case being written...

Also this is why you said that `element.asNode()` was not working on the tests, it's because you're mocking the node class and the instance is a proxy.... Just by using real "nodes/edges" that will work properly. 
Please keep in mind trying to refactor those tests once possible, although this test is not new and you just updated, would be great donig that work all toghether,  during whatever other task where the test results updated. 

Should be hard doing this refactor here? Other test cases are doing this properly, so just do a quick find on the project to see how easy is to create new instances on tests :)

Thx!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/conditionEditor/ConditionEditorGeneratorService.java
```diff
@@ -25,4 +25,7 @@
 public interface ConditionEditorGeneratorService
         extends KogitoClientService<Condition, GenerateConditionResult> {
 
+    default boolean isClient() {
```
romartin: The idea behind this method works for me, but what about renaming it to something more easy to "understand", what about just `isAvailable()`, or similar, this way:
- kogito - will return `false`
- BC - will return `true`
So the code will be read as something like this:
`if (service.isAvailable()) { ... } else { .... }`
WDYT?

romartin:
hey @inodeman , great job again! just added a couple of review comments, but nothing really blocker... thanks!

kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/conditionEditor/ConditionEditorGeneratorService.java
inodeman: Hi @romartin yes, makes sense isAvailable(), will change accordingly

kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/test/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidgetTest.java
inodeman: Hi @romartin yes, totally makes sense, I did not realize it was that easy just to use the real class, changed it in the test I added

inodeman:

inodeman:
Jenkins retest please

inodeman:

inodeman:
Jenkins retest please

inodeman:

inodeman:
Jenkins retest please

romartin:

inodeman:

inodeman:

inodeman:

inodeman:

inodeman:

inodeman:

inodeman:
Hi @romartin @domhanak can you review this PR?

inodeman:
Hi @domhanak can you review this PR?

romartin:
@domhanak plz missing your review, thanks!

romartin:
Thanks @domhanak for your feedback, good catch!

Let you know that @inodeman will be off for a week, so let's keep this in "pause" for a while please, meanwhile he's back...

Thanks guys!

lazarotti:
@domhanak @romartin @inodeman delete issue is not related to this commit. It is a sporadic problem that also happens to me while using Kogito on VSCode. I can say a specific way to reproduce it, but when it happens the only way to remove the activity or the sequece flow, is clicking on the bin shortcut.

lazarotti:
@domhanak just talked to @inodeman and he will fix this behavior by the Jira https://issues.redhat.com/browse/KOGITO-752, but it should not prevent this one to be merged. Is it ok for you?

inodeman:
Hi @domhanak @lazarotti just fixed the alignment issue, basically if it detects that there is only one option it will flush it to the left / will have zero left margin, otherwise will keep 20 pixels as before to show 2 options (business central), let me know if ok @domhanak

inodeman:

inodeman:
Jenkins retest please

romartin:

inodeman:
Jenkins retest please

inodeman:

romartin:
@domhanak plz missing your review, thanks!

hasys:

domhanak:
Checked BC, I was able to choose to either use expression or condition.
Noticed, that after I use expression and modify it I wasn't able to go back to condition unless I reopened the properties panel. Is that expected? I can report it as minor UX issue.
Building VSCode part now..

domhanak:
VSCode works aswell.
I was able to add Condition Expression to sequence flow.
I was able to change it and the changes stayed after saving and reopening.
Let's merge.

PR #3135, opened 2020-01-31
RHDM-1016: [DMN Designer] Scroll Bars are not updated when keyboard is used for navigation

danielzhe:
Part of an ensemble:
https://github.com/kiegroup/kie-wb-common/pull/3135
https://github.com/kiegroup/appformer/pull/885


manstis:
@jomarko Did you want to assign either yourself or Anna for review please.

yesamer:
@manstis @danielzhe Hi! It seems the bug it's not present in ScenarioTest editor. I just tried with current master code. I can investigate more on the reason of this difference. Anyway, I think we should agree and use the same approach

danielzhe:
@yesamer Did you investigate it further? It may not be that simple to found the differences between both panels, as we talked yesterday in private.

yesamer:
@danielzhe Unfortunately I didn't investigate deeper on it. Considering I don't want to block this PR, please continue to review and test it. Althought I don't expect to have an impact on test scenario module, please ask our QE colleague to include it in the manual testing process as well.
Thanks!

danielzhe:
@kiegroup/gatekeepers Can someone merge this too, please? Notice that this is part of an ensemble.

PR #3137, opened 2020-01-31
DROOLS-4969 DMN v1.3 instanceof support itemDef resolution

tarilabs:
Refs https://github.com/kiegroup/drools/pull/2741


tarilabs:
not sure the real need to use the parser, but this reflects signature changes in upstream.

danielzhe:
I approved since it's just a signature change. Waiting for the real hard code review in https://github.com/kiegroup/drools/pull/2741.

PR #3146, opened 2020-02-06
DROOLS-5017: [DMN Designer] Validation issues are not shown in Alerts panel

manstis:
See https://issues.redhat.com/browse/DROOLS-5017

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/validation/canvas/CanvasDiagramValidator.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/validation/canvas/CanvasDiagramValidator.java
@@ -17,7 +17,9 @@
 package org.kie.workbench.common.stunner.core.client.validation.canvas;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.enterprise.context.Dependent;
@@ -76,11 +78,11 @@ public class CanvasDiagramValidator<H extends AbstractCanvasHandler> {
 
         // See {@link Stream#anyMatch(Predicate)}. "...If the stream is empty then {@code false} is returned..."
         // Therefore we also need to check if there are any violations at all.
-        final boolean valid =
-                getElementViolationsStream(elementViolations).count() == 0 ||
-                        getElementViolationsStream(elementViolations)
-                                .map(v -> applyViolation(canvasHandler, v))
-                                .anyMatch(Boolean.FALSE::equals);
+        final Stream<ElementViolation> violationsStream = getElementViolationsStream(elementViolations);
+        final List<ElementViolation> violationsList = violationsStream.collect(Collectors.toList());
+        final boolean valid = violationsList.size() == 0 || violationsList
+                .stream()
+                .noneMatch(v -> applyViolation(canvasHandler, v));
 
         if (valid) {
             validationSuccessEvent.fire(new CanvasValidationSuccessEvent(uuid, name, title));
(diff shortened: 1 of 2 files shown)


manstis:
When validating a DMN file with WARN (and INFO) violation messages none were being shown in the Alerts panel and Stunner reported validation was successful. On examination use of `anyMatch(Boolean.FALSE::equals)` in the old code lead to `valid` being `true` IF there was at least one INFO violation (even though there were ERRORs or WARNs too).

manstis:
@romartin @jomarko Ready for your reviews, please.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/validation/canvas/CanvasDiagramValidator.java
```diff
@@ -76,11 +78,11 @@ private void checkViolations(final H canvasHandler,
...
-        final boolean valid =
-                getElementViolationsStream(elementViolations).count() == 0 ||
-                        getElementViolationsStream(elementViolations)
-                                .map(v -> applyViolation(canvasHandler, v))
-                                .anyMatch(Boolean.FALSE::equals);
+        final Stream<ElementViolation> violationsStream = getElementViolationsStream(elementViolations);
+        final List<ElementViolation> violationsList = violationsStream.collect(Collectors.toList());
+        final boolean valid = violationsList.size() == 0 || violationsList
+                .stream()
+                .noneMatch(v -> applyViolation(canvasHandler, v));
```
jomarko: Javadoc of noneMatch
```
Returns whether no elements of this stream match the provided predicate. May not evaluate the predicate on all elements if not necessary for determining the result. If the stream is empty then true is returned and the predicate is not evaluated.
```

Due to above I think we can simplify as
```java
        final boolean valid = violationsList
                .stream()
                .noneMatch(v -> applyViolation(canvasHandler, v));
```
WDYT?

manstis: Good catch. Updated in new commit (and removed old comment about use of `anyMatch(..)`).

manstis:
@romartin Could you please try to review when you get a chance.

romartin:
Just to double check, as this change is on the "generic" side, maybe it applies to BPMN as well..  so do you @LuboTerifaj want to be a reviewer as well?

jomarko:
@manstis I can confirm the amount of shown messages is increased, however found this, related to parallel work, please tell me if I should report new jira:

# Start business-central
# Open MySpace
# Add project **a-project**
# Add asset DMN **a-model**
# Add a Decision node **A-Decision**, do not specify its expression
# Save the **a-model**
# Return to MySpace
# Add project **b-project**
# Add asset DMN **b-model**
# Add a Decision node **B-Decision**, do not specify its expression
# Save the **b-model**

The issue is if you save or validate **b-model** while previously **a-model** was created, alerts panel contain messages about both models.

Screenshot from 2020-02-14 11-39-01

manstis:
Hi @jomarko can you please make a new JIRA as it is unrelated to this PR.

Most assets (GRE, GDT, DRL etc) are validated as part of the project build process in Business Central and the validation results of ALL assets are shown in the Alerts Panel. DMN, BPMN and CM are currently different and are only validated when saved (but not as a result of an incremental build as with other assets) or the Validate button clicked.

I therefore suspect, for consistency with the whole of Business Central, DMN, BPMN and CM should be validated as part of the build process in Business Central and validation issues for all DMN/BPMN/CM files would be in the Alerts Panel along with those for all other assets. The validate button in DMN, BPMN and CM would also need changing to operate the same as other assets (that show issues in a popup) rather than reporting to the Alerts Panel...

So, in short, we'd need to discuss with @romartin too.

jomarko:
@manstis before I report the jira, I am not sure if you understood my description of the issue. To me seems an issue that alerts contains issues of both projects, while you , if I understand correctly, says the validation need to be bind on build button, not on validate button?

manstis:
Hi @jomarko I'm glad my previous comment was clear  ;-)

ALL other ASSETS (GDT, GRE, DRL etc) are compiled when the project in Business Central is built and validation messages for them collected and shown in the Alerts panel. When a file is saved an incremental compilation of the project occurs the validation messages may change and the Alerts panel updated.

ALL other EDITORS (GDT, GRE, DRL etc) have a validate button that shows validation errors in a popup.

DMN, BPMN, CM do not add their validation messages to the those gathered when the project is built in Business Central. Consequentially the Alerts panel does not show any issues for these assets when a project is built. Instead Roger decided to have the validate button report them in the Alerts panel (as opposed to the same popup as used by all other assets).

Therefore, in my opinion, DMN, BPMN and CM should be brought inline with all other assets and have their validation messages added to the collection gathered when a project is built that are then shown in the Alerts panel. The validate button being changed to show results in the same popup as other editors. 

- The Alerts panel should show ALL compilation and validation messages for ALL assets.
- DMN, BPMN and CM should use the same in-editor validation popup as other editors.

jomarko:
The found issue reported as https://issues.redhat.com/browse/RHDM-1243

manstis:
@LuboTerifaj Did you want to review?

LuboTerifaj:

PR #3155, opened 2020-02-12
DROOLS-5053: Remove unused dependencies of kie-wb-common-server-ui-cl…

jomarko:
…ient module

For more details see https://issues.redhat.com/browse/DROOLS-5053

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/pom.xml
+++ b/kie-wb-common-screens/kie-wb-common-server-ui/kie-wb-common-server-ui-client/pom.xml
@@ -86,10 +86,6 @@
       <artifactId>uberfire-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.uberfire</groupId>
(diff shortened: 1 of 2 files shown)


jomarko:

jomarko:

mareknovotny:
Why only you do clean up of deps, @jomarko ? ;) I would expect that people who will change something in the code and this is projected into not needed dependency will do it in related PR.
Or if we do PR review we should probably check that there and to not bring new dependency which is not needed.
I get only if this is cleaning procedure for very long existing dependency. WDYT?

mareknovotny:
There is 
`java.lang.ClassCastException: org.kie.dmn.feel.lang.impl.MapBackedType cannot be cast to org.kie.dmn.feel.lang.types.GenListType`
in all failed tests in kie-wb-common screens module tests

mareknovotny:
ok, noticed there is fising PR https://github.com/kiegroup/drools/pull/2765 for this issue

mareknovotny:

jomarko:
@mareknovotny is green now

PR #3158, opened 2020-02-13
KOGITO-989: VSCode - BPMN modeler can not delete global variables

handreyrc:

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/variablesEditor/VariablesEditorWidget.html
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/variablesEditor/VariablesEditorWidget.html
@@ -59,11 +59,11 @@
                                 <input type="text" class="form-control" data-field="customDataType"/>
                             </td>
 
-                            <td style="width:4%">
-                                <input type="checkbox" class="center-block" data-field="kpi"/>
+                            <td style="width:4%" data-field="kpiTD">
+                                <input type="checkbox" style="margin-left:15px; margin-top:10px" data-field="kpi"/>
                             </td>
 
-                            <td style="width:4%" data-field="kpiTD">
+                            <td style="width:4%">
                                 <button type="button" class="center-block" data-field="deleteButton"></button>
                             </td>
                         </tr>


handreyrc:
@romartin, @inodeman Could you please review this PR?

romartin:

handreyrc:
Hey @domhanak, in this case the problem was a misplaced  "data-field" in a <TD>. it was causing the "Delete" button to be removed from screen. It was a punctual HTML adjustment only. 
This other issue seems something else with the innerworks of the widget now that the "Delete" button is back in place. 

@domhanak, @romartin Should we open a new issue or treat it as part of the problem?

domhanak:
Hi @handreyrc if you have some idea about the root cause, please report it. I am not sure I have enough info. Retesting the delete functionality...

domhanak:

handreyrc:
Hey @domhanak, it is curious but the full downstream build is not running. I suspect it doesn't run when we have changes only in HTML/CSS.  I tryied to run it before too and nothing happens.

domhanak:
@handreyrc  Yea it might be the case, rechacking.

domhanak:
Not working properly:
1. Open a BPMN process (Start -> UserTask -> End) + 2 global variables
2. Delete one global variable
3. Press Ctrl+S and close the file
4. Reopen the file again and check global variable; there should be one but there is no variable

romartin:
Hey @domhanak @handreyrc 

Thanks guys for your updates. As already commented before, I would propose:

* About the issue with Ctrl+S -> We think it's not related to this PR, so maybe it can be reported as a new ticket. There are other issues already reported which relate to the save feature in VSCode, and this looks a similar one. We're starting to work on all those issues during this sprint, so I would propose to just report this and check the bug in a few days, where it will probably be fixed by this work.  WDYT @domhanak ?

* About the issue with the variable definition -> can you please verify @domhanak if tihs only happens for kogito? Is it relevant for BC? Anyway notice gyus that Global Variables are NOT used actually for the kogito runtimes, so even the editor provides the ability to manage Global Variables for the process, those will have no effect. SO if this works on BC, but fails only for kogito, I would just report the issue and not spend time now on it, they're "unuseful" anyway at that point

WDYT guys?

domhanak:

domhanak:
@romartin Why is this issue marked as KOGITO issue when it does not make sense for it? Am I missing something?

domhanak:
Anyway business central works as expected. Please change the description of the PR to match issue ticket for BC, this is confusing.

handreyrc:
> This does not solve the issue for KOGITO. When global variable is added it won't be deleted for real, only visually.

handreyrc:
@domhanak I've checked this problem on kogito, and we have other issue taking place here. We don't have a flush data in the forms yet (however it's being implemented @romartin ). it means that to apply changes to the model, the component must lose focus. If you just click on any other field before cheking the xml the changes will be there.
@domhanak  Please, could you check it again?

romartin:
@domhanak this issue applies to both BC and kogito, so it should be working fine on both sides. Please take above considerations from @handreyrc in mind, let's see if you can make it work. Otherwise please let us know!
Thx!

domhanak:
@handreyrc Okay, I tested losing focus helped. Could you please share the issue @romartin is working on?

domhanak:
Ok. looks like it was shown because editor was still loading a file and I was already modifying Global Variables and there was one variable already defined. @handreyrc @romartin Ill try to reproduce further and report it, but it should be unrelated to this.

PR #3163, opened 2020-02-17
KOGITO-841: Documentation Links popup is not readable in DMN VS Code Extension

danielzhe:
This fix also will impact all popups, which I think it is a good thing.


danielzhe:
I noticed that `AbstractPopoverViewImpl` doesn't have any test at all. Do you think I should cover the class with new tests? I think this class will be deprecated soon since it uses deprecated `org.jboss.errai.common.client.dom.Div`, so I didn't do it because I don't know if this effort is valid now. @manstis

jomarko:

danielzhe:
@manstis I decided to make the test for `AbstractPopoverViewImpl` with a small refactor, since  `NameAndUrlPopoverViewImplTest.testShow()` would required some refactors too, since it calls `superShow()` and we don't have access to the options there.

jomarko:
### Manual tests
- [ ] Firefox
- [ ] Chrome
- [ ] VS Code

PR #3174, opened 2020-02-20
DROOLS-5097: [DMN Designer] BKM Description max length

manstis:
See https://issues.redhat.com/browse/DROOLS-5097


jomarko:

manstis:
@jomarko It looks like the line-breaks, whilst present in the _documentation_ are not being rendered (I suspect we simply need to change `\r` to be `<br/>` in the HTML). A new JIRA would be nice as the problem exists before the change to use a `TextArea`. Thanks.

Screenshot from 2020-02-25 12-35-39

jomarko:
@manstis done, see your inbox.

PR #3188, opened 2020-02-26
KOGITO-1219: Forms - Ability to flush the form state into the model on demand

pefernan:
Adding to `DynamicFormRenderer` the ability of flushing changes on UI to the model on demand.

@adrielparedes @caponetto  would you mind take a look?

### DIFF
--- a/kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/main/java/org/kie/workbench/common/forms/dynamic/client/DynamicFormRenderer.java
+++ b/kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/main/java/org/kie/workbench/common/forms/dynamic/client/DynamicFormRenderer.java
@@ -206,7 +206,7 @@ public class DynamicFormRenderer implements IsWidget,
     }
 
     public Object getModel() {
-        if (formHandler != null) {
+        if (isInitialized()) {
             return formHandler.getModel();
         }
         return null;
@@ -245,6 +245,12 @@ public class DynamicFormRenderer implements IsWidget,
         }
 
         return null;
+    }
+
+    public void flush() {
+        if(isInitialized()) {
+            formHandler.maybeFlush();
+        }
     }
 
     public boolean isInitialized() {
--- a/kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/main/java/org/kie/workbench/common/forms/dynamic/client/DynamicFormRenderer.java
+++ b/kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/main/java/org/kie/workbench/common/forms/dynamic/client/DynamicFormRenderer.java
@@ -248,7 +248,7 @@ public class DynamicFormRenderer implements IsWidget,
     }
 
     public void flush() {
-        if(isInitialized()) {
+        if (isInitialized()) {
             formHandler.maybeFlush();
         }
     }
--- a/kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/FormHandler.java
+++ b/kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/FormHandler.java
@@ -138,4 +138,7 @@ public interface FormHandler<T> {
     void setReadOnly(boolean readOnly);
 
     Form getForm();
+
+    void maybeFlush();
+
 }
--- a/kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/FormHandler.java
+++ b/kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/FormHandler.java
@@ -17,7 +17,6 @@
 package org.kie.workbench.common.forms.processing.engine.handling;
 
 import org.jboss.errai.databinding.client.api.Converter;
-import org.jboss.errai.databinding.client.api.DataBinder;
 
 /**
  * Provides API to handle the input changes allowing to add callbacks and run
@@ -26,41 +25,17 @@ import org.jboss.errai.databinding.client.api.DataBinder;
 public interface FormHandler<T> {
 
     /**
-     * Sets Up the FormHandler for the given DataBinder. It configures all the field
-     * callbacks required to process the field changes but doesn't bind the field
-     * Widgets to the DataBinder. Binder mustn't be null.
-     * 
-     * @param binder
-     *            The binder, must not be null.
-     */
-    void setUp(DataBinder<T> binder);
-
-    /**
-     * Sets up the FormHancler for the given DataBinder. It configures all the field
-     * callbacks required to process the field changes and binds the field Widgets
-     * to the DataBinder depending on the bindingInputs param.
-     * 
-     * @param binder
-     *            The binder, must not be null.
-     * @param bindInputs
-     *            Determines if the form widgets must be binded to the DataBinder
-     *            when registered or not,
-     */
-    void setUp(DataBinder<T> binder, boolean bindInputs);
-
-    /**
      * Sets up the FormHandler for the given model, Creates a DataBinder instance
      * for the given model, configures all the field callbacks required to process
      * the field changes and binds the field Widgets to the DataBinder
-     * 
-     * @param model
-     *            The form model, it must not be null.
+     *
+     * @param model The form model, it must not be null.
      */
     void setUp(T model);
 
     /**
      * Retrives the model used on the Form
-     * 
+     *
      * @return
      */
     T getModel();
@@ -71,9 +46,8 @@ public interface FormHandler<T> {
      * depending on how the FormHandler has been setUp.
      * <p>
      * Any of the setUp method's must be executed before register any FormField
-     * 
-     * @param formField
-     *            The FormField, it must not be null.
+     *
+     * @param formField The FormField, it must not be null.
      */
     void registerInput(FormField formField);
 
@@ -83,24 +57,22 @@ public interface FormHandler<T> {
      * depending on how the FormHandler has been setUp.
      * <p>
      * Any of the setUp method's must be executed before register any FormField
-     * 
-     * @param formField
-     *            The FormField, it must not be null.
-     * @param converter
-     *            The value converter used in data binding, if necessary
+     *
+     * @param formField The FormField, it must not be null.
+     * @param converter The value converter used in data binding, if necessary
      */
     void registerInput(FormField formField, Converter converter);
 
     /**
      * Validates all the form fields.
-     * 
+     *
      * @return
      */
     boolean validate();
 
     /**
      * Validates a specific Field of the form.
-     * 
+     *
      * @param propertyName
      * @return
      */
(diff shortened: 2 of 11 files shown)


pefernan:

romartin:
Hey @pefernan 

There is something still not clear for me, rather than the impl details. Lemme ask - once calling `flush`, for instance from Stunner, how the model will get updated? Same listeners will be called to update the field as once a change in being done from the UI? If so, should we expect multiple calls on that listeners just after calling `flush` (eg: once per field)? 

thanx!!

pefernan:
@romartin exactly, the mechanism uses the same listeners but there will be only one call to update the modified field. So no multiple call updates to the model. If you want I can provide a war that triggers a flush every 30 seconds so you can give it a try. WDYT?

romartin:
@pefernan 

I mean, actually each time a single field changes, [this callback](https://github.com/kiegroup/kie-wb-common/blob/f15bf48a4c96a0e21687ad62f83899203c46b008/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/widgets/FormPropertiesWidget.java#L256) is being fired, but noticed it just manages a single field update, not several. So once flushing, not sure if actually is possible, but maybe makes sense considering several fields could fire the change?

jomarko:

kie-wb-common-forms/kie-wb-common-dynamic-forms/kie-wb-common-dynamic-forms-client/src/main/java/org/kie/workbench/common/forms/dynamic/client/DynamicFormRenderer.java
```diff
@@ -247,6 +247,12 @@ public Form getCurrentForm() {
         return null;
     }
 
+    public void flush() {
+        if(isInitialized()) {
```
jomarko: minor: official kiegroup formatting would be: `if (isInitialized()) {`

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/FormHandler.java
```diff
@@ -138,4 +138,7 @@
     void setReadOnly(boolean readOnly);
 
     Form getForm();
+
+    void maybeFlush();
```
jomarko: Could we add javadoc as it is an interface?

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/NeedsFlush.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.forms.processing.engine.handling;
+
+public interface NeedsFlush {
+
+    void flush();
```
jomarko: Wr could add short java doc, maybe not what is `flush` but what classes are intended to implement it ?

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImpl.java
```diff
@@ -37,6 +37,7 @@
 import org.kie.workbench.common.forms.processing.engine.handling.FormHandler;
 import org.kie.workbench.common.forms.processing.engine.handling.FormValidator;
 import org.kie.workbench.common.forms.processing.engine.handling.IsNestedModel;
+import org.kie.workbench.common.forms.processing.engine.handling.NeedsFlush;
 
 public class FormHandlerImpl<T> implements FormHandler<T> {
```
jomarko: Not sure where to comment, but we should simplify the `setUp` variants to have just two (because of the passed values):

```java
@Override
    public void setUp(DataBinder<T> binder) {
        Assert.notNull("DataBinder cannot be null",
                       binder);

        clear();

        this.form = new Form();

        this.binder = binder;
    }

    @Override
    public void setUp(T model) {
        Assert.notNull("Model cannot be null",
                       model);

        setUp(getBinderForModel(model));
    }

```

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImpl.java
```diff
@@ -140,17 +144,26 @@ public void registerInput(FormField formField,
...
             unsubscribeHandlers.add(unsubscribeHandle);
         }
     }
 
+    protected void notifyFieldChange(String fieldName, Object newValue) {
+        fieldChangeManager.notifyFieldChange(fieldName, newValue);
+    }
+
+    protected void processFieldChange(FormField formField, Object newValue) {
+        if(enabledOnChangeValidations) {
```
jomarko: minor: official kiegroup formatting is: `if (enabledOnChangeValidations) {`

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImpl.java
```diff
@@ -239,7 +252,51 @@ public void clear() {
...
+        form.getFields().stream()
+                .filter(formField -> formField.getWidget() instanceof NeedsFlush)
+                .map(formField -> (NeedsFlush) formField.getWidget())
+                .forEach(NeedsFlush::flush);
+    }
+
+    @Override
+    public void maybeFlush() {
+
+        if(binder == null) {
```
jomarko: minor: kiegrooup formatting

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImpl.java
```diff
@@ -239,7 +252,51 @@ public void clear() {
...
+
+            binder.setModel(model, StateSync.FROM_UI, true);
+
+            if (!validate()) {
+                binder.setModel(backupModel, StateSync.FROM_MODEL, true);
+                validate();
+            }
+        } finally {
+            setEnabledOnChangeValidations(true);
+        }
```
jomarko: no methods seems be throwing exception, can't be as below?:
```java
            flushNestedForms();

            BindableProxy<T> proxy = (BindableProxy<T>) getModel();

            T model = proxy.deepUnwrap();

            T backupModel = proxy.deepUnwrap();

            binder.setModel(model, StateSync.FROM_UI, true);

            if (!validate()) {
                binder.setModel(backupModel, StateSync.FROM_MODEL, true);
                validate();
            }
        
            setEnabledOnChangeValidations(true);
```

jomarko:
Found unused dependencies, will handle in separate jira https://issues.redhat.com/browse/JBPM-9042

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/NeedsFlush.java
pefernan: Basically it should be implemented by nested forms. Nested forms have their own renderer and need to propagate the flush call. I decided to have a separate interface instead of reusing `IsNestedModel` just in case we need that functionallity for other custom form controls (stunner have some dedicated to the propreties panels)

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImpl.java
pefernan: Well, there might be cases that Exceptions could be thrown during the form validation, kind of unlikely but possile. That's why I added the _try-finally_ to make sure the `FormHandler` goes back to the normal state

kie-wb-common-forms/kie-wb-common-forms-commons/kie-wb-common-forms-processing-engine/src/main/java/org/kie/workbench/common/forms/processing/engine/handling/impl/FormHandlerImpl.java
pefernan: Let me take a look, honestly I'm not sure here. The renderer is being used in two different ways (static models like stunner properties panels & dynamic models on jbpm task list) that's the reason for both setup methods, but I cannot remember the details now.

pefernan: Ok, I think the `setUp` receiving a binder is legacy code from the old LiveSpark days... removing it. Thanks for the pointer.

manstis:
I'm not going to review but I'd like to understand the reasoning for this PR and how it's to be invoked (and how it may affect DMN) as the JIRA doesn't really say much (and I assume there's more _integration_ work with Stunner to follow).

Having read both the JIRA and nosed through this PR it appears that the Form(s) could contain updated state not in the (BPMN) model and therefore you want to programmatically flush the Form(s) content into the model.

This implies that the mere process of rendering a Form (for, I assume, the Diagram as this would be "open" by default when opening a BPMN process) is setting some domain properties and these need to be flushed back to the model.

If true, that all sounds mighty crazy.

pefernan:
@manstis the point is that forms are propagating the property changes into the model when they detect a change event fired by the widget (_errai-data-binding_ world!), on textboxes/text areas this happens when the widget looses the focus. The _big_ issue is that users on vscode can make changes on the properties panel and save the diagram without making the change event to happen. 

This flush mechanism is only to force a flush to happen BEFORE the save on vscode is finally executed. So any pending change will be pushed into the model prior to saving the diagram.

manstis:
@pefernan Ah, OK... so clicking VSCode's "Save" button (or keyboard shortcut) is not triggering the `blur/change` events on our HTML. Understood. Thanks.

pefernan:

pefernan:
jenkins execute full downstream build please

romartin:
@manstis mmm your Qs above meant that the requirement I wrote in [KOGITO-1219](https://issues.redhat.com/browse/KOGITO-1219) was not really clear... :) So I just updated the description, thanks for letting us know, and yeah feel free to complain when something is not clear! :+1:

romartin:
Approving it, althoguh I did not tested the integration with any of the BPMN/DMN editors. I expect tests ensuring the callbacks are properly working for changed properties, once  flushing. 

So anyway let's keep in touch @pefernan , once starting the integration with the edtiors will let you know!

Thanks, great job!

PR #3191, opened 2020-02-27
RHPAM-2748 :Update locale selection drop down to include only the sup…

RishiRajAnand:
- Removed the code that includes non-translated locales

Part of an ensemble:
- https://github.com/kiegroup/kie-wb-common/pull/3191
- https://github.com/kiegroup/kie-wb-distributions/pull/1018


RishiRajAnand:
I have just removed the locales names from the Constants.java, but haven't removed the .properties file as of now. @dgutierr Do we need to remove them as well ?

manstis:
@RishiRajAnand Thank-you.. IMO they should be removed from (at least) `CommonConstants.properties` as this will be used by our i18n team to determine the translations needed. If the entry remains in that file, they will continue to translate it. If it is removed from that file they will not continue to translate it!

RishiRajAnand:
@manstis Makes sense, removed the keys from CommonConstants.properties.

RishiRajAnand:
Jenkins retest this please

barboras7:

mareknovotny:
we need also backport of these PRs to 7.33.x today

RishiRajAnand:
@mareknovotny Sure, will do

barboras7:
Hi,
I tried to put some of the previously available locales into the URL. See the behavior on the screenshot. Is this what we want to happen? (See the German words mixed with English
Screenshot from 2020-02-28 13-46-18
)

manstis:
Thank-you for the check @barboras7.

Firstly, IIUC there should no longer be ANY German translations. Can you confirm you used the WAR generated from [this](https://github.com/kiegroup/kie-wb-distributions/pull/1018/) PR? Secondly, it can be _normal_ that missing translations for a locale fallback to English. So that alone is less of a worry (than there being ANY German translations!)

RishiRajAnand:
@manstis @barboras7 I am looking into it now, Looking like I missed a change somewhere

barboras7:
@manstis yes, I am using the correct .war file, I just double checked. 
@RishiRajAnand sure. Otherwise, it seems to work correctly, the above mentioned is my only concern

RishiRajAnand:
@manstis This behavior is explained in https://issues.redhat.com/browse/ERRAI-1101, the description clearly states the following 
"With locale=de query param
This is the case when using a locale that has a translation in the properties bundle but wasn't added to the locale property in GWT module. UI code that uses TransationService displays German translation and fallback for untranslated keys works." 

So looks like we have to remove the .properties file too make it work. @manstis wdyt ?

manstis:
@RishiRajAnand 
> So looks like we have to remove the .properties file too make it work. @manstis wdyt ?

i18n never fails to surprise me! I agree.. it looks like the now sprurious files need removing too.

manstis:
@RishiRajAnand Please see https://github.com/kiegroup/kie-wb-common/pull/3194. Perhaps @mbiarnes has already done a clean-up and, if @barboras7, rebuilds the FBD the new WAR will be OK.

mbiarnes:
Jenkins execute full downstream

RishiRajAnand:

mareknovotny:
I will merge this as the kiegroup/kie-wb-distributions#1018 is merged and 7.33.x branch PRs were merged too.
if there will be any issue with this change it will affect 7.33.x branch too, so we need to have fix on both.

PR #3196, opened 2020-02-27
DROOLS-3648: [DMN Designer] Boxed expressions - Decision Table - Allow Literal Expressions in the Decision Table header

manstis:
See https://issues.redhat.com/browse/DROOLS-3648

This PR refactors the generic popover to support:
- Different _values_ (currently `Name` or `Text`) in the Value/TypeRef popover
- The ability to provide different _labels_ for the _value_ field in the popover
- The ability to clean/normalise values entered by Users
- Refactoring of how the popover title is provided 

Tests and JavaDocs are being added before this PR moves from _draft_.


manstis:
@jomarko Absolutely :-)

Before this PR the "Input **expression**" was treated the same as "Name".

However this PR corrects the handling of "Input Clause" expressions and preserves the entered text as typed by the user with no trimming on normalising/sanitising. Who knows _what_ the expression should be.. so best to leave it exactly as typed (if you know 100% what sanitising we _could_ do I'm happy to add.. but atm it purposefully does no cleaning).

manstis:
@danielzhe Could you please review (before it gets stale)?

PR #3208, opened 2020-03-04
KOGITO-1337: [DMN Designer] Author and filename are always non defined

manstis:
See https://issues.redhat.com/browse/KOGITO-1337

The DMN _model name_ was already in the `DMNDocumentation` class, so I just removed `fileName`.

I've also [moved](https://github.com/kiegroup/kie-wb-common/pull/3208/files#diff-2beebe31108b6489cb3b435185072878) the existing _kogito_ version of `DMNDocumentationViewButtonsVisibilitySupplier` (and test) to the same sub-package path as the one it replaces for kogito for consistency. So there are a couple of other (sort of) related changes in this PR.

**TO TEST**

- Compile `kie-wb-common-dmn`
- Run `kie-wb-common-dmn-webapp-standalone`
- The "Documentation" tab will show both model name (Generated from) and user name (Generated by). The user name will be ANONYMOUS. If you run `drools-wb-webapp` you'll see a _real_ user name.
- Run `kie-wb-common-dmn-webapp-kogito-testing`
- The "Documentation" tab will show only model name (Generated from).

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/common/DMNDocumentationFactory.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/common/DMNDocumentationFactory.java
@@ -29,14 +29,11 @@ import org.kie.workbench.common.dmn.client.graph.DMNGraphUtils;
 import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
 import org.kie.workbench.common.stunner.core.client.canvas.util.CanvasFileExport;
 import org.kie.workbench.common.stunner.core.diagram.Diagram;
-import org.kie.workbench.common.stunner.core.diagram.Metadata;
-import org.uberfire.backend.vfs.Path;
 import org.uberfire.client.views.pfly.widgets.Moment;
 import org.uberfire.rpc.SessionInfo;
 
 import static org.kie.workbench.common.dmn.client.resources.i18n.DMNEditorConstants.DMNDocumentationFactory_Constraints;
 import static org.kie.workbench.common.dmn.client.resources.i18n.DMNEditorConstants.DMNDocumentationFactory_ListYes;
-import static org.kie.workbench.common.dmn.client.resources.i18n.DMNEditorConstants.DMNDocumentationFactory_NoFileName;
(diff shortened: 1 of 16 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/documentation/common/DMNDocumentationFactory.java
```diff
@@ -81,31 +77,25 @@ public DMNDocumentation create(final Diagram diagram) {
                                        getDocumentationI18n());
     }
 
-    private List<DMNDocumentationDRD> getDrds(final Diagram diagram) {
+    protected List<DMNDocumentationDRD> getDrds(final Diagram diagram) {
```
manstis: I changed the methods used by the `DMNDocumentation.create(...)` method to `protected` to allow them to be overridden more easily in the future.. if ever needed. At the moment I only override `getCurrentUserName()`.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editors/documentation/common/DMNDocumentationFactory.java
```diff
@@ -0,0 +1,51 @@
...
+        super(canvasFileExport,
+              translationService,
+              drdsFactory,
+              sessionInfo,
+              graphUtils);
+    }
+
+    @Override
+    protected String getCurrentUserName() {
+        return null;
```
jomarko: Due to change like on this line, I would expect other place in PR doing:
```
if (currentUserName == null) {
  // ...
}
```

Not needed?

manstis: We _sort of_ [do](https://github.com/kiegroup/kie-wb-common/pull/3208/files#diff-330975457b3de47273401f558e527cebR285).... Mustache (the template library used to generate the documentation) has some _lovely_ [syntax](v) to handle `null`.

danielzhe:
Very l337 the JIRA number.

danielzhe:
@manstis Sorry, it was just a joke. I'm reviewing right now.

PR #3210, opened 2020-03-04
DROOLS-5125: [DMN Designer] Node data type is lost after drag and drop in data type editor

danielzhe:
See: [https://issues.redhat.com/browse/DROOLS-5125](https://issues.redhat.com/browse/DROOLS-5125)

Some context of this change:
During the drag and drop, if you drag `typeB` over `typeA`, `typeB` is destroyed and recreated in the new position. The issue is that if `typeB` is selected as type output in some node, its selection is lost when `typeB` was destroyed because properties panel is notified that `typeB` doesn't not exist anymore and should be replaced by `<Undefined>`
With changes proposed in this PR, the properties panel will be not notified  when the custom data type is deleted and recreated **during a drag and drop operation**.
The properties panel will only be notified when the data type is deleted for real, and not to be replaced later.

### DIFF
--- a/kie-wb-common-ala/kie-wb-common-ala-build-maven/pom.xml
+++ b/kie-wb-common-ala/kie-wb-common-ala-build-maven/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>org.kie.workbench</groupId>
     <artifactId>kie-wb-common-ala</artifactId>
-    <version>7.34.0-SNAPSHOT</version>
+    <version>7.35.0-SNAPSHOT</version>
   </parent>
   <artifactId>kie-wb-common-ala-build-maven</artifactId>
   <name>KIE Workbench Common ALA :: Build Maven</name>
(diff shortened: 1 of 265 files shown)


manstis:
@danielzhe I'll take a look when @jomarko findings are resolved.

danielzhe:
@jomarko Thank you! Yes, I think I can fix this other issue in this PR too. I'll let you know when is done.

danielzhe:
@jomarko Changes applied. :)

jomarko:
@danielzhe thank you, I can confirm the corner case I reported is fixed now if user just move the custom data type. However if user selected  a node of a custom type (e.g. **tCustom**), then renamed the **tCsutom** to **tCustomRenamed**, returned to main editor tab, the selected node will have its output reference set as **Nothing Selected**. When user re-select the diagram node, it has correct reference to **tCustomRenamed**

Please let me know if this can be incorporated as part of this PR or we need separate.

danielzhe:
@jomarko I think we can handle this in another JIRA, since it only happens in very specific cases. I noticed it only happens if you rename 2 or more nodes. Wdyt?

danielzhe:
@yesamer This is the fix.

danielzhe:
@manstis You can re-review this when you want. :)

manstis:
@danielzhe I think we have sufficient review of this PR, so removed myself

manstis:
I looked at SonarCloud's test coverage analysis and I think we're good to go...

PR #3204, opened 2020-03-02
KOGITO-1014: DMN editor doesn't show Ace editor on invalid DMN content

manstis:
See https://issues.redhat.com/browse/KOGITO-1014

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/917
- https://github.com/kiegroup/kie-wb-common/pull/3204

PR #3216, opened 2020-03-11
KOGITO-437: Revert of change made to DMN since editor does not open in VSCode with change

manstis:
Hi @handreyrc with your change for the captioned JIRA the DMN editor no longer opens (any file) in VSCode. Removal of the line in this PR allows the DMN editor to open again however whatever the issue was for KOGITO-437 possibly now exists again in DMN too.

You may need to reconsider your fix for KOGITO-437. Did you want the JIRA re-opened?

@romartin @ederign FYI.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editor/AbstractDMNDiagramEditor.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/editor/AbstractDMNDiagramEditor.java
@@ -391,8 +391,8 @@ public abstract class AbstractDMNDiagramEditor extends AbstractDiagramEditor {
 
     @Override
     @SetContent
-    public void setContent(final String path, final String value) {
-        superOnClose();
+    public void setContent(final String path,
+                           final String value) {
         diagramServices.transform(path,
                                   value,
                                   new ServiceCallback<Diagram>() {


manstis:
Closing.. weird.. rebuilt things etc etc to verify my issue however all seems OK now :-)

manstis:
Re-opened as it does appear to break things...

This is what happens with two calls to `setContent(...)`:

1. `setContent("", "")` from `kogito-tooling`
2. `onClose()` is called (by change for KOGITO-437). This calls `AbstractDiagramEditor.doClose()` that calls `AbstractDiagramEditorCore.destroySession()`. However since no session has been created at this stage everything is OK.
3. The first (empty) content is then parsed by `KogitoClientDiagramService.transform(..)` and since it is valid content (eventually) a `ManagedSession` is created that uses `SessionLoader` to initialise itself. `SessionLoader` calls `SessionInitializer` to initialise the session. One of them (the only one?) `LienzoImageStripLoader` makes an **asynchronous** call (follow the code, but it eventually uses `ImageLoader` that loads images asynchronously).
4. Meanwhile the second call to `setContent("", "xml")` is made from `kogito-tooling`.
5. `onClose()` is called (by change for KOGITO-437). This calls `AbstractDiagramEditor.doClose()` that calls `AbstractDiagramEditorCore.destroySession()`. A session however now exists from the first call to `setContent(..)` which leads to `AbstractSessionPresenter.destroy()` to be called, that calls `SessionEditorImpl.destroy()` that sets `diagramSupplier = null`....
6. The **asynchronous** call now completes and the associated callback (setup in `AbstractSessionPresenter.open(..)`) executes. This calls `AbstractSessionViewer.doOpen(..)` that tries to call `getDiagram()` however `diagramSupplier` is `null` (from the first call) and a NPE thrown.

Stack trace:
```
Error caused by: JavaScriptException: (TypeError) : Cannot read property 'get_3_g$' of null
    at ExP_g$.IxP_g$ [as getDiagram_0_g$] (SessionEditorImpl.java:99)
    at ExP_g$.MwP_g$ [as doOpen_1_g$] (AbstractSessionViewer.java:109)
    at ExP_g$.$wP_g$ [as open_15_g$] (AbstractSessionViewer.java:40)
    at ExP_g$.YwP_g$ [as open_13_g$] (AbstractSessionViewer.java:38)
    at myP_g$.ZuP_g$ [as open_21_g$] (AbstractSessionPresenter.java:118)
    at myP_g$.RuP_g$ [as lambda$0_323_g$] (AbstractSessionPresenter.java:111)
    at qwP_g$.rwP_g$ [as accept_4_g$] (AbstractSessionPresenter.java:111)
    at utx_g$ (GlobalSessionManager.java:86)
    at $OP_g$._OP_g$ [as execute_3_g$] (GlobalSessionManager.java:86)
    at w1Q_g$.V1Q_g$ [as lambda$5_40_g$] (ManagedSession.java:181)
    at p3Q_g$.q3Q_g$ [as execute_5_g$] (ManagedSession.java:170)
    at U3Q_g$ (SessionLoader.java:64)
    at v4Q_g$.w4Q_g$ [as execute_3_g$] (SessionLoader.java:63)
    at P3Q_g$.W3Q_g$ [as loadInitializer_0_g$] (SessionLoader.java:98)
    at P3Q_g$.T3Q_g$ [as lambda$1_182_g$] (SessionLoader.java:93)
    at r4Q_g$.s4Q_g$ [as execute_3_g$] (SessionLoader.java:92)
    at ytO_g$.ztO_g$ [as run_2_g$] (LienzoImageStrips.java:68)
    at DO_g$.NO_g$ [as register_1_g$] (ImageStrips.java:56)
    at ZO_g$.$O_g$ [as run_2_g$] (ImageStrips.java:50)
    at bP_g$.cP_g$ [as run_2_g$] (ImageStrips.java:69)
    at iM_g$.kM_g$ [as onImageElementLoad_0_g$] (ImageElementProxy.java:34)
    at iM_g$.bM_g$ [as doImageElementLoadAndRetry_0_g$] (ImageLoader.java:166)
    at oM_g$.pM_g$ [as onLoad_1_g$] (ImageLoader.java:60)  
```
The NPE is being caught by `kogito-tooling`'s `GwtEditorWrapper` [here](https://github.com/kiegroup/kogito-tooling/blob/master/packages/kie-bc-editors/src/GwtEditorWrapper.tsx#L77).

manstis:
@tiagobento FYI.. another issue with calling `setContent(..)` twice in quick succession :-(

manstis:
@jomarko I've put a possible fix in https://github.com/kiegroup/kie-wb-common/pull/3204/commits/f064239e538e17e027d069bbb4c63fb7cacfa9a4. Could you please retest https://github.com/kiegroup/kie-wb-common/pull/3204 and verify it still works!?

romartin:
Hey @manstis @jomarko @ederign @tiagobento @handreyrc 

Once me and @handreyrc were looking at the fix for https://issues.redhat.com/browse/KOGITO-437 we noticed that the `setContent` method was called several times, at least for the GitHub extension, once swicthing between _original_ and _changes_ tabs. This changes the lifecycle for the editors we'd been using until this point - actually the editor just opens once (`setContent`), and save can happen several times (`getContent`). On the other hand, I was assuming that in the GitHub extension, as the webpage is not being reloaded once switching between tabs, it was making sense so start calling `setContent` several times as well.

So we've differnt ways to move forward:
1) Commit this PR and re-do KOGITO-437
2) Test & potentially merge quick fix from @manstis https://github.com/kiegroup/kie-wb-common/commit/f064239e538e17e027d069bbb4c63fb7cacfa9a4

But anyway, I would like to also clarify which is the expected lifecycle for the editors - can we expect several calls to `setContent`/`getContent` during same asset authoring? Also, as @manstis said, why it's DMN's `setContent` called twice on VSCode? Why this doesn't also applies to BPMN?

Thanks guys!

manstis:
Hi @romartin it is `kogito-tooling` that calls `setContent(..)` twice in quick succession when opening a file. Once with empty content and again with the real file content. I've linked the class (on the KOGITO-1014 JIRA IIRC). It is calling it twice in quick succession that causes most (possibly all) of the issues I've been working on due to the asynchronous initialisation of sessions (image loading). Removal of the two calls to `setContent(..)` when opening a file would fix many things.

manstis:
`kogito-tooling` calls `setContent(..)` twice for all editors and presumably all channels.

romartin:
@manstis agree. Also we've to consider the github extension where you can navigate between versions, where setContent then is called even more than twice.

We'd been never using this lifecycle before, about re-using same editor with different contents, so we're now facing issues, and probably more issues will appear.

So it would be easier for the editor side to assume only a single call to setContent, even on different channels or when navigating between versions, but I'm not sure if that's assumible at framework level, WDYT @ederign @tiagobento ??

Thanks guys!

manstis:
@romartin The issues I'm encountering all seem to be caused by [this](https://github.com/kiegroup/kogito-tooling/blob/master/packages/microeditor-envelope/src/EditorEnvelopeController.tsx#L61) in `kogito-tooling` where two calls to `setContent(..)` are made in _quick succession_ (before the asynchronous operation, loading images, for the first call completes).

I've not encountered any issues calling `setContent(..)` numerous times "more slowly" (i.e. at "user speed") performing interactive operations - simply because these happen much slower than the programmatic approach linked above and hence the asynchronous operation has time to complete before subsequent invocations of `setContent(..)`.

manstis:
Closed as I have fixed a number of issues relating to the successive calls to `setContent(..)` when initially opening a file in https://github.com/kiegroup/kie-wb-common/pull/3204. [KOGITO-1366](https://issues.redhat.com/browse/KOGITO-1366) remains to (ideally) remove the need for successive calls.

It's plausible other issues still exist, especially when calling `setContent(..)` again and again interactively with invalid (i.e. XML editor) and then valid (i.e. stunner editor) content but let's tackle those as and when they appear.

tiagobento:
Hi guys, thanks for the detailed conversation. I appreciate your involvement and interest in the 'behind the scenes' of `kogito-tooling`. Indeed, calling `setContent()` in a quick succession is not an ideal solution, but one that we've came up with so that we could have a "predictable" initialisation time for the editor UI. Not making an empty `setContent` call can lead to variable rendering time, due to the differences between diagram complexity and size, and this is not as pleasing for users. Specially on VSCode, where things feel native but are in fact a very fluid web app.

This is a decision we've made when we first prototyped kogito-tooling. However, I understand that now priorities may have shifted. As much as I understand the difficulties that this approach is causing, specially for you, @manstis, I think we should not simply try and remove the "double" `setContent()` call, as it would result in a less fluid experience.

As a suggestion, I would say we could always open the editor with an empty diagram (this would be controlled on the editor side). This way, all the properties panels, docks and whatnot would be properly rendered _before_ the loading screen could disappear. To initialise the envelope lifecycle, we wait for the editor to call a specific function (`appFormerGwtFinishedLoading`), so if we can assume that the editor would be "fully assembled" (i.e. only lacking the diagram) when this function is called, I think we would be fine if we removed the empty setContent call. WDYT?

@romartin @manstis @ederign

manstis:
@tiagobento Instead of first setting empty content and the editor signalling when it's ready to receive the _real_ content (as I think you're proposing?) at which point `kogito-tooling`hides the loading spinner; why not just have `kogito-tooling` provide the real content (with a single call to `setContent(..)`) and the editor signal when it's finished loading so the loading spinner can be hidden? Or provide a callback to the `setContent(..)` method the editor calls to hide the spinner? Or could we have `setContent(..)` return a `Promise` that hides the spinner?

tiagobento:
@manstis We have to know when the editor first finishes loading, because that's when the envelope _asks_ for the real content. When the content comes back (from the channel, to the envelope), we then do the double setContent. If setContent returns a Promise, we could in fact do what you say and hide the Loading Screen once the setContent promise resolves. The only thing to be aware of is that the setContent promise can only resolve when the diagram is fully setup, so users don't see a flicker or a UI shift etc. Do you think it's possible to make setContent return a promise?

romartin:
@manstis @tiagobento thanks for the clarification guys. I'm not against calling multiple times the setContent, but ONLY if that really makes sense to do... for exxample, switching between versions in the github extensions. 

On the other hand, if the current double call being done to setContent is just for managing the lifecycle of the editor (first call for to loading the editor itself, second call for setting the "real" content), I don't think it's the best way to go. Not sure if returning a Promise will really solve your issues, but we can also consider adding another method on the API just for loading the editor (apart from the actual setContent/getContent).

Well guys, even @manstis has been hopefully fixing this in the best way possible, it's potentially probalble to face with more issues related to this, so would appreciate to clarify how the editor will be managed, from the different  channels and also considering BC, and so we can take the necessary actions if necessary, before receving bugs.

Lemme know if I can help!!
Thanks!

romartin:
Hey @tiagobento @manstis 

I'd been testing to call setContent twice, via code as @manstis suggested, and it's also causing problems with the BPMN editor, see https://issues.redhat.com/browse/KOGITO-1425.

So the double setContent calls is also causing issues on VSCode with the BPMN editor. I'll try to fix the issue but I'm not sure if will be easy or even possible. Once calling the setContent for first time the editors are running into some tasks, and aynchronous code gets into the picture too, so it's not save assuming that after that call everything will be ready.

So adding also @ederign and @porcelli to see what to do. I would really appreciate guys if we can get rid of that double consecutive call to `setContent`, it's causing several issues around. Also, it looks only specific to VSCode?? CC @manstis @tiagobento 

thanks guys!

tiagobento:
@romartin @ederign @manstis Guys, just created this JIRA https://issues.redhat.com/browse/KOGITO-1436. Let's continue our discussion there so we don't spread content around in GitHub and email threads. Things usually get lost this way. Thx!

PR #3225, opened 2020-03-17
DROOLS-5170: [DMN Designer] Properties panel can not be expanded

manstis:
See https://issues.redhat.com/browse/DROOLS-5170

This is a patch for the regression added by https://issues.redhat.com/browse/RHPAM-1954

If the powers that be don't want to revert RHPAM-1954 this PR fixes the reported DMN problem.

There may be other, undiscovered, issues... but for now this is the only one I could find (in DMN).


manstis:
https://issues.redhat.com/browse/RHPAM-1954 has been reverted.

This PR is no longer needed.

PR #3227, opened 2020-03-19
DROOLS-5178: [DMN Designer] Generate unnecessary typerRef on output tag

manstis:
See https://issues.redhat.com/browse/DROOLS-5178


manstis:
SonarCloud is complaining about test coverage for the client-side marshaller changes. Unfortunately this is only testable with the new integration test that is not instrumented by SonarCloud to check for coverage.

PR #3230, opened 2020-03-19
DROOLS-5045: [DMN Designer] Marshaller does not support DMN1.2 DecisionTable RuleAnnotation

danielzhe:
See: [https://issues.redhat.com/browse/DROOLS-5045](https://issues.redhat.com/browse/DROOLS-5045)

1. If you open a file with rule that have the obsolete column "Description" it will be converted do the new rule annotation clause.
2. There's nothing in the spec that says that we need at least one rule annotation column, but we decided to use this approach to make it easier to the user, otherwise, when/where/how would we show the context menu to add annotation column?

demo


jomarko:
@danielzhe sonar report 26 code smells, could you please have a look on those reported as critical, they seems be quite useful tips, what do you think?

jomarko:
Also tests needs to be updated

danielzhe:
@jomarko The smells are about the name of the variables, but I'm following the pattern we use. Also there another smells that would need a very big refactor to be fixed, like the use of deprecated classes like HTMLElement, that we use because of our Popover that needs to be refactored, so I think there's nothing to be done now.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/RuleAnnotationClause.java
```diff
@@ -0,0 +1,96 @@
...
+import org.kie.workbench.common.stunner.core.definition.annotation.definition.Category;
+import org.kie.workbench.common.stunner.core.definition.annotation.definition.Labels;
+import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
+import org.kie.workbench.common.stunner.core.factory.graph.NodeFactory;
+
+import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.processing.fields.fieldInitializers.nestedForms.AbstractEmbeddedFormsInitializer.COLLAPSIBLE_CONTAINER;
+import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.processing.fields.fieldInitializers.nestedForms.AbstractEmbeddedFormsInitializer.FIELD_CONTAINER_PARAM;
+
+@Portable
+@Bindable
```
manstis: `@Bindable`, `@Definition` and `@FormDefinition` and `DomainObject` all relate to instances of these classes appearing in the "Properties Panel". However since `DecisionTableGrid.doAfterSelectionChange(..)` and `DecisionTableGrid.doAfterHeaderSelectionChange(..)` are not calling `BaseGrid.fireDomainObjectSelectionEvent(..)` both `RuleAnnotationClause` and `RuleAnnotationClauseText` can have all these annotations/interfaces removed.

IF you wanted `RuleAnnotationClause` to appear in the "Properties Panel" you would keep the annotations/interfaces and change `DecisionTableGrid.doAfterHeaderSelectionChange(..)` to call `fireDomainObjectSelectionEvent(..)` for the `ANNOTATION_CLAUSES` _section_.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DecisionTablePropertyConverter.java
```diff
@@ -39,6 +41,21 @@ public static DecisionTable wbFromDMN(final org.kie.dmn.model.api.DecisionTable
         result.setDescription(description);
         result.setTypeRef(typeRef);
 
+        if (dmn.getRule().stream().anyMatch(r -> !StringUtils.isEmpty(r.getDescription()))
```
manstis: This only adds a `RuleAnnotationClause` IF at least one existing row has a non-empty description.

What happens if a DMN model is imported without any descriptions? We'd have no `RuleAnnotationClause` added and none could be added in the UI (as we need at least one `RuleAnnotationClause` column so that the context menu to add/remove others can be shown). I think we need to create a `RuleAnnotationClause` IF none exist; and copy the _description_ to it (even if empty).

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddRuleAnnotationClauseCommand.java
```diff
@@ -0,0 +1,177 @@
...
+        this.ruleAnnotationClause = ruleAnnotationClause;
+        this.uiModel = uiModel;
+        this.uiModelColumnSupplier = uiModelColumnSupplier;
+        this.uiColumnIndex = uiColumnIndex;
+        this.uiModelMapper = uiModelMapper;
+        this.executeCanvasOperation = executeCanvasOperation;
+        this.undoCanvasOperation = undoCanvasOperation;
+    }
+
+    public Optional<RuleAnnotationClauseColumn> getUiModelColumn() {
```
manstis: Can this be package protected? It's only used by this class and the test.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddRuleAnnotationClauseCommand.java
```diff
@@ -0,0 +1,177 @@
...
+        return uiModelColumn;
+    }
+
+    public String getName() {
```
manstis: Can this be `private` (or at least package protected)? It's only used by this class.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddRuleAnnotationClauseCommand.java
manstis: In fact why do we need it at all!? 

If it's only use is this class or the test, the value is know by the test already as it constructs the instance.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddRuleAnnotationClauseCommand.java
```diff
@@ -0,0 +1,177 @@
...
+        if (StringUtils.isEmpty(name)) {
```
manstis: IIRC there's good reason the other _add_ commands derive the name in the _constructor_ and not at _execution_ time. I also notice that add, add, rename, rename, undo, undo, undo, redo, redo, redo does not appear to work properly when I tried setting an expression to a Decision Table, add Annotation, add Annotation etc in the UI.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableDefaultValueUtilities.java
```diff
@@ -54,6 +59,21 @@ private static int getMaxUnusedInputClauseIndex(final DecisionTable dtable) {
         return maxIndex + 1;
     }
 
+    public static String getNewRuleAnnotationClauseName(final DecisionTable decisionTable) {
```
manstis: Being picky, it'd be nice to have these 2 new methods at the end of the class so that the method order is _input_ related methods, _output_ related methods and _annotations_ related methods.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinition.java
```diff
@@ -73,7 +74,8 @@ public DecisionTableEditorDefinition(final DefinitionUtils definitionUtils,
                                          final TranslationService translationService,
                                          final HitPolicyPopoverView.Presenter hitPolicyEditor,
                                          final ManagedInstance<ValueAndDataTypePopoverView.Presenter> headerEditors,
-                                         final DecisionTableEditorDefinitionEnricher enricher) {
+                                         final DecisionTableEditorDefinitionEnricher enricher,
+                                         final ManagedInstance<ValuePopoverView.Presenter> valueEditors) {
```
manstis: I still wonder whether to have "in cell" editing of names vs popup. WDYT? Did you ask UX?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableUIModelMapperHelper.java
```diff
@@ -84,4 +83,32 @@ public static int getOutputEntryIndex(final DecisionTable dtable,
...
+        }
+        if ((_columnIndex = _columnIndex - outputClauseColumnCount) < 0) {
+            throw new IllegalArgumentException("columnIndex referenced an OutputEntry column. Should be a valid AnnotationEntry column.");
+        }
+        if (_columnIndex > annotationClauseColumnCount - 1) {
+            throw new IllegalArgumentException("columnIndex did not reference a valid AnnotationEntry column.");
+        }
+
+        return _columnIndex;
+//        final int annotationIndex = columnIndex - inputClauseColumnCount - outputClauseColumnCount - 1;
```
manstis: Please delete commented block.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddRuleAnnotationClauseCommand.java
```diff
@@ -0,0 +1,177 @@
...
+            @Override
+            public CommandResult<RuleViolation> execute(final GraphCommandExecutionContext context) {
+                decisionTable.getComponentWidths().add(uiColumnIndex, null);
+
+                final int clauseIndex = getClauseIndex();
+                decisionTable.getAnnotations().add(clauseIndex, ruleAnnotationClause);
+                ruleAnnotationClause.getName().setValue(getName());
+
+                decisionTable.getRule().forEach(rule -> {
+                    final RuleAnnotationClauseText le = new RuleAnnotationClauseText();
```
manstis: Not really a `le` here.. just a copy and paste _artifact_!

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinition.java
danielzhe: Yes, I think you have a good point. I commented in the chat right now to see what UX thinks (I'm replying here too for the case other peoples reads the PR don't thinks that I'm ignoring your suggestion)

jomarko:
I am waiting with review until Michael's points are addressed.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/definition/v1_1/DecisionTablePropertyConverter.java
danielzhe: You're right, I changed the 'DecisionTableEditorDefinitionEnricher' to handle new DecisionTable, but it doesn't handle imported DMN files without description. I'm making all proposed changes including the inline text editor. Thank you!

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddRuleAnnotationClauseCommand.java
danielzhe: We need to test [this](https://github.com/kiegroup/kie-wb-common/pull/3230/files#diff-2b6600ba8944ad9e98f876552ac6ebacR233), because `uiModelColumn` is only known after [this](https://github.com/kiegroup/kie-wb-common/pull/3230/files#diff-4f5d36356c40116ce91684b98b403c78R145).

danielzhe:
Changes applied! I was unable to reproduce the issue you reported about the drag and drop columns. Looks like everything is working now:
changes

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/resources/org/kie/workbench/common/dmn/client/resources/i18n/DMNClientConstants.properties
```diff
@@ -120,6 +126,7 @@ DataTypeListView.ReadOnlyMessage2=You can edit them by opening the original file
 DataTypeListView.ImportDataObject=Import Data Object
 NameAndDataTypePopover.NameLabel=Name
 NameAndDataTypePopover.DataTypeLabel=Data Type
+ValuePopover.ValueLabel=Value
```
manstis: Is this still needed?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/resources/i18n/DMNEditorConstants.java
```diff
@@ -271,6 +289,9 @@
     @TranslationKey(defaultValue = "")
     public static final String NameAndDataTypePopover_DataTypeLabel = "NameAndDataTypePopover.DataTypeLabel";
 
+    @TranslationKey(defaultValue = "")
+    public static final String ValuePopover_ValueLabel = "ValuePopover.ValueLabel";
```
manstis: Is this still needed?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/resources/i18n/DMNEditorConstants.java
```diff
@@ -172,6 +172,9 @@
     @TranslationKey(defaultValue = "")
     public static final String FunctionEditor_SelectFunctionKind = "FunctionEditor.SelectFunctionKind";
 
+    @TranslationKey(defaultValue = "")
+    public static final String DecisionTableEditor_SelectRuleAnnotationName = "DecisionTableEditor.SelectRuleAnnotationName";
```
manstis: Is this still needed?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/resources/i18n/DMNEditorConstants.java
```diff
@@ -172,6 +172,9 @@
...
+
     @TranslationKey(defaultValue = "")
     public static final String DecisionTableEditor_DescriptionColumnHeader = "DecisionTableEditor.DescriptionColumnHeader";
```
manstis: I don't think this is needed either.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/RuleAnnotationClauseText.java
```diff
@@ -0,0 +1,81 @@
...
+import org.kie.workbench.common.stunner.core.definition.annotation.definition.Labels;
+import org.kie.workbench.common.stunner.core.factory.graph.NodeFactory;
+
+import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.processing.fields.fieldInitializers.nestedForms.AbstractEmbeddedFormsInitializer.COLLAPSIBLE_CONTAINER;
+import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.processing.fields.fieldInitializers.nestedForms.AbstractEmbeddedFormsInitializer.FIELD_CONTAINER_PARAM;
+
+@Portable
+@Bindable
+@Definition(graphFactory = NodeFactory.class, nameField = "text")
+@FormDefinition(policy = FieldPolicy.ONLY_MARKED,
```
manstis: You can probably remove all the _form_ type stuff from this class too (like you did for `RuleAnnotationClause`).

You don't have support for [selecting the annotation cells](https://github.com/kiegroup/kie-wb-common/blob/dc3836d7c358ac60b1372fa4db4c3738d0bf460d/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableGrid.java#L801) in `DecisionTableGrid` and hence their properties can never be shown in the Properties Panel and hence all the _form binding_ stuff is not necessary.

See [here](https://github.com/kiegroup/kie-wb-common/pull/3230#discussion_r396486564) where I give more detail.

manstis:
@jomarko My comments are pretty much resolved and I expect the next commit to be the last.

Therefore feel free to now review.

jomarko:
### Comment 1
I am able to 'merge' rule annotations by naming them with same name, is that expected/allowed?
Screenshot from 2020-03-26 07-49-47

jomarko:
### Comment 2
User is not able to save changes in annotation data cell. See [video](https://issues.redhat.com/secure/attachment/12468488/annotation-save.webm).

jomarko:
### Comment 3
Putting value to the annotation header is not consistent with other cells. In other cells new value can be confirmed with clicking to other gird cell. In annotation header cell this doesn't work. See the attached [video](https://issues.redhat.com/secure/attachment/12468489/annotation-putting-value.webm).

jomarko:
### Comment 4
Column widths are discarded and even after save and reopen they are not preserved.

- master 
Screenshot from 2020-03-26 08-02-36

- pr 
Screenshot from 2020-03-26 08-20-05

manstis:
> Comment 1

This should not be possible. Column headers are merged if they _share_ meta-data. See [here](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/util/ColumnIndexUtilities.java#L50). The inherited `equals(..)` method in `RuleAnnotationClauseColumnHeaderMetaData` should be ensuring different columns are not equal and hence headers are not merged; but that is where I'd first look.

> Comment 4

This is probably caused by [this](https://github.com/kiegroup/kie-wb-common/pull/3230#discussion_r398129118). When opening a file a check is made that the number of columns' widths in the model is equal to the number of columns in the model. If different they are cleared and reset.

kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/DecisionRule.java
```diff
@@ -89,6 +97,9 @@ public boolean equals(final Object o) {
         if (inputEntry != null ? !inputEntry.equals(that.inputEntry) : that.inputEntry != null) {
             return false;
         }
+        if (annotationEntry != null ? !annotationEntry.equals(that.annotationEntry) : that.annotationEntry != null) {
```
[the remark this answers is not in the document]
danielzhe: I'm trying to follow the pattern used in the equals method, for other values. I agree with that simplification, but should I ignore the others right above the line I wrote? Or should I change the others?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/DeleteRuleAnnotationClauseCommand.java
```diff
@@ -0,0 +1,196 @@
...
+import org.kie.workbench.common.dmn.client.commands.VetoUndoCommand;
+import org.kie.workbench.common.dmn.client.commands.util.CommandUtils;
+import org.kie.workbench.common.dmn.client.editors.expressions.types.dtable.DecisionTableUIModelMapper;
+import org.kie.workbench.common.dmn.client.editors.expressions.types.dtable.DecisionTableUIModelMapperHelper;
+import org.kie.workbench.common.stunner.core.client.canvas.AbstractCanvasHandler;
+import org.kie.workbench.common.stunner.core.client.canvas.command.AbstractCanvasCommand;
+import org.kie.workbench.common.stunner.core.client.canvas.command.AbstractCanvasGraphCommand;
+import org.kie.workbench.common.stunner.core.client.command.CanvasCommandResultBuilder;
+import org.kie.workbench.common.stunner.core.client.command.CanvasViolation;
+import org.kie.workbench.common.stunner.core.command.Command;
```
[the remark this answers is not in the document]
danielzhe: Yes, it is used. Sorry, I don't get this comment. What should I change?

danielzhe:
Thank you guys for your review! All reported issues are pushed now, read for your review when you can. @manstis @jomarko 
About the merge columns, should we keep this behavior or change for this specific case?

manstis:
@danielzhe Column headers should not be merged.

danielzhe:
@manstis Done. Columns headers are not merged anymore. Thank you! :)

jomarko:
@danielzhe thank you for updates. My comments seems to e resolved but **Comment 2** seems be still there. If the only change since last save operation is changed value of rule annotation header/data cell, such change can not be saved.

manstis:
@danielzhe Re: @jomarko last comment about **Comment 2**; I think you'll need to add `equals(..)` and `hashCode()` implementations to both `RuleAnnotationClause` and `RuleAnnotationClauseText`.

danielzhe:
Thank you, @jomarko and @manstis ! Changes applied.

jomarko:
@danielzhe please have a look on failing tests, seems failures are caused by change we introduce

danielzhe:
@manstis I'm having issues with [this](https://github.com/kiegroup/kie-wb-common/blob/87db6625fe01579ce6f6ce049818adb7dc0b7a1b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddInputClauseCommandTest.java#L247) and [this](https://github.com/kiegroup/kie-wb-common/blob/87db6625fe01579ce6f6ce049818adb7dc0b7a1b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddOutputClauseCommandTest.java#L270) test. Both the same issue:

Before [this change](https://github.com/kiegroup/kie-wb-common/pull/3230/files#diff-97705a3cea25acf894c8d2d9b25974c1R33) in the `DecisionTable`, the exception expected in those tests [was raised here](https://github.com/kiegroup/kie-wb-common/blob/61ee2b7a0776f3ba0e63fc493a179c5a5024ed63/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddInputClauseCommand.java#L112) (`ArrayIndexOutOfBoundsException`).

But now with this change, the exception is raised a [few lines below, here](https://github.com/kiegroup/kie-wb-common/blob/61ee2b7a0776f3ba0e63fc493a179c5a5024ed63/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/expressions/types/dtable/AddInputClauseCommand.java#L115), and is `IndexOutOfBoundsException`, which reminds me the [DROOLS -3777 commented in those tests](https://issues.redhat.com/browse/DROOLS-3777) specific ignoring they when we are running in JDK11.

So my question if is there any issue that you know changing the test to expect `IndexOutOfBoundsException` and probably closing the DROOLS-3777.

jomarko:
If we agree the changed exception is expected due to changes in your PR, I am fine with change however I would vote for using `assertj` testing library that allows to test if specific line throws an exception. The current junit was (annotation over whole test) is not so accurate test. WDYT?

manstis:
Hi @danielzhe I did a bit of poking around:

If you put a break-point in `MoveColumnsCommand.moveClauses(..)` on `L#213` you'll see that (for example) `...dtable.MoveColumnsCommandTest.testMoveSingleAnnotationColumnRight()` that `clauses.removeAll(clausesToMove)` removes ALL `RuleAnnotationClauseText` that leads to the RTE. Only one instance should be removed. This points to an issue with the `equals(..)` method on `RuleAnnotationClauseText`. Indeed, if you set each instance of `RuleAnnotationClauseText` in the test to have different text values the tests all pass OK.

I think we should probably remove the `RuleAnnotationClauseText.equals(..)` (and `hashCode()`) method all together since even if the _text_ are of the same value we don't want the instances to be equals. `UnaryTests` and `LiteralExpression` used by `InputClause` and `OutputClause` in a decision table by chance are not affected by the same problem as they both contain a UUID (their `Id`) and hence `equals(..)` returns `false` for two instances with the same text value. 

(Or we add an _internal_ UUID to `RuleAnnotationClauseText` and use it in the `equals(..)` and `hashCode()` methods.. i.e. we don't have a getter for it).

I'm looking at the other failures.. and will comment in a bit.

manstis:
@danielzhe You can change both `AddInputClauseCommand` and `AddOutputClauseCommandTest` to expect `IndexOutOfBoundsException` and remove the check for JDK11, closing DROOLS-3777 as you propose. `ArrayList` throws `IndexOutOfBoundsException`.... which makes me wonder how the tests ever passed (unless an older JDK used to have `ArrayList` throw `ArrayIndexOutOfBoundsException`!!)

danielzhe:
@manstis Hi! Thank you, I noticed this too. In fact it's fixed and pushed now, I don't know if my changes is good enough but it's working now. 

Ready for you and @jomarko continue the review.

Thank you, guys!

jomarko:
User is unable to drag annotation column if two has same name

Screenshot from 2020-04-02 10-16-25

danielzhe:
@jomarko Well spotted. Change applied. Thank you!

manstis:
Good spot @jomarko.

The changes made to both `DecisionRulePropertyConverter` and `DecisionTablePropertyConverter` will need to be equally applied to the client-side marshaller [here](https://github.com/kiegroup/kie-wb-common/tree/master/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/model). An integration test could also be written [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/test/java/org/kie/workbench/common/dmn/showcase/client/selenium/DMNDesignerKogitoSeleniumIT.java) (we're not able to unit test the client-side marshaller).

danielzhe:
Good spot, @jomarko ! Thank you for point me the right path to follow, @manstis :)
Changes are applied. I updated the Selenium tests too.

PR #3237, opened 2020-03-22
Stunner - Update maven profiles for bpmn kogito webapp

romartin:
Hey @hasys 

Just a quick fix on the maven profiles for the BPMN kogito runtime webapp:
* Removed the `dev` profile 
* Fixed the `sources` profile

So let's keep it easy - there only exists a couple of profiles: `default` and `sources`:
* `default` - optimized / not source code
* `sources` - not optimized, prettyfied, source code (& sourcemaps) included

Thanks!


romartin:
Jenkins please retest this

PR #3249, opened 2020-03-26
KOGITO-395 BPMN Serialization reusing ID

hasys:
Hi @romartin, @LuboTerifaj, @domhanak, @krisv 

this PR removes all ID without use from BPMN file in KOGITO (Business Central is not affected we have separate jira issue for it). Two ids are still generates on every save, but they are used somewhere, so I didn't remove it, we will handle it as separate jira issue.

### This PR removes IDs for following elements
* Documentaion
* LaneSet
* ioSpecification
* inputSet
* outputSet
* dataInputAssociation
* DataOutputAssociation
* Assignment
* TimerEventDefinition
* CompensateEventDefinition
* SignalEventDefinition
* errorEventDefinition
* completionCondition
* messageEventDefinition
* escalationEventDefinition
* conditionalEventDefinition
* terminateEventDefinition
* MultiInstanceLoopCharacteristics
* Potential Owner (User Task → Actor property)
* FormalExpression tags such as:
    * CompletionCondition
    * ConditionExpression
    * Condition
    * TimeDate
    * TimeDuration
    * from
    * to

ID's removed because they are not used and regenerated on every save and it creates a lot of nose and make impossible to see difference between two BPMN files.

You can check that IDs for mentioned elements never used in whole BPMN file. Just locate any mentioned element, copy ID and search it in the BPMN file. You will find just one occurrence 

### Element where IDs still regenerated on every save because they used somewhere else

* Definition tag (root tag), also used in 
    * Definition → Relationship → Target (XML)
    * Definition → Relationship → Source (XML)
* Message tag, also used in
    * messageEventDefinition tag as messageRef attribute

Listed elements will be fixed as separate jira issue.

I tested this PR with https://github.com/kiegroup/kogito-travel-agency-tutorial/tree/master/01-kogito-travel-agency I opened all 3 files in Kogito Webapp (VS Code plugin doesn't work to me again) and slightly modified it (just moved a node). Than saved new file versions without IDs to the test project and was able to pass all tests, compile and start the application. Also I did some REST request and all looks good to me.

@krisv can I ask you to confirm that if IDs are not used anywhere in the file - they are useless and can be removed?

So this PR is ready for review. Thank you!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-emf/src/main/java/org/eclipse/emf/ecore/xmi/resource/xml/XMLSave.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-emf/src/main/java/org/eclipse/emf/ecore/xmi/resource/xml/XMLSave.java
@@ -29,7 +29,26 @@ import com.google.gwt.xml.client.Attr;
 import com.google.gwt.xml.client.Document;
 import com.google.gwt.xml.client.Element;
 import com.google.gwt.xml.client.Node;
+import org.eclipse.bpmn2.Assignment;
 import org.eclipse.bpmn2.BaseElement;
+import org.eclipse.bpmn2.CompensateEventDefinition;
+import org.eclipse.bpmn2.ConditionalEventDefinition;
+import org.eclipse.bpmn2.DataInputAssociation;
+import org.eclipse.bpmn2.DataOutputAssociation;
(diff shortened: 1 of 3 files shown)


hasys:

domhanak:

domhanak:
Looks good to me.
I created two new identical BPMN files in VSCode.
1. saved both
2. edited second one and diffed them.
3. Only id change I see is source and target as mentioned in comment.

Looks good to me. Thanks @hasys free to merge when FDB passes.

hasys:
Hi @domhanak, it is green already. Thank you!

krisv:
@domhanak went over the list of removed ids and I agree, I believe these are not referenced (at least not in the use cases we currently target afaik).

PR #3251, opened 2020-03-26
KOGITO-1244: [DMN Designer] Data Types: Save/cancel icons not visible when there are many constraints and RHS dock open

karreiro:
See https://issues.redhat.com/browse/KOGITO-1244

Before:
before

After:
after


jomarko:

PR #3265, opened 2020-04-14
RHPAM-2881 Notification Form in Stunner is limiting the Body content

hasys:
Hi @romartin, @LuboTerifaj,

this one was easy one. But it is good question, do we have any limits on the body? I didn't found any info about it. Maybe @MarianMacik @cristianonicolai have any info?

Also, I found two further possible issues to fix:
* all content after `@` symbol is cutted
* it is not possible to "enlarge" the text area. It is easy to fix, if needed. See on the picture below how it can be. Yes, it looks little bit ugly, but it can be useful and "better than nothing". I found it useful when I looked why part of the body still cut (see previous point).

obrazek


hasys:

MarianMacik:
@hasys As you can see here:

https://github.com/kiegroup/jbpm/blob/f70a465dd2621f5598f043631b8eb991142e05ba/jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/EmailNotificationHeaderImpl.java#L54

This will limit the column when using auto generation, i.e. not DDL scripts, which is discouraged in production.

And also here:

https://github.com/kiegroup/jbpm/blob/master/jbpm-db-scripts/src/main/resources/db/ddl-scripts/db2/db2-jbpm-schema.sql#L457

Although not all databases have this limit, see for example Oracle scripts, I would limit it anyway to properly support all of them, so 65 535 is enough I think.

hasys:
Great, thank you @MarianMacik, I upgraded to `65 535`.

hasys:

hasys:
Hi @romartin, @LuboTerifaj,

all is green and ready for review. Thank you!
obrazek

PR #3268, opened 2020-04-15
KOGITO-1196: Editor goes back horizontally when diagram is bigger than default

handreyrc:
@romartin , @inodeman : Could you review this PR please?
Ps: This PR depends on changes made in the lienzo-core to work properly.
PRs:
*[lienzo-core PR#108](https://github.com/kiegroup/lienzo-core/pull/108)
*[lienzo-tests PR#92](https://github.com/kiegroup/lienzo-tests/pull/92)


handreyrc:

handreyrc:
Hey @romartin, I agree it would be better to have it disabled by default from Stunner point of view. However, my point here is to avoid impact on other systems cosuming lienzo-core API that may expect this behavior enabled by default.
Anyway,  it  is a simple change and I'll be glad to do it if you are for.
WDYT?

handreyrc:
It was decided to change only the lienzo-core. The drag mediator will be disabled by default.

PR #3274, opened 2020-04-21
AF-2469 :kie-wb-common-ala/kie-wb-common-ala-services-rest tests requires docker image kitematic/hello-world-nginx

RishiRajAnand:
Ignored test until we find a permanent fix.


kie-wb-common-ala/kie-wb-common-ala-services-rest/src/test/java/org/guvnor/ala/services/rest/tests/RestRuntimeProvisioningImplTest.java
```diff
@@ -140,6 +141,7 @@ public static void beforeClass() {
         Assume.assumeFalse(SystemUtils.IS_OS_WINDOWS);
     }
 
+    @Ignore
```
mareknovotny, line 144: can you add here also the JiRA which explains why we ignore it?
```suggestion
     @Ignore("https://issues.redhat.com/browse/AF-2469") 
```

RishiRajAnand: link to Jira added

PR #3277, opened 2020-04-22
RHPAM-2674 :Exception thrown when a test scenerio containing a tag is renamed without saving

RishiRajAnand:
This exception was thrown in the build pipeline stage due to the asynchronous processing of events in incremental builds. During "Save and rename" operation for any resource,  Though the resource was moved, Resource modified event for original resource was queued in the build execution pipeline,  which result in NoSuchFileException. 

Please note that this exception happens in all resource types. To reproduce
1. Open resource (eg. .drl) 
2. Change any content
3. Click on rename
4. Select "Save and rename"

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/core/Builder.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/core/Builder.java
@@ -317,8 +317,15 @@ public class Builder implements Serializable {
     }
 
     public IncrementalBuildResults addResource(final Path resource) {
-        return addResource(resource,
-                           ioService.newInputStream(resource));
+        IncrementalBuildResults results = new IncrementalBuildResults(projectGAV);
+        // This validation prevents faulty update resource operation that may come during save and rename operation
+        // For more information see https://github.com/kiegroup/kie-wb-common/pull/3277
+        if (ioService.exists(resource)) {
+            results = addResource(resource,
+                                  ioService.newInputStream(resource));
+        }
+
+        return results;
     }
 
     private IncrementalBuildResults addResource(final Path resource,
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/builder/core/BuilderTest.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/test/java/org/kie/workbench/common/services/backend/builder/core/BuilderTest.java
@@ -30,6 +30,7 @@ import com.google.common.io.Resources;
 import org.drools.core.rule.TypeMetaInfo;
 import org.guvnor.common.services.project.builder.model.BuildMessage;
 import org.guvnor.common.services.project.builder.model.BuildResults;
+import org.guvnor.common.services.project.builder.model.IncrementalBuildResults;
 import org.guvnor.common.services.project.model.GAV;
 import org.guvnor.common.services.project.model.Module;
 import org.guvnor.common.services.project.model.POM;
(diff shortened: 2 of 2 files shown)


RishiRajAnand:

barboras7:
Hi @RishiRajAnand,

I tested the fix on multiple kinds of resources and following are my observations:

- **DRL** - everything seems to work fine (changes in editor, adding tags, adding description and afterwards renaming the asset without saving -> no error is thrown)
- **Data Object** - I am still seeing NoSuchFileException in logs (To reproduce the issue, create a new Data Object in some project, add a tag in this DO (Overview -> Metadata) and then rename the data object without previous saving of your change)
- **Business process** - I am getting an error modal, see attached screenshot (To reproduce this issue, create a Business Process, add tag in this business process (Overview -> Metadata) and then rename the business process without previous saving of your change)

Screenshot from 2020-04-27 14-48-39

RishiRajAnand:
@barboras7 thanks for the update, I'll get back to you with more information.

RishiRajAnand:
Hi @ederign , The cause of error are different in all scenerios:

1. In case of Data objects, it doesn't uses rename service implementation that other resources like Test scenario are using. Due to this, it always tries to save the original file first before moving it, which further results in indexer error since indexer sees it as a change. Replacing it with RenameService should solve this, I could see that it is already in our todos (https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-backend/src/main/java/org/kie/workbench/common/screens/datamodeller/backend/server/DataModelerServiceImpl.java#L800). Do you know if someone is already working on it ?

2. In case of business processes, it is due to rpc error. It doesn't give any error in UI/Backend logs.

While this fix can prevent the exception in all assets that uses https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-backend/src/main/java/org/uberfire/ext/editor/commons/backend/service/SaveAndRenameServiceImpl.java, for others we have to deal with case by case basis. wdyt ?

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/core/Builder.java
```diff
@@ -317,8 +317,13 @@ private void updateDependenciesClassLoader(final Module project,
     }
 
     public IncrementalBuildResults addResource(final Path resource) {
-        return addResource(resource,
-                           ioService.newInputStream(resource));
+        IncrementalBuildResults results = new IncrementalBuildResults(projectGAV);
+        if(ioService.exists(resource)) {
+            results = addResource(resource,
```
adrielparedes, line 322: What if this method throws an exception? Is that catch inside it? Shouldn't we try to prevent errors in there and include it in IncrementalBuildResults?

RishiRajAnand: @adrielparedes Ideally we should try to handle the exception and include the errors in IncrementalBuildResults. but in this particular case, I intend to filter out the faulty modify resource event which results in  NoSuchFileException, IMO This should not be included as part of IncrementalBuildResults. (since user expects to have one build after a SaveAndRename operation, this gives two). All other exceptions  will be handled by the caller method.

RishiRajAnand:
@adrielparedes I have added a test case, mainly to just test that method doesn't throw NoSuchFileException and returns empty build results.

RishiRajAnand:
@barboras7 As discussed, we can raise separate issues for other resources (like Data object, Business processes etc) which have problem with Save and rename operation. The best thing would be to create one issue for all resources having similiar logs.
@mareknovotny ^ We need this because errors in other resources are not related to this, since they have different implementations which throws different exceptions.

RishiRajAnand:

barboras7:
@RishiRajAnand approving this PR as it's fixing the original issue. 
I am going through all kinds of assets and I am checking the behavior of "Rename" action. I'll create Jiras when I have data about all the asset types.

RishiRajAnand:
Old FDB seems to be green. https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/kie-wb-common-downstream-pullrequests/1204/.However, new is failing for all the other PR's as well. @ederign ^.

ederign:
Merging based on last @RishiRajAnand comment.

PR #3283, opened 2020-04-23
AF-2467 : Commit history in metrics does not get refreshed when using other than master branch

akumar074:
* Fetching VersionRecords from all branches and filtering it on client-side

 * Space metrics show commit history of the master branch by default

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-contributors/kie-wb-common-contributors-backend/src/main/java/org/kie/workbench/common/screens/contributors/backend/dataset/ContributorsManager.java
+++ b/kie-wb-common-screens/kie-wb-common-contributors/kie-wb-common-contributors-backend/src/main/java/org/kie/workbench/common/screens/contributors/backend/dataset/ContributorsManager.java
@@ -45,6 +45,7 @@ import org.guvnor.structure.organizationalunit.OrganizationalUnitService;
 import org.guvnor.structure.organizationalunit.RemoveOrganizationalUnitEvent;
 import org.guvnor.structure.organizationalunit.RepoAddedToOrganizationalUnitEvent;
 import org.guvnor.structure.organizationalunit.RepoRemovedFromOrganizationalUnitEvent;
+import org.guvnor.structure.repositories.Branch;
 import org.uberfire.backend.server.util.Paths;
 import org.uberfire.commons.services.cdi.Startup;
 import org.uberfire.ext.editor.commons.backend.version.VersionRecordService;
@@ -59,6 +60,7 @@ import org.uberfire.workbench.events.ResourceUpdatedEvent;
 
 import static org.kie.soup.commons.validation.PortablePreconditions.checkNotNull;
 import static org.kie.workbench.common.screens.contributors.model.ContributorsDataSetColumns.COLUMN_AUTHOR;
+import static org.kie.workbench.common.screens.contributors.model.ContributorsDataSetColumns.COLUMN_BRANCH;
 import static org.kie.workbench.common.screens.contributors.model.ContributorsDataSetColumns.COLUMN_DATE;
 import static org.kie.workbench.common.screens.contributors.model.ContributorsDataSetColumns.COLUMN_MSG;
 import static org.kie.workbench.common.screens.contributors.model.ContributorsDataSetColumns.COLUMN_ORG;
@@ -104,6 +106,7 @@ public class ContributorsManager implements DataSetGenerator {
             .label(COLUMN_ORG)
             .label(COLUMN_REPO)
             .label(COLUMN_PROJECT)
+            .label(COLUMN_BRANCH)
             .label(COLUMN_AUTHOR)
             .text(COLUMN_MSG)
             .date(COLUMN_DATE)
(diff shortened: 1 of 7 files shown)


akumar074:
Jenkins execute full downstream build please

kie-wb-common-screens/kie-wb-common-contributors/kie-wb-common-contributors-backend/src/main/java/org/kie/workbench/common/screens/contributors/backend/dataset/ContributorsManager.java
```diff
@@ -150,32 +154,37 @@ public DataSet buildDataSet(Map<String, String> params) {
                 for (final WorkspaceProject project : projects) {
                     final String repoAlias = project.getRepository().getAlias();
```
adrielparedes, line 155: You can extract project.getRepositories() here and line 157

kie-wb-common-screens/kie-wb-common-contributors/kie-wb-common-contributors-backend/src/main/java/org/kie/workbench/common/screens/contributors/backend/dataset/ContributorsManager.java
```diff
@@ -150,32 +154,37 @@ public DataSet buildDataSet(Map<String, String> params) {
...
-                            Date date = record.date();
-                            dsBuilder.row(org,
-                                          repoAlias,
-                                          projectName,
-                                          author,
-                                          msg,
-                                          date);
+                    Collection<Branch> branches = project.getRepository().getBranches();
+
+                    branches.stream().forEach(branch -> {
```
adrielparedes, line 159: ok, so let me understand this part since I don't know that screen very well. We will have a new row with statistics for each branch, right? So if we have 100 branches this will be big right and may be resource consuming? I don't mean that all the projects that will be used in BC are going to have that number, but if branches are not deleted we will have a lot of information. Let's think smaller, what if we have 20 branches... how the screen will behave?
Or this means that everytime the screen is loaded we are going to query all that information without using it?
Just asking to add more background to what I understand :)

kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/test/java/org/kie/workbench/common/screens/library/client/screens/project/ProjectMetricsScreenTest.java
```diff
@@ -55,18 +56,18 @@
...
-                        {"org1", "repo", "project1", "user2", "04/03/19 12:00", "Commit 3"},
-                        {"org1", "repo", "project1", "user2", "06/04/19 12:00", "Commit 4"},
-                        {"org2", "repo", "project2", "user3", "07/05/19 12:00", "Commit 5"},
-                        {"org2", "repo", "project2", "user3", "09/06/19 12:00", "Commit 6"},
-                        {"org2", "repo", "project2", "user4", "11/07/19 12:00", "Commit 7"},
-                        {"org2", "repo", "project2", "user4", "02/08/20 12:00", "Commit 8"},
-                        {"emptyOrg", null, null, null, null, null, null}});
+                        {"org1", "repo", "project1", "master", "user1", "01/01/19 12:00", "Commit 1"},
+                        {"org1", "repo", "project1", "master", "user1", "03/02/19 12:00", "Commit 2"},
+                        {"org1", "repo", "project1", "master", "user2", "04/03/19 12:00", "Commit 3"},
```
adrielparedes, line 64: I think it will be nice to have a test with several branches, wdyt?

adrielparedes:
@barboras7 @akumar074 My concern about this is how that screen will behave when a project that is imported could contain several branches a several commits. If we try this with a new project this will work, but what if you have something really big? If you have time, can you perform that kind of test?

jesuino:
As discussed with @akumar074, the point raised by @adrielparedes is a valid one. Thousands of commits with 10+ branches will slowdown the displayers that uses the mentioned dataset. Two solutions were possible:

1) limit the number of rows;
2) create per branch dataset.

Abhishek will work on 2 and will update this PR soon. Another concern raised by Abhishek was regarding the memory occupied by multiple dataset definitions. IMO it will not likely impact the performance, because the rows are only fetched when a given dataset is looked up. However, cache may be an issue, if this is the case we can adjust the dataset definition cache settings.

barboras7:
Hi guys, 
Savitha (@sthundat) is going to work on this PR instead of me. If you need to discuss anything, please, reach out to her :slightly_smiling_face:

akumar074:
Hi @jesuino , @ederign , I explored and tried to resolve the issue with option 2, but the complexity is high which makes it difficult to implement without any change in dataset design. We are unable to create dynamic DatSetDef before building the project or branch-based dataset. Only one DataSetDef _"git_contriutors"_ is created when an instance of ContributorsManager(ApplicationScoped) is created. As William mentioned, the idea of a branch or project-based dataset is an improvement in the performance of Metrics, but the impact of change is very high. On the other hand, limiting the row of the dataset would be an easier option but considering sorting or any other operation there may be some performance issue here.
wdyt?

domhanak:
This is a stale PR, please advice how to proceed with this PR otherwise it will be closed.

PR #3290, opened 2020-04-29
DROOLS-5253: [DMN Designer] Copied node should be distinguished from original one

vpellegrino:
*Please refer to:* https://issues.redhat.com/browse/DROOLS-5253

*Issue description:* If user copy a DMN diagram node, the copy has exactly the same name as the original node. It causes validation issues as the node names have to be unique. 

*Proposed solution:* Every time we copy a node, a new node will be created with a suffix `-X`, where `X` it is an incremental numeric value.
Logic on uniqueness is performed. For instance, when copying the node `InputData-1`, if, for some reason, `InputData-2` is already present, then the name of the copied node will be `InputData-3`.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
@@ -18,7 +18,6 @@ package org.kie.workbench.common.dmn.client.commands.clone;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.function.Predicate;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
@@ -42,6 +41,7 @@ import org.kie.workbench.common.dmn.api.property.dmn.Id;
 import org.kie.workbench.common.dmn.api.property.dmn.Name;
 import org.kie.workbench.common.dmn.api.property.dmn.NameHolder;
 import org.kie.workbench.common.dmn.api.property.dmn.Text;
+import org.kie.workbench.common.dmn.client.property.dmn.DefaultValueUtilities;
 import org.kie.workbench.common.stunner.core.api.FactoryManager;
 import org.kie.workbench.common.stunner.core.client.api.SessionManager;
 import org.kie.workbench.common.stunner.core.definition.adapter.AdapterManager;
@@ -61,7 +61,6 @@ public class DMNDeepCloneProcess extends DeepCloneProcess implements IDeepCloneP
 
     private static final Logger LOGGER = Logger.getLogger(DMNDeepCloneProcess.class.getName());
     private static final RegExp NAME_SUFFIX_REGEX = RegExp.compile("[?!-]\\d+$");
-    private static final String CLONED_DEFAULT_SUFFIX = "-1";
     private static final String HYPHEN = "-";
     private final SessionManager sessionManager;
 
@@ -135,30 +134,47 @@ public class DMNDeepCloneProcess extends DeepCloneProcess implements IDeepCloneP
         target.setText(new Text(uniqueNodeName));
     }
 
-    String composeUniqueNodeName(final String name) {
+    protected String composeUniqueNodeName(final String name) {
         final String originalName = Optional.ofNullable(name).orElse("");
-        String uniqueName = originalName + CLONED_DEFAULT_SUFFIX;
 
         try {
             final MatchResult nameSuffixMatcher = NAME_SUFFIX_REGEX.exec(originalName);
             if (nameSuffixMatcher != null) {
-                uniqueName = buildNameWithIncrementedSuffixIndex(originalName, nameSuffixMatcher);
+                return buildNameWithIncrementedSuffixIndex(originalName, nameSuffixMatcher);
             }
         } catch (Exception e) {
             LOGGER.warning("There was an issue while parsing node with name " + originalName + " - A fallback will be used for it");
         }
 
-        return ensureNodeNameUniqueness(uniqueName);
+        return joinPrefixWithIndexedSuffix(originalName);
     }
 
-    private String ensureNodeNameUniqueness(final String uniqueName) {
-        return StreamSupport.stream(getGraphNodes().spliterator(), true)
+    private String buildNameWithIncrementedSuffixIndex(final String nameValue, final MatchResult matchResult) {
+        final String suffix = matchResult.getGroup(0);
+        final String prefix = Optional.ofNullable(nameValue.split(suffix)[0]).orElse("");
+        return joinPrefixWithIndexedSuffix(prefix);
+    }
+
+    private String joinPrefixWithIndexedSuffix(final String originalName) {
+        final String originalNameWithHyphen = originalName + HYPHEN;
+        return originalNameWithHyphen + getMaxUnusedIndexByNamePrefix(originalNameWithHyphen);
+    }
+
+    private int getMaxUnusedIndexByNamePrefix(final String namePrefix) {
+        final List<String> nodeNameList = StreamSupport.stream(getGraphNodes().spliterator(), true)
                 .map(this::nodeNamesMapper)
                 .filter(Objects::nonNull)
-                .filter(Predicate.isEqual(uniqueName))
-                .findAny()
-                .map(this::composeUniqueNodeName)
-                .orElse(uniqueName);
+                .collect(Collectors.toList());
+        return DefaultValueUtilities.getMaxUnusedIndex(nodeNameList, namePrefix);
+    }
+
+    private Iterable<Node<View, Edge>> getGraphNodes() {
+        return sessionManager
+                .getCurrentSession()
+                .getCanvasHandler()
+                .getDiagram()
+                .getGraph()
+                .nodes();
     }
 
     private String nodeNamesMapper(final Node<View, Edge> node) {
@@ -171,23 +187,6 @@ public class DMNDeepCloneProcess extends DeepCloneProcess implements IDeepCloneP
             return hasText.getText().getValue();
         }
         return null;
-    }
-
-    private Iterable<Node<View, Edge>> getGraphNodes() {
-        return sessionManager
-                .getCurrentSession()
-                .getCanvasHandler()
-                .getDiagram()
-                .getGraph()
-                .nodes();
-    }
-
-    private String buildNameWithIncrementedSuffixIndex(final String nameValue, final MatchResult matchResult) {
-        final String suffix = matchResult.getGroup(0);
-        int suffixIndex = Integer.parseInt(suffix.substring(1));
-        final String nameValueWithoutSuffix = Optional.ofNullable(nameValue.split(suffix)[0]).orElse("");
-        final String computedSuffix = HYPHEN + (++suffixIndex);
-        return nameValueWithoutSuffix + computedSuffix;
     }
 
     private void cloneTypeRefInfo(final IsInformationItem srcInformationItem, final IsInformationItem targetInformationItem) {
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
@@ -122,43 +122,43 @@ public class DMNDeepCloneProcess extends DeepCloneProcess implements IDeepCloneP
     }
 
     private void cloneDRGElementBasicInfo(final DRGElement source, final DRGElement target) {
-        final String distinguishedName = composeDistinguishedNodeName(source.getName().getValue());
+        final String uniqueNodeName = composeUniqueNodeName(source.getName().getValue());
         target.setId(new Id());
-        target.setNameHolder(new NameHolder(new Name(distinguishedName)));
+        target.setNameHolder(new NameHolder(new Name(uniqueNodeName)));
         target.setDescription(source.getDescription().copy());
         target.setParent(source.getParent());
         target.getLinksHolder().getValue().getLinks().addAll(cloneExternalLinkList(source));
     }
 
     private void cloneTextElementBasicInfo(final HasText source, final HasText target) {
-        final String distinguishedName = composeDistinguishedNodeName(source.getText().getValue());
-        target.setText(new Text(distinguishedName));
+        final String uniqueNodeName = composeUniqueNodeName(source.getText().getValue());
+        target.setText(new Text(uniqueNodeName));
     }
 
-    String composeDistinguishedNodeName(final String name) {
+    String composeUniqueNodeName(final String name) {
         final String originalName = Optional.ofNullable(name).orElse("");
-        String distinguishedName = originalName + CLONED_DEFAULT_SUFFIX;
+        String uniqueName = originalName + CLONED_DEFAULT_SUFFIX;
 
         try {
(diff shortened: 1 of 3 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -87,15 +118,78 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
...
         }
 
-        return super.clone(source, target);
+        return target;
     }
 
     private void cloneDRGElementBasicInfo(final DRGElement source, final DRGElement target) {
-        target.setName(source.getName().copy());
-        target.setNameHolder(source.getNameHolder().copy());
+        final String uniqueNodeName = composeUniqueNodeName(source.getName().getValue());
```
manstis, line 125: Just a thought, but please see `DefaultValueUtilities.updateNewNodeName(..)` that is currently used to ensure new elements have unique names. It might be re-usable here.

vpellegrino: I will check.
Regards

vpellegrino: @manstis I briefly looked at `DefaultValueUtilities` implementation.
The method `updateNewNodeName` seems to be not suitable for the *copy* case, since it is working only with default names (i.e. names assigned when node gets created).
In the *copy* case you can change a node name to whatever value, and then suffix sequence must work as well.
 There are other two methods `getExistingNodeNames` and `getMaxUnusedIndex` that maybe can work well with this specific use-case.
I need to perform some trial and check if the result is the same.

manstis: OK, fair point. It was a suggestion, re-use if and as appropriate.

vpellegrino:
jenkins retest this

vpellegrino:

vpellegrino:
Failure in full downstream seems not related to this PR:

https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/upstream-downstream/job/kie-wb-common.downstream/50/testReport/junit/org.kie.workbench.common.services.backend.compiler.rest.client/MavenRestClientTest/org_kie_workbench_common_services_backend_compiler_rest_client_MavenRestClientTest/

Error Message
Could not invoke deployment method: public static org.jboss.shrinkwrap.api.Archive org.kie.workbench.common.services.backend.compiler.rest.client.MavenRestClientTest.getDeployment() throws java.lang.Exception
Stacktrace
java.lang.RuntimeException: Could not invoke deployment method: public static org.jboss.shrinkwrap.api.Archive org.kie.workbench.common.services.backend.compiler.rest.client.MavenRestClientTest.getDeployment() throws java.lang.Exception
	at org.jboss.arquillian.container.test.impl.client.deployment.AnnotationDeploymentScenarioGenerator.invoke(AnnotationDeploymentScenarioGenerator.java:181)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -87,15 +117,78 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
...
+                .getCurrentSession()
+                .getCanvasHandler()
+                .getDiagram()
+                .getGraph()
+                .nodes();
+    }
+
+    private String nodeNamesMapper(final Node<View, Edge> node) {
+        if (node.getContent().getDefinition() instanceof NamedElement) {
+            NamedElement namedElement = (NamedElement) node.getContent().getDefinition();
```
dupliaka, line 182: Could be final as well

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -87,15 +117,78 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
...
+            return namedElement.getName().getValue();
+        }
+        if (node.getContent().getDefinition() instanceof HasText) {
+            HasText hasText = (HasText) node.getContent().getDefinition();
```
dupliaka, line 186: This could be final

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -87,15 +117,78 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
...
+            return hasText.getText().getValue();
+        }
+        return null;
```
dupliaka, line 189: Just a suggestion :
You can keep one return to have only one exit point.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
dupliaka, line 186: hasText named like boolean variable that checks for text. Probably we could provide more context - wise declaration?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
vpellegrino: In this case, for readability I prefer to have multiple returns, instead of a non-final variable to keep the result to return

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -87,15 +117,78 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
...
+
+    protected String composeUniqueNodeName(final String name) {
+        final String originalName = Optional.ofNullable(name).orElse("");
+
+        try {
+            final MatchResult nameSuffixMatcher = NAME_SUFFIX_REGEX.exec(originalName);
+            if (nameSuffixMatcher != null) {
+                return buildNameWithIncrementedSuffixIndex(originalName, nameSuffixMatcher);
+            }
+        } catch (Exception e) {
```
dupliaka, line 145: Small suggestion to investigate how could we avoid a generic exception since that would be hard to inspect if something would happen if exec will throw any exception like IndexOutOfBounds or buildNameWithIncrementedSuffixIndex throw some exception(bcs it would probably just skipped)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -69,10 +92,17 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
     @Override
```
dupliaka, line 92: Could you please update javadoc on it as well?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
vpellegrino: @dupliaka thank you for your question.
Let me explain a bit why I put this generic catch block.
I put this try-catch block because of the regex matcher.
This exception in reality should never happen.
It is only a way to way to unblock cloning mechanism in case any issue with the matcher occurs.
The strategy here, **only in the remote case an exception gets thrown*  is:
1) Softly print exception details
2) Return a fallback name for the copied node

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
vpellegrino: Which information would you update?
Nothing seems to be changed from doc perspective

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcessTest.java
```diff
@@ -261,4 +345,70 @@ private void setLinks(final DRGElement drgElement, final String... links) {
...
+        assertThat(dmnDeepCloneProcess.composeUniqueNodeName(INPUT_DATA_NAME))
+                .isEqualTo(INPUT_DATA_NAME + FIRST_INDEX_IN_SUFFIX);
+    }
+
+    @Test
+    public void testComposingUniqueNodeNameWhenItAlreadyContainsIndexedSuffix() {
+        mockSingleNodeInTheGraph();
+
+        assertThat(dmnDeepCloneProcess.composeUniqueNodeName(INPUT_DATA_NAME + FIRST_INDEX_IN_SUFFIX))
+                .isEqualTo(INPUT_DATA_NAME + "-2");
```
dupliaka, line 360: Suggestion:
Would you mind to remove magic -2 to the SECOND_INDEX_IN_SUFFIX?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
dupliaka: I mean that for new node name. It is not clear that this should comply with the validation mechanisms

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
dupliaka: Ok

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
dupliaka: Ok

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/test/java/org/kie/workbench/common/dmn/showcase/client/selenium/DMNDesignerKogitoSeleniumIT.java
```diff
@@ -250,6 +250,7 @@ public void testCopyAndPaste() throws Exception {
         expandDecisionNavigatorDock();
```
dupliaka, line 250: This test is a reproducer for https://issues.redhat.com/browse/DROOLS-4424
If your change affect this then we can remove the comment and close the bug.
Let me only check it manually first.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcessTest.java
```diff
@@ -261,4 +348,70 @@ private void setLinks(final DRGElement drgElement, final String... links) {
                                          .add(new DMNExternalLink(link, "description"))
                 );
     }
+
```
dupliaka, line 351: Suggestion
Would you mind to add a test for case if the matcher throws an exception in composeUniqueNodeName()?

vpellegrino: Looked at code of `com.google.gwt.regexp.shared.RegExp#exec` method.
There is no explicit exception thrown, so I am going to remove the try-catch block.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/test/java/org/kie/workbench/common/dmn/showcase/client/selenium/DMNDesignerKogitoSeleniumIT.java
dupliaka: I retested it manually I think you can remove the comment so now it is not a reproducer. Just a test.

dupliaka:
- [ ] Any dmn component or group of components copy pasts with incremented title title_copy-n
- [ ] If user has Terminator-3 and copies Terminator-2 node then new one has name Termonator-4
- [ ] If user got several copies Terminator-1 Terminator-3 then for the next copy  it would increment title (without creation of Terminator-2)
- [ ] If I copy paste from the context menu of vs code it does not copy past anything
- [ ] https://issues.redhat.com/browse/DROOLS-4424 does not reproduces

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
dupliaka: Ok

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/test/java/org/kie/workbench/common/dmn/showcase/client/selenium/DMNDesignerKogitoSeleniumIT.java
```diff
@@ -269,7 +267,12 @@ public void testCopyAndPaste() throws Exception {
         XmlAssert.assertThat(actual)
                 .withNamespaceContext(NAMESPACES)
                 .valueByXPath("count(//dmn:businessKnowledgeModel[@name='" + nodeName + "'])")
-                .isEqualTo(2);
+                .isEqualTo(1);
+
+        XmlAssert.assertThat(actual)
+                .withNamespaceContext(NAMESPACES)
+                .valueByXPath("count(//dmn:businessKnowledgeModel[@name='" + copiedNodeName + "'])")
+                .isEqualTo(1);
```
jomarko, line 275: Thank you for updating the test @vpellegrino @dupliaka !

PR #3301, opened 2020-05-13
KOGITO-2166 VSCode editor - Package property: default is not a valid Java package name

evacchi:
Steps:

-    just create a new process without changing the package name, it will be default

Description

 -   This breaks Kogito as the package name generates an invalid class


hasys:
Hi @evacchi, new commit "fixing" process ID, not process Package. So, after the second change full name of the process will be `com.example.com.example` instead of `com.example.default`. We can change default process ID to something more meaningful than `default`. For example at least make it with upper-case like `Default` though.

romartin:
Oh good catch @hasys , I didn't noticed!!! !:(

hey @Josephblt , maybe you can help here... could you please let @evacchi which is the right property for changing the default package value? Or feel free to create a PR for it in case too

romartin:
@Josephblt ping ^ can you create a new PR or point us to the right String to be changed plz? This is quite urgent.. thx!

romartin:
Closing as it's being replaced by:

https://github.com/kiegroup/kie-wb-common/pull/3306

thanks @evacchi  for your help!!! :+1:

evacchi:
yw, I just wanted to drop a tiny PR to show where the problem was, thanks for completing it and figuring out the issue :)

PR #3310, opened 2020-05-20
KOGITO-1553: [DMN Designer] Open a DMN file with an included model - Included Models tab

danielzhe:
This PR is part of an ensemble:
https://github.com/kiegroup/kie-wb-common/pull/3310
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1323
https://github.com/kiegroup/drools-wb/pull/1375

This PR is not part of the ensemble, but is required. It have to be merged alone to generate the new Gwt Jsonix Schema Compiler build:
https://github.com/kiegroup/gwt-jsonix-schema-compiler/pull/8

The following JIRAS are included in this PR:
KOGITO-1553: [DMN Designer] Open a DMN file with an included model - Included Models tab
KOGITO-1776: [DMN Designer] Create a "Load Projects From Client" button
KOGITO-1552: [DMN Designer] Open a DMN file with an included model - DRG Elements

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNIncludedNodeFactory.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNIncludedNodeFactory.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.webapp.kogito.common.client.converters;
+
+import org.kie.workbench.common.dmn.api.definition.HasVariable;
+import org.kie.workbench.common.dmn.api.definition.model.DRGElement;
+import org.kie.workbench.common.dmn.api.definition.model.InformationItemPrimary;
+import org.kie.workbench.common.dmn.api.definition.model.IsInformationItem;
+import org.kie.workbench.common.dmn.api.editors.included.DMNIncludedNode;
+import org.kie.workbench.common.dmn.api.editors.included.IncludedModel;
+import org.kie.workbench.common.dmn.api.property.dmn.Id;
+import org.kie.workbench.common.dmn.api.property.dmn.Name;
+import org.kie.workbench.common.dmn.api.property.dmn.QName;
+import org.kie.workbench.common.stunner.core.util.FileUtils;
+
+import static org.kie.workbench.common.dmn.api.editors.types.BuiltInTypeUtils.isBuiltInType;
+
+public class DMNIncludedNodeFactory {
+
+    DMNIncludedNode makeDMNIncludeNode(final String path,
+                                       final IncludedModel includeModel,
+                                       final DRGElement drgElement) {
+        return new DMNIncludedNode(FileUtils.getFileName(path), drgElementWithNamespace(drgElement, includeModel));
+    }
+
+    DRGElement drgElementWithNamespace(final DRGElement drgElement,
+                                       final IncludedModel includeModel) {
+
+        final String modelName = includeModel.getModelName();
+
+        drgElement.setName(createName(drgElement, modelName));
+        drgElement.setAllowOnlyVisualChange(true);
+
+        if (drgElement instanceof HasVariable) {
+            final HasVariable hasVariable = (HasVariable) drgElement;
+            final IsInformationItem variable = hasVariable.getVariable();
+            final QName qName = variable.getTypeRef();
+
+            if (qName != null && !isBuiltInType(qName.getLocalPart())) {
+                final QName typeRef = createTypeRef(modelName, qName);
+                setVariable(hasVariable, variable, typeRef);
+            }
+        }
+
+        return drgElement;
+    }
+
+    QName createTypeRef(final String modelName,
+                        final QName qName) {
+        return new QName(qName.getNamespaceURI(),
+                         modelName + "." + qName.getLocalPart(),
+                         qName.getPrefix());
+    }
+
+    Name createName(final DRGElement drgElement,
+                    final String modelName) {
+        return new Name(modelName + "." + drgElement.getName().getValue());
+    }
+
+    @SuppressWarnings("unchecked")
+    void setVariable(final HasVariable hasVariable,
+                     final IsInformationItem variable,
+                     final QName typeRef) {
+        if (variable instanceof InformationItemPrimary) {
+            hasVariable.setVariable(new InformationItemPrimary(variable.getId(), variable.getName(), typeRef));
+        }
+    }
+}
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNIncludedNodeFactory.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNIncludedNodeFactory.java
@@ -22,7 +22,6 @@ import org.kie.workbench.common.dmn.api.definition.model.InformationItemPrimary;
 import org.kie.workbench.common.dmn.api.definition.model.IsInformationItem;
 import org.kie.workbench.common.dmn.api.editors.included.DMNIncludedNode;
 import org.kie.workbench.common.dmn.api.editors.included.IncludedModel;
-import org.kie.workbench.common.dmn.api.property.dmn.Id;
 import org.kie.workbench.common.dmn.api.property.dmn.Name;
 import org.kie.workbench.common.dmn.api.property.dmn.QName;
 import org.kie.workbench.common.stunner.core.util.FileUtils;
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogito.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogito.java
@@ -15,12 +15,33 @@
  */
 package org.kie.workbench.common.dmn.webapp.kogito.common.client.converters;
 
+import java.util.List;
+import java.util.Map;
+
+import elemental2.promise.Promise;
+import org.kie.workbench.common.dmn.api.definition.model.ItemDefinition;
+import org.kie.workbench.common.dmn.api.editors.included.DMNIncludedModel;
+import org.kie.workbench.common.dmn.api.editors.included.DMNIncludedNode;
+import org.kie.workbench.common.dmn.api.editors.included.IncludedModel;
 import org.kie.workbench.common.dmn.api.marshalling.DMNMarshallerImportsHelper;
 import org.kie.workbench.common.dmn.webapp.kogito.marshaller.js.model.dmn12.JSITDRGElement;
 import org.kie.workbench.common.dmn.webapp.kogito.marshaller.js.model.dmn12.JSITDefinitions;
 import org.kie.workbench.common.dmn.webapp.kogito.marshaller.js.model.dmn12.JSITImport;
 import org.kie.workbench.common.dmn.webapp.kogito.marshaller.js.model.dmn12.JSITItemDefinition;
+import org.kie.workbench.common.stunner.core.client.service.ServiceCallback;
+import org.kie.workbench.common.stunner.core.diagram.Metadata;
 
 public interface DMNMarshallerImportsHelperKogito extends DMNMarshallerImportsHelper<JSITImport, JSITDefinitions, JSITDRGElement, JSITItemDefinition> {
 
+    Promise<Map<JSITImport, JSITDefinitions>> getImportDefinitionsAsync(final Metadata metadata,
+                                                                        final List<JSITImport> imports);
+
+    void getImportedItemDefinitionsByNamespaceAsync(final String modelName,
+                                                    final String namespace,
+                                                    final ServiceCallback<List<ItemDefinition>> callback);
+
+    void loadNodesFromModels(final List<DMNIncludedModel> includedModels,
+                             final ServiceCallback<List<DMNIncludedNode>> callback);
+
+    void loadModels(final ServiceCallback<List<IncludedModel>> callback);
 }
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
@@ -19,6 +19,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -406,7 +407,7 @@ public class DMNMarshallerImportsHelperKogitoImpl implements DMNMarshallerImport
                     }
                 }
             }
-
+            result.sort(Comparator.comparing(o -> o.getName().getValue()));
             callback.onSuccess(result);
             return promises.resolve(result);
         });
--- a/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
@@ -17,6 +17,7 @@ package org.kie.workbench.common.dmn.webapp.kogito.common.client.converters;
 
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -24,12 +25,12 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Vector;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
-import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 import javax.xml.namespace.QName;
 
@@ -71,7 +72,6 @@ import static org.kie.workbench.common.dmn.api.editors.types.BuiltInTypeUtils.is
 import static org.kie.workbench.common.dmn.webapp.kogito.common.client.converters.model.ImportedItemDefinitionPropertyConverter.withNamespace;
 
 @ApplicationScoped
-@Dependent
 public class DMNMarshallerImportsHelperKogitoImpl implements DMNMarshallerImportsHelperKogito {
 
     private final KogitoResourceContentService contentService;
@@ -102,8 +102,8 @@ public class DMNMarshallerImportsHelperKogitoImpl implements DMNMarshallerImport
         if (!imports.isEmpty()) {
             return loadDMNDefinitions().then(otherDefinitions -> {
                 final Map<JSITImport, JSITDefinitions> importDefinitions = new HashMap<>();
-                for (int i = 0; i < otherDefinitions.size(); i++) {
-                    final JSITDefinitions def = Js.uncheckedCast(otherDefinitions.get(i));
+                for (final Map.Entry<String, JSITDefinitions> entry : otherDefinitions.entrySet()) {
+                    final JSITDefinitions def = Js.uncheckedCast(entry.getValue());
                     findImportByDefinitions(def, imports).ifPresent(anImport -> {
                         final JSITImport foundImported = Js.uncheckedCast(anImport);
                         importDefinitions.put(foundImported, def);
@@ -116,15 +116,18 @@ public class DMNMarshallerImportsHelperKogitoImpl implements DMNMarshallerImport
     }
 
     Promise<Map<String, JSITDefinitions>> loadDMNDefinitions() {
-        final Map<String, JSITDefinitions> otherDefinitions = new HashMap<>();
-        return contentService.getFilteredItems(DMN_FILES_PATTERN, ResourceListOptions.assetFolder()).then(list -> {
-            final List<Promise<JSITDefinitions>> fileLoadPromises = new ArrayList<>();
-            for (final String file : list) {
-                final Promise<JSITDefinitions> promise = loadDefinitionFromFile(file, otherDefinitions);
-                fileLoadPromises.add(promise);
-            }
-            return promises.all(fileLoadPromises.toArray(new Promise[]{})).then(v -> promises.resolve(otherDefinitions));
-        });
+
+        return contentService.getFilteredItems(DMN_FILES_PATTERN, ResourceListOptions.assetFolder())
+                .then(list -> {
+                    if (list.length == 0) {
+                        return promises.resolve(new HashMap<>());
+                    } else {
+                        final Map<String, JSITDefinitions> otherDefinitions = new ConcurrentHashMap<>();
+                        return promises.all(Arrays.asList(list),
+                                            (String file) -> loadDefinitionFromFile(file, otherDefinitions))
+                                .then(v -> promises.resolve(otherDefinitions));
+                    }
+                });
     }
 
     @Override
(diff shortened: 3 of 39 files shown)


danielzhe:

karreiro:
Jenkins please retest this.

danielzhe:

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-marshaller/src/main/resources/org/kie/workbench/common/dmn/webapp/kogito/marshaller/js/MainJs.js
```diff
@@ -86,21 +86,25 @@ MainJs = {
         callback(toReturn);
     },
 
-    marshall: function (value, defaultNamespace, callback) {
+    marshall: function (value, namespaces, callback) {
+        var namespacesValues = {};
+        if (namespaces instanceof String) {
```
manstis, line 91: Won't this always be `false` for DMN, since we now provide a map?

danielzhe: Yes, but then it doesn't break for Scesim that stills using String.
The idea is: if is String, use the old behavior (for Scesim).
If is map, use the new behavior (for DMN).
Isn't that right? @manstis

karreiro: @danielzhe I'm not sure if I'm missing something. But I believe SceSim _never_ calls the marshall method.

manstis: @karreiro _this_ marshall method is in the DMN marshaller so is never called by _scesim_. It does however call its own generated marshall method [here](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/editor/ScenarioSimulationEditorKogitoWrapper.java#L373).

danielzhe: OK, I'm confused now. Isn't that what you was pointing [here](https://github.com/kiegroup/gwt-jsonix-schema-compiler/pull/8#issuecomment-632128604), @manstis ?

danielzhe: Ok, I get it now, @manstis .

danielzhe:

jomarko:

jomarko:
jenkins please retest this

danielzhe:

danielzhe:
This was failing because of [this](https://github.com/kiegroup/gwt-jsonix-schema-compiler/pull/8#discussion_r429835655).

danielzhe:
jenkins please retest this

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerKogitoUnmarshaller.java
```diff
@@ -559,25 +567,27 @@ private Node getRequiredNode(final Map<String, Entry<JSITDRGElement, Node>> elem
...
-        return dmnShapes
-                .stream()
-                .map(shape -> {
-                    final String dmnElementRef = getDmnElementRef(shape);
-                    final Optional<JSITDRGElement> ref = getReference(importedDRGElements, dmnElementRef);
-                    return ref.orElse(null);
-                })
-                .filter(Objects::nonNull)
-                .collect(toList());
+        //  final Map<String, String> indexByUri = NameSpaceUtils.extractNamespacesKeyedByUri(dmnXml);
```
karreiro, line 570: Should this block be removed? Considering: https://github.com/kiegroup/kie-wb-common/pull/3287.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerKogitoUnmarshaller.java
```diff
@@ -559,25 +567,27 @@ private Node getRequiredNode(final Map<String, Entry<JSITDRGElement, Node>> elem
...
+        // updateIDsWithAlias(indexByUri, importedDRGElements);
```
karreiro, line 571: I believe we can remove the non-used `updateIDsWithAlias` method as well.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogito.java
```diff
@@ -15,12 +15,33 @@
...
+                                                                        final List<JSITImport> imports);
+
+    void getImportedItemDefinitionsByNamespaceAsync(final String modelName,
+                                                    final String namespace,
+                                                    final ServiceCallback<List<ItemDefinition>> callback);
+
+    void loadNodesFromModels(final List<DMNIncludedModel> includedModels,
+                             final ServiceCallback<List<DMNIncludedNode>> callback);
+
+    void loadModels(final ServiceCallback<List<IncludedModel>> callback);
```
jomarko, line 46: - Do I understand correctly all these methods works asynchronously?
- What is the difference between those suffixed "Async" and those not suffixed
- What is the reason somewhere we return Promise and somewhere we use Callback

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,253 @@
...
+import org.kie.workbench.common.stunner.core.util.FileUtils;
+import org.kie.workbench.common.stunner.core.util.StringUtils;
 import org.uberfire.backend.vfs.Path;
+import org.uberfire.client.promise.Promises;
+
+import static org.kie.workbench.common.dmn.api.editors.types.BuiltInTypeUtils.isBuiltInType;
+import static org.kie.workbench.common.dmn.webapp.kogito.common.client.converters.model.ImportedItemDefinitionPropertyConverter.withNamespace;
 
 @ApplicationScoped
+@Dependent
```
jomarko, line 73: Could you please explain combination of @Dependent with @ApplicationScoped? For me the second one represnts singlotene per app while the first multiple instances available per app. I am not sure how we can combine them.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,253 @@
...
+
+    Promise<List<JSITDefinitions>> loadDMNDefinitions() {
+        final List<JSITDefinitions> otherDefinitions = new Vector<>();
+        return contentService.getFilteredItems(DMN_FILES_PATTERN).then(list -> {
+            final List<Promise<JSITDefinitions>> fileLoadPromises = new ArrayList<>();
+            for (final String file : list) {
+                final Promise<JSITDefinitions> promise = loadDefinitionFromFile(file, otherDefinitions);
+                fileLoadPromises.add(promise);
+            }
+            return promises.all(fileLoadPromises.toArray(new Promise[]{})).then(v -> promises.resolve(otherDefinitions));
```
jomarko, line 125: Shouldn't we say the array size as we know it?
```java
return promises
                    .all(fileLoadPromises.toArray(new Promise[fileLoadPromises.size()]))
                    .then(v -> promises.resolve(otherDefinitions));
```

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,253 @@
...
+                                                                                   final Optional<DMNIncludedModel> diagramImport = getDiagramImport(item, includedModels);
+                                                                                   final List<DMNIncludedNode> nodes = diagramImport
+                                                                                           .map(dmn -> diagramUtils
+                                                                                                   .getDRGElements(item)
+                                                                                                   .stream()
+                                                                                                   .map(node -> includedModelFactory.makeDMNIncludeNode(path, dmn, node))
+                                                                                                   .collect(Collectors.toList()))
+                                                                                           .orElse(new ArrayList<>());
```
jomarko, line 148: Here we can use `.orElse(Collections.emptyList())` as we are not manipulating with `nodes` afterwards.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,253 @@
...
+                                                                                   LOGGER.log(Level.SEVERE, error.getMessage());
+                                                                               }
+                                                                           });
+                                                                           return promises.resolve();
+                                                                       }
+            );
+            fileLoadPromises.add(promise);
+        }
+
+        promises.all(fileLoadPromises.toArray(new Promise[]{})).then(p -> {
```
jomarko, line 164: Similar question as above . shouldn't we leverage known size of `fileLoadPromises` during new array creation?

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,253 @@
...
+                    diagramService.transform(fileContent, new ServiceCallback<Diagram>() {
+                        @Override
+                        public void onSuccess(final Diagram item) {
+                            final String modelPackage = "";
+                            final Diagram<Graph, Metadata> diagram = (Diagram<Graph, Metadata>) item;
+                            final String namespace = diagramUtils.getNamespace(diagram);
+                            final String importType = DMNImportTypes.DMN.getDefaultNamespace();
+                            final int drgElementCount = diagramUtils.getDRGElements(diagram).size();
+                            final int itemDefinitionCount = diagramUtils.getDefinitions(diagram).getItemDefinition().size();
+                            // We get files only from the same folder, so doesn't need relative or full path.
```
jomarko, line 188: Could we mentions this aslo in javadoc of the added public API?

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,253 @@
...
+                            LOGGER.log(Level.SEVERE, error.getMessage());
+                        }
+                    });
+                    return promises.resolve();
+                });
+
+                fileLoadPromises.add(fileLoad);
+            }
+
+            return promises.all(fileLoadPromises.toArray(new Promise[]{})).then(v -> promises.resolve());
```
jomarko, line 210: Array instantiation question as above.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-marshaller/src/main/resources/org/kie/workbench/common/dmn/webapp/kogito/marshaller/js/MainJs.js
```diff
@@ -86,21 +86,13 @@ MainJs = {
...
+    marshall: function (value, namespacesValues, callback) {
         // Create Jsonix context
-        var namespaces = {};
-        namespaces[defaultNamespace] = "";
-        namespaces["http://www.omg.org/spec/DMN/20180521/MODEL/"] = "dmn";
-        namespaces["http://www.omg.org/spec/DMN/20180521/DI/"] = "di";
-        namespaces["http://www.drools.org/kie/dmn/1.2"] = "kie";
-        namespaces["http://www.omg.org/spec/DMN/20180521/DMNDI/"] = "dmndi";
-        namespaces["http://www.omg.org/spec/DMN/20180521/DC/"] = "dc";
-        namespaces["http://www.omg.org/spec/DMN/20180521/FEEL/"] = "feel";
```
jomarko, line 98: Is this change because we now parse real namespaces form given file?

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testing/src/main/java/org/kie/workbench/common/dmn/showcase/client/alternatives/KogitoResourceContentService.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Inject
+    public KogitoResourceContentService(final ResourceContentService resourceContentService,
+                                        final Promises promises) {
+        super(resourceContentService, promises);
+    }
+
+    @Override
+    public void getFilteredItems(final String pattern,
+                                 final RemoteCallback<List<String>> callback,
```
jomarko, line 45: Could we please rename to sucessCallback or something more explicitely saying that is callback in case no issues during computation happened?

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testing/src/main/java/org/kie/workbench/common/dmn/showcase/client/alternatives/KogitoResourceContentService.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public Promise<String> loadFile(final String fileUri) {
+        if (BASE_MODEL.equals(fileUri)) {
+            return Promise.resolve(DMNClientModels.BASE_FILE);
+        } else if (MODEL_WITH_IMPORTS.equals(fileUri)) {
+            return Promise.resolve(DMNClientModels.MODEL_WITH_IMPORTS);
+        } else {
+            return Promise.resolve("NOT FOUND!");
+        }
+    }
```
jomarko, line 78: From my point of view in the default `else` branches it is more 'not supported' or 'illegal state'. because new comer can add his file into `DMNClientModels` class. In such case message about not found can be confusing. What do you think?

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testing/src/main/java/org/kie/workbench/common/dmn/showcase/client/navigator/DMNDiagramsNavigatorScreen.java
```diff
@@ -86,6 +96,34 @@ public void edit() {
...
+        return null;
+    }
+
+    private RemoteCallback<List<String>> getItems() {
+        return response -> {
+            GWT.log("PATHS:");
+            for (final String s : response) {
+                GWT.log(s);
+            }
+            vfsService.openFile("something.dmn", DMNClientModels.MODEL_WITH_IMPORTS);
```
jomarko, line 123: Could you please explain what effect has this 'something.dmn' value? I mean is that reflected somehow in UI or just the `DMNClientModels.MODEL_WITH_IMPORTS` is important?

kie-wb-common-kogito/kie-wb-common-kogito-webapp-base/src/main/java/org/kie/workbench/common/kogito/webapp/base/client/workarounds/KogitoResourceContentService.java
```diff
@@ -104,4 +105,12 @@ public void getFilteredItems(final String pattern,
                     return null;
                 });
     }
+
+    public Promise<String[]> getFilteredItems(final String pattern) {
+        return resourceContentService.list(pattern);
+    }
+
+    public Promise<String> loadFile(final String fileUri) {
+        return resourceContentService.get(fileUri);
```
jomarko, line 114: to be consistent, we could add javadoc

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/util/FileUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+package org.kie.workbench.common.stunner.core.util;
+
+public class FileUtils {
+
+    private static final char UNIX_SEPARATOR = '/';
+    private static final char WINDOWS_SEPARATOR = '\\';
+
+    /**
+     * Gets the file name without the Windows extension.
```
jomarko, line 25: Is 'Windows' correct in this javadoc?

jomarko:
On my machine. files are listed twice, using `dmn-quarkus-demo` folder from `kogito-examples`

Screenshot from 2020-05-27 20-31-15

jomarko:
Even if I included non empty file, the panel right bottom says nothing included
Screenshot from 2020-05-27 20-39-01

jomarko:
I need to stop review as I am not able to use nodes of the included model. Could you please have a look if you can reproduce?

jomarko:
@danielzhe the 'bug' reported by sonar cloud seems valid to me. We call get(i) where i is int, while we declared the keys of given map as String. Do I miss sth?

danielzhe:
@jomarko Hi! I'm still working on the PR, just pushed to move it to another machine. I let you know when it's all ready for your re-review. Thank you!

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNIncludedNodeFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static org.kie.workbench.common.dmn.api.editors.types.BuiltInTypeUtils.isBuiltInType;
+
+public class DMNIncludedNodeFactory {
+
+    DMNIncludedNode makeDMNIncludeNode(final String path,
+                                       final IncludedModel includeModel,
+                                       final DRGElement drgElement) {
+        return new DMNIncludedNode(FileUtils.getFileName(path), drgElementWithNamespace(drgElement, includeModel));
+    }
```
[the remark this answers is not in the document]
danielzhe: It doesn't matter if it is relative or absolute. We just need the filename.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogito.java
danielzhe: 1. Yes, they are async. I didn't know exactly how to differentiate one method from other, so I used the same pattern that I see in some C# projects (using "async" prefix to differentiate async from sync methods). I don't know if this is the best name here. I'm open to suggestions.

2. The difference is that the methods without "async" run synchronously and are not supported in Kogito, specially VSCode, because VSCode is multithread which lead to some issues that we don't have in singlethread environments like in Business Central. The root of all this is in Content API, that is asynchronous. So the main issue is when loading imported models during unmarshalling: I'm running a thread and I can continue only when the threads that load all the imported models are done. Which bring us to the question about Promise and Callback:

3. We're using callback when is not required to wait for some thread to finish. But Promise gives us the ability to call some code and continue the main flow only when that called code are finished. In other hand, with callbacks, the main flow doesn't wait for it to be finished.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-marshaller/src/main/resources/org/kie/workbench/common/dmn/webapp/kogito/marshaller/js/MainJs.js
danielzhe: Yes. We have to make this change because the included models create new namespaces that are not hardcoded here.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testing/src/main/java/org/kie/workbench/common/dmn/showcase/client/navigator/DMNDiagramsNavigatorScreen.java
danielzhe: It can be anything. This method is just to load the test DMN models defined in `DMNClientModels` by the "Load Models From Client" button in our test project.

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/util/FileUtils.java
danielzhe: I thought it was, but by your question I think is not.

danielzhe:
@karreiro @jomarko Thank you for your review! Ready for you to continue the review. Unfortunately I had to make some refactors in[ this class](https://github.com/kiegroup/kie-wb-common/pull/3310/files#diff-178b684b4ef05ea71d906fe077517647R75), so you may want to check it again.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,257 @@
...
+                    final JSITDefinitions def = Js.uncheckedCast(entry.getValue());
+                    findImportByDefinitions(def, imports).ifPresent(anImport -> {
+                        final JSITImport foundImported = Js.uncheckedCast(anImport);
+                        importDefinitions.put(foundImported, def);
+                    });
+                }
+                return promises.resolve(importDefinitions);
+            });
+        }
+        return promises.resolve(new HashMap<>());
```
jomarko, line 115: Seems to me we can use `Colelctions.emptyMap()` as the resolved map is just iterated.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,257 @@
...
+    }
+
+    Promise<Map<String, JSITDefinitions>> loadDMNDefinitions() {
+
+        return contentService.getFilteredItems(DMN_FILES_PATTERN, ResourceListOptions.assetFolder())
+                .then(list -> {
+                    if (list.length == 0) {
+                        return promises.resolve(new HashMap<>());
```
jomarko, line 123: maybe also here we could use `Collections.emptyMap()`

jomarko:
@danielzhe user can not resize column width of the imported model see https://issues.redhat.com/secure/attachment/12474023/imported-node-column-width.webm

jomarko:
### Manual review 01
https://issues.redhat.com/browse/KOGITO-2347

jomarko:
### Manual review 02
Loaded models are not ordered. Could we fix that ir that is not possible due to async load?
Screenshot from 2020-06-02 11-22-10

jomarko:
### Manual review 03
I am not sure about this scenario. use VS code to follow it.
- Create a model **a.dmn** with one node inside, save and close **a.dmn**
- In same folder create a model **b.dmn**
- Include **a.dmn** into **b.dmn**
- Use the node of **a.dmn** in the **b.dmn** 
- Save and close **b.dmn**
- Remove **a.dmn**
- Reopen **b.dmn**

The good thing is that user is not blocked from using the editor even if included model was removed. However user is not informed that something happened. The previously used node is not shown anymore and the included model card reports 0 data types and nodes inside. 

What are your thoughts please?

jomarko:
### Manual review 04
https://issues.redhat.com/browse/KOGITO-2348

danielzhe:
Thank you for your review, @jomarko !
About the issue 2: I think we can sort it alphabetical in our side, after all nodes are loaded. Wdyt?

About the issue 3: I think this is a problem. We can't prevent the user from delete the file, but what we can do is inform the user that a imported model was not found. What do you think?

@karreiro, wdyt too?

danielzhe:
@jomarko Thank you. Changes applied! Now the list of files are ordered. Also as discussed in Kogito channel, we will cover only VSCode in this PR, since it is the only channel that supports read "external" files at this moment. @manstis @karreiro

jomarko:
@danielzhe thank you for fix of sorting, do we have please jira for point 3?

danielzhe:
Thank you, @jomarko. JIRA created: https://issues.redhat.com/browse/KOGITO-2382
Also I fixed the failing test.

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,256 @@
...
+                    .then(existingDefinitions -> promises.all(includedModels, model -> loadNodes(existingDefinitions, model, result))
+                            .then(p ->
+                                  {
+                                      callback.onSuccess(result);
+                                      return promises.resolve();
+                                  }));
+        }
+    }
+
+    private Promise loadNodes(final Map<String, JSITDefinitions> existingDefinitions,
```
karreiro, line 150: IntelliJ is warning about this method, I think we can solve it by using:
```suggestion
    private Promise<List<DMNIncludedNode>> loadNodes(final Map<String, JSITDefinitions> existingDefinitions,
```

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-common/src/main/java/org/kie/workbench/common/dmn/webapp/kogito/common/client/converters/DMNMarshallerImportsHelperKogitoImpl.java
```diff
@@ -16,29 +16,256 @@
...
+                        }
+                    });
+                    return promises.resolve();
+                })).then(v -> {
+                    callback.onSuccess(models);
+                    return promises.resolve();
+                }));
+    }
+
+    Promise loadDefinitionFromFile(final String file,
```
karreiro, line 223: IntelliJ is warning about this method, I think we can solve it by using:
```suggestion
    Promise<Void> loadDefinitionFromFile(final String file,
```

karreiro:
@danielzhe I've noticed that nodes with custom data types are losing their types with included models. See:

Node in the original model:
Screen Shot 2020-06-04 at 18 10 09

Node included into another model:
Screen Shot 2020-06-04 at 18 10 01

danielzhe:
@karreiro Thank you! Changes applied! It also fixed the issue we found talking in private chat about files from other vendors.

A few comments, that I think it is interesting to share:
1. About Issue 1, it was a very interesting issue. Apparently, every time you move a node or do any small changes in the canvas, VSCode calls "save". What does it means? It means that the marshaling is called. And what is the problem about that? The problem is that the Kogito marshaling has the same logic used in standalone marshaling, which removes the imported items definitions during marshaling, but the Kogito marshaling does that using the concrete object, the real current diagram, which means that the imported type [was being cleared](https://github.com/kiegroup/kie-wb-common/pull/3310/files#diff-a827063512efe99192ec2110035dc86cL151). Now we are not removing, [just ignoring it](https://github.com/kiegroup/kie-wb-common/pull/3310/files#diff-5b051f08ac22de13bb5b29fe7c1a0442R159).
We should have the same issue in standalone after you save the file, but I was unable to reproduce. I didn't investigate further, but I think that standalone uses a copy of the real diagram OR re-adds the imported items definitions somewhere.

2. This issue was also spotted by @jomarko [here](https://github.com/kiegroup/kie-wb-common/pull/3310#issuecomment-637434756) and we decided to open a [JIRA](https://issues.redhat.com/browse/KOGITO-2382) for it.

manstis:
I suspect VSCode calling save after every interaction related to support for undo/redo in VSCode. @ederign can you advise/confirm? It sure feels heavy weight (for large files) but if that's VSCode's strategy then I doubt there's much we can about it.

PR #3316, opened 2020-05-29
DROOLS-3685: [DMN Designer] Boxed expressions - Decision Table - Output clause - Creating Decision Table, sub-columns are auto-created for structured type

vpellegrino:
*Please refer to:* https://issues.redhat.com/browse/DROOLS-3685

*Improvement description:* Creating Decision Table, sub-columns should be auto-created for structured type

*Proposed solution:* Changing `DecisionTableEditorDefinitionEnricher.java` in order to consider Decision's type for the output columns creation

*Outcome:*

Custom data type:
image
Decision type:
image
Decision table:
image

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java
@@ -63,6 +63,7 @@ import org.kie.workbench.common.stunner.core.graph.Node;
 import org.kie.workbench.common.stunner.core.graph.content.definition.Definition;
 
 import static org.kie.workbench.common.dmn.api.property.dmn.QName.NULL_NS_URI;
+import static org.kie.workbench.common.stunner.core.util.StringUtils.isEmpty;
 
 @ApplicationScoped
 public class DecisionTableEditorDefinitionEnricher implements ExpressionEditorModelEnricher<DecisionTable> {
@@ -98,12 +99,6 @@ public class DecisionTableEditorDefinitionEnricher implements ExpressionEditorMo
             inputClause.setInputExpression(literalExpression);
             dtable.getInput().add(inputClause);
 
-            final OutputClause outputClause = new OutputClause();
-            outputClause.setName(DecisionTableDefaultValueUtilities.getNewOutputClauseName(dtable));
-            final HasTypeRef hasTypeRef = TypeRefUtils.getTypeRefOfExpression(dtable, hasExpression);
-            outputClause.setTypeRef(!Objects.isNull(hasTypeRef) ? hasTypeRef.getTypeRef() : BuiltInType.UNDEFINED.asQName());
-            dtable.getOutput().add(outputClause);
-
             final RuleAnnotationClause ruleAnnotationClause = new RuleAnnotationClause();
             ruleAnnotationClause.getName().setValue(DecisionTableDefaultValueUtilities.getNewRuleAnnotationClauseName(dtable));
             dtable.getAnnotations().add(ruleAnnotationClause);
@@ -113,9 +108,7 @@ public class DecisionTableEditorDefinitionEnricher implements ExpressionEditorMo
             decisionRuleUnaryTest.getText().setValue(DecisionTableDefaultValueUtilities.INPUT_CLAUSE_UNARY_TEST_TEXT);
             decisionRule.getInputEntry().add(decisionRuleUnaryTest);
 
-            final LiteralExpression decisionRuleLiteralExpression = new LiteralExpression();
-            decisionRuleLiteralExpression.getText().setValue(DecisionTableDefaultValueUtilities.OUTPUT_CLAUSE_EXPRESSION_TEXT);
-            decisionRule.getOutputEntry().add(decisionRuleLiteralExpression);
+            buildOutputClausesByDataType(hasExpression, dtable, decisionRule);
 
             final RuleAnnotationClauseText ruleAnnotationEntry = new RuleAnnotationClauseText();
             ruleAnnotationEntry.getText().setValue(DecisionTableDefaultValueUtilities.RULE_DESCRIPTION);
@@ -125,11 +118,9 @@ public class DecisionTableEditorDefinitionEnricher implements ExpressionEditorMo
 
             //Setup parent relationships
             inputClause.setParent(dtable);
-            outputClause.setParent(dtable);
             decisionRule.setParent(dtable);
             literalExpression.setParent(inputClause);
             decisionRuleUnaryTest.setParent(decisionRule);
-            decisionRuleLiteralExpression.setParent(decisionRule);
             ruleAnnotationEntry.setParent(dtable);
 
             if (nodeUUID.isPresent()) {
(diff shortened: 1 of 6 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java
```diff
@@ -152,6 +145,77 @@ public void enrich(final Optional<String> nodeUUID,
...
+    }
+
+    private boolean typeRefDoesNotMatchAnyDefinition(final QName typeRef) {
+        return !typeRefMatchesBuiltInType(typeRef) &&
+                dmnGraphUtils.getDefinitions().getItemDefinition()
+                        .stream()
+                        .noneMatch(typeRefIsCustom(typeRef));
+    }
+
+    private OutputClause buildOutputClause(final DecisionTable dtable, final DecisionRule decisionRule, final QName typeRef, final String text) {
```
manstis, line 205: I have issue with this method; since it returns an `OutputClause` that modifies the column collection _outside_ this method; however it adds a `LiteralExpression` to the `DecisionRule` _inside_ this method. i.e. IMO the mutations should either _both_ happen inside this method or _both_ outside this method. This is my opinion. If others feel less so; then we can disregard it.

vpellegrino: You are right. Good catch.

vpellegrino:
> If the `DecisionTable` is nested with a `ContextEntry` the `OutputClause` columns are not created as I'd expect (i.e. they should match the _type_ of the _parent_ which in this case is the `ContextEntry` and not a `Decision`). The same applies for use in a `BusinessKnowledgeModel`.
> 
> **ContextEntry**
> ContextOutputTypeRef
> 
> This is probably caused by [this](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java#L150) call that overwrites the columns this PR creates with those [here](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java#L278).
> 
> **BusinessKnowledgeModel**
> BusinessKnowledgeModelOutputTypeRef
> 
> Happy for these to be addressed as another JIRA if need be.
> 
> Otherwise LGTM.

Thanks for pointing them out
No problem in solving those cases in this PR.
I would just need a bit of time to understand the better strategy to accomplish them.

vpellegrino:
@karreiro thank you for your contribution.
I agree with you. 
I'd say that, in order to have incremental value, it would be better to merge this PR and manage those issues into separate JIRAs.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java
```diff
@@ -260,6 +325,21 @@ void addInputClauseRequirement(final ItemDefinition itemDefinition,
...
+    private boolean typeRefMatchesBuiltInType(final QName typeRef) {
+        for (BuiltInType bi : BuiltInType.values()) {
+            for (String biName : bi.getNames()) {
+                if (Objects.equals(biName, typeRef.getLocalPart())) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
jomarko, line 337: This can be simplified:
```java
private boolean typeRefMatchesBuiltInType(final QName typeRef) {
        return BuiltInTypeUtils.isBuiltInType(typeRef.getLocalPart());
    }
```

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/dtable/DecisionTableEditorDefinitionEnricher.java
```diff
@@ -260,6 +325,21 @@ void addInputClauseRequirement(final ItemDefinition itemDefinition,
...
+
+    private Predicate<ItemDefinition> typeRefIsCustom(final QName typeRef) {
+        return itemDef -> itemDef.getName().getValue().equals(typeRef.getLocalPart());
+    }
```
jomarko, line 341: Could we please reimplement it using `Objects.equals`? That will not throw an NPE in case `itemDef.GetName().getValue()` or `typeRef.getLocalPart()` are null.

```java
private Predicate<ItemDefinition> typeRefIsCustom(final QName typeRef) {
        return itemDef -> Objects.equals(itemDef.getName().getValue(), typeRef.getLocalPart());
    }
```

jomarko:
### Manual review
Question about the behavior in [jira comment](https://issues.redhat.com/browse/DROOLS-3685?focusedCommentId=14130275&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14130275).

vpellegrino:
@karreiro I just fixed point 3)

image
image

karreiro:
Excellent stuff @vpellegrino. I've noticed you fixed for input clauses as well, thanks! :-)

jomarko:
@vpellegrino we have one failing test, probably needs updates due to our changes.
```
at org.kie.workbench.common.dmn.client.editors.types.common.DataTypeManagerTest.testMakeDataTypeFromItemDefinitionWithAnImportedItemDefinition(DataTypeManagerTest.java:398)
```

vpellegrino:
@karreiro @manstis @jomarko in the last commit I accomplished also the case of decision table included in context:
image
image
What is missing, is the case of table included in BKM, that it seems not easy to accomplish.
If PR is ok for you, could you please merge it? I could trace a JIRA for the missing point of BKM.

jomarko:
@vpellegrino ok, I thought you are going to do it separately, need to re-review then, thank you.

vpellegrino:
@jomarko sorry for the confusion. In the meanwhile, @manstis gave me the hint of setting the missing parentship relationship btw bkm and function

manstis:
@vpellegrino It is just the use of Decision Tables nested in a Function that causes the [issue](https://github.com/kiegroup/kie-wb-common/pull/3316#pullrequestreview-424761202) reported by @karreiro 

I noticed that the output columns of a Decision Table when chosen as the _default_ for a `Context` _could_ possibly be automatically created, but that's another edge case and can be the subject of another JIRA.

This image shows a `Context`. The `ContextEntry` has a `Function` as it's expression type; that in turn has a `DecisionTable` (so the decision table is nested in the `Function`) and the _default_ value of the `Context` has also been set to a `DecisionTable` (I suspect the output columns could have been created to match the typeRef of the `Decision` i.e. `tPerson`).

Screenshot from 2020-06-05 08-33-41

vpellegrino:
> Thanks for the updates, @vpellegrino!
> 
> I've noticed this minor issue:
> 
> Screen Shot 2020-06-04 at 16 24 25

The issue seems to be not related to my PR:
image

Traced a new JIRA defect for it. Feel free to change description: https://issues.redhat.com/browse/DROOLS-5416

vpellegrino:
@kiegroup/gatekeepers can you please merge this PR?

PR #3340, opened 2020-06-22
JBPM-9198 Error during startup

bxf12315:


domhanak:
This is a stale PR @bxf12315 please close or advice how to proceed with this PR.

caponetto:
IMO this PR can be closed. 
We had to upgrade the JGIT library as part of the JDK11 upgrades, 
so I believe the warning message in question is no longer being thrown.

domhanak:
@caponetto Thanks, I will close as out of date. If that is not correct, please reopen the PR again.:)

PR #2659, opened 2019-05-22
JBPM-7117 Stunner - Inline text editor & usability improvements

treblereel:
JIRA : https://issues.jboss.org/browse/JBPM-7117

Depends on https://github.com/kiegroup/kie-wb-common/pull/2659

PR #3342, opened 2020-06-23
KOGITO-2548: Focus name input for node renaming

jomarko:


kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/TextEditorMultiLineBoxView.java
```diff
@@ -80,6 +80,7 @@ public void show(final String name) {
         nameField.setTextContent(name);
         nameField.setRows(2);
         nameField.setCols(25);
+        nameField.focus();
```
manstis, line 83: Hi @jomarko this is in essence what makes the current (Node) "name" editors User experience target for complaints. IIRC the popup editor is not always easy to display nor is its `TextBox` set to have the focus reliably. There may be some defer command executing somewhere that causes the problems. IDK. This is why these editors were being redesigned and rewritten. Here's the original [PR](https://github.com/kiegroup/kie-wb-common/pull/2659) that references the JIRA relating to it too.

If you need the `focus()` to be called (presumably there's a `onBlur()` handler somewhere to de-register keyboard shortcut handling?) then you could try adding it to the _show_ command executed [here](https://github.com/kiegroup/kie-wb-common/blob/430c6ad8594ad464234b5cb11488275d6a612612/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/main/java/org/kie/workbench/common/stunner/client/widgets/canvas/actions/AbstractTextEditorBoxView.java#L66) and invoked by `scheduleDeferredCommand(...)` e.g. 
```
this.showCommand = () -> {
    this.getElement().getStyle().setProperty("display", "block");
    scheduleDeferredCommand(()-> AbstractTextEditorBoxView.this.setFocusToEditableElement());
}
```
You'd also need to add an abstract method `protected abstract void setFocusToEditableElement()` that is implemented by both sub-classes of `AbstractTextEditorBoxView` and set focus on the `TextBox` or `TextArea` as applicable.

It's possibly worth a try.

jomarko:
Will be addressed by wider changes coming soon.

PR #3427, opened 2020-09-22
JBPM-9352: Stunner - Warning when getContent is called while the model is invalid

romartin:
Hey @paulovmr @ederign 

As discussed with @porcelli , fixing the actual blocker issue by completely disabling validation in the bpmn editor.

See comments in https://issues.redhat.com/browse/JBPM-9352

BTW You know who should be the QE tester for this, if any? Can't see anyone on the ticket yet.

Thanks!

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-kogito-runtime/src/main/java/org/kie/workbench/common/stunner/kogito/client/editor/BPMNDiagramEditor.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-kogito-runtime/src/main/java/org/kie/workbench/common/stunner/kogito/client/editor/BPMNDiagramEditor.java
@@ -15,9 +15,6 @@
  */
 package org.kie.workbench.common.stunner.kogito.client.editor;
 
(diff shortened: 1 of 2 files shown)


romartin:
Jenkins do fdb

romartin:
Jenkins do cdb

porcelli (approved):

porcelli (commented):
maybe add a test that will make sure that no validation is emitted on the getContent method call?

romartin:
@porcelli in fact I removed the code that calls validation, even the validator injections, in order to don't have dead code (we can inject later on as well, once fixing it). Also removed the validator instance from the test case, so it's for sure not calling validation now, I don't think I can ensure validation is not being called because of the class is now not using any validator.

LuboTerifaj:
Hello @domhanak would you mind to have a look at this PR? Thank you!

romartin:
hey @ederign @paulovmr @domhanak @LuboTerifaj @porcelli 
For me this fix is ready, please let me know some feedback asap. 
Thanks guys!

domhanak:
Thanks @romartin I tried to reproduce the warning but I can't with these changes.
I was able to save a BPMN process in invalid state.

Only time I couldn't was when I added sequence flow without target node. A notification that save operation was cancelled pops up. I believe this is correct behavior as sequence flow without target does not make sense. 
Approved.

domhanak (approved):

paulovmr (approved):

PR #3434, opened 2020-09-23
RHDM-1346: Enhace alert messages for asset validation

karreiro:
**JIRA**: [RHDM-1346](https://issues.redhat.com/browse/RHDM-1346): RHDM-1346: Enhace alert messages for asset validation

This change extends the `ClientTranslationService`  with the `DiagramElementNameProvider` interface (keeping the "old" behaviour as the default implementation).

**Artifact (it happens on Business Central only)**: [WAR file](https://www.dropbox.com/s/yvpt9lzs9fkrt7b/business-central-RHDM-1346.war?dl=0)

**Before:**

**After:**

---

<pre>
To retest a PR or trigger a specific build please add a comment:

</pre>


karreiro:
Jenkins do fdb.

romartin (commented):
Hey @karreiro 

Thanks for this work, messages are looking much better!

On the other hand, at the code side, I think there is no need for the new `DiagramElementNameProvider` and all the implementation and usages around it. Actually is possible from code to obtain the "name" for any Stunner domain, consider:

- Stunner domain is annotated in a way that, for each type in the domain, to obtain the "name" field for it, for example see https://github.com/kiegroup/kie-wb-common/blob/3beea666d6a8b9db8617bc24ac111982d1a95972/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/TextAnnotation.java#L73
- Once Stunner knows the field that represents the name for a type, you can for example use https://github.com/kiegroup/kie-wb-common/blob/3bac96154ad436471b707bf89137e71d18589dff/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/DefinitionUtils.java#L69

Could it be possible to avoid then creating a new layer (around `DiagramElementNameProvider`) on top of the "name" stuff, by just using actual API? I would prefer not adding more complexity on top of domain operations... WDYT? 

Thanks!!
Roger

Rikkola:
Is it possible to have two nodes with the same name in the same file? 
Even if it is, I doubt there is a way to make a clean way to know the difference with the single description line. Maybe it is enough to  let the user search by name and problem description.

karreiro:
Thank you for your review, @romartin and @Rikkola!

@romartin, the approach you've described solves most of cases. However, this mechanism assumes that all UUIDs are [associated with a graph node](https://github.com/kiegroup/kie-wb-common/blob/3bac96154ad436471b707bf89137e71d18589dff/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/i18n/ClientTranslationService.java#L88-L102). The DMN model validation can have UUIDs associated with elements that are not present in the graph like: Decision Table columns, Contexts, Literal Expressions, Data Types, and so on. Thus, for theses use cases, I believe we need to rely on this kind of extension. Wdyt?

@Rikkola, it's possible to have two nodes with the same name, but both nodes would represent the same DRG element. So users are able to fix the issue no matter which node they select :)

@danielzhe @jstastny-cz I've updated the PR code and the description. It had a bug related to the name of DMN elements. Given the current DMN validation errors, the name above the message was redundant. So, I've removed the first line. Wdyt?

Thank you all guys!

karreiro:
Jenkins do fdb.

Rikkola:
The feature is much needed, thank you for implementing. 👍

danielzhe (approved):
Very good addition! Thank you, @karreiro !

jstastny-cz (approved):
LGTM

romartin:
hey @karreiro ,

Thanks for the explanation, it makes sense having some new method for this, but I think we could do some last effort trying merging the existing and your new abilities around element's "name". Lemme explain.
So if I understand properly, there are elements which name is not being managed by the Stunner's core annotations, so it makes sense to have some api like `DiagramElementNameProvider`, but should be good "joining" this new ability with the actual API in some way, because actually once you need the "name" for some element you can just use the `DefinitionAdapter` API as: `definitionAdapter.getMetaPropertyField`, and from the field you can obtain the value for the name. In fact there exist a helper method for this (see https://github.com/kiegroup/kie-wb-common/blob/3bac96154ad436471b707bf89137e71d18589dff/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/util/DefinitionUtils.java#L69), but the idea is that there exist a single "entry point" for obtaining a name for an element.
Now I see we end up on 2 classes to check for obtaining a name for an element: `DefinitionAdapter` & `DiagramElementNameProvider`, so maybe we could create some new single method (somewhere) that manages this duplicity and check both adapter & provider. Otherwise maybe we'll end up on issues later on if somebody tries to get the name for an element but not caring about this duplicity.
WDYT?

romartin:
Also another Q plz @karreiro - I see only one DMN implementation for `DiagramElementNameProvider`, and it's not even implementing the `getElementName` method, so it shoud still return the default `Optional.empty()` from parent...  so what's the goal for this? we don't still have any customization that applies here or am I missing something?

Thx for clarifying!!

LuboTerifaj (approved):
Hello @karreiro , @jstastny-cz ,

I didn't find any new issues in Process designer.

However, just fyi, I have reported one present on master:
[JBPM-9408](https://issues.redhat.com/browse/JBPM-9408) _Stunner - Change "Unable to parse xml" validation messages in Alerts panel_

Thanks!

karreiro:
Hey @romartin, thanks for your answer :-)

Currently, we're returning the UUID of the element associated with the error when we **cannot find the name of the element**:

https://github.com/kiegroup/kie-wb-common/blob/e7d0b2a9197ebaad6dd6f464512f9378a3b4f04a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/i18n/ClientTranslationService.java#L88-L102

https://github.com/kiegroup/kie-wb-common/blob/e7d0b2a9197ebaad6dd6f464512f9378a3b4f04a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/i18n/CoreTranslationMessages.java#L136-L143

The goal of the `DiagramElementNameProvider` is to create a new layer that ables API users to decide (I) how to get the element name associated with the error message and (II) even if they want to provide the name for that element.

In the DMN editor context, we don't want to have the element name as the title of the alert:

Screen Shot 2020-10-06 at 11 12 18

Also, we're still keeping actual behaviour (of **returning the UUID when cannot find the name of the element**) if the editor doesn't have a provider:

https://github.com/kiegroup/kie-wb-common/blob/7a919db87685835139f93b4f3d31ec8f64619a66/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/i18n/ClientTranslationService.java#L100-L114

As an alternative, we could remove the `DiagramElementNameProvider`, but **we would need to change the current behaviour of returning the UUID as the default name of the node**. So this PR would be reduced to this change:

From this:

https://github.com/kiegroup/kie-wb-common/blob/e7d0b2a9197ebaad6dd6f464512f9378a3b4f04a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-core-common/src/main/java/org/kie/workbench/common/stunner/core/i18n/CoreTranslationMessages.java#L136-L143

To something like this:

```java
return Optional.of(message)
        .filter(StringUtils::nonEmpty)
        .map(msg -> translationService
                .getElementName(uuid)
                .map(name -> translationService.getValue(ELEMENT, name, msg))
                .orElse(msg));
```

However, I'm not sure if that's acceptable from the BPMN perspective. Wdyt? It would simplify this PR a lot and we would achieve the same result :-)

Thanks for the answer!

romartin (approved):

PR #3437, opened 2020-09-25
JBPM-9381 : Stunner - Data Object's Name with new line characters breaks the process

inodeman:
Hi @romartin @LuboTerifaj Can you check this PR. Fixes the issue in both BC and Kogito.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/util/StringUtils.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-core-api/src/main/java/org/kie/workbench/common/stunner/core/util/StringUtils.java
@@ -131,9 +131,6 @@ public class StringUtils {
                 case ' ':
                     sb.append("-");
                     break;
-                case '\n':
-                    sb.append("&newLine;");
-                    break;
                 default:
                     sb.append(c);
                     break;
@@ -155,8 +152,7 @@ public class StringUtils {
         return value.replace("&lt;", "<")
                 .replace("&gt;", ">")
                 .replace("&amp;", "&")
-                .replace("&quot;", "\"")
-                .replace("&newLine;", "\n");
+                .replace("&quot;", "\"");
     }
 
     /**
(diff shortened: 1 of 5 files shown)


inodeman:
jenkins do fdb

inodeman:
Hi @LuboTerifaj @romartin Artifacts uploaded to: https://drive.google.com/drive/folders/1im-f0Fk-PO3XiB-Dr_A9GqX8cHVgQenS?usp=sharing

LuboTerifaj:
Hello @inodeman , VSCode extension doesn't work for me locally. Did you test it or is there just an issue in my local environment? Thanks!

domhanak:
@LuboTerifaj @inodeman Same here, provided VSIX file does not load editors, only scesim is loaded.

LuboTerifaj (commented):
Hello @inodeman ,

I didn't test Kogito yet since it doesn't wok. See my previous [comment](https://github.com/kiegroup/kie-wb-common/pull/3437#issuecomment-698836247).

I have found two issues:
- Server log contains error messages.
- It is not possible to select Data Object in assignments: an unexpected error occurs. 
I think we can postpone them since by this fix we won't get broken processes.

I have reported issues above in a single jira:
- [RHPAM-3205](https://issues.redhat.com/browse/RHPAM-3205) Data Object has different Name on Canvas than in Properties panel

I will approve the PR once the Kogito is tested as well. Can you please have a look at the extension and let me know if the issue is only on my local machine?

Thank you!

hasys:
Hi @inodeman, shouldn't we just remove new line from the name? If I correct Data Objects name it is like variable name and is used as ID in the engine, so new line have no meaning there in my opinion. Anyway &newLine; is not valid XML escape value so no other product will handle it as well, if it is necessary to keep the new line in the name I suggest to replace it by `&#10;` WDYT?

inodeman:
Hi @hasys removing it so it gets merged. I think it is best to remove for now as you suggest. DO behave special as they have to share the same id and name and internally the Eclipse classes seem to break if there is something in them. Will test both scenarios

inodeman:
Hi @LuboTerifaj working on it. We will remove the newline character for now. The problem is the Eclipse classes. Next week I can work on a more permanent solution.

inodeman:
jenkins do fdb

inodeman:
Hi @LuboTerifaj @lazarotti @romartin sending PR. Will upload artifacts shortly. Verified works on BC and Kogito (WebApp and VSCode)

inodeman:
Hi @LuboTerifaj @romartin @lazarotti please find artifacts here.  https://drive.google.com/drive/folders/1im-f0Fk-PO3XiB-Dr_A9GqX8cHVgQenS?usp=sharing

romartin (approved):
Great, looks good to me! thanks @inodeman

LuboTerifaj (approved):
Hello @inodeman ,

I have found one issue, but it is present also on master:
[JBPM-9391](https://issues.redhat.com/browse/JBPM-9391) _Data Objects in Assignments disappear once the process is saved_

It looks good to me.

Thanks!

romartin:
thanks a lot @LuboTerifaj for the early review!!

PR #3442, opened 2020-09-29
AF-2644 : Package list fails to populate default package for imported projects

Rikkola:
**JIRA**: [AF-2644](https://issues.redhat.com/browse/AF-2644)

Problem is that there are two empty packages. One shows up as "nothing selected" and the other as "<default>". Removing one of them.

### DIFF
--- a/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/project/KieModuleServiceImpl.java
+++ b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/project/KieModuleServiceImpl.java
@@ -135,15 +135,17 @@ public class KieModuleServiceImpl
                                      PackageItem.DEFAULT_PACKAGE_NAME));
 
         for (final String packageName : packageServiceLoader.find(activeModule.getRootPath())) {
-            final StringBuilder packageBuilder = new StringBuilder();
-            for (final String token : packageName.split("\\.")) {
-                if (packageBuilder.length() != 0) {
-                    packageBuilder.append(".");
-                }
-                packageBuilder.append(token);
+            if (!packageName.trim().isEmpty()) {
+                final StringBuilder packageBuilder = new StringBuilder();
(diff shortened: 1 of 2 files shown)


jstastny-cz:
jenkins do fdb

kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/project/KieModuleServiceImpl.java
```diff
@@ -135,15 +135,17 @@ public KieModulePackages resolveModulePackages(Module activeModule) {
                                      PackageItem.DEFAULT_PACKAGE_NAME));
 
         for (final String packageName : packageServiceLoader.find(activeModule.getRootPath())) {
-            final StringBuilder packageBuilder = new StringBuilder();
-            for (final String token : packageName.split("\\.")) {
-                if (packageBuilder.length() != 0) {
-                    packageBuilder.append(".");
+            if (!packageName.trim().isEmpty()) {
```
Rikkola, line 138: The change is this if-line.

jstastny-cz:
jenkins do fdb

jstastny-cz:
Summary from our discussion:
* populated value is `<default>` for imported projects (using Try Samples project import)
* we will incorporate the fix within this PR

Rikkola:
@jstastny-cz Ok so looks like the problem is just with the DMN example. This is because the default package is formed from GAV and in this project it differs with the existing packages. I guess for this reason it is hard to know what the default package would be and fix might not be needed.

jstastny-cz:
Confirming that for imported projects the behavior is following:
* when package and GAV are in sync then the value in dialog is the expected package
* when package and GAV are not in sync then the value in dialog is <default>

jstastny-cz (approved):

vpellegrino (approved):
LGTM, thanks @Rikkola

Rikkola:
Jenkins run fdb

Rikkola:
@kiegroup/gatekeepers Can I get a merge please.

PR #3449, opened 2020-10-07
JBPM-9254 Stunner - BPMN Editor freezes in business central

hasys:
**JIRA**: [JBPM-9254](https://issues.redhat.com/browse/JBPM-9254)

**Business Central**: [here](https://drive.google.com/file/d/1jH__H8IhyHAhXlLWnqtqmL55wJ-omdU2/view?usp=sharing)

**VS Code**: [here](https://drive.google.com/file/d/1l292hIuhDJQwMdv-DIzm9d0_nZUBbHjD/view?usp=sharing)

<details>
<summary>
Hi @romartin, there is a fix for the jira, I tested it with business central and I can't reproduce it anymore. See details under the cut.
</summary>

<br>

<b>A side effect</b>: now if user will manually declare twice the same variable like `userVar[String]->processVar` all duplicates will be removed. But it won't remove variables if only one or two of tree parts (variable name, variable type, variable association) are the same, like `userVar[String]->processVar1` and `userVar[String]->processVar2` will be preserved.

Fix can be more elegant than adding addition `Set` to keep names of processed variables but it will need a bit more deep rework of marshallers due to missing overriding of `equals` method in EMF library. Since we are planning to replace marshallers by new one, I think this rework is unnecessary and this tmp `Set` should be enough. WDYT? If you agree with this solution I will duplicate solution for Kogito version of marshallers and attach Binaries for further testing. Thank you!

</details>

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/dataio/AssignmentsInfo.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/property/dataio/AssignmentsInfo.java
@@ -16,6 +16,8 @@
 
 package org.kie.workbench.common.stunner.bpmn.definition.property.dataio;
 
+import java.util.Objects;
+
 import org.jboss.errai.common.client.api.annotations.Portable;
 import org.jboss.errai.databinding.client.api.Bindable;
 import org.kie.workbench.common.forms.adf.definitions.annotations.metaModel.FieldDefinition;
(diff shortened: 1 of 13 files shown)


romartin (approved):
Thanks @hasys , this solution looks good enough to move forward, so let's go ahead with this PR!! :+1:

hasys:
jenkins do fdb

romartin:
ok @hasys thanks for the updates. Lemme just clarify plz - so I understand this issue finally affects both backend (BC) and client (kogito) side marshallers, so the fix is done on both sides on the code too. Is that correct?

hasys:
jenkins do fdb

hasys:
Hi @romartin, there are two issues:
1. Generated huge file
2. in Business Central validation/indexation mechanism causes freeze of Business central

The first one causes the second one and present in both Kogito and Business Central. But the second issue appears in Business Central only.

This PR fixing 1-st issue for both Kogito and Business Central which is the root cause of the second issue (from the issue title).

So in short: freeze not happens in Kogito, but the root cause of the freeze was present in both in Kogito and in Business Central and now fixed also in both places :)

hasys:
jenkins do fdb

hasys:
Sonar cloud failing due to duplications in Kogito and Business Central marshallers.

domhanak (approved):
Looks good to me. Tested in business central and VSCode editor.

Files linked to JIRA can be now fixed in BC or VSCode.
Duplicate data outputs are removed upon saving.

PR #3500, opened 2020-11-23
Stunner - Drop the showcase directory again, which was recreated by error

romartin:
Hey @mbiarnes 

As already discussed, this directory should be no longer present, it was recreated by error during the release process.

Thanks!


mbiarnes (approved):
+1

romartin:
I'll wait for the biuld to be ok and mergee, thanks!

PR #3508, opened 2020-12-02
Revert "[JBPM-9311] KIE server will not always deploy services, stay …

elguardian:
This is not in 7.44. removing it from master.


pefernan (approved):
+1 when green

PR #3558, opened 2021-02-05
Temporary disabling stunner's project showcase, as it's causing build…

romartin:
Hey @mareknovotny @tiagobento 

As this showcase is producing build issues in our CI, I would suggest to disable the gwt compilation for it in a temporary way, until we find the rigth solution. At least this way we'll be not blocking others' work.

WDYT?


romartin:
jenkins do fdb

mareknovotny (approved):
it can be good temporary solution until we find the real fix.

romartin:
looks now we're having same issues on the workbench distros builds guys, let's see how to address the fix asap please

caponetto:
I'd say we can revert this change since our builds are fixed now.
WDYT @romartin @mbiarnes?

romartin:
hey  @caponetto 

Thanks for the pointer above, in fact yes, we could revert this, but anyway this showcase, even compiling, is not running properly at runtime, so I created this ticket to manage it:

https://issues.redhat.com/browse/JBPM-9596

On the meanwhile, IMO there is no need to revert this until fixing the issues above, because either way we can't run it until fixing issues. So let's go ahead with the ticket and once providing the PR for it, we'll also revert this situation.

Hope it makes sense for you.

Thanks!

caponetto:
Thanks, @romartin! Looks good to me!

PR #3540, opened 2021-01-13
DROOLS-5911 - It's not possible to save arrow edits on DMN Editor

Josephblt:
DMN marshallers/unmarshaller were not prepared to handle auto-magnets.
Those will be handled as Double.Nan values (<di:waypoint x="NaN" y="NaN" />).

**JIRA**: [DROOLS-5911](https://issues.redhat.com/browse/DROOLS-5911)

**Business Central**: [WAR file](https://drive.google.com/file/d/1TBt05A-cEfzXN5udnqrv7TLflBK-O_-B/view?usp=sharing)

**VS Code**: [plugin](https://drive.google.com/file/d/1o2oWp6z5pW5ObjT7lVciPgWXBqiySq8z/view?usp=sharing)

PR #3571, opened 2021-02-19
KOGITO-4500: [DMN Designer] DMN schema/model validation errors when model has AUTO-SOURCE or AUTO-TARGET connections

karreiro:
**JIRA**: [KOGITO-4500](https://issues.redhat.com/browse/KOGITO-4500)

This PR fixes a small issue we introduced [here](https://github.com/kiegroup/kie-wb-common/pull/3540).

If we create a model with AUTO-SOURCE or AUTO-TARGET connections and we try to execute it (like this: [learn-dmn-in-15-minutes.com/learn/execution](https://learn-dmn-in-15-minutes.com/learn/execution)) the validation fails, and we can't execute the model.

With this change, users can successfully execute and use AUTO-SOURCE or AUTO-TARGET connections.

**VS Code**: [plugin before this PR](https://www.dropbox.com/s/2o8g1c2y3bm8wop/KOGITO-4500-before.vsix?dl=0) -- [plugin after this PR](https://www.dropbox.com/s/ze4wvpber34jusz/KOGITO-4500-after.vsix?dl=0)

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshallerStandalone.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshallerStandalone.java
@@ -144,9 +144,9 @@ public class DMNMarshallerStandalone implements DiagramMarshaller<Graph, Metadat
 
     private static final double CENTRE_TOLERANCE = 1.0;
 
-    private static final String AUTO_SOURCE_CONNECTION = "#AUTO-SOURCE";
+    private static final String AUTO_SOURCE_CONNECTION = "-AUTO-SOURCE";
 
-    private static final String AUTO_TARGET_CONNECTION = "#AUTO-TARGET";
+    private static final String AUTO_TARGET_CONNECTION = "-AUTO-TARGET";
 
     private XMLEncoderDiagramMetadataMarshaller diagramMetadataMarshaller;
     private FactoryManager factoryManager;
--- a/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerStandaloneTest.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-backend/src/test/java/org/kie/workbench/common/dmn/backend/DMNMarshallerStandaloneTest.java
@@ -2311,7 +2311,7 @@ public class DMNMarshallerStandaloneTest {
     @Test
     public void testIsAutoConnection() {
         String id = "DMNEdge-ID";
-        String autoConnectionID = "#AUTO-CONNECTION";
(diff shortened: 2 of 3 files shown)


karreiro:
Jenkins do fdb.

Josephblt (approved):

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshallerStandalone.java
```diff
@@ -144,9 +144,9 @@
 
     private static final double CENTRE_TOLERANCE = 1.0;
 
-    private static final String AUTO_SOURCE_CONNECTION = "#AUTO-SOURCE";
+    private static final String AUTO_SOURCE_CONNECTION = "-AUTO-SOURCE";
 
-    private static final String AUTO_TARGET_CONNECTION = "#AUTO-TARGET";
+    private static final String AUTO_TARGET_CONNECTION = "-AUTO-TARGET";
```
jomarko, line 149: I think this is backward incompatible change. If I understood correctly some models could be already created using '#' character. That is I think we should update also `isSourceAutoConnectionEdge` and `isTargetAutoConnectionEdge` to check both '#' and '-'. What do you think?

jomarko (commented):
Question about backward compatibility.

kie-wb-common-dmn/kie-wb-common-dmn-backend/src/main/java/org/kie/workbench/common/dmn/backend/DMNMarshallerStandalone.java
karreiro: @jomarko Well spotted, I agree. However, the only consequence is that we lose the "AUTO-TARGET" information and users can still use the model and update it. Given this https://github.com/kiegroup/kie-wb-common/pull/3540 was merged 18 days ago, and we have only one community release since that, I don't believe that we need to introduce extra logic to handle backward-compatibility.

jomarko: Isn't the issue more serious? If user created "auto" connection using the release you mentioned, then he will be never able to run such model without manual updating XML, even after upgrading to latest kogito tooling release, am I right @karreiro ?

karreiro: I don't think so, @jomarko. I created this DMN model with the old plugin version (listed in the PR description) with the `"#AUTO-TARGET"`, and I can use the new version of the plugin to set a `"-AUTO-TARGET"`:



Please, let me know if I'm missing something :-)

jomarko (approved):
Thank you @karreiro

PR #3613, opened 2021-04-14
KOGITO-4530: [DMN Designer] Boxed Expressions - Decision Tables - HiDPI is not working as expected

danielzhe:
**JIRA**: [KOGITO-4530](https://issues.redhat.com/browse/KOGITO-4530)

**VS Code**: [plugin](https://drive.google.com/file/d/1Ytt4Rmq2gUwqySWdOhXiuXfwSG1g5Ur1/view?usp=sharing)

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshaller.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-backend-api/src/main/java/org/kie/workbench/common/stunner/core/definition/service/DiagramMarshaller.java
@@ -51,7 +51,6 @@ public interface DiagramMarshaller<G extends Graph, M extends Metadata, D extend
      * @return
      */
     default MarshallingResponse<G> unmarshallWithValidation(final MarshallingRequest<InputStream, M> request) {
-
         try {
             final G result = unmarshall(request.getMetadata(), request.getInput());
             return MarshallingResponse.builder()
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/screens/DiagramEditorPropertiesScreen.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-client/src/main/java/org/kie/workbench/common/stunner/forms/client/screens/DiagramEditorPropertiesScreen.java
@@ -128,9 +128,11 @@ public class DiagramEditorPropertiesScreen {
             // Show the loading view.
             view.showLoading();
             // Open the forms properties widget for the current session.
+
             formPropertiesWidget
                     .bind(session)
                     .show(view::hideLoading);
+
             done = true;
         }
         if (!done) {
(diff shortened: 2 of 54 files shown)


romartin:
Hey   @danielzhe 

Thanks for the fix!

Anyway it's not yet clear to me why adding this call again, the canvas abstraction in stunner already sets the HiDPI once being initalized, see https://github.com/kiegroup/kie-wb-common/blob/1082e1994c9ed231384984b8bb8b9da2205b56fe/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/LienzoCanvas.java#L69

So maybe you're just missing to pass the right canvas setting flag for HiDPI? See https://github.com/kiegroup/kie-wb-common/blob/1082e1994c9ed231384984b8bb8b9da2205b56fe/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/LienzoCanvas.java#L67

WDYT?

danielzhe:
Because of the order of the loading. 
The flag must be set before the controls are loaded here: https://github.com/kiegroup/kie-wb-common/pull/3613/files#diff-85b5e66d7877c0ddefb922bd2167ebb3efb6c029e744ef36105f756d5ee083a6R178

That's the issue, not the flag being not set.
@romartin

karreiro (approved):
Thank you, @danielzhe! LGTM and works fine too! Also, thanks for the clarification :-)

karreiro:
Jenkins retest this.

karreiro:
Jenkins retest this.

karreiro:
Jenkins retest this.

karreiro:
Jenkins retest this.

karreiro:
Jenkins retest this.

danielzhe:
jenkins do fdb

romartin:
@danielzhe mmmm makes sense... lemme check code

romartin:
hey @danielzhe @karreiro good catch. Just looking at lienzo, how it handles the HiDPI flag, and it's being used at the construction of the canvas object. So makes sense that we're setting this flag in the wrong order... although then I don't understand why it even works most of the times :/

If the root cause is the order of execution of those lines of code, it's fine, but actual PR is duplicating code - I mean, we're adding another API method for setting the flag, which happens in the right order, but we're not refactoring the exisitng call, so adding this piece of code, it will end up on two calls to setHiDPIenabled happening at different points, which makes everything confusing. 

Also the `CanvasSetting` object does not just handle the HiDPI flag, it also manages other settings, such as canvas size, wich can be used at some point (although acutally our main canvas for editors is flexible and it fits to available size). So as this PR only creates the setter for the HiDPI flag, maybe we're also missing to set other settings, so an option is just passing the whole `CanvasSetting` object as well.

Soo sorry for bothering a bit guys, but I don't think this code, even working, is the right fix yet. It creates duplicity and also socpes for a subset of the canvas settings available fields.  Not sure how urgent is this, but we can we try to spend a few more time guys to try to get to some other fix that avoids duplicity at least? 

I'm actually looking at code, once having some idea I'll share... please lemme know if also have some clue on this. Anyway thanks guys, I don't want to discriminate this work because it's great, the analysis and finding of the root cause was not easy, but I would try to work on a more "suitable" fix... :) WDYT?

Thanks!!

karreiro:
@danielzhe alternatively, we could try to apply changes in the DMN side only (as the boxed expression editor will be removed/replaced by the new implementation in some weeks, we wouldn't need to be too strict with the approach).

danielzhe:
We can work in a more "suitable" fix, but it will require a *lot* of refactor.... @romartin 
About doing it on DMN side, it will not be easy too, because when we have the control at DMN side, the widgets are already loaded in memory with HiDPI off. @karreiro

romartin:
@danielzhe sorry I don't want to just drop this work, but I think we should get into some better fix, if possible. how urgent is this guys? Just schedule a call with me once this needs to be done and let's figure some solution together, by looking at code CC @karreiro

romartin:
oh sorry, just looking at this now! https://github.com/kiegroup/kie-wb-common/pull/3627 looks good!

PR #3432, opened 2020-09-23
RHPAM-3119 : [Data Modeler] Unsaved changes dialog appears after rename

akumar074:
**Issue**: When we change a package or fileName in the general properties of DataModeller and save it, the DataObject is reloaded. If we try to close the data object before reload is complete we see unsaved changes dialog. 

**Reason**: DataObject context and source are updated only during reloading.

**Solution**: The context and source hash are updated before reload. The DataObject is sent to the server after any change in DataModeller or the source to handle the inconsistencies in source and DataObject Model.

**JIRA**: [RHPAM-3119](https://issues.redhat.com/browse/RHPAM-3119)

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* Retest PR: <b>jenkins retest this</b>
* A full downstream build: <b>jenkins do fdb</b>
* A compile downstream build: <b>jenkins do cdb</b>
* A full production downstream build: <b>jenkins do product fdb</b>
* An upstream build: <b>jenkins do upstream</b>
</details>

PR #3628, opened 2021-04-28
JBPM-9715 - Data Object editor removes imports when saving

Josephblt:
Source content was not being parsed on server, this fixes the problem.

**JIRA**: [JBPM-9715](https://issues.redhat.com/browse/JBPM-9715)
**JIRA**: [RHPAM-3544](https://issues.redhat.com/browse/RHPAM-3544)

**Referenced Pull Requests**:
* https://github.com/kiegroup/kie-wb-common/pull/3629

**Business Central**: [WAR file](https://donwnload.here/something)

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
@@ -586,14 +586,8 @@ public class DataModelerScreenPresenter
             @Override
             public void execute(final String commitMessage) {
 
-                final DataObject[] modifiedDataObject = new DataObject[1];
-                if (isDirty()) {
-                    /* when DataObject editor or source is changed we
-                     * need to send the DataObject to the server  in order
-                     *  to let the source to be updated prior to save.
-                     */
-                    modifiedDataObject[0] = context.getDataObject();
-                }
+                final DataObject modifiedDataObject = getModifiedDataObject();
+
                 view.showSaving();
 
                 if (newTypeInfo != null) {
@@ -601,7 +595,7 @@ public class DataModelerScreenPresenter
                                         new DataModelerErrorCallback(Constants.INSTANCE.modelEditor_saving_error())).saveSource(
                             getSource(),
                             path,
-                            modifiedDataObject[0],
+                            modifiedDataObject,
                             metadata,
                             commitMessage,
                             newTypeInfo.getPackageName(),
@@ -611,7 +605,7 @@ public class DataModelerScreenPresenter
                                         new DataModelerErrorCallback(Constants.INSTANCE.modelEditor_saving_error())).saveSource(
                             getSource(),
                             path,
-                            modifiedDataObject[0],
+                            modifiedDataObject,
                             metadata,
                             commitMessage);
                 }
@@ -669,6 +663,13 @@ public class DataModelerScreenPresenter
                     versionRecordManager.reloadVersions(currentPath);
             }
         };
+    }
+
+    protected DataObject getModifiedDataObject() {
+        if (isDirty() && context.isEditorChanged()) {
+            return context.getDataObject();
+        }
+        return null;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Josephblt:
Jenkins run cdb

Josephblt:
Jenkins run fdb

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
```diff
@@ -671,6 +665,15 @@ public void callback(GenerationResult result) {
         };
     }
 
+    protected DataObject getModifiedDataObject() {
+        if (isDirty()) {
+            if (context.isEditorChanged()) {
```
Rikkola, line 670: Minor change, but this could be on one line just to make it faster to read.

Rikkola (commented):
Looks good, one formatting change.

Josephblt:
Jenkins run fdb
Jenkins run cdb

kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/java/org/kie/workbench/common/screens/datamodeller/client/DataModelerScreenPresenter.java
Josephblt: Agreed. Done.

jstastny-cz:
I am seeing slight change in behavior of what happens when saving non-valid file (missing import let's say).
After this change editor saves even a non-valid contents.
Before it threw a generic error:
```
Error: "An error occurred during saving". The server error is: "Source file couldn't be updated, path: default://master@MySpace/jstastny/src/main/java/com/myspace/jstastny/fgasd.java, dataObject: com.myspace.jstastny.fgasd."
```
I will re-check if it's caused by this change or is already like this in master/7.52.x.

Rikkola (approved):

Rikkola:
Validation before saving can take a while to run. This is if we need to run the full backend call for it. However if we can make a light check for the problem, that would be fine.

Josephblt:
Hello @jstastny-cz, @Rikkola. Toni, the changes are always parsed on the server, so the validation is already happening.

About saving invalid files, in my opinion it should be possible. The user should be able to submit the changes and pick up from where he left off at his will. In the matter of code corrections i believe they should be done in the sources tab. Regardless of my opinions, this was how it worked before the [PR](https://github.com/kiegroup/kie-wb-common/pull/3432) that introduced the problems.

When saving invalid data objects an alert message is added to the alerts windows. Those messages are easy to miss, specially when a green "Item successfully saved" popup is displayed at the same time. The way I see this are our options:
- Prompt the user about proceeding/cancelling save while bringing attention to the alert message.
- Display a information dialog (similar to the one that pops up when you change from source to model and there are errors) and proceed with the save. 
- Rely exclusively on the alert message and move on with the changes as they are.

Please let me know what you guys think. Thanks.

jstastny-cz:
@Josephblt you're right there are alerts. So I think it is fine to go ahead with your changes here and just track the change in behavior (I will take care of that).

Cause there's a slight difference in Alerts after saving non-valid contents vs. opening those:
1. after saving there are alerts for each occurrence of missing import type and field reference
  ```
  dateList cannot be resolved or is not a field
  List cannot be resolved to a type
  dateList cannot be resolved or is not a field
  List cannot be resolved to a type
  dateList cannot be resolved or is not a field
  List cannot be resolved to a type
  List cannot be resolved to a type
  ```
  and stacktrace in logs:
  ```
  Caused by: java.lang.ClassNotFoundException: Unable to find class 'List'
	at deployment.business-central.war//org.kie.soup.project.datamodel.commons.types.ClassTypeResolver.resolveType(ClassTypeResolver.java:268)
	at deployment.business-central.war//org.kie.soup.project.datamodel.commons.types.ClassTypeResolver.resolveType(ClassTypeResolver.java:153)
	at deployment.business-central.war//org.kie.workbench.common.services.datamodeller.util.DriverUtils.isManagedType(DriverUtils.java:149)
	... 16 more
  ```
2. after reopen there's just a single alert:
  ```
  It was not possible to create or load DataObject: "com.myspace.pr1.WithDate" . : Class could not be resolved for name: List. Unable to find class 'List' : Unable to find class 'List'
  ```
  and error in logs:
  ```
Caused by: java.lang.ClassNotFoundException: Unable to find class 'List'
	at deployment.business-central.war//org.kie.soup.project.datamodel.commons.types.ClassTypeResolver.resolveType(ClassTypeResolver.java:268)
	at deployment.business-central.war//org.kie.soup.project.datamodel.commons.types.ClassTypeResolver.resolveType(ClassTypeResolver.java:153)
	at deployment.business-central.war//org.kie.workbench.common.services.datamodeller.util.DriverUtils.isManagedType(DriverUtils.java:149)
	... 84 more
  ```

And as you can notice the root cause is the same, so I will probably file an issue reporting this discrepancy if it were only for documentation purposes.

jstastny-cz (approved):
LGTM.

PR #3677, opened 2021-08-23
AF-2891 : Space's default archetype is not respected when creating a new project

akumar074:
* if the default template is included for a space checkbox will be active.

**Thank you for submitting this pull request**

**JIRA**: 
[AF-2891](https://issues.redhat.com/browse/AF-2891)
[BAPL-1902](https://issues.redhat.com/browse/BAPL-1902)

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/AddProjectPopUpPresenter.java
+++ b/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/project/AddProjectPopUpPresenter.java
@@ -131,6 +131,8 @@ public class AddProjectPopUpPresenter {
         String getSelectedTemplate();
 
         void enableBasedOnTemplatesCheckbox(boolean isEnabled);
+
+        void enableTemplatesSelect(boolean isEnabled);
     }
 
     private Caller<LibraryService> libraryService;
@@ -263,14 +265,14 @@ public class AddProjectPopUpPresenter {
                 .collect(Collectors.toList());
 
         final boolean templatesAvailable = !templates.isEmpty();
+        final String defaultSelection = preferences.getDefaultSelection();
+        final boolean defaultTemplateIncluded = templates.contains(defaultSelection);
 
         if (templatesAvailable) {
-            final String defaultSelection = preferences.getDefaultSelection();
             final int selectedTemplateIdx = IntStream.range(0, templates.size())
                     .filter(i -> templates.get(i).equals(defaultSelection))
                     .findFirst()
                     .orElse(0);
-
             view.setTemplates(templates, selectedTemplateIdx);
         } else {
             view.setTemplates(Collections.singletonList(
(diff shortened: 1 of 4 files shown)


akumar074:
Jenkins run fdb

adrielparedes (approved):

akumar074:
Jenkins run fdb

barboras7:
Hi @akumar074. I checked the BC from FDB. From the description from the Jira, what the customer expects is: 

> Currently, you can set a default archetype for new projects in a given space (post with more details), but you still need to click on advanced options and based on a template.
The requirement is to be able to create on BC all new projects based on a default template without having to go on advanced options, and select "Based from template"

With the fix from this PR, the customer won't have to check the "Based on template" but they have to click the "Advanced options" anyway. IMHO, what the customer is expecting is something like checkbox with "Based on default template" in the new project dialog (without clicking the Advanced options). WDYT?

akumar074:
@barboras7 as per my understanding, the jira explains the behaviour that if there's default template for space, then we should be able to create new projects based on selected default template in space settings.
This fix provides this ability to create project with default template available to the space without selecting it again in advance options.

barboras7:
I have one concern with the current approach. Let me share a scenario:
- if I add some archetypes into the BC, there will be some archetype that will be considered "default"
- when I create a new space, the default archetype of the space will be automatically the one that is default for the whole BC

What happens now is that every time I create a new space, then new project, automatically, the project will be created based on the archetype (and not as an empty project). This changes behaviour of BC for all users, not just for those who want to use archetypes. 
Like this, each user that has some archetypes in the BC will automatically get new project based on template instead of an empty one.

akumar074:
@barboras7 
Usually, the default template should be used, without any other interference. Currently, to use the default template we need to select the checkbox similar to any other templates. 
This fix changes the behaviour for all the users, they will need to deselect the checkbox to create a new project without any template.
If we have to comply with previous behaviour, then we'll need to provide other implementations such as,
1. Create a new project preference for `Always use default template` -> with this user can turn on/off this behaviour.
or
2. Provide an option to keep templates in space settings without any default template.

@caponetto can you please share your thoughts on the above concern.

caponetto:
Hi @akumar074 @barboras7,

I gave it a try, just one comment:
With these changes, the `Based on template` checkbox is enabled even when there are no templates available.
This leads to an error when creating projects if users check the checkbox when there's no template.
Can you check that, @akumar074?

Regarding the concerns,
I completely understand the points that @barboras7 has raised.
In fact, the feature was designed this way to avoid such confusion.
The whole "create a project" flow that users were used to could not be changed when this feature was introduced, 
that's why users need to pro-actively go to the advanced options and select the template in order to be aware of what's going on.

On the other hand, adding an archetype and seeing it as "default" is also a little bit confusing.
One could think that "hey, I set a default one so every project should be created with this template with no other action".

With that being said, IMO this change is valid and we can only rely on the admin that they will know what they are doing when adding archetypes. 

Personally, I would move the `Based on template` checkbox out of the advanced options so it can be always visible.

akumar074:
Jenkins run fdb

barboras7 (approved):
Looks good, thanks @akumar074!

PR #3700, opened 2021-10-29
KOGITO-6036: Use html <input type='color'> picker

jomarko:
For more details see:
- https://issues.redhat.com/browse/KOGITO-6036

This is a backport of:
- kiegroup/kogito-editors-java#147

For 7.x stream `ColorPickerDialog` can not be removed from codebase, as it is used in `appformer` project on other place.

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-api/src/main/java/org/kie/workbench/common/stunner/forms/model/ColorPickerFieldDefinition.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-forms/kie-wb-common-stunner-forms-api/src/main/java/org/kie/workbench/common/stunner/forms/model/ColorPickerFieldDefinition.java
@@ -29,9 +29,10 @@ import org.kie.workbench.common.stunner.forms.meta.definition.ColorPicker;
 public class ColorPickerFieldDefinition extends AbstractFieldDefinition {
 
     public static final ColorPickerFieldType FIELD_TYPE = new ColorPickerFieldType();
+    public static final String COLOR_REGEXP = "^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$";
 
     @ColorPicker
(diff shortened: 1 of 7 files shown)


jomarko:
Jenkins please retest this

jomarko:
Jenkins do fdb

jomarko:
Jenkins please retest this

barboras7:
jenkins do fdb

barboras7:
Hi @jomarko,

I tested your implementation in Chrome, version 95. Here's how the UI looks for me when I click the "Background color":
Screenshot from 2021-11-11 11-55-17
My concern is that with this, it might not be obvious to the user that the hexa code for the colour of his choice is available for him. IMHO, it would be better to use some colorpicker that shows the hexa code to the user more directly. 
Also, another issue is that when I try to select the hexa code and copy it (Ctrl+C) and then paste it with Ctrl+V, it does not work. It would be better if the hexa code could be copied and then pasted into another hexa code input. 

Please, let me know your thoughts on this :slightly_smiling_face:

jomarko:
@barboras7 thank you for yor review.

I agree with you that it is not so straightforward to copy and paste color hex value from this color picker. however it is possible, as user can change the input mode by clicking bottom right button.

then on chrome the color picker looks like:
Screenshot from 2021-11-11 12-38-49


`<input type="color"/>` causes browser implementation is used. So on firefox is the dialog different, but also allows to copy and paste hex color value.


If you click plus icon in firefox
Screenshot from 2021-11-11 12-42-21

then you will get
Screenshot from 2021-11-11 12-42-27


Assuming the original DMN editor colorpicker didn't allow to copy and paste hex color value at all, I think we can proceed with this PR. What do you think?

Maybe I should create also some documentation about this change? Not sure, the color picker is quite common component also in other tools.

barboras7:
Thanks @jomarko for the explanation. Yes, I noticed that it is possible to display the hex value, however, I just wanted to check if there's no way that the user would see it without having to do any action to get there. Anyway, the PR looks good :tada:

barboras7 (approved):

jomarko:
@kiegroup/gatekeepers please merge

